PR #24, opened 2021-05-14
Adds Roe flux example, updates initial conditions for bickley jet, and uses different plotting

sandreza:
A few updates to the Bickley jet
- Roe Fluxes
- Balance Color Map
- Incompressible Initial Condition


sandreza:
closing since main is up to date :)

PR #25, opened 2021-05-14
Add column mesh

charleskawczynski:
Add a sketch of the column mesh.

### DIFF
--- a/src/mesh/column_mesh.jl
+++ b/src/mesh/column_mesh.jl
@@ -1,0 +1,804 @@
+#####
+##### Column mesh
+#####
+
+export n_cells, n_faces
+export ColumnMesh, coords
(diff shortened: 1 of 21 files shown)


src/mesh/column_mesh.jl
```diff
@@ -0,0 +1,242 @@
...
+end
+
+"""
+    LocalStencil
+
+A local stencil, used for interpolations or
+finite-differences. This is effectively a
+single row of the global operator.
+"""
+struct LocalStencil{T}
```
simonbyrne, line 47: I think a stencil should encode in the type space:
- whether it is _from_ a centered or face field
- whether it is _to_ a centered or face field
- width and offset: this could be encoded as a "hi, lo" so that we can easily determine the required ghost cells

PR #32, opened 2021-06-01
Fix support for Julia down to 1.4

charleskawczynski:
Not that we'll be running on lower versions of Julia, but it seems that this is all we need to support Julia versions down to 1.4.

### DIFF
--- a/.github/workflows/OS-UnitTests.yml
+++ b/.github/workflows/OS-UnitTests.yml
@@ -12,7 +12,16 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        version:
+          - '1.4'
+          - '1.5'
+          - '1.6'
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        arch:
+          - x64
 
     runs-on: ${{ matrix.os }}
 
--- a/bors.toml
+++ b/bors.toml
@@ -1,7 +1,13 @@
 status = [
-  "test-os (ubuntu-latest)",
-  "test-os (windows-latest)",
-  "test-os (macos-latest)",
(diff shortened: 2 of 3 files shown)


.github/workflows/OS-UnitTests.yml
```diff
@@ -13,6 +13,10 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        version:
+          - '1.4'
+          - '1.5'
+          - '1.6.0'
```
kpamnany, line 19: There's a 1.6.1 now. IIRC, simply making this '1.6' should include this version too...

kpamnany (approved):

charleskawczynski:
bors try

jakebolewski:
why do we need to support more than one version?

bors.toml
```diff
@@ -1,7 +1,13 @@
 status = [
-  "test-os (ubuntu-latest)",
-  "test-os (windows-latest)",
-  "test-os (macos-latest)",
+  "test-os 1.4 - (ubuntu-latest)",
```
jakebolewski, line 2: all of this seems completely unnecessary tbh

charleskawczynski:
We don't need to, but I am having to switch versions back and forth when developing on CM and CMCore

bors.toml
charleskawczynski: Ok, should I just add 1.5 and remove 1.4?

jakebolewski: yes, let's focus energies on 1 well supported version at a time

jakebolewski:
that doesn't seem to be a big issue?

simonbyrne:
I agree it's not worth trying to support <1.6 at the moment (for one thing, the CUDA versions are incompatible)

charleskawczynski:
Ok, we can always return to this if needed.

PR #35, opened 2021-06-01
Add field constructor, add heat equation example to buildkite

charleskawczynski:
This PR:
 - Adds a `CentField` and `FaceField` constructor that accepts `FiniteDifferenceSpace`s
 - Adds the heat equation example to buildkite (+ an animation)
 - Adds tests for `∇_cent_to_cent`
 - Moves all column operator tests to a single file

Now all we need is some convergence tests for the higher order operators


kpamnany (commented):
Looks okay to me, but better if Jake also takes a look.

I'm not sure why there needs to be a `CentField` and a `FaceField`; can they not just be `Field`s?

jakebolewski:
The addition of the buuildkite test looks good but I think we should hold
off on the  API changes.

charleskawczynski:
> I'm not sure why there needs to be a CentField and a FaceField; can they not just be Fields?

They were added just as outer constructors since they have different sizes. The object returned is still just a field.

> we should hold off on the  API changes.

Which API changes? Just the `CentField`/`FaceField` outer constructors?

jakebolewski:
Superseded by #40

PR #57, opened 2021-06-22
Define Base.axes(::Field) to return the space, instead of a tuple

simonbyrne:
This turns out to be easier than I originally thought. It also means we can remove `Fields.space`, as it is no longer needed.


simonbyrne:
bors try

jakebolewski (approved):
LGTM

PR #61, opened 2021-06-23
Change name SetValue -> SetInput

charleskawczynski:
This PR renames `SetValue` to `SetInput`, since it removes the ambiguity for gradients: `SetValue` could mean set value of the gradient, whereas `SetInput` excludes this potential ambiguity.

This also helps with clarity on setting fluxes when using `SetInput` on gradient fields (or products of scalars and gradients).

### DIFF
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -25,7 +25,7 @@ GradientC2F
 ### Boundary conditions
 
 ```@docs
-SetValue
+SetInput
 SetGradient
 Extrapolate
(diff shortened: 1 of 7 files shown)


charleskawczynski:
Can we merge this?

jakebolewski:
I would say let's get the PRs with functionality merged in before renaming stuff.

jakebolewski:
I think we are sticking with the current names for the time being

PR #63, opened 2021-06-30
Make FD operators callable

### LINKED ISSUE #53: Support callable operators
We just need to define a generic call -> broadcast(arg) fallback

jakebolewski:
Closes #53


simonbyrne:
I'm still not 100% convinced that this is a good idea:
1. it won't fuse intermediate stencils, leading to potential performance issues when used unwittingly
2. in Julia generally, there are only a handful of cases where `f(x)` and `f.(x)` do the same thing (namely scalar multiplication and vector addition/subtraction), and some where they do very different things (e.g. `exp(x)` vs `exp.(x)`).

jakebolewski:
This is mostly a UI issue from feedback by end users, applying simple interpolation stencils when iterating on a problem should not require broadcasting.  Fusing operators won't work anyways unless using broadcasting for simple expressions because how the BC change.  The behavior of calling an individual stencil is the same here as a callable or broadcasted so I'm not sure the semantics here really would change at all.  Since the use-case is kind of limited I don't think the performance issue is that much more of a concern.

PR #65, opened 2021-07-06
rename package to ClimaCore.jl

### LINKED ISSUE #64: Rename to ClimaCore.jl
ClimateMachineCore.jl was a temporary name. The software call agreed to go with ClimaCore.jl.

jakebolewski:
Closes #64

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -2,7 +2,7 @@ status = [
   "test-os (ubuntu-latest)",
   "test-os (windows-latest)",
   "test-os (macos-latest)",
-  "buildkite/climatemachinecore-ci",
+  "buildkite/climacore-ci",
   "format",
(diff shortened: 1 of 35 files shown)


bors.toml
```diff
@@ -2,7 +2,7 @@ status = [
   "test-os (ubuntu-latest)",
   "test-os (windows-latest)",
   "test-os (macos-latest)",
-  "buildkite/climatemachinecore-ci",
+  "buildkite/ClimaCore-ci",
```
charleskawczynski, line 5:
```suggestion
  "buildkite/climacore-ci",
```
I think?

jakebolewski: it does seem like it is case sensitive

PR #73, opened 2021-07-13
overload the default behavior of materialize! to check that the dest space aligns with the inferred stencil result space

### LINKED ISSUE #56: Better error message when tryingto broadcast into the wrong type of Field
Ex: 

```
operator = Operators.InterpolateC2F(
    left = Operators.SetValue(0.0),
    right = Operators.SetValue(0.0),
)
center_field .= operator.(cent_field, weights_field)
```

only throws:

```
ERROR: LoadError: AssertionError: idx == length(space.Δh_c2c)
Stacktrace:
 [1] stencil_right_boundary
   @ ~/clima/ClimateMachineCore.jl/src/Operators/finitedifference.jl:247 [inlined]
 [2] getidx
...
```

It would be good to catch that the `copyto!` field space is incorrect early before the stencil is applied

jakebolewski:
Closes #56


simonbyrne:
This seems fine.

FWIW, `Base.Broadcast` does this check by
1. [remaking the `Broadcasted` object](https://github.com/JuliaLang/julia/blob/v1.6.1/base/broadcast.jl#L894) with the `axes` the same as `dest`
2. Calling `instantiate`, which then calls [`check_broadcast_axes(axes(dest), bc.args...)`](https://github.com/JuliaLang/julia/blob/v1.6.1/base/broadcast.jl#L269)

We can't do the same as the axes of the input args don't match the output.

simonbyrne (approved):

jakebolewski:
Yeah unfortunately `Base` logic doesn't match what we need so we need to overload `materialize!`  I added some comments along that describes what you stated as to why we have to do this so hopefully it is clear.

Issue #75, opened 2021-07-16
Mesh warping


simonbyrne:
Add infrastructure for generic (non-regular, warped) meshes

1. 2D non-regular mesh (e.g. example from Taylor & Fournier, section 3.6)
2. Cubed-sphere mesh, equiangular, S2 geometry

valeriabarra:
Just putting down here some ideas that @jkozdon suggested and some questions that came up when we chatted about this yesterday, so that we can use these as a starting point for our discussion today.

Questions/ideas:
- We need to define a data structure to store the mesh in (are we interested in reading the mesh from file, generating it internally, or both?)
- Edge connectivity (2 elements/cells are connected to an edge in 2D)
- Corner connectivity (must support arbitrary number of connected elements in 2D or only quads?)

Steps:
- Define a data structure to store the mesh in 
- Define an `element2vertex` map, then from the `element2vertex` we build the connectivity
- Additional utilities/query functions, eg, for each edge return the pair of vertices `v1` and `v2` it connects (with orientation, perhaps `v1 < v2` --- in Julia we can use tuples for this)
- To test: Start with a simple structured mesh and deform the internal vertices by a sinusoidal, user-defined warping function (leave identity as default warping function)

simonbyrne:
The required interfaces are specified in https://github.com/CliMA/ClimaCore.jl/blob/main/src/Topologies/Topologies.jl

1. Need a way to determine face pairs: [map (elem, face) to neighbouring (nbrelem, nbrface, reversed)](https://github.com/CliMA/ClimaCore.jl/blob/4375878b7ae0010dcce8ff0baed319a64c195f55/src/Topologies/Topologies.jl#L59).

    Can do this two ways:
 a.  ClimateMachine.jl way: for each (elem, face), store an arrays of `elemtoelem`, `elemtoface`, `elemtoordr` (which basically matches our `reversed`): https://github.com/CliMA/ClimateMachine.jl/blob/74964c5f54af2f4d22e9229736f9d353e5ef2c0e/src/Numerics/Mesh/Topologies.jl#L118-L137
 b. store all mesh faces as (elem, face, nbrelem, nbrface, reversed)


2. A way to iterate over boundary faces: the easiest option is to store a vector of vectors: for each boundary tag, store a vector of (elem, face) pairs that are on that boundary

3. Iterate over mesh vertices, and the (elem, vert) pairs that are at that point (to support arbitrary vertex valence):
[...]

PR #82, opened 2021-07-30
Tensor-product meshes

valeriabarra:
Some development of things mentioned in issue #75 . 

This adds support for tensor-product meshes. They are still meshes with an underlying rectangular topology (internal vertices with valence 4), but not necessarily equispaced points. 

The tests were ported over from the `EquispacedRectangleMesh` case.


jakebolewski:
Don't we want to test functionality on something that is not an equispaced mesh?  This seems mostly to re-implement what we have in a slightly different way.

src/Meshes/Meshes.jl
```diff
@@ -108,4 +111,113 @@ struct EquiangularCubedSphereMesh{FT} <: AbstractMesh{FT}
...
+    end
+
+    if isnothing(coordinates)
+
+        coordinates = Vector{Cartesian2DPoint{FT}}(undef, (n1 + 1) * (n2 + 1))
+        # Default equispaced vertex coordinates, if the user has not specified their locations
+        range1 = range(domain.x1min, domain.x1max; length = n1 + 1)
+        range2 = range(domain.x2min, domain.x2max; length = n2 + 1)
+
+        # Coordinates array, row-major storage
```
jakebolewski, line 207: why are we storing these in row major?  shouldn't we use the same ND `DataLayouts` storage backing as the existing coordinate representation (and what the rest of the array ds use)

valeriabarra:
> Don't we want to test functionality on something that is not an equispaced mesh? This seems mostly to re-implement what we have in a slightly different way.

The discussion in #75 was to initially reproduce the same exact regular meshes we have, with a data structure that supports unstructured meshes, to make sure indeed to replicate existing behavior. The intermediate step of going tensor-product first, instead of fully unstructured, was preferred for simplicity and in view of what Tapio suggested. So yes, for now, this was a test to replicate the existing behavior with a different, slightly more flexible data structure. As it is outlined in the issue, the next step would be to test an actual, warped mesh.

src/Meshes/Meshes.jl
valeriabarra: Probably yes, I just wasn't familiar enough to use the existing DataLayouts. This was probably going to change anyways, once we had defined how and where to define the warping function. So this was only a quick implementation to test the existing behavior. This block of code might change anyways in view of the latest notes in #75

valeriabarra: When I chatted with @simonbyrne he said that we can leave this as is for now, since it is more pressing to get to a cubed-sphere mesh, for which this PR is an intermediate step than optimizing code. Unless you have a strong preference about this -- in that case, I might need some walk-through the existing DataLayouts storage to get familiar with it :)

valeriabarra:
This is ready for review, thanks!

test/tensorproductspectraloperators.jl
```diff
@@ -0,0 +1,246 @@
+using Test
+using StaticArrays
+import ClimaCore.DataLayouts: IJFH, VF
+import ClimaCore:
+    Geometry, Fields, Domains, Topologies, Meshes, Spaces, Operators
+using LinearAlgebra, IntervalSets
+
+
+domain = Domains.RectangleDomain(
```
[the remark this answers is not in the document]
valeriabarra: Done

valeriabarra:
bors try

src/Meshes/Meshes.jl
```diff
@@ -132,4 +133,134 @@ struct EquiangularCubedSphereMesh{FT} <: AbstractMesh{FT}
...
+    n2::Int64 # number of elements in x2 direction
+    faces::Vector{Tuple{Int64, Int64, Int64, Int64, Bool}}
+    coordinates::Vector{Cartesian2DPoint{FT}}
+end
+
+function TensorProductMesh(
+    domain::RectangleDomain{FT},
+    n1,
+    n2,
+    coordinates = nothing,
```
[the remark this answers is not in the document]
valeriabarra: We need to pass coordinates since in this type of mesh vertices are not assumed to be equispaced

valeriabarra:
bors try

src/Domains/Domains.jl
```diff
@@ -125,6 +125,12 @@ end
 coordinate_type(::RectangleDomain{FT}) where {FT} =
     Geometry.Cartesian2DPoint{FT}
 
+struct WarpedDomain{RD, F} <:
```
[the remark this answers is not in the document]
valeriabarra: Ok, `WarpedDomain` and `WarpedMesh` have been removed from this PR.

src/Topologies/tensorproductmesh.jl
```diff
@@ -0,0 +1,344 @@
...
+    mesh::M
+    boundaries::B
+end
+
+"""
+    TensorProductTopology(n)
+
+A tensor-product topology of `n` unstructured elements
+"""
+function TensorProductTopology(mesh::TensorProductMesh)
```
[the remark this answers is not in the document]
valeriabarra: Thanks for your feedback. I removed the redundant type and parametrized the Topology type on the underlying mesh type in [7252328](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/7252328b79e5ad50d5615fe53a03241b09bc0492). 
cc: @simonbyrne , this should address and solve what I was referring to earlier today

valeriabarra:
bors try

src/Topologies/Topologies.jl
```diff
@@ -166,10 +166,318 @@ struct Vertex{T <: AbstractTopology, V}
...
+    n2 = topology.mesh.n2
+    return n1 * n2
+end
+
+eachslabindex(topology::TensorProductTopology{M}) where {M <: AbstractMesh} =
+    1:nlocalelems(topology)
+
+
+# InteriorFaceIterator
+function Base.length(fiter::InteriorFaceIterator)
```
valeriabarra, line 222: As per @charleskawczynski 's suggestion on Slack, this and the following methods for the iterators could work as well with the following signature:
```
Base.length(fiter::InteriorFaceIterator{T}) where {M <: AbstractMesh, T <: TensorProductTopology{M}}
```
@simonbyrne , Since both Charlie's way and the one that I used here (i.e., without any type specification) work and pass all tests, is it advisable to change it for a better type safety, or it is really not that important and can be left as is now?

valeriabarra: I decided to go the safest way in [a1607c7](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/a1607c7e26483e592f57847facbd8288e7fb6ff1) and followed @charleskawczynski 's suggested way :)

valeriabarra:
bors try

src/Meshes/Meshes.jl
```diff
@@ -132,4 +134,113 @@ struct EquiangularCubedSphereMesh{FT} <: AbstractMesh{FT}
     n::Int64
 end
 
+"""
+    TensorProductMesh(domain::RectangleDomain, n::Integer)
+
+A tensor-product `AbstractMesh` of `domain` with `n1` elements in dimension 1, and `n2`
+in dimension 2.
+"""
+struct TensorProductMesh{FT, RD <: RectangleDomain{FT}} <: AbstractMesh{FT}
```
simonbyrne, line 143: can we move all the TensorProduct-specific stuff to a new file? I'd like the keep `src/Meshes/Meshes.jl` so that it only contains code for generic meshes

src/Topologies/Topologies.jl
```diff
@@ -166,10 +166,340 @@ struct Vertex{T <: AbstractTopology, V}
     num::V
 end
 
+"""
+    TensorProductTopology(mesh)
+
+A generic tensor-product topology defined on either an equispaced, regular rectangular
+structured grid or irregular one.
+"""
+struct TensorProductTopology{M <: AbstractMesh, B} <: AbstractTopology
```
simonbyrne, line 175: again, can we move tensor-product specific code to a new file?

src/Meshes/Meshes.jl
valeriabarra: Yes, but this would just be another generic mesh constructor, like we have already the `EquispacedLineMesh`, `EquispacedRectangleMesh` ones in the same file, and their corresponding `show` methods. Why would this fit in another category?

src/Topologies/Topologies.jl
valeriabarra: In commit [7252328](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/7252328b79e5ad50d5615fe53a03241b09bc0492), I renamed `GridTopology` -> `TensorProductTopology` since their constructors and behaviors were the same. I moved the common methods to this file, and separated only the methods that needed to be dispatched based on the distinct, concrete types in the corresponding files `grid.jl` and `tensorproductmesh.jl`. Where would the common code reside then?

valeriabarra: Like the common methods in a separate common file, say `Topologies/grids.jl` and the specific methods (basically only `vertex_coordinates` and `opposing_face` are left to be distinct)  in `Topologies/regulargrid.jl` (what is now `Topologies/grid.jl`) and `Topologies/tensorproduct.jl`?

valeriabarra: Okay, I've done this in [`945407c`](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/945407c14707b0234f7887b90c68dbc4643db408). I hope it is clearer now.

src/Meshes/Meshes.jl
simonbyrne: Because they're fairly small. This adds 200 lines, which probably makes sense to be in a different file.

src/Topologies/grids.jl
```diff
@@ -1,14 +1,15 @@
+
 """
-    GridTopology(n1,n2)
+TensorProductTopology(mesh)
```
simonbyrne, line 3: Should be indented 4 spaces so it renders as a code block in markdown
```suggestion
    TensorProductTopology(mesh)
```

src/Topologies/tensorproductgrids.jl
```diff
@@ -0,0 +1,32 @@
+function vertex_coordinates(
```
simonbyrne, line 1: why does this need to be in a separate file? would make sense to keep in `grids.jl`

src/Topologies/equispacedgrids.jl
```diff
@@ -0,0 +1,93 @@
...
+    z2, z1 = fldmod(elem - 1, n1)
+
+    c1 = Geometry.Cartesian2DPoint(range1[z1 + 1], range2[z2 + 1])
+    c2 = Geometry.Cartesian2DPoint(range1[z1 + 2], range2[z2 + 1])
+    c3 = Geometry.Cartesian2DPoint(range1[z1 + 1], range2[z2 + 2])
+    c4 = Geometry.Cartesian2DPoint(range1[z1 + 2], range2[z2 + 2])
+    return (c1, c2, c3, c4)
+end
+
+function opposing_face(
```
simonbyrne, line 38: does this not work for all `TensorProductTopology`s?

src/Topologies/grids.jl
simonbyrne: My main concern with the name `TensorProductTopology` is that we are going to use the term "tensor" elsewhere for a very different object (`AxisTensor`). I don't want to get too bogged down in naming stuff (we can always rename things later), so it don't let it block this, but do you have any other ideas (would `ProductTopology` be sufficient?)

simonbyrne: And perhaps rename this file accordingly (e.g `tensorproduct.jl`/`product.jl`)?

valeriabarra: Thanks for the feedback. No, unfortunately the term `ProductTopology` would not mean anything. How about `RegularGrid` and `IrregularGrid`? Or `UniformGrid` and `NonUniformGrid`?

src/Meshes/Meshes.jl
valeriabarra: okay

src/Topologies/equispacedgrids.jl
valeriabarra: It would, but all the following algorithm becomes a one-liner in the `TensorProductTopology` case (i.e., `return topology.mesh.faces[(elem - 1) * 4 + face][3:5]`), because of how the data structure is set up in its constructor. So I thought it was better to take advantage of that. But if you think we want to unify this and do not dispatch on the concrete type, I can get rid of the special implementation, and just leave this redundant way for that data structure

src/Topologies/tensorproductgrids.jl
valeriabarra: Okay, I thought it was more organized to have specific, concrete-typed methods in separate files, one for the regular case and one for the irregular one. I can keep them both in one file. 
Do you also prefer all the common methods of the abstract type together with the redefined methods for the concrete types all in one unique file?

valeriabarra: Done in [`82f89ed`](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/82f89ed04ce33546240e0489473b174aad391484). I hope it is clearer now.

src/Topologies/grids.jl
valeriabarra: In [`82f89ed`](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/82f89ed04ce33546240e0489473b174aad391484), I renamed back the original `GridTopology`, as it was, and put all implementations (dispatching on either type of concrete Mesh) in the same original file `grids.jl`. I hope it is clearer now. Thanks

src/Topologies/Topologies.jl
valeriabarra: In [`82f89ed`](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/82f89ed04ce33546240e0489473b174aad391484), I put both `GridTopology` implementations (ie, the one dispatching on the `EquispacedRectangularMesh` and the one dispatching on the `TensorProductMesh` back into the same `Topologies/grids.jl` file. I hope it is clearer now.

src/Meshes/Meshes.jl
valeriabarra: Done in [`82f89ed`](https://github.com/CliMA/ClimaCore.jl/pull/82/commits/82f89ed04ce33546240e0489473b174aad391484)

valeriabarra:
@simonbyrne , would you please let me know what is the current plan for this PR? Thanks

simonbyrne (approved):
Okay, I think we can merge this. @valeriabarra can you squash the commits?

valeriabarra:
@simonbyrne thank you! I reduced this to a few semantically distinct commits

PR #84, opened 2021-08-04
Unify LocalGeometry for spectral and finite difference spaces

simonbyrne:


src/Spaces/spectralelement.jl
```diff
@@ -16,8 +15,8 @@ end
 Topologies.nlocalelems(Space::AbstractSpace) =
     Topologies.nlocalelems(Space.topology)
 
-undertype(::Type{Geometry.LocalGeometry{FT, M}}) where {FT, M} = FT
-undertype(Space::AbstractSpace) = undertype(eltype(Space.local_geometry))
+undertype(space::AbstractSpace) =
+    DataLayouts.basetype(eltype(space.local_geometry))
```
jakebolewski, line 19: looks like this needs to be `local_geometry_data()`

PR #90, opened 2021-08-12
Add FieldVector wrapper

### LINKED ISSUE #87: Handling of fields on different spaces
When we have fields on different spaces (e.g. center and faces in a finite difference field), we currently use [`ArrayPartition` from RecursiveArrayTools.jl](https://github.com/SciML/RecursiveArrayTools.jl#arraypartition). The downside however is that you need to extract values via integer indices.

@ChrisRackauckas suggested [ComponentArrays.jl](https://github.com/jonniedie/ComponentArrays.jl), which lets you extract via names. We could also roll our own.

simonbyrne:
This wraps multiple fields as an OrdinaryDiffEq-compatible vector, replacing PartitionArray. Is compatible with implicit solvers.

Fixes #87.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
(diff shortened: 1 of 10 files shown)


jakebolewski (approved):

simonbyrne:
@Zhengyu-Huang @OsKnoth @dennisYatunin This should make it easier to to use the OrdinaryDiffEq implicit solvers with ClimaCore. The performance is far from optimal for now, but it should at least give you larger step sizes.

It's a bit rough around the edges, but please open issues or ask on #software if stuff doesn't work.

PR #92, opened 2021-08-13
Improve performance of `bickleyjet_dg_reference.jl` benchmark

kpamnany:
Not much improvement in `add_face_ref!`, mostly in `volume_ref!`. Previous:

old_volume

With this PR:

volume

PS> How do I make the Y axes use the same ticks?


jakebolewski:
There doesn't seem to be any significant improvement? `ylim=(mi, mx)`

http://docs.juliaplots.org/latest/generated/attributes_axis/

kpamnany:
Improvement is barely noticeable with a 16x16 grid. Also, this is using 1 thread -- many of these changes are aimed at improving scaling.

jakebolewski:
It would be good to start adding some scaling benchmarks / output if that is what we want to start measuring.

.buildkite/pipeline.yml
```diff
@@ -216,10 +216,21 @@ steps:
...
     key: "cpu_benchmark_bickleyjet_dg"
     command:
-      - "julia --color=yes --project=benchmarks/bickleyjet benchmarks/bickleyjet/runbenchmarks.jl"
+      - "julia --color=yes --project=benchmarks/bickleyjet benchmarks/bickleyjet/core_vs_ref.jl"
+    artifact_paths:
+      - "benchmarks/bickleyjet/*.png"
+    agents:
+      config: cpu
+      queue: central
+      slurm_ntasks: 1
```
jakebolewski, line 228: any slurm job cli flag  setting is passed through with the `slum_` prefix.  underscores get converted into hyphens so we don't have to escape, ex: `--mem-per-gpu=val` is `slum_mem_per_gpu: val` 

you'll need to adjust the number of tasks, or if you want to reserve a full node for perf testing

kpamnany:
Here's current thread scaling performance up to 16 threads:
vol_thread_scaling

.buildkite/pipeline.yml
kpamnany: Should I also add a `--exclusive`?

jakebolewski: I guess I did that in the past by reserving the whole node, but you can do that with `slurm_exlusive:  ""`

kpamnany:
bors try

jakebolewski:
it looks like the slurm manager killed it, can you launch what you are trying to do as a plain old slurm job with the settings you want?

kpamnany:
bors try

kpamnany:
bors try

kpamnany:
bors try-

kpamnany:
bors try

jakebolewski:
LGTM

kpamnany:
> LGTM

Great. Will leave it to you or @simonbyrne to merge.

PR #96, opened 2021-08-16
[WIP] finite difference operators to use vectors

simonbyrne:
change gradient to return a vector
add finite difference divergence


simonbyrne:
bors try

PR #135, opened 2021-08-31
Add hydrostatic balance test with total energy variable

Zhengyu-Huang:
With Oswald, we add the explicit/implicit solver test with total energy variables
hydrostatic_energy_end


jakebolewski:
Should we keep adding these examples to this repo? I'm not sure what additional functionality it is testing for the dycore.

Zhengyu-Huang:
> consolidate

Zhengyu-Huang:
> Should we keep adding these examples to this repo? I'm not sure what additional functionality it is testing for the dycore. I guess we can consolidate some of the existing examples.

That's a good point, we did not test any functionality, just tested the idea of using total energy.

PR #137, opened 2021-08-31
fix vertex iterator

### LINKED ISSUE #122: test/spectraloperators.jl: most tests fail when number of elem is different in x1 and x2 directions
While running tests for the `TensorProductMesh` development in #82, I realized that most tests in `test/spectraloperators.jl` fail when exercised with a different number of elements in the different directions, e.g., with

```
mesh = Meshes.EquispacedRectangleMesh(domain, 15, 16)
```
instead of the current value (16,16).

The only three in the file that pass are the `gradient` , `curl-curl`, and the `div` tests.

simonbyrne:
Fixes #122

cc: @valeriabarra 

need to add a test to `test/grid.jl`


valeriabarra:
This will close #122

PR #145, opened 2021-09-02
Vertical fd operator support on hybrid domains

### LINKED ISSUE #142: Add functionality allowing for broadcasting vertical operators over 1D, 2D extruded domains

jakebolewski:
Closes #142

### DIFF
--- a/test/hybrid.jl
+++ b/test/hybrid.jl
@@ -17,31 +17,68 @@ import ClimaCore.Domains.Geometry: Cartesian2DPoint
 function hvspace_2D()
     FT = Float64
     vertdomain =
-        Domains.IntervalDomain(FT(-π), FT(π); x3boundary = (:bottom, :top))
+        Domains.IntervalDomain(FT(0), FT(4π); x3boundary = (:bottom, :top))
     vertmesh = Meshes.IntervalMesh(
         vertdomain,
-        Meshes.ExponentialStretching(π / 2),
-        nelems = 20,
+        #Meshes.ExponentialStretching(π / 2),
+        nelems = 64,
     )
     vert_center_space = Spaces.CenterFiniteDifferenceSpace(vertmesh)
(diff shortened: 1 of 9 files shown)


jakebolewski:
bors try

bischtob:
Can we r+ this @simonbyrne ?

test/hybrid.jl
```diff
@@ -54,13 +91,13 @@ end
         # horizontal divergence operator applied to all levels
         hdiv = Operators.Divergence()
         @. dudt = hdiv(u * Geometry.Cartesian1Vector(1.0))
```
bischtob, line 93: Do we support this horizontal operator with Cartesian13Vectors like in the test above? It seems like we should be able to apply a spectral divergence to the full field, @simonbyrne, no?

test/hybrid.jl
```diff
@@ -17,31 +17,68 @@ import ClimaCore.Domains.Geometry: Cartesian2DPoint
...
+
+    hv_center_space, hv_face_space = hvspace_2D()
+    V =
+        Geometry.Cartesian13Vector.(
+            zeros(Float64, hv_face_space),
+            ones(Float64, hv_face_space),
+        )
 
+    function rhs!(dudt, u, _, t)
+        A = Operators.AdvectionC2C(
```
bischtob, line 57: Here it seems that that `AdvectionC2C` operates on `Cartesian13Vector`, should we expect the same behavior from the spectral operators, @simonbyrne ?

bischtob: Ok, I can see how we can leverage the implicit Ekman example for the vertical velocity equation here.

test/hybrid.jl
bischtob: Or at least we'd want to advect a `Cartesian3Vector` by a `Cartesian1Vector`?

PR #146, opened 2021-09-02
Add cube surface mesh.

sriharshakandala:


src/Meshes/BoxMesh.jl
```diff
@@ -0,0 +1,86 @@
+module BoxMesh
```
simonbyrne, line 1: This doesn't need to be a separate module, it's fine to live in `Meshes`/

src/Meshes/CubedSphereMesh.jl
```diff
@@ -0,0 +1,246 @@
+module CubedSphereMesh
```
simonbyrne, line 1: Same here

src/Meshes/Meshes.jl
```diff
@@ -132,4 +132,96 @@ struct EquiangularCubedSphereMesh{FT} <: AbstractMesh{FT}
...
+    "# of elements in the mesh"
+    nelems::I
+    "# of zones in the mesh"
+    nbndry::I
+    "x₁, x₂, ... coordinates of nodes `(nnodes, dim)`, dim can be greater than 2 for 2D manifolds embedded in higher dimensional space"
+    coordinates::FTA2D
+    "face node numbers `(nfaces, 2)`"
+    face_nodes::IA2D
+    "boundary elems for each face `(nfaces, 2)`"
+    face_neigh::IA2D
```
simonbyrne, line 184: Okay to be more verbose
```suggestion
    face_neighbors::IA2D
```

src/Meshes/BoxMesh.jl
```diff
@@ -0,0 +1,85 @@
+"""
+    rectangular_mesh(x1c, x2c, per = (false, false))
+
+This function builds a 2D rectangular mesh with points located at x1c and x2c in x1 and x2 directions respectively.
+"""
+function rectangular_mesh(x1c, x2c, per = (false, false))
```
valeriabarra, line 6: We have already a rectangular mesh. How is this different? And they are located in Meshes/Meshes.jl. Probably no need for a separate file

src/Meshes/Meshes.jl
```diff
@@ -132,4 +132,98 @@ struct EquiangularCubedSphereMesh{FT} <: AbstractMesh{FT}
...
+    "x₁, x₂, ... coordinates of nodes `(nnodes, dim)`, dim can be greater than 2 for 2D manifolds embedded in higher dimensional space"
+    coordinates::FTA2D
+    "face node numbers `(nfaces, 2)`"
+    face_nodes::IA2D
+    "boundary elems for each face `(nfaces, 2)`"
+    face_neighbors::IA2D
+    "face zones for each face `(nfaces, 1)`"
+    face_bndry::IA1D
+    "node numbers for each elem `(nelems, 4)`"
+    elem_nodes::IA2D
```
valeriabarra, line 188: I think these `nnodes` and `elem_nodes` can lead to confusion. We have vertex/vertices for the corners of a quadrilateral in a mesh. Nodes are the Finite Element nodes, i.e., there are p+1 of them, for a polynomial degree p finite element space. But the corners of a quad face are called vertices. See the API for [`vertex_coordinates`](https://github.com/CliMA/ClimaCore.jl/blob/f24aa819613346197cc4e6a8123a60b62ebf6b22/src/Topologies/Topologies.jl#L56)

src/Meshes/BoxMesh.jl
sriharshakandala: The rectangle_mesh function generates output in the Mesh2D format (conformal, unstructured). The advantage is that all 2D unstructured meshes (ex: cube surface mesh) can be handled by the same code. 
The interface `rectangular_mesh(x1c, x2c, per = (false, false))`, covers all 4 sided polygons (assuming warping function is applied to the rectangular domain). Equispaced, stretched or even manually spaced point locations can all be covered as wrappers around this function, where the wrappers feed in the respective x1c and x2c.

src/Meshes/BoxMesh.jl
```diff
@@ -1,3 +1,16 @@
+function equispaced_rectangular_mesh(
```
valeriabarra, line 1: How is this different from the existing [`EquispacedRectangleMesh`](https://github.com/CliMA/ClimaCore.jl/blob/0375673a843f263f9a5186db886e6cc32cc8b068/src/Meshes/Meshes.jl#L110). If it is not that different, then why not simply use the existing one?

src/Meshes/BoxMesh.jl
simonbyrne: If it's for comparison purposes, then it would probably make sense to (a) have a method to convert the structured to the unstructured, and (b) a unit test that tests them

jakebolewski:
Sorry it looks like I clicked the wrong button and commented on the commit :-/

sriharshakandala:
> I think we can eventually get rid of this in the PR and switch to the unstructured mesh as the default mesh for all 2D examples.

Yes. I believe, we need to update the interpolation and plotting code to work with unstructured mesh format.

sriharshakandala:
> we should use / define an accessor method here for the mesh coordinates
Yes. But, this is a different approach where we directly read the generated unstructured mesh.

simonbyrne (approved):
Thanks, tests look good. May need to be updated once #82 is in.

sriharshakandala:
bors try

PR #147, opened 2021-09-03
add hybrid advection example

bischtob:
**context**
We need to add a comprehensive diagonal advection test case for the hybrid domain.

**content**
 A simple example of a tracer that is advected by a constant velocity field is added. Many questions remain, but the example runs.


bischtob:
bors try

Project.toml
```diff
@@ -12,6 +12,7 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GaussQuadrature = "d54b0c1a-921d-58e0-8e36-89d8069c0969"
 IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
valeriabarra, line 15: Is this intended in the top-level, too, since it's also in the specific example environment?

examples/hybrid/advection.jl
```diff
@@ -0,0 +1,99 @@
...
+    horzdomain = Domains.RectangleDomain(
+        -1..1,
+        -0..0,
+        x1periodic = true,
+        x2boundary = (:a, :b),
+    )
+    horzmesh = Meshes.EquispacedRectangleMesh(horzdomain, 10, 1)
+    horztopology = Topologies.GridTopology(horzmesh)
+
+    quad = Spaces.Quadratures.GLL{8}()
```
valeriabarra, line 37: General question. Is this setting the polynomial degree, the number or finite element nodes, or the number of quadrature points?

bischtob: @simonbyrne, this is for you.

bischtob: 8 Refers to the number of quadrature points

valeriabarra: Thanks. And where is the polynomial degree set?

jakebolewski:
I think this can be closed?

PR #159, opened 2021-09-08
add transform function

simonbyrne:
Transforms the first axis of vectors and 2-tensors to a different basis


jakebolewski:
LGTM, our vectors will be <= 9 in length in practice so it shouldn't bog down the compiler too bad :)  we might want to have an assertion in the generated functions to bail on malformed inputs over a certain size.

PR #161, opened 2021-09-08
Fix sgeom⁺ in SpectralElementSpace2D

valeriabarra:
This was spotted while I was testing spectral operators with an irregular, warped mesh.


valeriabarra:
bors try

PR #165, opened 2021-09-11
add vector advection tests

simonbyrne:
Still needs:
- [x] concrete `@test`s: compare with analytic expression
- [x] remove diffeq solver (just test rhs evaluation


bischtob:
cc @valeriabarra, @LenkaNovak, @kmdeck

test/hybrid2d.jl
```diff
@@ -127,44 +133,216 @@ end
...
+    function rhs!(f)
+
+        divuf = Fields.FieldVector(h = zeros(eltype(f), hv_center_space))
+        h = f.h
+        dh = divuf.h
 
         # vertical advection no inflow at bottom
         # and outflow at top
-        Ic2f = Operators.InterpolateC2F(top = Operators.Extrapolate())
-        divf2c = Operators.DivergenceF2C(
```
valeriabarra, line 137: @simonbyrne , in PR #180 you mentioned to add convergence tests for `DivergenceC2F`, but I thought that here we found the problem with the boundary value of `DivergenceF2C` . Do both needs checking then?

test/Project.toml
```diff
@@ -11,6 +11,7 @@ Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
simonbyrne, line 14: You should add Revsie to your global environment (the one you get by starting Julia without `--project`), not ClimaCore (Julia "stacks" environments, so anything in the global environment is accessible elsewhere)

test/hybrid2d.jl
```diff
@@ -48,77 +48,83 @@ function hvspace_2D(
...
-    # so if you you have a periodic domain of size [-π, π]
-    # at time t, the solution is f(x - c * t, y)
-    # here c == 1, integrate t == 2π or one full period
+    # Divergence Operator
+    # ∂_x (c f)
+    # for this test, we use f(x) = sin(x) and c = 1
+    # => ∂_x (c f) = cos(x)
 
-    function rhs!(dudt, u, _, t)
+    function rhs!(f)
```
simonbyrne, line 104: the Julia convention is that the `!` suffix should only be used when mutating arguments: https://docs.julialang.org/en/v1/manual/style-guide/#bang-convention

Since this no longer mutates the arguments, it should drop the `!` (and similar elsewhere)

test/hybrid2d.jl
```diff
@@ -48,77 +48,83 @@ function hvspace_2D(
...
+        adv = zeros(eltype(f), hv_center_space)
         A = Operators.AdvectionC2C(
-            bottom = Operators.SetValue(sin(-t)),
+            bottom = Operators.SetValue(0.0),
             top = Operators.Extrapolate(),
         )
-        return @. dudt = -A(V, u)
+        return @. adv = A(c, f)
```
simonbyrne, line 84: Although it works, it is better practice to pass `C` as a function argument rather than rely on it being available as a global variable.

test/Project.toml
valeriabarra: Agreed. This was a slip! I didn't mean for this to be pushed

test/hybrid2d.jl
valeriabarra: Agree. I am still not done cleaning up these tests. I will rename `rhs!` to something meaningful of the operator being tested in the test, as I have done in the following ones

test/hybrid2d.jl
valeriabarra: I agree. I had kept that construction because it was like this in the original version with the ODE

valeriabarra: I applied the suggested change wherever possible. Some instances were not possible because of the broadcasting issue that was found.

valeriabarra:
Not ideal but for now I have bumped up the tolerance for the vector advection test. 
We can keep investigating. @LenkaNovak also mentioned he wanted to join the investigation :) so we can keep looking into it.

test/hybrid2d.jl
simonbyrne: Yes, that is a separate issue.

PR #173, opened 2021-09-12
Add getindex support for column (VF data layouts)

### LINKED ISSUE #66: Add indexing support for vertical columns
Alternatively if we don't want to provide full indexing, get a way to get the values at the bottom,top cell / face.

charleskawczynski:
This PR adds `getindex` support for column (VF data layouts) for `Field`s

Closes #66.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -34,7 +34,7 @@ Field(::Type{T}, space::S) where {T, S <: AbstractSpace} =
 const SpectralElementField2D{V, S} =
     Field{V, S} where {V <: AbstractData, S <: Spaces.SpectralElementSpace2D}
 
-const FiniteDifferenceField{V, S} =
+const ColumnFiniteDifferenceField{V, S} =
     Field{V, S} where {V <: AbstractData, S <: Spaces.FiniteDifferenceSpace}
 const FaceFiniteDifferenceField{V, S} =
     Field{V, S} where {V <: AbstractData, S <: Spaces.FaceFiniteDifferenceSpace}
@@ -71,6 +71,13 @@ Base.propertynames(field::Field) = propertynames(getfield(field, :values))
 
 @inline Base.getproperty(field::Field, name::Integer) =
     Field(getproperty(field_values(field), name), axes(field))
+
+Base.@propagate_inbounds function Base.getindex(
+    field::ColumnFiniteDifferenceField,
+    i::Integer,
+)
+    Base.getindex(field_values(field), i)
+end
 
 Base.eltype(field::Field) = eltype(field_values(field))
 Base.parent(field::Field) = parent(field_values(field))
--- a/src/Plots/plots.jl
+++ b/src/Plots/plots.jl
@@ -36,7 +36,7 @@ function UnicodePlots.heatmap(
 end
 
 function UnicodePlots.lineplot(
-    field::Fields.FiniteDifferenceField;
+    field::Fields.ColumnFiniteDifferenceField;
     name::Union{Nothing, Symbol} = nothing,
     width = 80,
     height = 40,
@@ -73,7 +73,7 @@ function UnicodePlots.lineplot(
     )
 end
 
-RecipesBase.@recipe function f(field::Fields.FiniteDifferenceField)
+RecipesBase.@recipe function f(field::Fields.ColumnFiniteDifferenceField)
     # unwrap the data to plot
     space = axes(field)
     xdata = parent(field)[:, 1]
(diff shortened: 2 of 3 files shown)


src/Fields/Fields.jl
```diff
@@ -72,6 +72,13 @@ Base.propertynames(field::Field) = propertynames(getfield(field, :values))
 @inline Base.getproperty(field::Field, name::Integer) =
     Field(getproperty(field_values(field), name), axes(field))
 
+Base.@propagate_inbounds function Base.getindex(
+    field::Field{V},
+    i::Integer,
+) where {V <: DataLayouts.VF}
```
jakebolewski, line 78: We probably want an alias for `ColumnField` to be more explicit (see above).

jakebolewski (commented):
I think this needs a bit more work, for example simple things like slicing `zf[1:2]` won't work as expected.

charleskawczynski:
> I think this needs a bit more work, for example simple things like slicing zf[1:2] won't work as expected. I'm not quite sure the requirements projects need in terms of functionality.

I'm able to run Bomex in TC.jl with just these changes, but I agree that we should fix slicing if it wouldn't work as expected.

jakebolewski:
We can start with bare minimum support if this is all that is required to be useful and just create an issue for indexing support that is missing, keeping it bare bones might be better overall anyways.

src/Fields/Fields.jl
charleskawczynski: Are you suggesting we write this as `getindex(field::ColumnField, i::Integer)` and then just define a `ColumnField` to specify the types we want this to work for?

jakebolewski: Yes, that way for any changes we just have to update one def. in the future.

src/Fields/Fields.jl
```diff
@@ -56,6 +56,9 @@ const CenterExtrudedFiniteDifferenceField{V, S} = Field{
     S,
 } where {V <: AbstractData, S <: Spaces.CenterExtrudedFiniteDifferenceSpace}
 
+const Column =
```
jakebolewski, line 59: Let's call this `ColumnField` to be consistent with all the other names

src/Fields/Fields.jl
```diff
@@ -72,6 +75,10 @@ Base.propertynames(field::Field) = propertynames(getfield(field, :values))
 @inline Base.getproperty(field::Field, name::Integer) =
     Field(getproperty(field_values(field), name), axes(field))
 
+Base.@propagate_inbounds function Base.getindex(field::ColumnField, i::Integer)
+    Base.getindex(field_values(field), i)
```
jakebolewski, line 79: If the goal is to eventually use `PlusHalf` types for typesafe indexing (noted here: https://github.com/CliMA/TurbulenceConvection.jl/issues/256#issuecomment-917803482) then you need to use the FiniteDifference `getidx` methods (`getidx(field_data, ::Interior, idx)`) operator instead of punching through to the datalayout getindex call.

charleskawczynski: Yeah, that's the eventual goal, although I think we're still a bit aways away. I'm fine with making this change if it works either way

charleskawczynski: Hmm, `Interior` (and `getidx`) is defined in the `Operators` module, which is included after the `Field` module. Should we define a `getidx` function stub and move `Interior` def into the `Field` module?

jakebolewski: Hmm, maybe we want to punt on this for now and discuss what exactly we need on the API side to support the changes you need.

charleskawczynski: Sounds good

charleskawczynski:
Anything else needed here? @jakebolewski

src/Fields/Fields.jl
```diff
@@ -56,6 +56,9 @@ const CenterExtrudedFiniteDifferenceField{V, S} = Field{
     S,
 } where {V <: AbstractData, S <: Spaces.CenterExtrudedFiniteDifferenceSpace}
 
+const ColumnField =
+    Field{V, S} where {V <: DataLayouts.VF, S <: Spaces.FiniteDifferenceSpace}
```
simonbyrne, line 60:
```suggestion
const ColumnField =
    Field{V, S} where {V <: DataLayouts.DataColumn, S <: Spaces.FiniteDifferenceSpace}
```
or maybe we could just rename `FiniteDifferenceField` above?

charleskawczynski: I'm fine with either one. But aren't we suppose to restrict this to `DataColumn`? `FiniteDifferenceField ` is only restricted to `AbstractData`.

charleskawczynski: I made the suggested change to `DataColumn`.

simonbyrne: we only support finite difference fields as columns at the moment

simonbyrne: I'm just not keen on having two aliases for the same thing

simonbyrne: Actually, lets rename the above one to `ColumnFiniteDifferenceField`

charleskawczynski: Done.

src/Fields/Fields.jl
simonbyrne: I would like to use `PlusHalf` for indexing faces: our experience so far is that it catches a _lot_ of off-by-one errors. Why don't we move the `PlusHalf` into the `Fields` module, and then define `Base.getindex`.

simonbyrne: We can leave `getidx` and `Interior` in `Operators` for now. We don't need those for this.

charleskawczynski: Ok, so are you proposing we only support `Base.getindex(field::FaceFiniteDifferenceField, ::PlusHalf)` and `Base.getindex(field::CenterFiniteDifferenceField, ::Int)`?

simonbyrne: Yes, exactly

charleskawczynski: Hmm, I'm not sure how we can do that, since [TC.jl's 255](https://github.com/CliMA/TurbulenceConvection.jl/pull/255) only replaces the grid to start with. The indices that we traverse ([here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/src/Grid.jl#L59)) are used both for the coordinates (which would be fine for our grid) and for field variables that aren't using ClimaCore.Field's yet (which would be problematic).

charleskawczynski: I also don't mind pirating these methods if that's preferred.

simonbyrne: I would go with pirating what you need for now.

charleskawczynski: Maybe I'll just do that for now. This is kind of holding up my primary task.

PR #180, opened 2021-09-14
FD operator fixes from bubble PR

simonbyrne:
- `return_eltype` of `GradientOperator` to use generic machinery
- `SetValue` bc of `DivergenceC2F`


simonbyrne:
I think we need a test for `DivergenceC2F`: should be able to adapt the convergence tests we use for `GradientC2F`.

simonbyrne:
These were incorporated as part of #185.

simonbyrne:
Still need a test for `DivergenceC2F`

PR #195, opened 2021-09-17
forcibly inline get_struct

simonbyrne:
When dealing with complicated structs (such as AxisTensors), we were hitting some sort of inlining limit that resulted in allocations. This forcibly inlines it by unrolling the generated function, and adds a test.


simonbyrne:
bors try

jakebolewski:
bors try

jakebolewski:
Looks like with these changes we have about a 30% improvement in rising bubble case (1 hour -> 40 min).  I'll take a look to see if we can propagate the suggestion by @vchuravy throughout the stencil code.

vchuravy:
I think the code can be made a faster still.

Remember my suggestion to rewrite `column` like `Base.Broadcast.preprocess` to make it easier on the compiler. 

Also we stack allocate large structs and if we avoid that by only passing the information we need to the error function as well as inline a bit more agressivly. We should be able to avoid large stack copies and the last allocations

vchuravy:
I just gave this another stab, but the recursive nature of `column` stumped me... I tried inroducing `preprocess_args` but that didn't help much more...

vchuravy:
FYI  besides Cthulhu I have been using:

```
julia> open("f.jll", "w") do io; show(io, @code_typed(debuginfo=:source, f!(divx, x))); end

julia> open("f.ll", "w") do io; code_llvm(io, f!, (typeof(divx), typeof(x)), debuginfo=:source); end
```

To test the impact of changes.

vchuravy:
- [ ] bounds-check elission in `setidx!`
- [ ] `stencil_interior` not inlined
- [ ] `get_boundary` can trigger error: https://github.com/CliMA/ClimaCore.jl/blob/432faf01acf9c01cf86045db14c1971932db5aa7/src/Operators/finitedifference.jl#L1602 
- [ ] `column_rec` not inlined
- [ ] `check_broadcast_shape` large arguments being passed on the stack -> no SROA
- [ ] `bc_error`  large arguments being passed on the stack -> no SROA

My last change https://github.com/CliMA/ClimaCore.jl/pull/195/commits/432faf01acf9c01cf86045db14c1971932db5aa7 seems to have dropped the allocations to zero, when including the file twice? That's rather strange.

```
julia> include("test/allocations.jl")
#= /home/vchuravy/src/ClimaCore/test/allocations.jl:35 =# @allocated(f!(divx, x)) = 480
Test Failed at /home/vchuravy/src/ClimaCore/test/allocations.jl:36
  Expression: #= /home/vchuravy/src/ClimaCore/test/allocations.jl:36 =# @allocated(f!(divx, x)) == 0
   Evaluated: 480 == 0
ERROR: LoadError: There was an error during testing
in expression starting at /home/vchuravy/src/ClimaCore/test/allocations.jl:36

julia> include("test/allocations.jl")
#= /home/vchuravy/src/ClimaCore/test/allocations.jl:35 =# @allocated(f!(divx, x)) = 0
Test Passed
  Expression: #= /home/vchuravy/src/ClimaCore/test/allocations.jl:36 =# @allocated(f!(divx, x)) == 0
   Evaluated: 0 == 0
```

vchuravy:
Yeah on the second `include` `column_rec` get's inlined... @aviatesk any ideas on how that could be?

aviatesk:
I'm still trying to understand why `abstract_call_method` bails out for this case (`ClimaCore.Operators.AbstractStencilStyle` is very complex !).

For now you can try this super hacky solution would be:
```diff
diff --git a/src/Operators/finitedifference.jl b/src/Operators/finitedifference.jl
index 59d357b..4cb7017 100644
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -1703,10 +1703,16 @@ end
 end
 
 # Doesn't remove self-referential of `column`, but is best way of writing it.
-column_rec(inds, arg) = (column(arg, inds...),)
-column_rec(inds, arg, args...) = (column(arg, inds...), column_rec(inds, args...)...)
+column_rec(inds, arg) = (column(arg, inds),)
+column_rec(inds, arg, args...) = (column(arg, inds), column_rec(inds, args...)...)
+if hasfield(Method, :recursion_relation)
+    dont_limit = (args...) -> true
+    for m in methods(column_rec)
+        m.recursion_relation = dont_limit
+    end
+end
 
-# This is self-referential when `bc.args` contains Broadcasted 
+# This is self-referential when `bc.args` contains Broadcasted
 @inline function column(
     bc::Base.Broadcast.Broadcasted{Style},
     inds...,
@@ -1765,5 +1771,5 @@ end
         Base.@_inline_meta
         @inbounds elem = tup[I]
         f(elem)
-    end 
+    end
 end
```

test/allocations.jl
```diff
@@ -0,0 +1,35 @@
...
+
+function f!(divx, x)
+    divf = Operators.DivergenceF2C()
+    # works without negation
+    @. divx = -divf(x)
+    return nothing
+end
+
+f!(divx, x)
+@test @allocated(f!(divx, x)) == 0
```
aviatesk, line 35: @vchuravy is it intentional to modify `divx` in the first call ?

vchuravy:
On 1.7-rc1 this now no longer allocates (thanks @aviatesk)

@simonbyrne 

```
40 ─│││││││        invoke ClimaCore.Operators.get_boundary($(QuoteNode(ClimaCore.Operators.DivergenceF2C{NamedTuple{(), Tuple{}}}(NamedTuple())))::ClimaCore.Operators.DivergenceF2C{NamedTuple{(), Tuple{}}}, $(QuoteNode(ClimaCore.Operators.RightBoundaryWindow{:top}()))::ClimaCore.Operators.RightBoundaryWindow{:top})::Union{}
```

Is the signature for `get_boundary` that could throw an error.

vchuravy:
@jakebolewski any reason why https://github.com/CliMA/ClimaCore.jl/blob/44c9cf6b9d6dccaba648f8084fb2185a2c1fa3ce/src/DataLayouts/DataLayouts.jl#L595 is not `@propagate_inbounds`?

jakebolewski:
@vchuravy I think all the get/setindex should be `Base.propagate_inbounds`, I don't think we gone through and updated these consistently

src/DataLayouts/DataLayouts.jl
```diff
@@ -590,8 +590,9 @@ end
     col[I[4]] = val
 end
 
-@inline function Base.setindex!(data::VF{S}, val, v::Integer) where {S}
-    set_struct!(view(parent(data), v, :), convert(S, val))
+@propagate_inbounds function Base.setindex!(data::VF{S}, val, v::Integer) where {S}
+    subdata = @inbounds view(parent(data), v, :)
+    set_struct!(subdata, convert(S, val))
```
jakebolewski, line 595: I think we can elide the view construction and just use pre-computed strides here, similar to the rewritten getstruct impl in the PR

vchuravy: Yeah but this works as well xD

test/allocations.jl
```diff
@@ -0,0 +1,40 @@
...
+    Topologies,
+    Spaces,
+    Fields,
+    Operators
+
+using Test
+
+Nv = 100
+
+vertdomain = Domains.IntervalDomain(0.0, 1000.0; x3boundary = (:bottom, :top))
```
jakebolewski, line 17: this needs to be `Domains.IntervalDomain(Geometry.ZPoint(0.0), Geometry.ZPoint(1000.0), boundary_tags=(:bottom, :top))`, should fix the test failure

jakebolewski:
bors try

jakebolewski:
Looks like a bunch of  performance regressions with these changes unfortunately, I think we probably have to tackle this more holistically and build better infrastructure measure impacts at different layers of abstraction.

jakebolewski:
One obvious difference is that we are not running 1.7 on CI so lots of the targeted changes in that direction do not apply

jakebolewski:
bors try

jakebolewski:
Closing as this has been mostly superseded by a minimal set of patches, but broadly following the course laid out in this PR.

jakebolewski:
In the end getting everything to inline correctly on 1.7 rc2 gives us the target ~10x speedup.

Issue #13, opened 2021-04-19
Add dimension objects to support named dimensions with units + metadata for domains


jakebolewski:
Domains should take labeled dimension objects with optional units for unit conversion between across the model interface.  Named dimensions allow for higher level indexing into model domains and can be passed through to support output formats that adhere to the CF metadata conventions.

Alternatively we could add support for this this as part of the `Domain` constructor and have dimensions is a hidden interface.

simonbyrne:
Would be useful to have some examples for this.

Issue #81, opened 2021-07-28
Coordinates and vectors


simonbyrne:
As we add 3D support, we need to modify how we represent coordinates and vectors. After speaking with a few people, here are my current thoughts:

# Coordinates

- Every domain should have a "canonical" global coordinate system.

- In both the box and spherical shell, we should call the vertical coordinate the same thing. `x3` doesn't really make sense for a sphere, so it probably think we should call it `z`. 

- If we call the vertical coordinate `z`, it doesn't make sense to call the horizontal coordinates `x1` and `x2`....

So, what I was thinking is
- `XYPoint(x, y)`: a horizontal point in Cartesian space
- `LatLongPoint(lat, long)`: a point on a sphere (in radians)
- `ZPoint(z)`: a point in the vertical (typically 0 = sea level)
- `XYZPoint(x,y,z)`: a point in a box
- `LatLongZPoint(lat, long, z)`: a point in a spherical shell

We can define a `CartesianPoint` for conversion purposes (e.g. plotting) but probably won't use it by default (it would match the `XYZPoint` in the case of a box).

# Vectors

- if we go with `x,y,z`, then we should call the vector components `u,v,w`:
- again, it would be ideal if we can use the same types consistently with the box & sphere.

So we end up with:
- `UVVector(u, v)`: these are the orthonormal horizontal components
  - in the box, `u` is aligned with `x`, `v` with `y`.
  - on a sphere these are the components of the lat/long basis vectors
- `WVector(w)`: the vertical components
- `UVWVector(u,v,w)`: both horizontal and vertical components

We would keep the `Covariant`/`Contravariant` types as currently defined.

Again, we can define a `CartesianVector` for conversion purposes, but would not be used by default.

# Storage

At each point in space, we would need to store:
a. canonical coordinate 
b. conversion matrix from `Covariant` to canonical vector component (e.g. `UVWVector`)
c. conversion matrix from canonical to `Contravariant` (is this just the transpose of b?)
d. covariant metric tensor
e. sqrt Jacobian determinant of covariant metric tensor (Riemann volume form)

For DSS vector quantites:
 - the easiest option probably to convert to the global vector basis, DSS, then convert back

# Open questions

- How do we apply this idea to the FD approach? 
- What do we need for numerical fluxes?

tapios:
Looks good. A couple comments:
* If it's not much trouble (e.g., in evaluating trigonometric functions), I suggest we use degrees for latitude and longitude.
[...]

PR #201, opened 2021-09-21
Domain coordinate types

jakebolewski:
part of #81


jakebolewski:
This will close #13  minus the metadata part, as the typed geometry point type is equivalent to a "labeled" dimension.

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
Needs a review and squash, I largely didn't touch the unstructured mesh components as they don't quite align with the previous abstractions and will need a bit more work (but largely are self contained).

jakebolewski:
bors try

src/Geometry/coordinates.jl
```diff
@@ -43,32 +52,75 @@ end
 
 @pointtype Cartesian12Point x1 x2
 @pointtype Cartesian123Point x1 x2 x3
+
 # TODO: deprecate these
```
simonbyrne, line 56: can remove the TODO

src/Geometry/coordinates.jl
```diff
@@ -43,32 +52,75 @@ end
...
-    Cartesian3DPoint{FT}(pt.x, pt.y, pt.z)
-
-components(p::AbstractPoint) = SVector(ntuple(i -> getfield(p, i), nfields(p)))
+Cartesian12Point(pt::XYPoint{FT}) where {FT} = Cartesian12Point{FT}(pt.x, pt.y)
+Cartesian123Point(pt::XYZPoint{FT}) where {FT} =
+    Cartesian123Point{FT}(pt.x, pt.y, pt.z)
+
+ncomponents(p::AbstractPoint) = nfields(p)
+components(p::AbstractPoint{FT}, i::Symbol) where {FT} = getfield(p, i)::FT
+components(p::AbstractPoint{FT}, i::Integer) where {FT} = getfield(p, i)::FT
```
simonbyrne, line 82: Should this be singular?
```suggestion
component(p::AbstractPoint{FT}, i::Symbol) where {FT} = getfield(p, i)::FT
component(p::AbstractPoint{FT}, i::Integer) where {FT} = getfield(p, i)::FT
```

src/Geometry/coordinates.jl
```diff
@@ -43,32 +52,75 @@ end
...
+    _coordinate_type(ptyp, Val(ax))
+coordinate_type(ptyp::Type{<:AbstractPoint}, ax::Integer) =
+    _coordinate_type(ptyp, Int(ax))
+
+coordinate(pt::AbstractPoint, ax::Int) = _coordinate(pt, Val(ax))
+coordinate(pt::AbstractPoint, ax::Integer) = _coordinate(pt, Int(ax))
+
+# TODO: we need to rationalize point operations with vectors
+Base.:(-)(p1::T, p2::T) where {T <: AbstractPoint} =
+    unionalltype(T)((components(p1) - components(p2))...)
```
simonbyrne, line 116: What does this return?

simonbyrne (approved):

src/Geometry/coordinates.jl
jakebolewski: This returns a point that promotes correctly. I'll see if this is still needed as I was trying to get the local geometry to work correctly with CoordinateTypes but there needs to be a more incremental approach as there are a lot of baked in assumptions currently.

Ideally this should return a vector type.

src/Geometry/coordinates.jl
jakebolewski: I was thinking coordinates is kind of like `Base.axes` and there are not singular versions for the later, I can change it though although the similarity of the names `comoponents` `component` might be a bit error prone.

jakebolewski:
bors try

src/Geometry/coordinates.jl
simonbyrne: A vector type would make sense in a box, but wouldn't e.g. on a sphere.

Where is addition/subtraction of coordinates used? Is it just for ranges?

jakebolewski:
I added a comment it is just for range support

PR #205, opened 2021-09-22
Embedded domain field into unstructured 2D mesh struct.

sriharshakandala:


jakebolewski:
@sriharshakandala I think this needs to be rebased after https://github.com/CliMA/ClimaCore.jl/pull/201 is merged.  Looking at it though I don't think it will be too bad, also #201 allows for the coordinate types defined in #81 to be applied for the sphere.

src/Meshes/Meshes.jl
```diff
@@ -184,7 +183,9 @@ https://p4est.github.io/papers/BursteddeWilcoxGhattas11.pdf
 # Fields
 $(DocStringExtensions.FIELDS)
 """
-struct Mesh2D{I, IA1D, IA2D, FT, FTA2D, SNT, NB} <: AbstractMesh{FT}
+struct Mesh2D{I, IA1D, IA2D, FT, FTA2D, SNT, NB, D} <: AbstractMesh{FT}
```
jakebolewski, line 186: usually the unique types are in the order of the fields to make it easy to reference

src/Domains/Domains.jl
```diff
@@ -125,14 +125,22 @@ end
 coordinate_type(::RectangleDomain{FT}) where {FT} =
     Geometry.Cartesian2DPoint{FT}
 
+abstract type AbstractSphere{FT} end
+struct EquiangularSphere{FT} <: AbstractSphere{FT} end
```
jakebolewski, line 129: isn't the sphere type more a property of the mesh than the domain?

jakebolewski: ex. an `EquiangularSphereMesh()` makes sense but a equiangular sphere domain does not really as the domain just defines the coordinates (LatLong) / coordinate extent (Radius).

sriharshakandala:
@jakebolewski : I rebased the code with master. I also added the bickley jet CG experiment with unstructured mesh to the buildkite so that changes relevant to unstructured mesh will also be carried over wherever newer PRs are added.

sriharshakandala:
bors try

jakebolewski:
Nice, seems to be a bit simpiler and is more in line with what we do for the vertical warping. LGTM

jakebolewski (approved):

sriharshakandala:
Bots r+

PR #211, opened 2021-09-24
rename quadrature rules, and flexible constructors

### LINKED ISSUE #163: Better naming for quadrature rules and user-defined polynomial degree
Can we consider renaming [`GLL`](https://github.com/CliMA/ClimaCore.jl/blob/cc105dbb7685aa57fe6fb6e6f6cf0fd62e26e6a0/src/Spaces/quadrature.jl#L45), `LGL`, and [`GL`](https://github.com/CliMA/ClimaCore.jl/blob/cc105dbb7685aa57fe6fb6e6f6cf0fd62e26e6a0/src/Spaces/quadrature.jl#L60), `LG`. These, or their long-but-abbreviated-forms,  Gauss-Lobatto and Gauss, respectively, (not sure why in their long forms the poor Legendre is dropped), are more widely used in the Finite Element communities. 

Also, can we consider allowing the user to specify not only the total number of quadrature points to use, but also the polynomial degree (usually denoted by  `p`)? I know the two are related, but it is often of interest to run `h`-refinement (ie, mesh refinement), `p`-refinement (ie, polynomial degree refinement), or a combination of the two, ie, `hp`-refinement studies, for convergence  or performance-benchmarking purposes. So it would be a quantity handy to have instead of having to derive it from the number of quadrature points (especially if one wants to run several experiments with different under/exact/over-integration strategies).

In this regard, one might want to specify a separate name for the number of finite element nodes (one more of the polynomial degree, perhaps `P = p+1`, `numP = p+1`, `N = p+1`, or `numN = p+1`). For the distinction between finite element nodes and quadrature (or integration) points, you can see this [reference](http://solidmechanics.org/Text/Chapter8_1/Chapter8_1.php) , Section 8.1.5. Thanks.

simonbyrne:
Fixes #163. Renames the quadarture rules (and aliases), and adds keyword constructors so that you can specify either number of points, or polynomial degree.

Would be good to get some input on the exact names of (a) the quadrature rules, and (b) the keyword arguments. @valeriabarra @sriharshakandala any thoughts?

### DIFF
--- a/src/Spaces/quadrature.jl
+++ b/src/Spaces/quadrature.jl
@@ -42,41 +42,135 @@ function quadrature_points end
 
 
 """
-    GLL{Nq}()
+    LegendreGaussLobatto
+    LGL
 
-Gauss-Legendre-Lobatto quadrature using `Nq` quadrature points.
+A Legendre-Gauss-Lobatto (LGL) quadrature rule.
+
+# Constructors
+
+    LegendreGaussLobatto(npoints=N)
+    LGL(npoints=N)
+
+Construct the quadrature rule with `N` quadrature points (`N ≥ 2`).
+
+    LegendreGaussLobatto(degree=p)
+    LGL(degree=p)
+
+Construct the quadrature rule with polynomial degree `p` (i.e. `p+1` quadrature
+points, with `p ≥ 1`).
+
+# Details
+
+A Legendre-Gauss-Lobatto rule has places quadrature points at the endpoints of
+the interval, and is able to exactly integrate polynomials of degree `2N-3`
+where `N` is the number of quadrature points.
+
+# External links
+
+- [Gauss-Lobatto rules on
+  Wikipedia](https://en.wikipedia.org/wiki/Gaussian_quadrature#Gauss%E2%80%93Lobatto_rules)
 """
-struct GLL{Nq} <: QuadratureStyle end
+struct LegendreGaussLobatto{Nq} <: QuadratureStyle end
+function LegendreGaussLobatto(;
+    npoints::Union{Int, Nothing} = nothing,
+    degree::Union{Int, Nothing} = nothing,
+)
+    if !isnothing(npoints)
+        LegendreGaussLobatto{npoints}()
+    elseif !isnothing(degree)
+        LegendreGaussLobatto{degree + 1}()
+    else
+        error("either npoints or degree option must be specified.")
+    end
+end
+const LGL = LegendreGaussLobatto
 
-polynomial_degree(::GLL{Nq}) where {Nq} = Nq - 1
-degrees_of_freedom(::GLL{Nq}) where {Nq} = Nq
+# TODO: deprecate
+const GLL = LegendreGaussLobatto
+polynomial_degree(::LegendreGaussLobatto{Nq}) where {Nq} = Nq - 1
+degrees_of_freedom(::LegendreGaussLobatto{Nq}) where {Nq} = Nq
 
-@generated function quadrature_points(::Type{FT}, ::GLL{Nq}) where {FT, Nq}
+@generated function quadrature_points(
+    ::Type{FT},
+    ::LegendreGaussLobatto{Nq},
+) where {FT, Nq}
     points, weights = GaussQuadrature.legendre(FT, Nq, GaussQuadrature.both)
     :($(SVector{Nq}(points)), $(SVector{Nq}(weights)))
 end
 
 """
-    GL{Nq}()
+    LegendreGauss
+    LG
 
-Gauss-Legendre quadrature using `Nq` quadrature points.
+A Legendre-Gauss (LG) quadrature rule.
+
+# Constructors
+
+    LegendreGauss(npoints=N)
+    LG(npoints=N)
+
+Construct the quadrature rule with `N` quadrature points
+
+    LegendreGauss(degree=p)
+    LG(degree=p)
+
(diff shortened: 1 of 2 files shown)


src/Spaces/quadrature.jl
```diff
@@ -42,41 +42,135 @@ function quadrature_points end
...
+    LegendreGaussLobatto(npoints=N)
+    LGL(npoints=N)
+
+Construct the quadrature rule with `N` quadrature points (`N ≥ 2`).
+
+    LegendreGaussLobatto(degree=p)
+    LGL(degree=p)
+
+Construct the quadrature rule with polynomial degree `p` (i.e. `p+1` quadrature
+points, with `p ≥ 1`).
```
valeriabarra, line 61: Thanks Simon. I think this docstring can lead to some confusion. If we are talking about constructing a _basis_ such as an H1 Lagrange basis, then I would agree with using a phrase  such as `Construct a finite element H1 Lagrange basis with polynomial degree `p` (i.e. `p+1` nodes, with `p ≥ 1`).` 
But a quadrature rule generally can have an arbitrary number of quadrature points, so I would use:
```suggestion
    LegendreGaussLobatto(qpoints=q)
    LGL(qpoints=q)
    
Construct the quadrature rule with quadrature points `q` (i.e. `q` quadrature
points, with `q` arbitrary).
```
`q` could be totally independent of `p`. It could be equal to `p`, it could be less than `p` for cases of underintegration and can be more than `p` for cases of overintegration. That's why is useful expressing these two concepts that are distinct with different variable names.

simonbyrne: I'm not sure I understand what you're proposing here: what's the difference between `npoints` and `qpoints`?

simonbyrne: > `q` could be totally independent of `p`. It could be equal to `p`, it could be less than `p` for cases of underintegration and can be more than `p` for cases of overintegration.

At the moment, that requires constructing two spaces with two different quadrature rules (one for the storage space, one for the integration space). 

The purpose of this PR to allow specifying either the number of points, or the degree of the polynomial that you can approximate with those points (which I thought was the point you were raising in #163).

sriharshakandala: > Fixes #163. Renames the quadarture rules (and aliases), and adds keyword constructors so that you can specify either number of points, or polynomial degree.
> 
> Would be good to get some input on the exact names of (a) the quadrature rules, and (b) the keyword arguments. @valeriabarra @sriharshakandala any thoughts?

The `degree`/'npoints' keyword argument looks good to me.

valeriabarra: They are conceptually different things and `qpoints` can be arbitrary, while the number of nodes, if you want to call them `npoints`, are not. `npoints = p+1` always, by definition. E.g., for a quadratic basis, you have `npoints = p + 1 \equiv 3` nodes. But you can decide to have lower, equal, or higher order integration (generally overintegration is preferred for obvious reasons, not underintegration) so that at the end when you do integration across all elements, you can have a desired order for the integration of non-linear terms (e.g., `u^2` terms) that does not necessarily coincide with the order chosen to represent the state variable `u`

sriharshakandala: @simonbyrne : One option we can think of is to evaluate the degree `d` polynomials on an user specified grid!

src/Spaces/quadrature.jl
```diff
@@ -42,41 +42,135 @@ function quadrature_points end
...
+
+# Details
+
+A Legendre-Gauss-Lobatto rule has places quadrature points at the endpoints of
+the interval, and is able to exactly integrate polynomials of degree `2N-3`
+where `N` is the number of quadrature points.
```
valeriabarra, line 67: I think below the parameter is named `Nq`, so I would be consistent here. 
```suggestion
A Legendre-Gauss-Lobatto rule has places quadrature points at the endpoints of
the interval, and is able to exactly integrate polynomials of degree `2Nq-3`
where `Nq` is the number of quadrature points.
```

src/Spaces/quadrature.jl
```diff
@@ -42,41 +42,135 @@ function quadrature_points end
...
+const GLL = LegendreGaussLobatto
+polynomial_degree(::LegendreGaussLobatto{Nq}) where {Nq} = Nq - 1
+degrees_of_freedom(::LegendreGaussLobatto{Nq}) where {Nq} = Nq
 
-@generated function quadrature_points(::Type{FT}, ::GLL{Nq}) where {FT, Nq}
+@generated function quadrature_points(
+    ::Type{FT},
+    ::LegendreGaussLobatto{Nq},
+) where {FT, Nq}
     points, weights = GaussQuadrature.legendre(FT, Nq, GaussQuadrature.both)
```
valeriabarra, line 98: Since effectively this function only creates a tabulation of points and weights (as a quadrature rule should do), I stand by comment above, saying that defining this function with the `LegendreGaussLobatto(degree=p)` degree argument instead of the `qpoints` or `Nq`, however you want to call them, would be wrong in my opinion. A quadrature rule it's only an integration rule and does not need to know the polynomial order of the finite element basis

jakebolewski:
Paul also calls it GLL, ex: https://github.com/ClimateGlobalChange/tempestremap/blob/31e51ff650419c6345614e3283f43bb7d5094acf/src/FiniteElementTools.cpp#L30

I like the idea of keywords with no default, which should disambiguate Uniform vs. GLL / LGL quadratures.   I think it's useful to keep in mind that defacto we only will use one.

PR #213, opened 2021-09-27
update JET

aviatesk:
From Julia v1.7 and higher, JETTest's functionalities are all moved to
JET.jl (and JETTest is just deprecated).
Now JET@0.5.0 is released, and let's just focus on running `JET.@test_opt`
on Julia v1.7 and higher.


vchuravy:
Thanks!

PR #214, opened 2021-09-27
Add tutorial notebook

simonbyrne:
We may want to move this to some sort of literate notebook, but for now it would be good to keep as part of the repo.


valeriabarra (approved):

jakebolewski:
I think checking in the notebook iteself might be a bit difficult to support, probably better to create a separate repo with the commit of ClimaCore used in the manifest so it's reproducible?  otherwise I suspect this will go stale relatively quickly.

charleskawczynski:
I agree with @jakebolewski on this. At the very least, couldn’t we go the Literate.jl route and just check in the Julia script?

jakebolewski:
There is this to try? https://github.com/oxinabox/ipynb2LiterateJulia

jakebolewski:
Actually running it

valeriabarra:
> Actually running it

Do you mean adding the machinery so that we can run this when building the docs?

jakebolewski:
yes or just run it at all as a sanity check

valeriabarra:
I just tried i locally. I reduced the time of the `ODEProblem`s so that it doesn't take too long (but it seems to still take very long to be executed -- I personally interrupted it at some point). It seemed that it was running fine and produced outputs (gifs). 
If anyone wants to double check that it runs  for them and does not actually stall, that'd be great. Thanks

simonbyrne:
I would like to hook this up to the docs, but that will slow down the build process. Any thoughts on the tradeoff?

PR #231, opened 2021-10-05
Scalar Poisson problem unit test in 2D to test horizontal diffusion operator

valeriabarra:
This PR modifies the current `test/diffusion2d.jl` test that accounted for time-dependence to be steady-state so that we test only the horizontal spatial 2D Laplacian. 

It also adds the test to `runtests.jl`, as it wasn't included and was outdated.


valeriabarra:
bors try

jiahe23 (approved):

PR #232, opened 2021-10-05
change vertex element ordering to counter-clockwise

### LINKED ISSUE #219: Use counter-clockwise ordering of vertices
Needs to update `vertex_coordinates` and bilinear interpolation to be consistent across all mesh types.

simonbyrne:
Fixes #219.

### DIFF
--- a/src/Geometry/coordinates.jl
+++ b/src/Geometry/coordinates.jl
@@ -156,43 +156,38 @@ end
 
 
 """
-    interpolate(coords::NTuple{4}, ξ1, ξ2)
+    bilinear_interpolate(coords::NTuple{4}, ξ1, ξ2)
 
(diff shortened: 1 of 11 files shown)


simonbyrne:
bors try

simonbyrne:
indeed. it does look like something is going wrong...

sriharshakandala:
We need to update the documentation for the Mesh2D struct in Meshes.jl.
Screen Shot 2021-10-05 at 1 59 38 PM

is typical, but if we choose to use a different, we still need to update the documentation.

simonbyrne:
bors try

simonbyrne:
bors try-

simonbyrne:
bors try

PR #238, opened 2021-10-06
Sphere diffusion unit test

valeriabarra:
Add a Poisson's problem on the sphere unit test, to test the spatial Laplacian only.


test/runtests.jl
```diff
@@ -1,26 +1,28 @@
 using Base: operator_associativity
 
-include("recursive.jl")
```
jakebolewski, line 3: I would revert this change as the original ordering  in runtests is useful for debugging because it better captures the rough dependency order of functionality for the tests

jakebolewski (changes requested):

test/runtests.jl
valeriabarra: Oh, I thought the alphabetical order was just better organized for people to look for existing tests and insert new ones. Also, the other day I found a test that was included twice (`diffusion2d.jl`) and I thought that if they were alphabetized it would be clearer to avoid double entries.

jakebolewski: it might be good to add a comment about why the ordering is the way it is for the future

test/sphere_diffusion.jl
```diff
@@ -0,0 +1,58 @@
...
+using StaticArrays, IntervalSets, LinearAlgebra
+
+@testset "Scalar Poisson problem - ∇⋅∇ = f on the cubed-sphere" begin
+    # Poisson equation on a sphere
+    # - ∇⋅(∇ u(ϕ, λ)) = f(ϕ, λ)
+
+    # With:
+    # ϕ = asin(z / R) # latitude (R radius of sphere)
+    # λ = atan2(y, x) # longitude ∈ [-π, π]
+    # True solution (eigenfunction): u(ϕ, λ) = sin(λ) * cos(ϕ)
```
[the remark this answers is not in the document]
valeriabarra: Yes, I know spherical harmonics are, but this is a test case that it's used (see https://www2.cisl.ucar.edu/sites/default/files/levy_poisson_0.pdf and other Finite Element packages that include this as a test)

test/runtests.jl
valeriabarra: Done.

jiahe23:
> > > @jiahe23 do you think this looks good? If yes, can approve it?
> > 
> > 
> > I think for sphere tests, we should use spherical harmonics as eigenfunctions as in [here](https://github.com/CliMA/ClimaCore.jl/blob/d6e480dbce58ff420820de4e511619afdf7ed9bb/test/hyperdiffusion_sphere.jl#L15)
> 
> I don't think it's really necessary in this case. I left a reference where this test was formalized and it's also used in other Finite Element packages. It works and I also checked the spectral superconvergence properties.

Sounds good!

jiahe23 (approved):

jakebolewski:
it would be good to squash commits going forward for PR's like this before merging

PR #249, opened 2021-10-12
DSS for AxisVector

### LINKED ISSUE #160: Make DSS tensor-aware
To support non-uniform coordinate systems we will need DSS to handle the appropriate vector basis transformations at the element boundaries

simonbyrne:
Transforms the AxisVector to the local orthogonal coordinates before doing DSS.

Fixes #160.

### DIFF
--- a/src/Spaces/dss.jl
+++ b/src/Spaces/dss.jl
@@ -1,9 +1,69 @@
 import ..Topologies
 using ..RecursiveApply
 
+
+function dss_transform(arg, local_geometry)
+    RecursiveApply.rmap(arg) do x
+        dss_transform(x, local_geometry)
+    end
+end
+dss_transform(arg::Number, local_geometry) = arg
+dss_transform(
+    arg::Geometry.AxisTensor{T, N, <:Tuple{Vararg{Geometry.CartesianAxis}}},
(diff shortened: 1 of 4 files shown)


src/Spaces/dss.jl
```diff
@@ -1,9 +1,59 @@
...
+dss_transform(
+    arg::Geometry.AxisTensor{T, N, <:Tuple{Vararg{Geometry.LocalAxis}}},
+    local_geometry::Geometry.LocalGeometry,
+) where {T, N} = arg
+
+function dss_transform(
+    arg::Geometry.AxisVector,
+    local_geometry::Geometry.LocalGeometry,
+)
+    ax = axes(local_geometry.∂x∂ξ, 1)
```
jakebolewski, line 24: Should this be?

```suggestion
    ax = axes(arg, 1)
```

src/Spaces/dss.jl
```diff
@@ -1,9 +1,69 @@
...
+    axfrom = axes(arg, 1)
+    if ax isa Geometry.Cartesian12Axis && (
```
jakebolewski, line 34: this seems fine as a workaround

PR #251, opened 2021-10-13
FV->FV Cartesian Remapping

jb-mackay:
Adds remapping operator construction for Finite Volume -> Finite Volume cartesian grids (2D).


src/Meshes/overlap_mesh.jl
```diff
@@ -0,0 +1,13 @@
...
+    x2c = y_coords(mesh1) ∪ y_coords(mesh2)
+    sort!(x1c)
+    sort!(x2c)
+    return Mesh2D(mesh1.domain, x1c, x2c)
+end
+
+x_coords(mesh::EquispacedRectangleMesh) = mesh.range1
+y_coords(mesh::EquispacedRectangleMesh) = mesh.range2
+x_coords(mesh::Mesh2D) = mesh.coordinates[:, 1]
+y_coords(mesh::Mesh2D) = mesh.coordinates[:, 2]
```
valeriabarra, line 13: I'm wondering if it would be better to define the `overlap_mesh` as its own type (i.e., `struct`). That way, you wouldn't have to redefine the `x_coords` and `y_coords`, but you could extract those by dispatching on the underlying mesh?

valeriabarra: It's not that important, I guess. Your way works as well. I was just commenting on the design that would be a bit different this way, compared to the existing API design we had so far for meshes and topologies. So far, Mesh objects have been just a data structure and we have left the functions to access those data (such as the coordinates field) in Topologies and they were dispatched according to the underlying mesh.

test/runtests.jl
```diff
@@ -20,6 +20,8 @@ include("fdspaces.jl")
 include("fielddiffeq.jl")
 include("hybrid2d.jl")
 include("hybrid3d.jl")
+#include("diffusion2d.jl")
```
valeriabarra, line 23: Not related to this PR, but I think the `diffusion2d.jl` test can be added to the suite to run

Project.toml
```diff
@@ -15,6 +15,7 @@ IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
valeriabarra, line 18: This needs to be added to the `test/Project.toml`. That's the environment that is used for the tests

valeriabarra: I think that's why CodeCov is complaining

valeriabarra:
Thanks Ben for your work! Looks good to me. 
Let me know if you need help with rebasing/squashing commits!

test/runtests.jl
valeriabarra: Oh, I see. This was the second, redundant include of that test. It's already been fixed on `main`, so if you rebase onto main, this will be fixed

valeriabarra (approved):

test/runtests.jl
valeriabarra: @jb-mackay this comment is still here, and it would introduce it back to `main`, which we don't want. Probably something missed by the rebase

jb-mackay: Good eye, will remove

PR #255, opened 2021-10-15
Solid body rotation on a sphere: Gaussian blob 

valeriabarra:
This PR adds another IC to the solid body example on the sphere. It also adds the possibility to specify the IC (or test case to run) via CL. The default one is still the cosine bell, i.e., the only existing case before. 

Since hyperdiffusion was not used in the solid body test case (advection equation only), it was decided that we remove the hyperdiffusion part in the operator. 

I also added a convergence study under mesh refinement.

### DIFF
--- a/examples/sphere/solidbody.jl
+++ b/examples/sphere/solidbody.jl
@@ -34,17 +34,26 @@ convergence_rate(err, Δh) =
 const R = 6.37122e6
 const h0 = 1000.0
 const r0 = R / 3
-const α0 = 0.0
 const u0 = 2 * pi * R / (86400 * 12)
 const center = Geometry.LatLongPoint(0.0, 270.0)
 const test_name = get(ARGS, 1, "cosine_bell") # default test case to run
+const test_angle_name = get(ARGS, 2, "alpha0") # default test case to run
 const cosine_test_name = "cosine_bell"
 const gaussian_test_name = "gaussian_bell"
+const alpha0_test_name = "alpha0"
+const alpha45_test_name = "alpha45"
+
+if test_angle_name == alpha45_test_name
+    const α0 = 45.0
+else # default test case, α0 = 0.0
+    const α0 = 0.0
+end
 
 # Plot variables and auxiliary function
 ENV["GKSwstype"] = "nul"
 Plots.GRBackend()
-dirname = "cg_sphere_solidbody"
+dirname = "cg_sphere_solidbody_$(test_name)"
+dirname = "$(dirname)_$(test_angle_name)"
 path = joinpath(@__DIR__, "output", dirname)
 mkpath(path)
 
@@ -59,7 +68,8 @@ end
 
 FT = Float64
 ne_seq = 2 .^ (2, 3, 4, 5)
-err, Δh = zeros(FT, length(ne_seq)), zeros(FT, length(ne_seq))
+Δh = zeros(FT, length(ne_seq))
+L1err, L2err, Linferr = zeros(FT, length(ne_seq)), zeros(FT, length(ne_seq)), zeros(FT, length(ne_seq))
 Nq = 4
 
 # h-refinement study
@@ -112,7 +122,7 @@ for (k, ne) in enumerate(ne_seq)
 
     # Solve the ODE
     T = 86400 * 12
-    dt = 30 * 60
+    dt = 20 * 60
     prob = ODEProblem(rhs!, h_init, (0.0, T), u)
     sol = solve(
         prob,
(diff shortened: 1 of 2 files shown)


valeriabarra:
Something seems off when looking at convergence rates. Might be the way I computed them (?)

```
$ julia --project=test examples/sphere/solidbody.jl gaussian_bell
```
Gives:
```
[ Info: ("Converge rates for this test case are: ", [5.451778823836831, 4.46496874108546, 1.2369369329411826])
```
Whereas, the cosine bell test case (the default one) 
```
$ julia --project=test examples/sphere/solidbody.jl
```
Gives:
```
[ Info: ("Converge rates for this test case are: ", [1.3041703846775095, 3.064524477040917, 2.4421918792565975])
```

For the Gaussian bell, I had reproduced all runs to compare with the ones shared during the dycore meeting. And it seems like we are diverging :/ :-1: 
[CliMA_results.pdf](https://github.com/CliMA/ClimaCore.jl/files/7350200/CliMA_results.pdf)

examples/sphere/solidbody.jl
```diff
@@ -5,112 +5,144 @@ import ClimaCore: slab, Fields, Domains, Topologies, Meshes, Spaces
...
+    sol = solve(
+        prob,
+        SSPRK33(),
+        dt = dt,
+        saveat = dt,
+        progress = true,
+        adaptive = false,
+        progress_message = (dt, u, p, t) -> t,
+    )
+    err[k] = norm(sol.u[end] .- h_init) / length(h_init)
```
jakebolewski, line 111: `h_init` will have unit length `1`, is this what you want to compute?

valeriabarra: You're right. I was looking at the same helper function in `test/fdspaces.jl` . I would think the relative error would just be `(exact - computed)/exact`, but in that case `length(parent(exact))` was used, so I got confused

valeriabarra: I'm going to leave it as it was, except for the `length(h_init)` bit (which wasn't going to have unit length, as we said above though. Now that I looked at it more closely)

examples/sphere/solidbody.jl
```diff
@@ -107,10 +57,91 @@ function linkfig(figpath, alt = "")
...
+    @info "Number of elements per cube panel: $(ne) x $(ne)"
+    @info "Solution norm at t = 0: ", norm(h_init)
+    @info "Solution norm at t = $(T): ", norm(sol.u[end])
+    @info "L2 error at t = $(T): ", err[k]
+
+    #----------------lat, long plots
+    Plots.png(Plots.plot(coords.lat), joinpath(path, "latitude.png"))
+    linkfig("output/$(dirname)/latitude.png", "latitude")
+    Plots.png(Plots.plot(coords.long), joinpath(path, "longitude.png"))
+    linkfig("output/$(dirname)/longitude.png", "longitude")
```
simonbyrne, line 138: I don't think we need a plot of the lat & long: it would probably make more sense to plot the error.
```suggestion
```

examples/sphere/solidbody.jl
```diff
@@ -107,10 +57,91 @@ function linkfig(figpath, alt = "")
...
+end
+
+# Plot the error
+Plots.png(Plots.plot(1:length(ne_seq), err), joinpath(path, "error.png"))
+
+linkfig("output/$(dirname)/error.png", "L2 error Vs ne")
```
simonbyrne, line 144: This needs to be the full path to display in the log.

valeriabarra: Ok, not a problem. But I'm afraid I cannot look at it before Tuesday. If you don't mind changing it, please go ahead

simonbyrne: No rush, it can wait. Have a good weekend.

examples/sphere/solidbody.jl
```diff
@@ -107,10 +57,91 @@ function linkfig(figpath, alt = "")
...
+
+    #----------------lat, long plots
+    Plots.png(Plots.plot(coords.lat), joinpath(path, "latitude.png"))
+    linkfig("output/$(dirname)/latitude.png", "latitude")
+    Plots.png(Plots.plot(coords.long), joinpath(path, "longitude.png"))
+    linkfig("output/$(dirname)/longitude.png", "longitude")
+end
+
+# Plot the error
+Plots.png(Plots.plot(1:length(ne_seq), err), joinpath(path, "error.png"))
```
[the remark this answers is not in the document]
valeriabarra: Regarding the log scales, if I use `xscale = :log` or `yscale = :log` then I see the following warning:
```
 Warning: scale log is unsupported with Plots.GRBackend(). Choose from: [:identity, :log10]
```
Is there a reason why we use this particular backend in these examples?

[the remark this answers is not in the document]
simonbyrne: ah, you probably want `log10`

examples/sphere/solidbody.jl
```diff
@@ -107,10 +66,125 @@ function linkfig(figpath, alt = "")
...
+        1:length(ne_seq),
+        L1err,
+        xscale = :log,
+        yscale = :log,
+        xlabel = "Nₑ",
+        ylabel = "L₁ err",
+    ),
+    joinpath(path, "L1error.png"),
+)
+linkfig("output/$(dirname)/L1error.png", "L₁ error Vs Nₑ")
```
simonbyrne, line 163:
```suggestion
linkfig(relpath(joinpath(path,"L1error.png"), "../.."), "L₁ error Vs Nₑ")
```
The DG one doesn't work either. The path should be relative to the top-level (ClimaCore.jl) directory.

examples/sphere/solidbody.jl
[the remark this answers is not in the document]
valeriabarra: > ah, you probably want `log10`

What if someone wants to plot with a different scale that is not log10? Do we have to use this backend necessarily? I don't know much about Plots backends.

examples/sphere/solidbody.jl
```diff
@@ -107,10 +66,131 @@ function linkfig(figpath, alt = "")
...
+        λ = coord.long
 
-#----------------lat, long plots
-Plots.png(Plots.plot(coords.lat), joinpath(path, "latitude.png"))
-linkfig("output/$(dirname)/latitude.png", "latitude")
-Plots.png(Plots.plot(coords.long), joinpath(path, "longitude.png"))
-linkfig("output/$(dirname)/longitude.png", "longitude")
+        uu = u0 * (cosd(α0) * cosd(ϕ) + sind(α0) * cosd(λ) * sind(ϕ))
+        uv = -u0 * sind(α0) * sind(λ)
+        Geometry.UVVector(uu, uv)
```
simonbyrne, line 112: So I think the issue we're seeing here is that this doesn't end up defining the vector correctly at the pole if `α0` is not zero.

As Paul suggested, we should probably just define it in Cartesian coordinates (we will have to add some functionality for that)

simonbyrne: From Paul:
> The solid body velocity field in 3D Cartesian is \vec{u} = - (\vec{Omega} \cdot \vec{x}) (\vec{Omega} x \vec{x}), where \Omega = (cos(alpha), 0, sin(alpha))

valeriabarra:
Much better convergence rates after rebasing this onto `main` with the #262 fix in. 
```
$ julia --project=test examples/sphere/solidbody.jl gaussian_bell alpha45
```
```
[ Info: ("Converge rates for this test case are: ", [5.108328182717773, 4.053364684528974, 2.896215173984146])
```
image

valeriabarra:
bors try

examples/sphere/solidbody.jl
```diff
@@ -5,99 +5,58 @@ import ClimaCore: slab, Fields, Domains, Topologies, Meshes, Spaces
...
-@info "L2 error at T = $(T): ", norm(sol.u[end] .- h_init)
 
+# Plot variables and auxiliary function
 ENV["GKSwstype"] = "nul"
-import Plots
 Plots.GRBackend()
-
-dirname = "cg_sphere_solidbody"
+dirname = "cg_sphere_solidbody_$(test_name)"
+dirname = "$(dirname)_$(test_angle_name)"
```
simonbyrne, line 56: need to update the path in the `.buildkite/pipeline.yml`

examples/sphere/solidbody.jl
```diff
@@ -107,10 +66,131 @@ function linkfig(figpath, alt = "")
...
+        collect(ne_seq),
+        Linferr,
+        yscale = :log10,
+        xlabel = "Nₑ",
+        ylabel = "log₁₀(L∞ err)",
+        label = "",
+    ),
+    joinpath(path, "Linferror.png"),
+)
+linkfig(relpath(joinpath(path, "Linferror.png"), "../.."), "L∞ error Vs Nₑ")
```
simonbyrne, line 196: I had this wrong: it should be
```suggestion
linkfig(relpath(joinpath(path, "Linferror.png"), joinpath(@__DIR__, "../..")), "L∞ error Vs Nₑ")
```

simonbyrne: and above as well

valeriabarra:
bors try

valeriabarra:
> @simonbyrne would it be better to squash it first?

Yes, I meant to do it after approval. I didn't know Simon was going to r+ right away. I'll do this in a bit.

simonbyrne:
> bots r-

'twas a typo.

No big deal, I'll be less trigger happy with my r+ in future.

PR #257, opened 2021-10-15
Document spectral element operators

simonbyrne:
This was long overdue...

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -415,16 +415,19 @@ end
     wdiv = WeakDivergence()
     wdiv.(u)
 
-Computes the "weak divergence" of a vector field `u`. This is defined as the
-scalar field ``\\theta`` such that for all ``\\phi``
+Computes the "weak divergence" of a vector field `u`.
+
+This is defined as the scalar field ``\\theta \\in \\mathcal{V}_0`` such that
+for all ``\\phi\\in \\mathcal{V}_0``
 ```math
 \\int_\\Omega \\phi \\theta \\, d \\Omega
 =
 - \\int_\\Omega (\\nabla \\phi) \\cdot u \\,d \\Omega
 ```
+where ``\\mathcal{V}_0`` is the space of ``u``.
 
-This is named as it arises as the contribution of the volume integral after by
-applying integration by parts to the weak form expression of the divergence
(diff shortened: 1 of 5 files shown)


src/Operators/spectralelement.jl
```diff
@@ -404,25 +412,38 @@ end
...
-
-    - J \\ (D₁' * W * J * v¹ + D₂' * W * J * v²)
-
+    wdiv = WeakDivergence()
+    wdiv.(u)
+
+Computes the "weak divergence" of a vector field `u`. This is defined as the
+scalar field ``\\theta`` such that for all ``\\phi``
+```math
+\\int_\\Omega \\phi \\theta \\, d \\Omega
```
valeriabarra, line 421: I would be careful with defining `\theta` here. Otherwise, don't just use another letter, since you already have `u`. I would just use a sample equation, say, a wave equation or a heat equation, where you have du/dt on the left hand side

valeriabarra: Then, if it's just to document the weak spatial operator, why do we need a LHS at all?

simonbyrne: > Since you define the weak derivative of u, I'd just use u and not introduce another letter

I was actually thinking of using theta for the weak derivative as well, since then it would be consistent

valeriabarra: Ok, sounds good. Yes, I see it now and it's better to be consistent

valeriabarra: Yes, that'd be good. Usually, in Finite Elements library that I've seen (e.g., libCEED and MFEM), it is customary to show at least one example that leads to one type of operator (i.e., just a L2 projection problem that leads to the Mass matrix; just an advection problem; and just a Laplace (Poisson) problem --- even just steady state. No need for the time derivative and linear solve, really). 
These 3 scenarios respectively lead to having the 3 different possible formulations: `interp-interp` (meaning, for an L2 projection, you end up with interpolation of your trial function `u` and test function `phi`); for a divergence only operator, `interp-grad` (meaning, you end up with interpolation of your trial function `u` and gradient of your test function `phi`); and the Laplacian leads to `grad-grad` (meaning, you end up with gradient of your trial function `u` and gradient of your test function `phi`)

PR #262, opened 2021-10-20
Define UV at the pole to be consistent with long=0

simonbyrne:
This should fix the even-numbered element issue from #255.

### DIFF
--- a/src/Spaces/spectralelement.jl
+++ b/src/Spaces/spectralelement.jl
@@ -334,6 +334,9 @@ function SpectralElementSpace2D(
                 ξ[2],
             )
             xl = Geometry.LatLongPoint(x)
+            # [∂x1/∂ξ¹ ∂x1/∂ξ²
+            #  ∂x2/∂ξ¹ ∂x2/∂ξ²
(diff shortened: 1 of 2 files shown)


src/Spaces/spectralelement.jl
```diff
@@ -346,8 +346,15 @@ function SpectralElementSpace2D(
                 )
             end
             if abs(xl.lat) ≈ oftype(xl.lat, 90.0)
-                # at the pole: choose u to line with x1, v to line with x2
-                ∂u∂ξ = ∂x∂ξ[SOneTo(2), :]
+                # at the pole we orient u and v by taking the limit approaching
+                # from the line long == 0 (i.e. along the line x2 = 0, x1 > 0)
+                # => u axis is aligned with x2, v is aligned with -x1
+                @assert xl.long ≈ 0
+                F = @SMatrix [
```
valeriabarra, line 353: I think while you were working you had a comment here, like 
```
# [du/dx1 du/dx2 du/dx3 
   dv/dx1 dv/dx2 dv/dx3]                                         
```
?
 Probably good to keep it?

PR #263, opened 2021-10-21
Warped mesh

### LINKED ISSUE #284: Topography
After some discussion with @Zhengyu-Huang and Paul, my general thought is to handle this as follows

1. Construct the horizontal space (i.e. a sphere or rectangle)
2. Construct the hybrid product space with a pseudo-vertical coordinate: this can either be η in (0,1), or it could be a reference z* in (0, `z_top`).
3. Construct a field `z_surface` on the horizontal space.
4. Construct a field `z_levels` on the product face space (this would be as a function of 3 and the pseudo-verical coordinate in 2.)
5. Take the horizontal covariant gradient of `z_levels` to get the metric terms.
6. Update the `LocalGeometry` terms to get the new space

At a user level, initially I think the process would be
1. Construct the horizontal space
2. Construct `z_surface` (either as a function, or loading from a data file: figuring out how to do this will be a separate issue)
3. Some sort of function that takes (`z_surface`, the vertical stretching, and the terrain adaptation), and returns the hybrid space

Zhengyu-Huang:
Working with @akshaysridhar @sriharshakandala 

The goal of the PR is to build 2D warped Cartesian mesh and 3D warped spherical mesh. 

The warped/3D-spherical mesh design plan:
1. For each single column, there is a reference element:  ξ1 in [-1, 1],  ξ2 in [-1, 1],  ξ3 in [0, nz], here nz is the number of vertical elements, and δξ3 = 1 corresponding to element vertical grid-spacing.
2. In the vertical direction, stretching + warping function will map ξ3 = 0, 1, 2, …., nz to different heights, 0 to the terrain, and  nz to the top, the function is r(ξ1,ξ2,ξ3),  r(ξ1,ξ2,0) or  r(ξ1,ξ2,0) - R is read from topography file. We can start with  for example: r(ξ1,ξ2,ξ3) = r(ξ1,ξ2,0) + (H - r(ξ1,ξ2,0) ) * sin(πξ3 / 2nz)   
3. Compute metric terms from the map (ξ1,ξ2,ξ3)-> (x,y,z) in each single column by auto diff,  at cell faces (ξ1, ξ2, i) and cell centers  (ξ1, ξ2, i+1/2)

For the implementation:
```
#Step1 create geometry information
domain = DomainFunc(Lx,Ly,Lz) or DomainFunc(R, H)
#Step2 create mesh information
mesh = MeshFunc(domain, nx=ny, nz)

#Step3 create bottom surface mesh without topography, get all quadrature points and build topography file.


#Step4 implement mesh, create mesh vertices (no quadrature points) based on the [MAP] 
grid_topology = Topology_Func(mesh, topography_file, stretching_function)

#Step5 compute quadrature points, metric terms based on the [MAP] 
quad = Spaces.Quadratures.GLL{Nq}()
space = Spaces.SpectralElementSpace2D(mesh, topography_file, stretching_function)
```


examples/hybrid/bubble-mountain.jl
```diff
@@ -0,0 +1,162 @@
...
+
+Z = map(Fields.coordinate_field(space)) do coord
+    z_top = 1000.0
+    a = 1/2
+    z_ref = coord.z
+    z_s = 8 * a^3 / (coord.x^2 + 4 * a^2)
+    return z_ref + (1-z_ref/z_top) * z_s
+end
+
+function vertical_warp!(space::Spaces.FaceExtrudedFiniteDifferenceSpace, Z::Fields.FaceExtrudedFiniteDifferenceField)
```
simonbyrne, line 85: @Zhengyu-Huang What do you think about this: you provide a new `Field` object `Z` with the object heights at the faces?

Zhengyu-Huang: Yes, I think that is a great idea

examples/hybrid/bubble-mountain.jl
```diff
@@ -0,0 +1,170 @@
...
+    # todo do we seperate hv_center_space & hv_face_space
+    # construct hv center/face spaces, recompute metric terms
+
+    hv_face_space =
+        Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_face_space)
+end
+
+# set up rhs!
+space = hvspace_2D((-500, 500), (0, 1000))
+
```
Zhengyu-Huang, line 75: @simonbyrne  Should we also add stretching here? we have them in interval.jl : for example NoStretching, ExponentialStretching,  GeneralizedExponentialStretching,

src/Meshes/interval.jl
```diff
@@ -87,6 +87,70 @@ function IntervalMesh(
...
+
+    # nondimensional vertical coordinate
+    ζ_n = LinRange(1, nelems, nelems) / nelems
+
+
+    f_s(h) = dz_s - cmax * exp_stretch(ζ_n[1], h)
+    h_s = find_zero(f_s, -dz_s / cmax / log(1 - 1 / nelems))
+    f_t(h) = dz_t - cmax * (1 - exp_stretch(ζ_n[end - 1], h))
+    h_t = find_zero(f_t, ((cmax - cmin) - dz_t) / cmax / log(nelems))
+
```
Zhengyu-Huang, line 143: This stretching requires Roots.jl , should I using Roots here ?

src/Meshes/interval.jl
```diff
@@ -87,6 +87,70 @@ function IntervalMesh(
...
+    stretch::NoStretching;
+    nelems,
+) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
+    cmin = Geometry.component(domain.coord_min, 1)
+    cmax = Geometry.component(domain.coord_max, 1)
+    Δc = (cmax - cmin) / (nelems)
+
+    faces = [CT(cmin + i * Δc) for i in 0:nelems]
+    IntervalMesh(domain, faces)
+end
```
simonbyrne, line 109: Isn't this just `Uniform` above?

src/Meshes/interval.jl
simonbyrne: If you need it, then you should add it to the project, and then `using Roots`

src/Meshes/interval.jl
Zhengyu-Huang: Yes, this is a uniform grid

examples/hybrid/bubble-mountain.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    z_surface = warp_fn.(Fields.coordinate_field(horzspace))
+
+    hv_face_space = Spaces.ExtrudedFiniteDifferenceSpace(
+        horzspace,
+        vert_face_space,
+        z_surface,
+        Topographies.LinearAdaption(),
+    )
```
simonbyrne, line 71: This is the rough high-level interface I was thinking of: basically add an extra argument for the surface field, and some mechanism for determining the higher faces.

Thoughts @Zhengyu-Huang @jakebolewski ?

Zhengyu-Huang: That's fantastic, I agree we need both the stretching function and the adaptation function! Thank you for the great work!

jakebolewski: I think it makes sense to include the face adaption bit in the extruded space definition, but IMO the z_surface field is really a property / metadata of the horizontal/surface space at each coordinate point.  I feel this will come up in other situations (ex. grid weights for coupling).  I don't know what the thinking was in wrapping this information in a modified topology object for only the product space.

simonbyrne:
TODO:
- [ ] Plotting / VTK
- [ ] Agnesi test case

src/Meshes/interval.jl
```diff
@@ -1,3 +1,5 @@
+using Roots
```
charleskawczynski, line 1: Not a big issue, but should we use `RootSolvers` instead?

Zhengyu-Huang: Do you recommend RootSolvers?

szy21:
Does #335 supersede this?

akshaysridhar:
> Does #335 supersede this?

Yes

PR #267, opened 2021-10-26
Add unit tests for operators on the sphere

jiahe23:
This PR adds unit tests and convergence tests for the strong and weak forms of gradient and divergence operators, and the strong form curl operator on the sphere. 

(@simonbyrne Weak form curl operator is commented out in the test file because dss fails to perform on the Convariant3Vector in [this line](https://github.com/CliMA/ClimaCore.jl/blob/jh/sph-operator-tests/test/sphere_curl.jl#L36). We could either wait till the dss is fixed or merge this first and add the weak curl test once the dss is ready.)


test/sphere_divergence.jl
```diff
@@ -0,0 +1,93 @@
+using Test
+using StaticArrays, IntervalSets
+import ClimaCore.DataLayouts: IJFH
+import ClimaCore:
+    Fields, Domains, Meshes, Topologies, Spaces, Operators, Geometry
+using StaticArrays, IntervalSets, LinearAlgebra
+
+FT = Float64
+
+radius = FT(1) #FT(6.37122e6)
```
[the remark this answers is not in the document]
jiahe23: ok, I made the changes.

jiahe23:
It looks like the gradient operator converges at `Nq-1` but the divergence and curl operators both have a convergence rate of `2`.  @simonbyrne @valeriabarra  Any idea why that is the case?

Rows from top to bottom are `Nq` from 3 to 6; First column is the rate between `Ne=4` and `Ne=8`, second column is between `8` and `16`:
```
julia> conv_rate
4×2 Matrix{Float64}:
 1.99051  1.99517
 3.00358  3.00115
 3.99784  3.9978
 5.00629  5.00143

julia> convergence_rate_sdiv
4×2 Matrix{Float64}:
 1.10332   1.04456
 1.00829   1.00013
 0.983406  0.995962
 0.978339  0.994748

julia> convergence_rate_scurl
4×2 Matrix{Float64}:
 1.11231   1.04111
 1.00909   1.00036
 0.983198  0.996038
 0.978649  0.994869
```

valeriabarra:
> 
> julia> convergence_rate_sdiv
> 4×2 Matrix{Float64}:
>  1.10332   1.04456
>  1.00829   1.00013
>  0.983406  0.995962
>  0.978339  0.994748
> 
> julia> convergence_rate_scurl
> 4×2 Matrix{Float64}:
>  1.11231   1.04111
>  1.00909   1.00036
>  0.983198  0.996038
>  0.978649  0.994869
> ```

From these two tables it looks more around 1 than 2, no?

jiahe23:
> > julia> convergence_rate_sdiv
> > 4×2 Matrix{Float64}:
> > 1.10332   1.04456
> > 1.00829   1.00013
> > 0.983406  0.995962
> > 0.978339  0.994748
> > julia> convergence_rate_scurl
> > 4×2 Matrix{Float64}:
> > 1.11231   1.04111
> > 1.00909   1.00036
> > 0.983198  0.996038
> > 0.978649  0.994869
> 
> From these two tables it looks more around 1 than 2, no?

Ah yes, you are correct! So the divergence and curl operators converge at the rate of `1` not `~~2~~`...

jiahe23:
The above mentioned issues in the convergence rates for the div and curl operators are due to false set-ups for the vector fields. This problem has been fixed and they all converge at the expected rates.

test/sphere_curl.jl
```diff
@@ -0,0 +1,107 @@
...
+                cosd(coords.long),
+            )
+
+            u = Geometry.transform.(Ref(Geometry.Covariant12Axis()), u_local)
+
+            curl_us = scurl.(u)
+            # Spaces.weighted_dss!(curl_us)
+
+            # curl_uw = wcurl.(u)
+            # Spaces.weighted_dss!(curl_uw)
```
simonbyrne, line 91: Is this still not working?

test/sphere_curl.jl
```diff
@@ -0,0 +1,107 @@
...
+"""
+convergence_rate(err, Δh) =
+    [log(err[i] / err[i - 1]) / log(Δh[i] / Δh[i - 1]) for i in 2:length(Δh)]
+
+@testset "convergence tests for the curl operator on the sphere" begin
+    Nes = [3, 9, 27]
+    Nqs = [4, 6]
+
+    for (Iq, Nq) in enumerate(Nqs)
+        # TODO!: The reference array needs to be changed once the dss is ready.
```
simonbyrne, line 68: This comment is no longer true.

test/sphere_curl.jl
jiahe23: It works now and I've uncommented it. Tests for weak curl are included now.

test/sphere_curl.jl
jiahe23: removed!

jiahe23:
bors try

jiahe23:
@simonbyrne  @valeriabarra  Let me know if you have any other concerns. Otherwise, I think this it good to go ;)

simonbyrne:
Fantastic, thanks!

valeriabarra:
Thanks @jiahe23 for your nice work! I wonder if at this point we should also use the same pattern for the diffusion and hyperdiffusion tests that were already added.

@simonbyrne is it worth it to mention, perhaps in a comment for the future reader, why there is no difference between the weak and strong forms of the operators?

simonbyrne:
> @simonbyrne is it worth it to mention, perhaps in a comment for the future reader, why there is no difference between the weak and strong forms of the operators?

What did you have in mind?

valeriabarra:
> > @simonbyrne is it worth it to mention, perhaps in a comment for the future reader, why there is no difference between the weak and strong forms of the operators?
> 
> What did you have in mind?

Nothing fundamental. We already mention the different formulations of strong and weak operators in our docs ([strong](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.Divergence) and [weak](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.WeakDivergence)), so probably nothing further needs to be specifically mentioned here :+1:

jiahe23:
> Thanks @jiahe23 for your nice work! I wonder if at this point we should also use the same pattern for the diffusion and hyperdiffusion tests that were already added.

That sounds good. Maybe in a separate PR?

PR #280, opened 2021-11-03
Shallow-water Test Case 3: Steady State Geostrophic Flow with Compact Support

valeriabarra:
This ~~WIP~~ PR adds Test Case 3 to our suite of Shallow-water test cases in `shallow_water.jl`. 

The description of the test case can be found in Williamson "A standard test set for numerical approximations to the shallow water equations in spherical geometry", 1992.

This modifies the Initial Condition only, not the governing equations set in the `rhs` function, which is the same for all test cases.


examples/sphere/shallow_water.jl
```diff
@@ -171,6 +184,76 @@ if test_name == rossby_haurwitz_test_name
...
+        # Set auxiliary function needed for initial state of velocity field
+        b(x) = x ≤ 0.0 ? 0.0 : exp(-x^(-1))
+
+        x(ϕprime) = xₑ * (ϕprime - ϕᵦ) / (ϕₑ - ϕᵦ)
+        uλprime(ϕprime) =
+            u0 * b(x(ϕprime)) * b(xₑ - x(ϕprime)) * exp(4.0 / xₑ)
+        uϕprime = 0.0
+
+        # Set integral needed for height initial state
+        hᵢₙₜ(γ) = abs(γ) < 90.0 ? (2 * Ω * sind(γ) + uλprime(γ) * tand(γ) / R) * uλprime(γ) : 0.0
```
charleskawczynski, line 224: This is what I see

Screen Shot 2021-11-03 at 3 08 38 PM


cc @simonbyrne

valeriabarra: Oh, interesting. I got the variable name from what we were using in a [similar test case in ClimateMachine.jl](https://github.com/CliMA/ClimateMachine.jl/blob/46b0828569134b8d7ceb68559e76f1abf15968f8/test/Numerics/DGMethods/compressible_navier_stokes_equations/sphere/benchmark_barotropic_instability.jl#L79)

valeriabarra: We can rename it, of course

simonbyrne: it renders correctly on mine, but looks a bit funny. lets just call it `h_int`

examples/sphere/shallow_water.jl
```diff
@@ -171,6 +184,78 @@ if test_name == rossby_haurwitz_test_name
...
+            u0 * b(x(ϕprime)) * b(xₑ - x(ϕprime)) * exp(4.0 / xₑ)
+        uϕprime = 0.0
+
+        # Set integral needed for height initial state
+        hᵢₙₜ(γ) =
+            abs(γ) < 90.0 ?
+            (2 * Ω * sind(γ) + uλprime(γ) * tand(γ) / R) * uλprime(γ) : 0.0
+
+        # Set initial state for height field
+        h = h0 - (R / g) * quadgk(γ -> hᵢₙₜ(γ), -90.0, ϕprime)[1]
```
simonbyrne, line 229: the anonymous function here is redundant:
```suggestion
        h = h0 - (R / g) * quadgk(hᵢₙₜ, -90.0, ϕprime)[1]
```

valeriabarra:
bors try

valeriabarra:
This works now, thanks to Oswald who caught the small bug!

valeriabarra:
I added 3 artifacts of plots of errors over time. Those are quantities of interest suggested in the reference paper.
These are for 10 days simulation. 

L1errorVstime

L1InferrorVstime

L2errorVstime

valeriabarra:
@simonbyrne , this is ready for review. 
If it looks good and approved, I can squash commits and move quickly to the Gawlesky's Test case, which is supposed to be a slight modification of this with a perturbed field

valeriabarra:
For comparison, the default case `steady_state`, which is Test Case 2 "Global Steady State Nonlinear Zonal Geostrophic Flow" errors:

These are for 5 days simulation and for `α = 0` degrees:

L1InferrorVstime_steadystate_alpha0_day5

L1errorVstime_steadystate_alpha0_day5


L2errorVstime_steadystate_alpha0_day5

And these are for 5 days simulation and for `α = 45` degrees:

L1InferrorVstime_steadystate_alpha45_day5


L1errorVstime_steadystate_alpha45_day5

L2errorVstime_steadystate_alpha45_day5

valeriabarra:
bors try

valeriabarra:
bors try

simonbyrne (approved):

simonbyrne:
Code looks good to me (I can't comment on the numerical results though: they look stable at least?)

valeriabarra:
> Code looks good to me (I can't comment on the numerical results though: they look stable at least?)

Yes, I think as these are steady-state runs, it is normal to see round-off errors accumulating after many time steps

simonbyrne:
would it make sense to plot relative norms for L1 and L2? or perhaps normalize by the unit norm?

valeriabarra:
> would it make sense to plot relative norms for L1 and L2? or perhaps normalize by the unit norm?

Yes, thanks for reminding me. I meant to use relative errors (that was also the recommended quantity in the reference paper, anyways). I just pushed it. If it looks good, I can go ahead and squash/merge

valeriabarra:
bors try

valeriabarra:
@simonbyrne artifact look good to me on buildkite (with relative errors) and I squashed commits. Ready to merge on my end.

PR #287, opened 2021-11-08
Add 1D FV-FV Remapping

jb-mackay:
Adds support for finite volume remapping between 1D intervals.

### DIFF
--- a/src/Operators/remapping.jl
+++ b/src/Operators/remapping.jl
@@ -1,6 +1,8 @@
-using ..Spaces: AbstractSpace, SpectralElementSpace2D, Quadratures
-using ..Topologies: GridTopology
+using ..Spaces:
+    AbstractSpace, SpectralElementSpace1D, SpectralElementSpace2D, Quadratures
+using ..Topologies: GridTopology, IntervalTopology
 using ..Fields: Field
+using ..DataLayouts
 using SparseArrays, LinearAlgebra
 
 struct LinearRemap{T <: AbstractSpace, S <: AbstractSpace, M <: AbstractMatrix}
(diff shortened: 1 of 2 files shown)


src/Operators/remapping.jl
```diff
@@ -108,6 +110,35 @@ function overlap_weights(
...
+        vertices_i = Topologies.vertex_coordinates(target_topo, i)
+        xmin_i, xmax_i = vertices_i[1].x, vertices_i[2].x
+        for j in 1:nx2
+            vertices_j = Topologies.vertex_coordinates(source_topo, j)
+            xmin_j, xmax_j = vertices_j[1].x, vertices_j[2].x
+            xmin_ov, xmax_ov = max(xmin_i, xmin_j), min(xmax_i, xmax_j)
+            x_overlap = xmax_ov > xmin_ov ? xmax_ov - xmin_ov : continue
+            X_ov[i, j] = x_overlap
+        end
+    end
```
jb-mackay, line 138: Some code duplication with 2D case. Could abstract and create function for this (the Y-direction case for 2D is the annoying part) if we feel this is worth it, but this all may be temporary with FV <-> Spectral coming.

src/Operators/remapping.jl
```diff
@@ -108,6 +110,35 @@ function overlap_weights(
     return kron(Y_ov, X_ov)
 end
 
+function overlap_weights(
+    target::T,
+    source::S,
+) where {
+    T <: SpectralElementSpace1D{<:IntervalTopology, Quadratures.GL{1}},
+    S <: SpectralElementSpace1D{<:IntervalTopology, Quadratures.GL{1}},
+}
```
valeriabarra, line 119: Just my little OCD speaking... what do you think about moving the 1D case up before the 2D case?

src/Operators/remapping.jl
valeriabarra: You're right. This code block that finds the overlap in x-dimension could be wrapped in a function and used both in the 1D and 2D cases. I don't think it will be unnecessary, even if we move to FV <-> SEM remapping. I think we could still keep the FV <-> FV prototype, no?

jb-mackay: Let me know what you think of the change here @valeriabarra!

valeriabarra: I like it! Much cleaner and better organized. Thanks!!

test/remapping.jl
```diff
@@ -72,149 +72,291 @@ end
...
-            # test consistent remap
-            target_field = remap(R, source_field)
-            @test vec(parent(target_field)) ≈
-                  ones(length(target.local_geometry))
-
-            # test simple remap
-            vec(parent(source_field)) .= [1.0; 2.0; 3.0; 4.0]
-            remap!(target_field, R, source_field)
-            @test vec(parent(target_field)) ≈ [1.0; 2.0; 3.0; 4.0]
+    @testset "1D Domains" begin
```
valeriabarra, line 225: Besides my little OCD, I think it's more natural and organized to put the 1D case (both in tests and in the API functions) first and then the higher-dim cases

valeriabarra (approved):

src/Operators/remapping.jl
```diff
@@ -54,60 +56,102 @@ function linear_remap_op(target::AbstractSpace, source::AbstractSpace)
...
+    # Calculate element overlap pattern
+    # X_ov[i,j] = overlap length between target elem i and source elem j
+    X_ov = fv_overlap_1D(target, source; dims = :x)
```
jakebolewski, line 76: I think it would be more natural to use the Geometry point types  here rather than `dims` symbol.

valeriabarra: That's a good point. It's better to be consistent with the existing API

jb-mackay: Yes that's a nice idea!

simonbyrne (approved):

PR #289, opened 2021-11-09
Add Biased operators

charleskawczynski:
This PR adds `LeftBiasedF2C` and `RightBiasedF2C` operators and tests for these and existing biased operators.

I think I've set up the tests correctly, but I'd definitely appreciate a second set of eyes.

### DIFF
--- a/test/fdspaces.jl
+++ b/test/fdspaces.jl
@@ -567,3 +567,63 @@ end
     @test conv_div_cos_f[1] ≤ conv_div_cos_f[2] ≤ conv_div_cos_f[3]
 
 end
+
+@testset "Biased interpolation" begin
+    FT = Float64
+    n_elems = 10
+
+    domain = Domains.IntervalDomain(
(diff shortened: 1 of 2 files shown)


jakebolewski (approved):

test/fdspaces.jl
```diff
@@ -567,3 +567,63 @@ end
     @test conv_div_cos_f[1] ≤ conv_div_cos_f[2] ≤ conv_div_cos_f[3]
 
 end
+
+@testset "Biased interpolation" begin
```
valeriabarra, line 571: Is the term "biased" very common to express one-sided (forward/backward) operators in this community? It's not that common in the applied math / scientific computing communities where biases are generally left for statics and ML contexts

charleskawczynski: I was just following the existing naming convention. I'm also used to using one-sided/forward/backward operators.

PR #292, opened 2021-11-10
Shallow-water Test Case: Barotropic Instability

valeriabarra:
This PR concludes the suite of recommended shallow-water test cases. 
This adds the Barotropic Instability test case (see refs: Galewsky et al, "An initial-value problem for testing numerical models of the global shallow-water equations", 2004 (also in Sec. 7.6 of Ullirch et al, "High-order ﬁnite-volume methods for the shallow-water equations on  the sphere", 2010). 
This is an unsteady test case, so no error checking and plotting in this case, but artifacts for the final height field for both  α = 0 and α = 30 degrees are produced.

### DIFF
--- a/examples/sphere/shallow_water.jl
+++ b/examples/sphere/shallow_water.jl
@@ -14,7 +14,7 @@ using TerminalLoggers: TerminalLogger
 global_logger(TerminalLogger())
 
 # This example solves the shallow-water equations on a cubed-sphere manifold.
-# This file contains four test cases:
+# This file contains five test cases:
 # - One, called "steady_state", reproduces Test Case 2 in Williamson et al,
 #   "A standard test set for numerical approximations to the shallow water
 #   equations in spherical geometry", 1992. This test case gives the steady-state
@@ -34,24 +34,38 @@ global_logger(TerminalLogger())
 # - A fourth one, called "rossby_haurwitz", reproduces Test Case 6 in the same
 #   reference paper. It represents the solution of the nonlinear barotropic
 #   vorticity equation on the sphere
+# - A fifth one, called "barotropic_instability", reproduces the test case in
+#   Galewsky et al, "An initial-value problem for testing numerical models of
+#   the global shallow-water equations", 2004 (also in Sec. 7.6 of Ullirch et al,
+#   "High-order ﬁnite-volume methods for the shallow-water equations on
+#   the sphere", 2010). This test case consists of a zonal jet with compact
+#   support at a latitude of 45°. A small height disturbance is then added,
+#   which causes the jet to become unstable and collapse into a highly vortical
+#   structure.
 
 # Physical parameters needed
 const R = 6.37122e6
 const Ω = 7.292e-5
 const g = 9.80616
 const D₄ = 1.0e16 # hyperdiffusion coefficient
+
+# Test case specifications
 const test_name = get(ARGS, 1, "steady_state") # default test case to run
 const test_angle_name = get(ARGS, 2, "alpha0") # default test case to run
 const steady_state_test_name = "steady_state"
 const steady_state_compact_test_name = "steady_state_compact"
 const mountain_test_name = "mountain"
 const rossby_haurwitz_test_name = "rossby_haurwitz"
+const barotropic_instability_test_name = "barotropic_instability"
 const alpha0_test_name = "alpha0"
+const alpha30_test_name = "alpha30"
 const alpha45_test_name = "alpha45"
 const alpha60_test_name = "alpha60"
 
 # Test-specific physical parameters
-if test_angle_name == alpha45_test_name
+if test_angle_name == alpha30_test_name
+    const α = 30.0
+elseif test_angle_name == alpha45_test_name
     const α = 45.0
 elseif test_angle_name == alpha60_test_name
     const α = 60.0
@@ -77,6 +91,17 @@ elseif test_name == steady_state_compact_test_name
     const ϕᵦ = -30.0
     const ϕₑ = 90.0
     const xₑ = 0.3
+elseif test_name == barotropic_instability_test_name
+    const u_max = 80.0
+    const xₑ = 0.3
+    const αₚ = 19.09859
+    const βₚ = 3.81971
+    const h0 = 10158.18617 # value for initial height from Tempest https://github.com/paullric/tempestmodel/blob/master/test/shallowwater_sphere/BarotropicInstabilityTest.cpp#L86
+    const h_hat = 120.0
+    const ϕ₀ = 25.71428
+    const ϕ₁ = 64.28571
+    const ϕ₂ = 45.0
+    const eₙ = exp(-4.0 / (deg2rad(ϕ₁) - deg2rad(ϕ₀))^2)
 else # default case, steady-state test case
     const u0 = 2 * pi * R / (12 * 86400)
     const h0 = 2.94e4 / g
@@ -101,7 +126,7 @@ function linkfig(figpath, alt = "")
 end
 
 # Set up discretization
-ne = 7 # the rossby_haurwitz test case's initial state has a singularity at the pole. We avoid it by using odd number of elements
+ne = 9 # the rossby_haurwitz test case's initial state has a singularity at the pole. We avoid it by using odd number of elements
 Nq = 4
 
 domain = Domains.SphereDomain(R)
@@ -202,7 +227,7 @@ elseif test_name == steady_state_compact_test_name
             λprime = λ
         else
             ϕprime = asind(sind(ϕ) * cosd(α) - cosd(ϕ) * cosd(λ) * sind(α))
-            λprime = asind(sind(λ) * cosd(ϕ) / cosd(ϕprime))
+            λprime = asind(sind(λ) * cosd(ϕ) / cosd(ϕprime)) # for alpha45, this experiences numerical precision issues. The test case is designed for either alpha0 or alpha60
 
             # Temporary angle to ensure λprime is in the right quadrant
             λcond = cosd(α) * cosd(λ) * cosd(ϕ) + sind(α) * sind(ϕ)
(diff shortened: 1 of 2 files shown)


valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

examples/sphere/shallow_water.jl
```diff
@@ -256,6 +281,69 @@ elseif test_name == steady_state_compact_test_name
...
+
+        if λ > 0.0
+            λ -= 360.0
+        end
+        if λ < -360.0 || λ > 0.0
+            @info "Invalid longitude value"
+        end
+
+        # Add height perturbation
+        h += h_hat * cosd(ϕ) * exp(-(λ^2 / αₚ^2) - ((ϕ₂ - ϕ)^2 / βₚ^2))
```
szy21, line 336: Is this from eq. 4 of the paper? I think λ and ϕ are radians in the paper but are degrees here.

examples/sphere/shallow_water.jl
```diff
@@ -256,6 +281,69 @@ elseif test_name == steady_state_compact_test_name
...
+        ϕ = coord.lat
+        λ = coord.long
+
+        if α == 0.0
+            ϕprime = ϕ
+            λprime = λ
+        else
+
+            ϕprime = asind(sind(ϕ) * cosd(α) - cosd(ϕ) * cosd(λ) * sind(α))
+            λprime = asind(sind(λ) * cosd(ϕ) / cosd(ϕprime)) # for alpha45, this experiences numerical precision issues. The test case is designed for either alpha0 or alpha30
```
szy21, line 299: It seems `λprime` is not used in the initial conditions. Maybe just remove it?

examples/sphere/shallow_water.jl
simonbyrne: The easiest fix would be to scale αₚ and βₚ appropriately.

szy21: Yeah, that works too.

valeriabarra: Uh! Thanks for the catch @szy21 !!

examples/sphere/shallow_water.jl
valeriabarra: Yes, thanks for the good eye!

examples/sphere/shallow_water.jl
valeriabarra: > The easiest fix would be to scale αₚ and βₚ appropriately.

Yes, I had scaled other quantities appropriately. I had just missed these two. Thanks for the catch!

valeriabarra: I fixed it in [51a87e8](https://github.com/CliMA/ClimaCore.jl/pull/292/commits/51a87e892b1eb49a4eea503de24e1fda81de2d3c). It didn't change the overall behavior though.

valeriabarra:
bors try

examples/sphere/shallow_water.jl
```diff
@@ -77,6 +91,17 @@ elseif test_name == steady_state_compact_test_name
...
+    const u_max = 80.0
+    const xₑ = 0.3
+    const αₚ = 19.09859
+    const βₚ = 3.81971
+    const h0 = 10158.18617 # value for initial height from Tempest https://github.com/paullric/tempestmodel/blob/master/test/shallowwater_sphere/BarotropicInstabilityTest.cpp#L86
+    const h_hat = 120.0
+    const ϕ₀ = 25.71428
+    const ϕ₁ = 64.28571
+    const ϕ₂ = 45.0
+    const eₙ = exp(-4.0 / (ϕ₁ - ϕ₀)^2)
```
szy21, line 104: Ah I didn't notice this yesterday. I think you need to change `eₙ` or `u_max` as well, as `ϕ₁` and `ϕ₀` are radians in the paper.

examples/sphere/shallow_water.jl
```diff
@@ -256,6 +281,55 @@ elseif test_name == steady_state_compact_test_name
...
+        if α == 0.0
+            ϕprime = ϕ
+        else
+            ϕprime = asind(sind(ϕ) * cosd(α) - cosd(ϕ) * cosd(λ) * sind(α))
+        end
+
+        # Set initial state of velocity field
+        uλprime(ϕprime) =
+            (u_max / eₙ) *
+            exp(1.0 / ((ϕprime - ϕ₀) * (ϕprime - ϕ₁))) *
```
szy21, line 302: Also here, `ϕprime`, `ϕ₀`, `ϕ₁` are all radians in the paper.

szy21: And I don't think scaling the parameters would work here. You may need to change the latitude variables. @simonbyrne?

examples/sphere/shallow_water.jl
valeriabarra: I'm not sure I follow what you're saying. I did change `ϕ₀` and `ϕ₁` to degrees. And `eₙ` is computed from them. Are you saying that we should change the formula for `eₙ`? 
In the previous case `steady_state_compact`, we didn't change `u0`, which is the analog of `u_max` here and we didn't change the formulation for `uλprime` that has a similar factor `exp(4.0 / xₑ)`. Are you saying we should have change that as well?

szy21: It's ok in the `steady_state_compact` because `x(ϕprime) = xₑ * (ϕprime - ϕᵦ) / (ϕₑ - ϕᵦ)`, and `u` is a function of x, so the results would be the same as long as the latitudes are all in degrees or all in radians. But here `eₙ = exp(-4.0 / (ϕ₁ - ϕ₀)^2)`, and `u` is affected by `eₙ`, so the results would be different for degrees and radians.

examples/sphere/shallow_water.jl
valeriabarra: Oh, I think now I know what you mean. I think we need a conversion factor in the formula, like we did on this other [line](https://github.com/CliMA/ClimaCore.jl/blob/51a87e892b1eb49a4eea503de24e1fda81de2d3c/examples/sphere/shallow_water.jl#L312) for the h integral. So this line would read:

```
uλprime(ϕprime) =
            (pi / 180.0) * (u_max / eₙ) *
            exp(1.0 / ((ϕprime - ϕ₀) * (ϕprime - ϕ₁))) *
            (ϕ₀ < ϕprime < ϕ₁)
```
I'll try in a bit and will let you know

valeriabarra:
Day 10 run, with α=0:

https://user-images.githubusercontent.com/39932030/141384381-13952bf3-ca14-4e76-9454-fc1db1735bcf.mp4

Day 10 run, with α=30:

https://user-images.githubusercontent.com/39932030/141384438-4acae84c-49a2-42e1-b9ee-cbc511a881da.mp4

examples/sphere/shallow_water.jl
szy21: It's not exactly the same. For the line you referred to, the term `quadgk(h_int, -90.0, ϕprime)[1]` when ϕprime is in radians is the same as  `(pi / 180.0) * quadgk(h_int, -90.0, ϕprime)[1]` when ϕprime is in degrees. But here, exp(1.0 / ((ϕprime - ϕ₀) * (ϕprime - ϕ₁))) when ϕ is in radians is **not** (pi / 180.0)*exp(1.0 / ((ϕprime - ϕ₀) * (ϕprime - ϕ₁))) when ϕ is in degrees, because ϕ is in the exponent. Same thing for eₙ.

valeriabarra: Oh, thank you for clarifying that. Now it is indeed more clear! I did not understand what you meant before. I'll look into it

szy21: Sounds good, thanks! Let me know if there is anything still unclear.

valeriabarra: I think I finally fixed in [6245abb](https://github.com/CliMA/ClimaCore.jl/pull/292/commits/6245abb90a9a1e17c4e66a5380efb892304c91dd). Let me know if you have any other comments! And thanks again! 
This decision of using degrees over radians and between [-180.0, 180.0] over [0, 360.0] it's giving more than a hiccup in debugging test cases and examples when comparing them with the literature :joy:

valeriabarra:
bors try

valeriabarra:
Day 10, with `α = 0`:

https://user-images.githubusercontent.com/39932030/141518632-d3e457b3-df3e-46e1-8545-c9961f517f88.mp4

Day 10, with `α = 30`:

https://user-images.githubusercontent.com/39932030/141518670-0ea0039f-69a0-4e8c-ad34-ff5ee5e03cfe.mp4

examples/sphere/shallow_water.jl
szy21: I think it's good now! Thanks!

szy21 (approved):

examples/sphere/shallow_water.jl
```diff
@@ -77,6 +91,17 @@ elseif test_name == steady_state_compact_test_name
...
+    const u_max = 80.0
+    const xₑ = 0.3
+    const αₚ = 19.09859
+    const βₚ = 3.81971
+    const h0 = 10158.18617 # value for initial height from Tempest https://github.com/paullric/tempestmodel/blob/master/test/shallowwater_sphere/BarotropicInstabilityTest.cpp#L86
+    const h_hat = 120.0
+    const ϕ₀ = 25.71428
+    const ϕ₁ = 64.28571
+    const ϕ₂ = 45.0
+    const eₙ = exp(-4.0 / (π * ϕ₁ / 180.0 - π * ϕ₀ / 180.0)^2)
```
jakebolewski, line 104: I think it would be clearer to use the function `deg2rad()` consistently doing the conversions in this example.

valeriabarra: Okay, I can do that!

jakebolewski:
Is there a common colormap range we can use for these examples?  having a varying colormap over simulation time makes it difficult to compare across runs, as it is output dependent.

valeriabarra:
> Is there a common colormap range we can use for these examples? having a varying colormap over simulation time makes it difficult to compare across runs, as it is output dependent.

Do you mean to fix the colormap limits across all test cases? They have quite different ranges of values, so that'll probably make the distinct features hard to see in some of them. 
We are not using the animations over time as artifcats by the way, but only the snapshot at the final simulation time. I was just posting the animations here to show the dynamic results, but they are not in the buildkite artifacts

valeriabarra:
I squashed commits as I think all review comments related to this test case in particular were addressed.

bischtob (approved):
Looks good to me

PR #295, opened 2021-11-10
Fixes a bug in opposing face

sriharshakandala:
Fixes a bug in opposing face, that potentially effects single element meshes with periodic boundary conditions.


sriharshakandala:
bors try

simonbyrne (approved):

PR #296, opened 2021-11-10
allow using scalars and arrays in FieldVectors

### LINKED ISSUE #139: Add scalar support to FieldVector
Not sure the best way to do this, but it should be possible to do something like
```
Y = FieldVector(x=(a=1.0,b=2.0))
``
and do either
```
Y.x = (a=1.0, b=3.0)
```
or 
```
Y.x.b = 3.0
```

### PR DESCRIPTION
**Pull request by simonbyrne**
Fixes #139.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -1,22 +1,91 @@
 import BlockArrays
 
+
+"""
+    FieldVector
+
+A `FieldVector` is a wrapper around one or more `Field`s that acts like vector
+of the underlying arrays.
+
+It is similar in spirit to [`ArrayPartition` from
+RecursiveArrayTools.jl](https://github.com/SciML/RecursiveArrayTools.jl#arraypartition),
+but allows referring to fields by name.
+
+# Constructors
+
+    FieldVector(;name1=field1, name2=field2, ...)
+
+Construct a `FieldVector`, wrapping `field1, field2, ...` using the names
+`name1, name2, ...`.
+"""
 struct FieldVector{T, M} <: BlockArrays.AbstractBlockVector{T}
     values::M
 end
 FieldVector{T}(values::M) where {T, M} = FieldVector{T, M}(values)
 
+
+"""
+    ScalarWrapper(val) <: AbstractArray{T,0}
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Fields/fieldvector.jl [hunk 1]
```diff
@@ -1,22 +1,92 @@
...
 
+
+"""
+    ScalarWrapper(val) <: AbstractArray{T,0}
+
+This is a wrapper around scalar values that allows them to be mutated as part of
+a FieldVector. A call `getproperty` on a `FieldVector` with this component will
+return a scalar, instead of the boxed object.
+"""
+mutable struct ScalarWrapper{T} <: AbstractArray{T, 0}
```
jakebolewski, line 34: Does Ref not currently work?  it seems to provide the same basic interface

src/Fields/fieldvector.jl
```diff
@@ -1,22 +1,92 @@
...
 _values(fv::FieldVector) = getfield(fv, :values)
-_parent_values(fv::FieldVector) = map(parent, _values(fv))
+
+"""
+    backing_array(x)
+
+The `AbstractArray` or `Ref` that is behind an object `x`, allowing it to be
+treated as a component of a `FieldVector`.
+"""
+backing_array(x) = x
```
jakebolewski, line 63: should we specialize this to just AbstractArray` and `Ref`?

src/Fields/fieldvector.jl
jakebolewski: Ah is it because all `FieldVector` backing types need to be an `AbstractArray` subtype?

simonbyrne: I tried using `Ref` originally, but it required too many special cases (e.g. `similar(::Ref)` isn't defined, and linear indexing doesn't work). In the end it was simpler to just define our own.

jakebolewski (approved):

src/Fields/fieldvector.jl
simonbyrne: Oh, I meant to get rid of the `Ref` mention (since that doesn't really work for the reasons I noted above).

The input can be anything, the output should be an `AbstractArray`.

PR #297, opened 2021-11-11
Allow basetype replacement

dennisYatunin:
The current version of `basetype` doesn't throw an informative error message when the input type has 0 size, and it can throw an error when the input type contains one or more fields with zero size (e.g., `ForwardDiff.Dual{Nothing, Float64, 0}`). This PR fixes both of these issues. It also gets rid of the `basetype` method defined for `NamedTuple`, which is unnecessary because the fields of a `NamedTuple` are the same as the fields of the underlying `Tuple`.

Note that this does not add the ability to have `basetype` return something other than an `AbstractFloat`. That will require further changes.

@simonbyrne @jakebolewski


jakebolewski:
I font think the implementation is correct (see tests), and the way the comparison is being done is not type stable and this method is called often.  Not sure if the optimizer is able to unbox the Union{Nothing, Type} type in our uses.

src/DataLayouts/struct.jl
```diff
@@ -14,21 +14,21 @@ import Base: @propagate_inbounds
 """
     basetype(S...)
 
-Compute the "base" floating point type of one or more types `S`. This will throw
+Compute the "base" floating point type of `S`. This will throw
 an error if there is no unique type.
 """
 basetype(::Type{FT}) where {FT <: AbstractFloat} = FT
-basetype(::Type{NamedTuple{names, T}}) where {names, T} = basetype(T)
```
jakebolewski, line 21: IIRC this was added for performance reasons

dennisYatunin:
I've added back the two method signatures I removed (sorry for not realizing they were necessary). I also switched the `nothing`s to `Union{}`s. I think that should make things type stable.

jakebolewski:
It would be good to add some tests.

dennisYatunin:
I have now extended this PR to allow the base type of a FieldVector to be replaced with any valid type. As long as the tendency `rhs!` does not allocate new fields with operator broadcasts (more on this later), this PR allows the use of matrix-free ODE solver methods with automatic differentiation, e.g.:
```
using OrdinaryDiffEq
using DiffEqOperators
f = ODEFunction(rhs!; jac_prototype = JacVecOperator(rhs!, Y));
sol = solve(
    ODEProblem(f, Y, (0.0, 1000.0)),
    TRBDF2(linsolve = LinSolveGMRES()),
);
```
Here is a summary of the changes I made in this PR:
- Added a call to `check_basetype` to every DataLayout constructor. This function checks that both the base type and the struct type of a DataLayout are `isbitstype`, and that the struct type can be stored using the base type. Since both the base type and the struct type are guaranteed to be `isbitstype`, I removed the code that checks for that in the `typesize` function. Also, since it is guaranteed that the struct type can be stored using the base type, and since every call to `setindex!` on a DataLayout converts the input to the struct type, I removed the code that checks for an unexpected argument type in the `set_struct!` function. In case users frequently construct new DataLayouts, I made `check_basetype` an optionally generated function. I also added unit tests for `check_basetype`.
- Removed the old `basetype` function, along with its unit tests. There were a few places where `basetype` was used to get the element type of the parent array of a DataLayout, so I replaced those with calls to `eltype`.
- Extended the `replace_basetype` function in struct.jl so that it can handle arbitrary types, not just Tuples and NamedTuples. I did this by replacing type parameters instead of field types, because the latter appears to be impossible. This has the downside of potentially causing issues when users utilize type parameters in ways they shouldn't, which I explained in detail in a comment. I also extended `replace_basetype` in DataLayouts.jl from just VF to all DataLayouts.
- Replaced `FT` with `T` in DataLayouts.jl, since the base type is no longer restricted to being an AbstractFloat. I did not do this for the entire codebase, as that would be a very large number of insignificant changes.
- Allowed broadcasts over DataLayouts with different base types (and container types). Specifically, when `BroadcastStyle` is called on two DataStyles with different array types `A1` and `A2`, the resulting DataStyle has an array type of `promote_container_type(A1, A2)`. The function `promote_container_type` separately promotes the container types (currently defaulting to `Array` or `CuArray`) and the element types (done with `promote_type`).
- Fixed a bug which occurred when `deepcopy` was called on FieldVectors or objects containing FieldVectors. This would previously result in the copied FieldVector having a distinct space from the original FieldVector, but now `deepcopy` preserves physical equality of spaces.
- Added some unit tests for replacing the base type of FieldVectors and calling `deepcopy` on them.

There is still another issue that needs to be fixed throughout the Operator broadcasting code: we need to remove the assumption that the container and element types of the DataLayouts used for the space of a Field are the same as those used for the values of a Field. This PR includes one example of such a change. Originally, the following code was in Fields.jl:
```
Base.similar(field::Field, ::Type{Eltype}) where {Eltype} =
     Field(Eltype, axes(field))
```
This has been replaced with
```
Base.similar(field::Field, ::Type{T}) where {T} =
     Field(similar(field_values(field), T), axes(field))
```
Right now, any broadcasts with Operators that allocate new Fields have the potential of allocating Fields with incorrect base types. For example, if `ρw` is a Field whose space is stored with a base type of Float64 but whose values are stored with a base type of Dual, then `Operators.InterpolateF2C().(ρw)` will allocate a new Field whose values are stored with a base type of Float64, even though the values themselves are Duals. I will make this a separate issue once this PR is merged in.

@jakebolewski

src/DataLayouts/DataLayouts.jl
```diff
@@ -893,6 +940,12 @@ function IV1JH2{S, Ni}(array::AbstractMatrix{S}) where {S, Ni}
     IV1JH2{S, Ni, typeof(array)}(array)
 end
 
+function replace_basetype(data::IV1JH2{S, Ni}, ::Type{T}) where {S, Ni, T}
```
jakebolewski, line 943: We don't really need this definition as we can only use this datalayout for output / plotting and Duals won't work there

src/DataLayouts/broadcast.jl
```diff
@@ -59,31 +59,35 @@ Base.Broadcast.BroadcastStyle(
...
+    ::IFHStyle{Ni, A2},
+) where {Ni, A1, A2} = VIFHStyle{Ni, promote_container_type(A1, A2)}()
 
 Base.Broadcast.BroadcastStyle(
-    ::VFStyle{A},
-    ::IJFHStyle{Nij, A},
-) where {Nij, A} = VIJFHStyle{Nij, A}()
+    ::VFStyle{A1},
+    ::IJFHStyle{Nij, A2},
+) where {Nij, A1, A2} = VIJFHStyle{Nij, promote_container_type(A1, A2)}()
```
jakebolewski, line 70: `promote_parent_type`? we refer to the parent backing array everywhere

src/DataLayouts/cuda.jl
```diff
@@ -19,9 +19,13 @@ Adapt.adapt_structure(to, data::IF{S, Ni}) where {S, Ni} =
...
     VF{S}(Adapt.adapt(to, parent(data)))
 
-parent_array_type(
-    ::Type{A},
-) where {A <: CUDA.GPUArrays.AbstractGPUArray{FT}} where {FT} = CUDA.CuArray{FT}
+parent_array_type(::Type{<:CUDA.GPUArrays.AbstractGPUArray{T}}) where {T} =
+    CUDA.CuArray{T}
+
+promote_container_type(
+    ::Type{<:CUDA.GPUArrays.AbstractGPUArray{T1}},
```
jakebolewski, line 26: this needs to have a condition that the dimensions and buffers are the same type:
```
::GPUArray{T1, Dim, Buffer}
::GPUArray{T2, Dim, Buffer}
where {T1, T2, Dim, Buffer}
```

src/DataLayouts/struct.jl
```diff
@@ -11,46 +11,92 @@
...
+replace_basetype(::Type{T}, ::Type{T′}, ::Type{S}) where {T, T′, S} =
+    length(S.parameters) == 0 ? S :
+    S.name.wrapper{replace_basetypes(T, T′, S.parameters...)...}
+replace_basetype(::Type{T}, ::Type{T′}, ::Type{<:T}) where {T, T′} = T′
+replace_basetype(::Type{T}, ::Type{T′}, value) where {T, T′} = value
+replace_basetypes(::Type{T}, ::Type{T′}, value) where {T, T′} =
+    (replace_basetype(T, T′, value),)
+replace_basetypes(::Type{T}, ::Type{T′}, value, values...) where {T, T′} =
+    (replace_basetype(T, T′, value), replace_basetypes(T, T′, values...)...)
+# TODO: This could potentially lead to some annoying bugs, since it replaces
```
jakebolewski, line 71: yeah there are probably better ways (or at least more checks we could add) to do this but this is a good writeup of the issues

src/DataLayouts/struct.jl
```diff
@@ -11,46 +11,92 @@
...
-# TODO: extract interface to overload for backends into separate file
+parent_array_type(::Type{<:AbstractArray{T}}) where {T} = Array{T}
+
+"""
+    promote_container_type(::Type{<:AbstractArray}, ::Type{<:AbstractArray})
+
+Returns the type of a container that could store the same values as both input
+container types.
+"""
+promote_container_type(
```
jakebolewski, line 96: `promote_parent_type`?

src/DataLayouts/struct.jl
```diff
@@ -11,46 +11,92 @@
...
+    ::Type{<:AbstractArray{T1}},
```
jakebolewski, line 97: I think this should just be enforced to be `Array{T1}` for now.  Like CUArrays we also need to enforce that the Dimension parameter is also correct ` where {T1, T2, Dim}`

src/Fields/fieldvector.jl
```diff
@@ -101,17 +101,27 @@ Base.similar(fv::FieldVector{T}) where {T} =
...
 Base.zero(fv::FieldVector{T}) where {T} = FieldVector{T}(map(zero, _values(fv)))
 
+# Ensure that Field axes are preserved when FieldVectors are deepcopied and when
+# objects containing FieldVectors are deepcopied.
+_deepcopy(x) = deepcopy(x)
+_deepcopy(x::Field) = Field(deepcopy(field_values(x)), axes(x))
+_deepcopy_internal(x, stackdict::IdDict) = Base.deepcopy_internal(x, stackdict)
+_deepcopy_internal(x::Field, stackdict::IdDict) =
+    Field(Base.deepcopy_internal(field_values(x), stackdict), axes(x))
+Base.deepcopy(fv::FieldVector{T}) where {T} =
```
jakebolewski, line 120: I think you only need to consistently  overload `deepcopy_internal` to modify behavior (deepcopy def here isn't needed)?

jakebolewski:
Just a few minor comments, but looks goods.  Needs a squash before merging.

dennisYatunin:
Did I squash correctly? I did the following:
```
git reset --hard HEAD~15
git merge --squash HEAD@{1}
git commit -m <commit message>
git pull
git push
```

jakebolewski:
No that squashes all commits into the topic branch into the current branch.  I would just use `git rebase -i` and do it manually to have control, then when it looks good locally force push to the remote.

dennisYatunin:
Good now?

jakebolewski (approved):

PR #302, opened 2021-11-15
Add JET opt tests for FD Stencil operators

jakebolewski:
* Add tests for all FD operators that do not depend on AxisTensor ops


jakebolewski:
bors try

PR #305, opened 2021-11-20
Add finite difference curl

simonbyrne:
This computes the finite difference (dimension 3) contribution of the curl.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -239,7 +239,6 @@ struct SetDivergence{S} <: BoundaryCondition
     val::S
 end
 
-
 """
     Extrapolate()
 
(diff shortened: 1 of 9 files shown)


simonbyrne:
bors try

src/Operators/finitedifference.jl
```diff
@@ -239,6 +239,14 @@ struct SetDivergence{S} <: BoundaryCondition
     val::S
 end
 
+"""
+    SetCurl(val)
+
+Set the curl at the boundary to be `val`.
+"""
+struct SetCurl{S} <: BoundaryCondition
```
jakebolewski, line 247: It doesn't look like this is used?

jakebolewski:
It would be good to add a stub optimization test in `test/fdspaces_opt.jl` as well.

src/Operators/finitedifference.jl
simonbyrne: Good point. It isn't yet.

src/Operators/finitedifference.jl
```diff
@@ -1760,6 +1768,86 @@ function stencil_right_boundary(
...
+```math
+\\begin{align*}
+C(v)[i]^1 = - \\frac{1}{J[i]} (v₂[i+\\tfrac{1}{2}] - v₂[i-\\tfrac{1}{2}]) \\\\
+C(v)[i]^2 = \\frac{1}{J[i]}  (v₁[i+\\tfrac{1}{2}] - v₁[i-\\tfrac{1}{2}])
+\\end{align*}
+```
+where ``v₁` and ``v₂`` are the 1st and 2nd covariant components of ``v``, and
+``J`` is the Jacobian determinant.
+"""
+struct CurlC2F{BC} <: CurlFiniteDifferenceOperator
```
charleskawczynski, line 1792: I'm a bit confused about the docs here. If this is the vertical contribution of the curl, why would we need BCs? Shouldn't it be using points at the same radius (on the sphere), and it should be a collocated operation, working across DG points.

If it's for the horizontal components, then I think this should be okay.

charleskawczynski: Actually, if it's for horizontal components, shouldn't this live in the DG section?

simonbyrne: Okay, perhaps the docs aren't clear here: it's the contribution of the derivative in the vertical direction, i.e. the terms involve d/d\xi_3. But it only applies to horizontal vectors (i.e. `Covariant12Vector`s)

simonbyrne: Curls are confusing.

jakebolewski: "compute the vertical contribution to the horizontal curl"?

charleskawczynski: Hmm. What is the direction of the output curl vector?

charleskawczynski: > "compute the vertical contribution to the horizontal curl"?

That sounds much more clear to me.

src/Operators/finitedifference.jl
```diff
@@ -1760,6 +1759,89 @@ function stencil_right_boundary(
...
+
+Compute the vertical-derivative contribution to the curl of a center-valued
+covariant vector field `v`, returning a face-valued contravariant vector field.
+Specifically it approximates:
+```math
+\\begin{align*}
+C(v)^1 &= -\\frac{\\partial}{\\partial \\xi^3} v₂ \\\\
+C(v)^2 &= \\frac{\\partial}{\\partial \\xi^3} v₁ \\\\
+\\end{align*}
+```
```
simonbyrne, line 1780: @charleskawczynski is this clearer?

src/Operators/finitedifference.jl
```diff
@@ -1760,6 +1759,89 @@ function stencil_right_boundary(
...
+
+abstract type CurlFiniteDifferenceOperator <: FiniteDifferenceOperator end
+return_eltype(::CurlFiniteDifferenceOperator, arg) =
+    Geometry.curl_result_type(Val((3,)), eltype(arg))
+
+"""
+    C = CurlC2F(;boundaryname=boundarycondition...)
+    C.(v)
+
+Compute the vertical-derivative contribution to the curl of a center-valued
```
charleskawczynski, line 1772:
```suggestion
Compute the vertical-derivative contribution to the horizontal curl of a center-valued
```

charleskawczynski: I think adding "horizontal" to describing the curl is minimally needed to make this clear, we could also specify that the input vector and output vectors are also horizontal, but I think that is inferable.

test/hybrid2d.jl
```diff
@@ -366,3 +366,53 @@ end
...
+        Geometry.transform.(
+            Ref(Geometry.Covariant3Axis()),
+            Geometry.WVector.(cos.(3 .* fcoords.x .+ 4 .* fcoords.z)),
+        )
+
+    curl = Operators.Curl()
+    curlC2F = Operators.CurlC2F(
+        bottom = Operators.SetValue(Geometry.Covariant1Vector(0.0)),
+        top = Operators.SetValue(Geometry.Covariant1Vector(0.0)),
+    )
```
charleskawczynski, line 394: Are these BCs for the boundary-normal, or boundary-tangent, component of the input vector? (or both?)

src/Operators/finitedifference.jl
valeriabarra: Is it really clearer? I'm actually a bit more confused now that I read it this way. Is the curl horizontal or the vectors it acts on?

simonbyrne: I would cal it the "vertical curl", in that it represents the vertical derivative contribution to the curl (in the same way that we define "vertical gradient" or "vertical divergence"). However it only acts on the horizontal covariant component (the vertical covariant component is in the null space of the operator), and it returns a horizontal contravariant  component.

charleskawczynski: I think that the description will be clear so long we document the direction of the input and output vectors. This is the way I typically view these operators.

 - Gradient: scalar -> vector, scalar to one direction
 - Divergence: vector -> scalar, all directions to scalar
 - Curl: vector -> vector, output direction is normal to two input directions

Also, it might be nice to also write how this operator would be written in a more general context with projections. For example, we could include vector identities that we plan to enforce by composing this operator with other complementary differential operators.

I don't want to slow things down, so I'm fine to just merge and open an issue about improving the docs.

valeriabarra: I also think of these differential operators this way. Hence, since given two vectors, the curl is supposed to return a vector normal to the plane spanned by the two input vectors, and here we're only concerned with its vertical component derivatives, I'm a bit confused about why it returns a contravariant horizontal component. But yes, we should improve documentation all of our vector representations and operators

test/hybrid2d.jl
simonbyrne: It's the boundary-tangent component.

simonbyrne: Or, more to the point, only the boundary-tangent component has any effect.

src/Operators/finitedifference.jl
simonbyrne: Another way of thinking about it: if you think of the curl as the cross product of a vector with a differential operator, then you are just taking the cross product with the differential operator in the vertical direction. Since this acts in the vertical covariant direction, the cross product will only depend on the horizontal covariant components of the vector, and the result will have vertical contravariant component 0 (i.e. it will be a horizontal contravariant vector)

valeriabarra: I see, thanks for explaining it this way!

PR #325, opened 2021-12-03
Specify boundary conditions as fields

simonbyrne:
This lets you specify the boundary fluxes as a field across the boundary. Required for coupling

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -132,21 +132,38 @@ interior_indices2 = 2:n-1
 
 import ..Utilities: PlusHalf, half
 
-left_idx(::Spaces.CenterFiniteDifferenceSpace) = 1
-right_idx(space::Spaces.CenterFiniteDifferenceSpace) =
-    length(space.center_local_geometry)
-left_idx(::Spaces.FaceFiniteDifferenceSpace) = half
-right_idx(space::Spaces.FaceFiniteDifferenceSpace) =
-    PlusHalf(length(space.center_local_geometry))
+left_idx(
+    space::Union{
+        Spaces.CenterFiniteDifferenceSpace,
+        Spaces.CenterExtrudedFiniteDifferenceSpace,
+    },
+) = left_center_boundary_idx(space)
+right_idx(
+    space::Union{
+        Spaces.CenterFiniteDifferenceSpace,
+        Spaces.CenterExtrudedFiniteDifferenceSpace,
+    },
+) = right_center_boundary_idx(space)
+left_idx(
+    space::Union{
+        Spaces.FaceFiniteDifferenceSpace,
+        Spaces.FaceExtrudedFiniteDifferenceSpace,
+    },
+) = left_face_boundary_idx(space)
+right_idx(
+    space::Union{
+        Spaces.FaceFiniteDifferenceSpace,
+        Spaces.FaceExtrudedFiniteDifferenceSpace,
+    },
+) = right_face_boundary_idx(space)
 
-left_face_boundary_idx(space::Spaces.FiniteDifferenceSpace) =
-    left_idx(Spaces.FaceFiniteDifferenceSpace(space))
-right_face_boundary_idx(space::Spaces.FiniteDifferenceSpace) =
-    right_idx(Spaces.FaceFiniteDifferenceSpace(space))
-left_center_boundary_idx(space::Spaces.FiniteDifferenceSpace) =
-    left_idx(Spaces.CenterFiniteDifferenceSpace(space))
-right_center_boundary_idx(space::Spaces.FiniteDifferenceSpace) =
-    right_idx(Spaces.CenterFiniteDifferenceSpace(space))
+left_center_boundary_idx(space::Spaces.AbstractSpace) = 1
+right_center_boundary_idx(space::Spaces.AbstractSpace) =
+    size(space.center_local_geometry, 4)
+left_face_boundary_idx(space::Spaces.AbstractSpace) = half
+right_face_boundary_idx(space::Spaces.AbstractSpace) =
+    size(space.face_local_geometry, 4) - half
+
 
 left_face_boundary_idx(arg) = left_face_boundary_idx(axes(arg))
 right_face_boundary_idx(arg) = right_face_boundary_idx(axes(arg))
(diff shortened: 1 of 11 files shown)


src/Operators/finitedifference.jl
```diff
@@ -2106,13 +2106,32 @@ function Base.Broadcast.materialize!(
...
+
+unwrap_ref(r::Ref) = r[]
+unwrap_ref(r) = r
+
+@inline function column_op(f::DivergenceF2C, inds...)
+    DivergenceF2C(map(bc -> column_bc(bc, inds...), f.bcs))
+end
+
+@inline function column_op(f::InterpolateC2F, inds...)
+    InterpolateC2F(map(bc -> column_bc(bc, inds...), f.bcs))
```
simonbyrne, line 2118: These need to be implemented for all operators. Not sure if there is a nicer way to do it.

src/Operators/finitedifference.jl
```diff
@@ -2106,13 +2106,32 @@ function Base.Broadcast.materialize!(
     )
 end
 
+
+unwrap_ref(r::Ref) = r[]
+unwrap_ref(r) = r
```
simonbyrne, line 2111: This is to work around the fact that `column(::SpectralElementField, inds...)` returns a `Ref`, but `column(::Real, inds...)` gives a `Real`. We should make these consistent.

src/Operators/finitedifference.jl
```diff
@@ -2106,13 +2106,32 @@ function Base.Broadcast.materialize!(
...
+
+@inline function column_op(f::InterpolateC2F, inds...)
+    InterpolateC2F(map(bc -> column_bc(bc, inds...), f.bcs))
+end
+@inline function column_bc(bc::SetValue, inds...)
+    SetValue(unwrap_ref(column(bc.val, inds...)))
+end
+@inline function column_bc(bc::Extrapolate, inds...)
+    bc
+end
```
simonbyrne, line 2125: Again, these would need to be defined for all BCs

src/Operators/finitedifference.jl
jakebolewski: It would be good to test if the optimizer works with these before maybe continuing.

`map -> tuplemap` might help a bit

src/Operators/finitedifference.jl
jakebolewski: we should  make Tensor / Vector ops act like scalars in broadcasting IMO

src/Operators/finitedifference.jl
jakebolewski: I feel we get some speedups and new even more indirection are added :)

simonbyrne:
The main thing I don't like about this approach is that the horizontal field has no knowledge of the vertical coordinate or the direction of the boundary normal. So we may want some sort of separate "boundary space", but this will be sufficient for immediate coupler test cases.

simonbyrne:
I think boundaries do need their own special Field or Space object:
- it can have a vertical coordinate (which the horizontal doesn't)
- it has a normal direction

Correspondingly, we can have a `NormalVector` type, which is only definable on such boundary fields.

simonbyrne:
I think this will work for now, we can handle the extra stuff in another PR.

simonbyrne:
bors try

simonbyrne:
@jakebolewski I ran the bubble examples locally (2d and 3d), and didn't see any slowdown

jakebolewski:
Nice, can we add at least one example to the JET tests (with a TODO to add all the others)?

simonbyrne:
@jakebolewski I added the tests, but they're broken even without this change.

jakebolewski:
Probably better to extract that then and to work on it in  a separate PR.  The difference here is that these are tested against ExtrudedSpace Fields?

examples/hybrid/bubble_2d_fieldbc.jl
```diff
@@ -0,0 +1,281 @@
...
+
+bc = (
+    ρθ = (bottom = ZeroFieldFlux(), top = ZeroFlux()),
+    ρu = nothing, # for now BCs are hard coded, except for ρθ
+)
+
+# ZeroFlux condition set as a Field
+sol2 = run!(init(; bc = bc)...)
+
+@test sol1.u == sol2.u
```
jb-mackay, line 281: @simonbyrne let me know where you would like to see this go from here.

jb-mackay:
Currently debugging `examples/hybrid/bubble_2d_fieldbc.jl` demo.

simonbyrne:
bors try

simonbyrne:
Looks like allocations issues...

jb-mackay:
> Looks like allocations issues...

Yep... let's discuss tomorrow. I also have allocation issues on #640

simonbyrne:
bors try

examples/hybrid/bubble_2d_fieldbc.jl
```diff
@@ -0,0 +1,321 @@
+push!(LOAD_PATH, joinpath(@__DIR__, "..", ".."))
```
simonbyrne, line 1: if this file isn't going to be part of a test, it should be deleted

jb-mackay: What do you think is appropriate? It does demonstrate the two ways to set BCs.

test/Operators/hybrid/2d.jl
```diff
@@ -188,74 +188,70 @@ end
...
+        divh = divergence(h, bottom_flux)
 
-        err[k] = norm(exact_divuf .- divuf.h)
+        err[k] = norm(exact_divh .- divh)
     end
     # Divergence convergence rate
+    @test err[3] ≤ err[2] ≤ err[1] ≤ 0.01
     conv_comp_div = convergence_rate(err, Δh)
-    @test err[3] ≤ err[2] ≤ err[1] ≤ 0.1
+    #= not sure these are right?
```
charleskawczynski, line 250: Can we confirm / update these tests?

examples/hybrid/bubble_2d_fieldbc.jl
simonbyrne: It would be better to have a unit test if possible.

test/Operators/hybrid/2d.jl
simonbyrne: @jb-mackay what do these values give?

jb-mackay:
```
julia> conv_comp_div = convergence_rate(err, Δh)
3-element Vector{Float64}:
 1.285293238893628
 0.7164519706218709
 0.5352264906826446
```

charleskawczynski: Doesn't this look non-converging?

test/Operators/hybrid/3d.jl
```diff
@@ -176,15 +176,31 @@ end
...
+    
+    struct ZeroFieldFlux <: BCtag end
+    function bc_divF2C_bottom!(::ZeroFieldFlux, dY, Y, p, t)
+        space = axes(Y.h).horizontal_space
+        FT = Spaces.undertype(space)
+        zeroflux = Fields.zeros(FT, space)
+        return Operators.SetValue(
+            Geometry.UVWVector.(zeroflux, zeroflux, zeroflux)
+        )
+    end
```
jb-mackay, line 193: @simonbyrne are there further unit tests we should add and is this an appropriate place to add this test?

test/Operators/hybrid/2d.jl
simonbyrne: Maybe we need a simpler example?

simonbyrne: Maybe base it on this one: https://github.com/CliMA/ClimaCore.jl/blob/main/test/Operators/finitedifference/column.jl#L687-L741
but make it spatially varying?

jb-mackay:
@simonbyrne: `err = [0.006196395231636458, 0.0019002452759326605, 0.0006113774005246, 0.0002043926280062839]`

jb-mackay:
bors try

PR #326, opened 2021-12-06
misc fixes when debugging compile times

jakebolewski:
Some misc fixes to restore functionality noticed when debugging compile times with @charleskawczynski 

* Edit it didn't seem like changing the optimization made the CI faster


charleskawczynski:
Hmm, I tried running with this commit and everything still runs but the perf didn't seem to change much. I'll have to retry using SnoopCompile to see if it has any affect on that high specialization point.

jakebolewski:
This won’t change anything, it just fixes the non generated fallback that
we added when debugging

PR #327, opened 2021-12-07
Bubble in vector invariant form

simonbyrne:


simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
This now runs the bubble for 500 seconds, which blows out the runtime to almost an hour, but it's helpful to see a longer movie.

jakebolewski:
We need to work on being able to optimize through the Curl Operator, that was a remaining TODO from the first pass.

jakebolewski:
We didn't have an issue to track this: #334

jakebolewski:
@szy21 it would be probably more helpful to have a true 3D bubble in this case rather than N slice models for testing the operators (so that there is non-zero mag for each component).

szy21:
> @szy21 it would be probably more helpful to have a true 3D bubble in this case rather than N slice models for testing the operators (so that there is non-zero mag for each vector component).

Sounds good. @akshaysridhar Is there a 3D bubble in the literature that we can follow? Or maybe I could simply add a perturbation in the y direction?

examples/hybrid/bubble3d_invariant.jl
```diff
@@ -0,0 +1,307 @@
...
+    @. dw -= vgradc2f(cp) / Ic2f(cρ)
+
+    cE = @. (norm(cuvw)^2) / 2 + Φ(coords.z)
+    @. duₕ -= hgrad(cE)
+    @. dw -= vgradc2f(cE)
+
+    # 3) potential temperature
+    @. dρθ -= vdivf2c(fw * Ic2f(cρθ))
+    @. dρθ -= vdivf2c(Ic2f(cuₕ * cρθ))
+    @. dρθ -= hdiv(cuₕ * Yc.ρθ)
```
szy21, line 254: Is this the same as `@. dρθ -= hdiv(cuₕ * cρθ)`?

examples/hybrid/bubble_invariant.jl
```diff
@@ -0,0 +1,325 @@
...
+        top = Operators.Extrapolate(),
+    )
+    cw = If2c.(fw)
+    fuₕ = Ic2f.(cuₕ)
+    cuw = Geometry.Covariant13Vector.(cuₕ) .+ Geometry.Covariant13Vector.(cw)
+
+    dw .= fw .* 0
+
+    # 1.a) horizontal divergence
+    dρ .-= hdiv.(cρ .* (cuw))
```
szy21, line 192: Why is the horizontal divergence term for ρ different from the one for ρθ (line 249)?

simonbyrne: Both work in this case, but this one is more correct: divergence operates on the contravariant components, and in the presence of topography the vertical covariant component may have a horizontal contravariant component.

simonbyrne:
bors try

szy21:
We need to add bubble3d_invariant to ci as well?

simonbyrne:
the plots triggered the out-of-memory!

simonbyrne:
bors try

simonbyrne:
bors try

examples/hybrid/bubble_2d.jl
```diff
@@ -327,3 +335,175 @@ Plots.png(
     joinpath(path, "energy.png"),
 )
 Plots.png(Plots.plot((Mass .- mass_0) ./ mass_0), joinpath(path, "mass.png"))
+
```
szy21, line 338: I guess we don't need this (from this line to the end)? Or do we need to change this file at all?

examples/hybrid/bubble3d_invariant.jl
```diff
@@ -0,0 +1,306 @@
...
+    )
+    @. dw -= vgradc2f(cp) / Ic2f(cρ)
+
+    cE = @. (norm(cuvw)^2) / 2 + Φ(coords.z)
+    @. duₕ -= hgrad(cE)
+    @. dw -= vgradc2f(cE)
+
+    # 3) potential temperature
+
+    @. dρθ -= hdiv(cuvw * Yc.ρθ)
```
szy21, line 253:
```suggestion
    @. dρθ -= hdiv(cuvw * cρθ)
```

PR #328, opened 2021-12-07
Switch (warped) slice models plot output to use triplots

jakebolewski:
So we can plot warped (topo) meshes, adds TriplotBase to dependencies


jakebolewski:
I've only tested the existing slice model setup but it (should) work with warped coordinates.  The output is not really distinguishable from what we have previously in the regular (stretched) grid space:
foo

jakebolewski:
bors try

PR #330, opened 2021-12-08
fix visual artifact in interpolated slice plots with topo

jakebolewski:
FIxes visual banding artifacts for triangulation when there are not shared vertex points under interpolation.

interp_error

![fixed_interp](https://user-images.githubusercontent.com/1157798/145154833-f1e3111f-617a-4233-b890-ac0348684d31.png


jakebolewski:
The default plot works fine it seems but we need to used closed uniform interpolation if you want something a bit nicer.

cc @akshaysridhar

PR #333, opened 2021-12-08
changing examples to consistently use FieldVectors

### LINKED ISSUE #272: Move to using `FieldVector` consistently in tests / examples
When variables are on different function spaces, we need to wrap them in an additional data structure to have a consistent interface for the ODE solve.  Currently we do that with a mix of `Fields` / RecursiveArray Vectors / and `FieldVector`s.  Switching to `FieldVector` would be simpiler and allow us to drop some of the current monkey patched method extensions for DiffEqBase and RecursiveArrayTools.

jakebolewski:
Closes #272

### DIFF
--- a/examples/column/ekman.jl
+++ b/examples/column/ekman.jl
@@ -1,6 +1,6 @@
 push!(LOAD_PATH, joinpath(@__DIR__, "..", ".."))
 
-import ClimaCore.Geometry, LinearAlgebra, UnPack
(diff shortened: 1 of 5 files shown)


jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
@simonbyrne  it looks like there is some perf hit moving from `ArrayPartition` to `FieldVectors` so it might be a good reference as we optimize the later.

PR #336, opened 2021-12-08
Add FD operator allocation tests

charleskawczynski:
After tracking allocations in TurbulenceConvection.jl, I was able to track down that most are coming from ClimaCore's FD operators. This PR adds several tests that highlight where we are doing well (0 allocations) and where we incur allocations. Tests (and `@test_broken`s) were added to indicate the allocations.

For the simple examples (but not all), I added tests for closures and when `getproperty` calls are inside the broadcasted expression, which occurs allocations but can be removed by hoisting the `getproperty` calls outside.

As @jakebolewski suggested, the next step is to analyze these cases with JET to find out what's going wrong.


test/fdspaces.jl
```diff
@@ -645,3 +645,398 @@ end
...
+        Ic = Operators.InterpolateF2C()
+        @. cz = ifelse(cx < cy, abs(Ic(fy)) * abs(Ic(fy)) * cϕ * cψ, 0) # Compile first
+        p = @allocated begin
+            @. cz = ifelse(cx < cy, abs(Ic(fy)) * abs(Ic(fy)) * cϕ * cψ, 0)
+        end
+        @test p == 1858176
+        @test_broken p == 0
+    end
+
+    println("##### Test allocs for nested expressions from real-world examples")
```
jakebolewski, line 1030: Can we remove the print output, or comment it out?

jakebolewski (dismissed):
It would be easier to work through these if they were defined in a separate file.

test/fdspaces.jl
charleskawczynski: Done.

charleskawczynski:
> were defined in a separate file.

Done.

jakebolewski:
Comment for future self:  we should constrain the PlusHalf / index types to just be Int's

charleskawczynski:
I don't remember exactly, but I thought we had example 11 with zero allocs, not sure if we got tricked by Revise or if I messed something up in the cleanup. Either way the fixes applied here are overall a nice improvement it seems (most easily seen in commit 2)

charleskawczynski:
> @charleskawczynski can you try one more variation removing the type 
> they might not be needed anymore

yeah, I’ll try continuing with this PR tomorrow. I’m interested to see what the test suite looks like now.

charleskawczynski:
bors try

charleskawczynski:
It turns out that the conditions leading to `PlusHalf`/`Int` types in `left_interor_window_idx`/`right_interor_window_idx` are not as simple as I originally thought (based on the results of the CI tests, which reveal contradicting cases), so I've reverted those changes. Now, we only have the added `has_boundary` check, which helps the compiler (verified using JET) and should be correct.

I'll tackle more of these, but lets merge in the tests and the one fix for now.

PR #339, opened 2021-12-09
See if types always match

charleskawczynski:
Running a smoke test with this to see if anything fails. This is a follow up to #336.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -2148,6 +2148,12 @@ function apply_stencil!(field_out, bc)
     lw = left_interor_window_idx(bc, space, lbw)
     ri = right_idx(space)
     rw = right_interor_window_idx(bc, space, rbw)
+    if !(typeof(ri) == typeof(rw))
+        error("oops")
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

src/Operators/finitedifference.jl
```diff
@@ -2148,6 +2148,12 @@ function apply_stencil!(field_out, bc)
     lw = left_interor_window_idx(bc, space, lbw)
     ri = right_idx(space)
     rw = right_interor_window_idx(bc, space, rbw)
+    if !(typeof(ri) == typeof(rw))
+        error("oops")
```
valeriabarra, line 2152: I am always in favor of more informative error messages :)

charleskawczynski: Haha, yeah, this is just to test if this condition is generally true or not. I'll be removing it :)

jakebolewski: this will always be true because the space is checked before the broadcasted expression is materialized

PR #345, opened 2021-12-10
ClimaCoreVTK writevtk for NamedTuples of time series fields

valeriabarra:
Currently we support named tuples for the single-snapshot version of the `writevtk` function. But not for the time series one.
I tried to fix this by overwriting the `writevtk` function so that we can output named tuples of field arrays in time series collections. But my attempt is not working at the moment and needs some more work. Any feedback would be appreciated, thanks.

### DIFF
--- a/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
+++ b/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
@@ -100,7 +100,7 @@ end
         basename::String,
         times,
         fields,
-        [ispace=vtk_space(fields)];
+        [ispace=vtk_space(first(fields))];
         vtkargs...
     )
 
@@ -123,6 +123,41 @@ function writevtk(
                 ispace;
                 vtkargs...,
             )
+        end
+    end
+end
+
+"""
+    writevtk(
+        basename::String,
+        times,
+        fields,
+        [ispace=vtk_space(first(first(fields)));
+        vtkargs...
+    )
+
+Write a sequence of NamedTuple fields `fields` at times `times` as a Paraview collection.
+"""
+function writevtk(
+    basename::String,
+    times,
+    fields::NamedTuple,
+    ispace = vtk_space(first(first(fields)));
+    vtkargs...,
+)
+    npad = ndigits(length(times); pad = 3)
+
+    for (key, val) in pairs(fields)
+        paraview_collection(basename) do pvd
+            for (n, time) in enumerate(times)
+                field = NamedTuple{(key,)}([val[n]])
+                pvd[time] = vtk_file(
+                    basename * "_" * string(n, pad = npad),
+                    field,
+                    ispace;
+                    vtkargs...,
+                )
+            end
         end
     end
 end
--- a/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
+++ b/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
@@ -147,10 +147,9 @@ function writevtk(
 )
     npad = ndigits(length(times); pad = 3)
 
-    for (key, val) in pairs(fields)
         paraview_collection(basename) do pvd
             for (n, time) in enumerate(times)
-                field = NamedTuple{(key,)}([val[n]])
+                field = NamedTuple(key => val[n] for (key, val) in pairs(fields))
                 pvd[time] = vtk_file(
                     basename * "_" * string(n, pad = npad),
                     field,
@@ -159,7 +158,6 @@ function writevtk(
                 )
             end
         end
-    end
 end
 
 
--- a/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
+++ b/lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
@@ -147,17 +147,17 @@ function writevtk(
 )
     npad = ndigits(length(times); pad = 3)
 
-        paraview_collection(basename) do pvd
-            for (n, time) in enumerate(times)
-                field = NamedTuple(key => val[n] for (key, val) in pairs(fields))
-                pvd[time] = vtk_file(
-                    basename * "_" * string(n, pad = npad),
-                    field,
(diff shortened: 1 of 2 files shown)


lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
```diff
@@ -100,11 +100,11 @@ end
         basename::String,
         times,
         fields,
-        [ispace=vtk_space(fields)];
+        [ispace=vtk_space(first(fields))];
         vtkargs...
     )
 
-Write a sequence of fields `fields` at times `times` as a Paraview collection.
```
jakebolewski, line 107: This format is specific to paraview

lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
```diff
@@ -127,5 +127,37 @@ function writevtk(
...
+        vtkargs...
+    )
+
+Write a sequence of NamedTuple fields `fields` at times `times` as a vtk collection.
+"""
+function writevtk(
+    basename::String,
+    times,
+    fields::NamedTuple,
+    ispace = vtk_space(first(first(fields)));
```
jakebolewski, line 145: it should be the first Field in the series

lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
```diff
@@ -127,5 +127,37 @@ function writevtk(
...
+    vtkargs...,
+)
+    npad = ndigits(length(times); pad = 3)
+
+    paraview_collection(basename) do pvd
```
jakebolewski, line 150: you need to create a paraview collection for each key value tuple pair in fields, so this should destructure the tuple and then pass the key value pairs to the previous defined function.

lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
simonbyrne: In that case, we should probably rename the function (`writepvd`?)

simonbyrne: we can do that in a separate PR.

jakebolewski: I think that might be less confusing, it looks like they are also pushing the new JSON series format, so that would be `write_vtk_series` or something similar.

simonbyrne:
Are you trying to write a `NamedTuple` of `Vector`s of `Field`s, or a `Vector` of `NamedTuple`s of `Field`s?

jakebolewski:
I think the issue here is that vorticity is a scalar field series, and so the default name is data.  One way to do this current is just to create a field with a name voriticity as an input instead, but I guess it would also be useful to add an optional name argument for scalar fields.

valeriabarra:
> I think the issue here is that vorticity is a scalar field series, and so the default name is data. One way to do this currently is to create a field with a name vorticity as an input instead, but I guess it would also be useful to add an optional name argument for scalar fields and be able to give a name to an anon field value when writing it out

initial_velocity_x

It's not only for scalar fields. Also for vector fields (see in this picture, I am showing velocity-x component and it shows as "data X". 
I saw we support named tuples so that the symbol is automatically used for the field name and it shows fine in Paraview, but we support it only for the single snapshot version at the moment (either for scalar or vector fields), not for the time collection one

valeriabarra:
> Are you trying to write a `NamedTuple` of `Vector`s of `Field`s, or a `Vector` of `NamedTuple`s of `Field`s?

The first one

lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
valeriabarra: I can put the docstring back as it was. Yeah, I was misled by the name of the function and automatically assumed that it wasn't Paraview-specific

valeriabarra:
bors try

valeriabarra:
I fixed it and adjusted the test for a scalar field series.

valeriabarra:
@simonbyrne did you have any chance to take a look at this?

lib/ClimaCoreVTK/src/ClimaCoreVTK.jl
```diff
@@ -127,5 +127,40 @@ function writevtk(
...
+                field = NamedTuple{(key,)}([val[n]])
+                pvd[time] = vtk_file(
+                    basename * "_" * string(n, pad = npad),
+                    field,
+                    ispace;
+                    vtkargs...,
+                )
+            end
+        end
+    end
```
simonbyrne, line 162: I don't think you want the `for` loop on the outside (that will create multiple collections, but they will overwrite each other as they have the same name). Do you want something like this:

```suggestion
        paraview_collection(basename) do pvd
            for (n, time) in enumerate(times)
                field = NamedTuple(key => val[n] for (key, val) in pairs(fields))
                pvd[time] = vtk_file(
                    basename * "_" * string(n, pad = npad),
                    field,
                    ispace;
                    vtkargs...,
                )
            end
        end
```
?

valeriabarra: Doh! I had tried the loop in the inside, but for some reason when I was testing this I thought I had gotten the expected behavior. Thanks so much!

valeriabarra:
@simonbyrne , I added a test for vector series on the sphere as well. Let me know if this is fully reviewed/approved and I can go ahead and squash commits. Thank you very much

simonbyrne (approved):

valeriabarra:
bors try

simonbyrne:
@valeriabarra did you want to squash this?

valeriabarra:
> @valeriabarra did you want to squash this?

I did. I separated into what I considered to be semantically different. One commit modified the library components. One added the tests (and applied formatter). There was just yours in between, that I kept to preserve your authorship contribution.
We can reduce further if we prefer.

Issue #9, opened 2021-04-15
Disconnect mesh construction floating point type from storage type


jkozdon:
One of the things that we saw before is that there was some benefit to building the operators in a higher precision and then casting down. Same for the metric calculations.

Would be nice to support this here.

kpamnany:
Closed by https://github.com/CliMA/ClimaCore.jl/pull/349.

Issue #323, opened 2021-12-02
Distributed topology objects


simonbyrne:
- divide domain into n chunks, would contain ClimaComm context

Consider the 1D case:
```
function DistributedIntervalTopology(mesh::IntervalMesh, commctx::ClimaComms.Context)
  # how many procs?
  # what pid am i?
  # do linear partition, figure out indices, neighbors
  ...
end
```

Can we just use `SpectralElementSpace1D`, or do we need a distributed version?
 

Need an interface to say:
- what are the element indices that need to get send to each process (stuff to copy to send buffer)
- what are the indices of the ghost elements belonging to each process (stuff to copy from receive buffer)

simonbyrne:
1. partition the mesh elements: what should the topology see?



key question: what information does a process need to construct the topology on that process?
 - mesh (global)
 - global element ordering / partitioning
   - option 1: lookup for `elem => procid`
   - option 2: give a `elemorder` and a sorted partitioning

storage: "local element ordering"
 - `1:nlocalelems`: local elements
 - `ghostelem_index[i]:ghostelem_index[i+1]-1`: ghost elements from process `i`

e.g. global element numbers:

1 2  | 5
3 4 | 6
----+-
7 8 | 9


proc 1:
```
local_elems = [1,2,3,4]
send_elems = [2,4,3,4,4]; send_elems_index = [1,1,3,5,6]
local_send_elems = [2,4,3,4,4]
ghost_elems=[5,6, 7,8, 9]; ghostelem_index = [1,1,3,5,6]

```
proc 2: 
```
local_elems = [5,6]; 
send_elems = [5,6,6,6]; send_elems_index = [1,3,3,4,5]
local_send_elems = [1,2,2,2]
ghost_elems=[2,4, 8, 9]; ghostelem_offset  = [1,3,3,4,5]
```
proc 3: 
```
local_elems = [7,8]
send_elems = [7,8,8,8]; send_elems_index = [1,3,4,4,5]
local_send_elems = [1,2,2,2]
ghost_elems=[3,4, 6, 9];  ghostelem_offset  = [1,3,4,4,5]
```
proc 4: 
```
local_elems = [9]; 
send_elems = [9,9,9]; send_elems_index = [1,2,3,4,4]
local_send_elems = [1,1,1]
ghost_elems=[4,6,8];  ghostelem_offset  = [1,2,3,4]`
```
 - `send_elem` should probably store the local element number?
 - `ghost_elem`


[...]

PR #349, opened 2021-12-14
Refactor mesh and topology code

simonbyrne:
This is a large refactor and simplification of the Meshes and Topologies modules:

- Meshes are now all very lightweight objects, currently supporting interval, rectangles and cubed spheres
  -  A `RectangleMesh` is now explicitly a product of two `IntervalMesh`s
  - `RectangleMesh` replaces both `EquispacedRectangleMesh` and `TensorProductMesh`
  - There are various `AbstractCubedSphereMesh`es, which now encode the warping functions
- Meshes now have a relatively well-specified API: basically the key function is `opposing_face` (all other properties are computed from this).
   - It internally now uses Base Julia `CartesianIndex` machinery, which gets rid of a lot of manual indexing we had to do previously.
   - Cubed spheres now use a different panel ordering (it matches the S2Geometry library and Oceananigans).
   - Provide convenience functions for computing face and vertex connectivity matrices.
- There is now only 1 2D Topology object, called (unimaginatively) `Topology2D`. This is a heavier object, in that it stores all the interior face pairs, unique vertices and boundary faces, but is the same for all meshes.
  - It has an option for different element orderings (e.g. to use space-filling curves)
  - This will allow adding a new `DistributedTopology2D` type (#323), which should be able to use the same mesh interface.


src/Topologies/topology2d.jl
```diff
@@ -0,0 +1,129 @@
...
+    Topology2D(mesh::AbstractMesh2D, elemorder=Mesh.elements(mesh))
+
+
+
+"""
+struct Topology2D{M <: Meshes.AbstractMesh{2}, EO, OI, BF} <: AbstractTopology
+    "mesh on which the topology is constructed"
+    mesh::M
+    "`elemorder[e]` should give the `e`th element "
+    elemorder::EO
```
jakebolewski, line 13: In a follow up PR I think it would be good to make the element ordering / order index an explicit type we can dispatch off of as part of the topology instead of two separate arguments.

simonbyrne:
bors try

simonbyrne:
Need to fix plotting, but otherwise should be ready to review.

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
A few final renames: `RectangleMesh` is now `RectilinearMesh`: see https://en.wikipedia.org/wiki/Regular_grid

simonbyrne:
bors try

simonbyrne:
Okay, I think this is ready to go. Any final comments before i squash?

src/Geometry/coordinates.jl
```diff
@@ -126,13 +157,23 @@ Base.:(*)(x::Number, p::AbstractPoint) = p * x
 Base.:(/)(p::T, x::Number) where {T <: AbstractPoint} =
     unionalltype(T)((components(p) / x)...)
 
+function Base.lerpi(
```
charleskawczynski, line 160: I don't see any docs in this in Julia `Base`, but can we add something here?

simonbyrne:
```suggestion
# we add our own method to this so that `BigFloat` coordinate ranges are computed accurately.
function Base.lerpi(
```

src/Domains/Domains.jl
```diff
@@ -1,56 +1,64 @@
...
 } where {FT}
     coord_min::CT
     coord_max::CT
-    boundary_tags::B
+    boundary_names::B
 end
 
+isperiodic(domain::IntervalDomain) = isnothing(domain.boundary_names)
+boundary_names(domain::IntervalDomain) =
+    isperiodic(domain) ? () : domain.boundary_names
```
jakebolewski, line 36: should this return `nothing` for periodic domains?

src/Geometry/coordinates.jl
```diff
@@ -1,9 +1,40 @@
 abstract type AbstractPoint{FT} end
 
+"""
+    float_type(T)
+
+Return the floating point type backing `T`: `T` can either be an object or a type.
+"""
+float_type(::Type{<:AbstractPoint{FT}}) where {FT} = FT
+float_type(::AbstractPoint) where {FT} = FT
```
jakebolewski, line 9: is this the new domain "undertype"? we should probably make it part of the public API then

docs/src/api.md
```diff
@@ -18,25 +18,72 @@ Geometry.Cartesian12Vector
...
 ```@docs
+Domains.AbstractDomain
 Domains.IntervalDomain
 Domains.RectangleDomain
+Domains.SphereDomain
+```
+
+### Interfaces
+```@docs
+Domains.boundary_names
```
jakebolewski, line 31: Add `Domains.float_type`?

jakebolewski (approved):

jakebolewski:
Closes https://github.com/CliMA/ClimaCore.jl/issues/9

src/Domains/Domains.jl
simonbyrne: no, an empty set makes more sense (it's the collection of all boundary names)

simonbyrne: maybe it should be `unique(domain.boundary_names)` in the other case though.

jakebolewski: if it is a set interpretation that would be a good change, or you could just check upon construction that the names are unique which seems better / less confusing

src/Geometry/coordinates.jl
simonbyrne: i'd like to leave that to another PR

src/Domains/Domains.jl
simonbyrne: It is now unique.

PR #350, opened 2021-12-14
Add ntuple field perf tests

charleskawczynski:


jakebolewski (approved):

jakebolewski:
Have you checked with an optimization test if we can infer these large tuples correctly?

charleskawczynski:
> Have you checked with an optimization test if we can infer these large tuples correctly?

Yeah, JET says everything is inferrable, but we're still somehow incurring allocations. What we had to do in ClimateMachine was propagate indexes down via `Val`. It looks like this could be affecting the non-FD stencils, too

charleskawczynski:
Hmm, that was when I commented everything else out aside from the new test. Need to take a second look

charleskawczynski:
Ah, I redefined a field, which had the same effect as redefining operators. Everything looks good.

PR #351, opened 2021-12-15
fix coordinates.jl for the 3d sphere mesh

szy21:
co-authored-by: @simonbyrne 

This PR fixes coordinates.jl for the 3d sphere mesh. It also adds a simple test for the sphere metric terms.
Will squash and rebase once it is approved.

### DIFF
--- a/src/Geometry/coordinates.jl
+++ b/src/Geometry/coordinates.jl
@@ -67,12 +67,16 @@ Cartesian123Point(pt::Cartesian13Point{FT}) where {FT} =
     Cartesian123Point(pt.x1, zero(FT), pt.x3)
 
 @pointtype LatLongPoint lat long
+@pointtype LatLongZPoint lat long z
 
 product_coordinates(xp::XPoint, yp::YPoint) = XYPoint(promote(xp.x, yp.y)...)
 product_coordinates(xp::XPoint, zp::ZPoint) = XZPoint(promote(xp.x, zp.z)...)
 
 product_coordinates(xyp::XYPoint, zp::ZPoint) =
     XYZPoint(promote(xyp.x, xyp.y, zp.z)...)
+
+product_coordinates(latlongp::LatLongPoint, zp::ZPoint) =
+    LatLongZPoint(promote(latlongp.lat, latlongp.long, zp.z)...)
 
 component(p::AbstractPoint{FT}, i::Symbol) where {FT} = getfield(p, i)::FT
 component(p::AbstractPoint{FT}, i::Integer) where {FT} = getfield(p, i)::FT
--- a/test/sphere_metric_terms.jl
+++ b/test/sphere_metric_terms.jl
@@ -1,5 +1,5 @@
 using LinearAlgebra, IntervalSets, UnPack
-import ClimaCore: Domains, Topologies, Meshes, Spaces
+import ClimaCore: Domains, Topologies, Meshes, Spaces, Geometry
 import ClimaCore.Meshes: EquiangularSphereWarp, Mesh2D
 
 using Test
@@ -9,7 +9,6 @@ using Test
     radius = FT(3)
     ne = 4
     Nq = 4
-    Nqh = 7
     domain = Domains.SphereDomain(radius)
     mesh = Mesh2D(domain, EquiangularSphereWarp(), ne)
     grid_topology = Topologies.Grid2DTopology(mesh)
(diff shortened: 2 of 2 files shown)


szy21:
bors try

szy21:
bors try

src/Geometry/coordinates.jl
```diff
@@ -67,13 +67,17 @@ Cartesian123Point(pt::Cartesian13Point{FT}) where {FT} =
...
+@pointtype LatLongZPoint lat long z
 
 product_coordinates(xp::XPoint, yp::YPoint) = XYPoint(promote(xp.x, yp.y)...)
 product_coordinates(xp::XPoint, zp::ZPoint) = XZPoint(promote(xp.x, zp.z)...)
 
 product_coordinates(xyp::XYPoint, zp::ZPoint) =
     XYZPoint(promote(xyp.x, xyp.y, zp.z)...)
 
+product_coordinates(latlongp::LatLongPoint, zp::ZPoint) =
+    LatLongZPoint(promote(latlongp.lat, latlongp.long, zp.z)...)
```
valeriabarra, line 79: What does exactly a `LatLongZPoint` represent? Like a point in spherical coordinates, where the third dimension is along the radial direction? Should it be called `LatLongRPoint` then?

sriharshakandala: Probably, elevation measure from planet surface!

test/sphere_metric_terms.jl
```diff
@@ -18,3 +17,33 @@ using Test
...
+
+    horzdomain = Domains.SphereDomain(radius)
+    horzmesh = Meshes.Mesh2D(horzdomain, Meshes.EquiangularSphereWarp(), helem)
+    horztopology = Topologies.Grid2DTopology(horzmesh)
+    quad = Spaces.Quadratures.GLL{Nq}()
+    horzspace = Spaces.SpectralElementSpace2D(horztopology, quad)
+
+    hv_center_space =
+        Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_center_space)
+    hv_face_space = Spaces.FaceExtrudedFiniteDifferenceSpace(hv_center_space)
```
valeriabarra, line 45: do you need this `hv_face_space` here? It seems that for the sum below you only use the center space

src/Geometry/coordinates.jl
szy21: I think z makes sense, as it represents height above the surface, but I'm fine either way. @simonbyrne What do you think?

test/sphere_metric_terms.jl
szy21: Thanks! I'll remove it.

src/Geometry/coordinates.jl
simonbyrne: Yes, it's the distance from the surface of the sphere: we could have used radius (i.e. distance from the origin, but that would be somewhat less convenient, and would mean that the box model has a different vertical coordinate. I tried to lay this out in #81.

test/sphere_metric_terms.jl
```diff
@@ -18,3 +17,33 @@ using Test
...
+
+    @test sum(ones(hv_center_space)) ≈ 4pi * radius^2 * (zlim[2] - zlim[1]) rtol =
+        1e-3
```
simonbyrne, line 48: Add a comment to explain the formula
```suggestion
    # "shallow" spherical shell: volume = surface area * height
    @test sum(ones(hv_center_space)) ≈ 4pi * radius^2 * (zlim[2] - zlim[1]) rtol =
        1e-3
```

test/sphere_metric_terms.jl
```diff
@@ -18,3 +17,33 @@ using Test
...
+    horzmesh = Meshes.Mesh2D(horzdomain, Meshes.EquiangularSphereWarp(), helem)
+    horztopology = Topologies.Grid2DTopology(horzmesh)
+    quad = Spaces.Quadratures.GLL{Nq}()
+    horzspace = Spaces.SpectralElementSpace2D(horztopology, quad)
+
+    hv_center_space =
+        Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_center_space)
+    hv_face_space = Spaces.FaceExtrudedFiniteDifferenceSpace(hv_center_space)
+
+    @test sum(ones(hv_center_space)) ≈ 4pi * radius^2 * (zlim[2] - zlim[1]) rtol =
```
sriharshakandala, line 47: It looks like shallow water approximation is being used here. May be, bumping up R relative to \Delta z should help!

simonbyrne (approved):
Thanks! Squash the commits once you're happy with it (or I can do it if you like)

test/sphere_metric_terms.jl
sriharshakandala: Looks good to me.

src/Geometry/coordinates.jl
valeriabarra: ok, but if one has in mind the usual picture for spherical coordinates (see for instance this one from Wiki) then I think the `z` would be confusing as it would mean something parallel to the global z-axis, which is not what this represents
image

test/sphere_metric_terms.jl
valeriabarra: "hollow sphere" or just "spherical shell"?

test/sphere_metric_terms.jl
szy21: Thanks! I have changed R to 128 so it's closer to the ratio of the earth (R~6400km and z~50km).

test/sphere_metric_terms.jl
szy21: I think "spherical shell" is ok: https://en.wikipedia.org/wiki/Spherical_shell

src/Geometry/coordinates.jl
szy21: Shall we move the discussion to #81 ? I think it will affect many things, not only this PR.

test/sphere_metric_terms.jl
valeriabarra: Yes, but it is more appropriate to specify thin in this case, otherwise the formula for its volume is different. Thanks

szy21: You mean "thin" instead of "shallow"? We use shallow atmosphere and deep atmosphere assumptions a lot in the literature. But if you think "thin" is better I can change it.

simonbyrne: Yes, I think "shallow atmosphere" and "deep atmosphere" correspond to what wikipedia calls thin.

szy21: I'll stick to it and use `"shallow atmosphere" spherical shell` then. That would be more clear to atmosphere science people I think.

PR #352, opened 2021-12-15
Copyright caltech

simonbyrne:
CliMA isn't an org that can hold copyright, so it should be Caltech.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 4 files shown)


valeriabarra (approved):

simonbyrne:
Maybe we should paste the whole license in?:
https://www.apache.org/licenses/LICENSE-2.0.txt

simonbyrne:
Reading the actual license, it looked like we were applying it incorrectly. I have updated it accordingly.

Issue #354, opened 2021-12-15
Add the ability to plot range slices from 3D sphere models


jakebolewski:
Ex. plot along a given longitude between a given latitude range

PR #358, opened 2021-12-15
Add the 3d deformation flow test case

szy21:
Co-authored-by: @Zhengyu-Huang @jiahe23 
Still needs
- [x] add plots
- [x] hook up CI
- [ ] run a high resolution simulation and make plots
- [x] squash and rebase

### DIFF
--- a/examples/3dsphere/deformation_flow.jl
+++ b/examples/3dsphere/deformation_flow.jl
@@ -1,0 +1,264 @@
+push!(LOAD_PATH, joinpath(@__DIR__, "..", ".."))
+
+using Test
+using StaticArrays, IntervalSets, LinearAlgebra, UnPack
+
+import ClimaCore:
+    ClimaCore,
+    slab,
+    Spaces,
+    Domains,
+    Meshes,
+    Geometry,
+    Topologies,
+    Spaces,
+    Fields,
+    Operators
+using ClimaCore.Geometry
+
+using Logging: global_logger
+using TerminalLoggers: TerminalLogger
+using OrdinaryDiffEq: ODEProblem, solve, SSPRK33
+
+global_logger(TerminalLogger())
+
+# 3D deformation flow (DCMIP 2012 Test 1-1)
+# Reference: http://www-personal.umich.edu/~cjablono/DCMIP-2012_TestCaseDocument_v1.7.pdf, Section 1.1
+
+const R = 6.37122e6 # radius
+const grav = 9.8 # gravitational constant
+const R_d = 287.058 # R dry (gas constant / mol mass dry air)
+const z_top = 1.2e4 # height position of the model top
+const p_top = 25494.4 # pressure at the model top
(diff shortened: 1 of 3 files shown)


examples/3dsphere/deformation_flow.jl
```diff
@@ -0,0 +1,214 @@
...
+
+    ∂ = Operators.DivergenceF2C(
+        bottom = Operators.SetValue(Geometry.WVector(0.0)),
+        top = Operators.SetValue(Geometry.WVector(0.0)),
+    )
+
+    hdiv = Operators.Divergence()
+
+    @. dydt.ρq1   = -∂(If(y.ρq1 .* w))
+    @. dydt.ρq1  -= hdiv(y.ρq1 .* uₕ)
```
szy21, line 169: @simonbyrne Is this correct for scalar advection? uh and w are physical velocities and are on cell centers.

szy21:
It runs but the result seems wrong. We would probably need some visualization (#354) for debugging.

jiahe23 (commented):
mod() function modifies -180~0 to 180~360 and leave 0~180 unchanged. why was it replace by +180?

szy21:
> mod() function modifies -180~0 to 180~360 and leave 0~180 unchanged. why was it replace by +180?

I was just trying to make it exactly the same as the test case in the paper (so longitude goes 0-360 rather than 180-360, 0-180). For a zonally symmetric simulation this probably doesn't matter, I'm not entirely sure though.

szy21:
bors try

examples/3dsphere/deformation_flow.jl
```diff
@@ -0,0 +1,262 @@
...
+    dt = dt,
+    saveat = dt,
+    progress = true,
+    adaptive = false,
+    progress_message = (dt, u, p, t) -> t,
+)
+
+q1_error =
+    norm(sol.u[end].ρq1 ./ ρ.(coords.z) .- y0.ρq1 ./ ρ.(coords.z)) /
+    norm(sol.u[end].ρq1 ./ ρ.(coords.z))
```
jiahe23, line 214: how about dividing the differences by the initial states?

jiahe23: same for the other tracers

szy21: done:)

jiahe23 (approved):

szy21:
bors try

valeriabarra:
Nice job! Should we add documentation of the formulation of this test case or provide a reference for where it is described?

szy21:
> Nice job! Should we add documentation of the formulation of this test case or provide a reference for where it is described?

Good idea. Just added it.

szy21:
bors try-

szy21:
bors try

szy21:
The artifacts look good. Thanks @jakebolewski for the plotting function!

Issue #360, opened 2021-12-16
Plot vertical levels of a spherical shell


simonbyrne:

jakebolewski:
Is this different than https://github.com/CliMA/ClimaCore.jl/issues/354?  I was thinking to start just say `plot(sphere, level=N)` and go from there, we could also have a `Fields.level_field()` function that would return a horizontal shell field at a given level

jakebolewski:
closed by #381

valeriabarra:
Just dropping an example of a spherical level vtk output here for reference (baroclinic wave test case, level = 3, w component plotted, final simulation time)
![w_t_end_level_3](https://user-images.githubusercontent.com/39932030/148141626-2597ca45-ec5b-431c-8c35-d6ba86fff5c2.png)

PR #363, opened 2021-12-17
add level interface function

simonbyrne:
It's a bit of a kludge, but it seemed  like the easiest way to handle #360.

### DIFF
--- a/src/ClimaCore.jl
+++ b/src/ClimaCore.jl
@@ -1,6 +1,7 @@
 module ClimaCore
 
 include("interface.jl")
+include("Utilities/Utilities.jl")
 include("RecursiveApply/RecursiveApply.jl")
 include("DataLayouts/DataLayouts.jl")
 include("Geometry/Geometry.jl")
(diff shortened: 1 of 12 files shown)


jakebolewski:
Isn't the space incorrect doing it this way?  I think you would have to level set the local geometry object as well and re-build a pure horizontal space, otherwise the coordinates will be incorrect.  If this is for plotting we only need the datalyout inferface.

simonbyrne:
Yes, I need to think a bit more about how we might use this, so it should be considered experimental for now. At the moment, the main purpose is to be able to plot things.

jakebolewski:
it's kind of weird because as soon as you take a level the global geometry changes as well, or else if the domain is not a pure 2D shell (ex with topography) you have some form of warped surface

simonbyrne:
Yeah, this is the same problem for boundary fields (#325).

jakebolewski:
It will come also in being able to compute derived level fields of ex. constant pressure

jakebolewski (approved):
LGTM (addresses by comments above), I guess i"m wondering how we will support this interface if / when we have nodal elements in the vertical?

simonbyrne:
Use two arts `level(field, v, k)`?

PR #369, opened 2021-12-20
Add plot artifact output generation to CI

jakebolewski:


jakebolewski:
Unfortunately for now it is not very usable on Github compared to Buildkite  (need to download a ZIP)

See: https://github.com/actions/upload-artifact/issues/14

jakebolewski:
Example: https://github.com/CliMA/ClimaCore.jl/actions/runs/1603869492

We need to setup some libraries on the cluster for this to work in general

PR #377, opened 2021-12-22
Fix product metric terms

simonbyrne:
The matrix was transposed, which caused strange behavior in the rotation test case.

Co-authored by: Zhaoyi Shen <pkuszy@gmail.com>


jakebolewski:
GIt blame points to me, I can never remember if the constructor assumes column major or row major ordering :-/

PR #379, opened 2021-12-22
add the nonhydrostatic gravity wave test case

szy21:
co-authored-by: @simonbyrne 

Still needs:
- [ ] Run a 4-day simulation and visualize the results (require slice along a latitude)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -445,6 +445,17 @@ steps:
       config: cpu
       queue: central
       slurm_ntasks: 1
+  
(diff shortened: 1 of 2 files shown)


szy21:
It will take roughly 12 hours walltime for a 4-day simulation (to compare with the literature) with the current coarse resolution. We may need to do it implicitly (for both gravity wave and sound wave?) for a faster turnaround, depending on how important this test case is.

szy21:
bors try

jiahe23 (approved):

PR #381, opened 2021-12-22
add extruded cubed sphere level plot

jakebolewski:
follow up to #363, this should get some default viz output going


jakebolewski:
CC @szy21

simonbyrne (approved):

PR #386, opened 2021-12-23
Fix hyperviscocity in 3D bubble case

jiahe23:
This PR corrects the u_h hyperdiffusion.


jiahe23:
@simonbyrne  https://github.com/CliMA/ClimaCore.jl/blob/main/examples/hybrid/bubble_3d.jl#L189
Should this also be corrected in the same way?

szy21:
cc @akshaysridhar

szy21 (approved):

PR #390, opened 2021-12-28
Explicitly return nothing & splt union method w convert

charleskawczynski:
I was [Cthulhu-ing](https://github.com/JuliaDebug/Cthulhu.jl) through some TurbulenceConvection code and saw that it wasn't able to infer some ClimaCore methods, in particular:
 - Overloaded `Base.copyto!` for `VF` (and potentially others), which seems to be due to the `Union` definition
 - A edge case bug in `set_struct!` (all explicitly `return nothing` except the "leaf" diagonal case)
 - and `setidx!`, which can't seem to be inferred because the resulting output type is equal to the input type (which currently left generic). There may be a different preferred way to do this, but I figured the proposed fix (returning `nothing`) would be simplest for now since we don't ever use the returned value.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -2073,12 +2073,10 @@ Base.Broadcast._broadcast_getindex_eltype(
 ) = eltype(bc)
 
 # check that inferred output field space is equal to dest field space
-@noinline inferred_stencil_spaces_error(
-    dest_space_type::Type,
-    result_space_type::Type,
-) = error(
-    "dest space `$dest_space_type` is not the same instance as the inferred broadcasted result space `$result_space_type`",
+inferred_stencil_spaces_error(::Type{A}, ::Type{B}) = error(
+    "dest space `$A` is not the same instance as the inferred broadcasted result space `$B`",
 )
+inferred_stencil_spaces_error(::T, ::T) where {T} = nothing
 
 function Base.Broadcast.materialize!(
     ::Base.Broadcast.BroadcastStyle,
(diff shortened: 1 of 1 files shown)


src/Operators/finitedifference.jl
```diff
@@ -2040,13 +2040,13 @@ end
...
     field_data[idx] = val
-    val
+    return nothing
 end
 
 function setidx!(bc::Fields.FaceFiniteDifferenceField, idx::PlusHalf, val)
     field_data = Fields.field_values(bc)
     field_data[idx.i + 1] = val
-    val
+    return nothing
```
jakebolewski, line 2049: we should return the bc variable here instead, to align with `Base.setindex!`

src/DataLayouts/struct.jl
```diff
@@ -227,6 +227,7 @@ end
     offset,
 ) where {S}
     array[offset + 1] = val
+    return nothing
```
jakebolewski, line 230: we should probably return the container value (similar to `setindex!` instead of `nothing`, but good catch.  I think this is fine for now.

src/DataLayouts/broadcast.jl
```diff
@@ -377,9 +377,21 @@ end
 
 # inline inner column(::DataColumn) copy
 @inline function Base.copyto!(
-    dest::VF{S},
-    bc::Union{VF{S, A}, Base.Broadcast.Broadcasted{VFStyle{A}}},
-) where {S, A}
+    dest::T,
+    bc::VF{S, A},
+)::T where {S, A, T <: VF{S}}
```
jakebolewski, line 382: I am skeptical about this one, but maybe the force inline is hiding the implicit specialization.

Is the same result achieved keeping the Union but writing it this way where the return type is declared to be the `dest type`, ex: `copyto!()::T`?

jakebolewski: One TODO is to make `VFStyle{S, A}` as well (to mirror the datalayout)

charleskawczynski:
Ok, I'm going to split this up a bit.

src/DataLayouts/struct.jl
charleskawczynski: So, just `return val` here and in the `else` block above?

src/DataLayouts/broadcast.jl
jakebolewski: Reminder: we looked at this in a call and it actually isn't an issue?

charleskawczynski:
Doh, overwrote this branch accidentally, I'll close and we can revisit

PR #392, opened 2021-12-31
start solid body rotation implicit

Zhengyu-Huang:


szy21:
@Zhengyu-Huang Could we close this?

PR #394, opened 2022-01-01
change baroclinic wave setup

szy21:
This PR changes the default polynomial order for baroclinic wave and balanced flow to 4 and dt to 5s. It also includes some cleanup. It actually seems less stable than polynomial order 3, but is more correct when using hyperdiffusion. Still needs to run a longer simulation. cc @bischtob

### DIFF
--- a/examples/3dsphere/baroclinic_wave.jl
+++ b/examples/3dsphere/baroclinic_wave.jl
@@ -99,6 +99,8 @@ elseif test_name == "balanced_flow"
     δv(λ, ϕ, z) = 0.0
     const κ₄ = 0.0
 end
+uu(λ, ϕ, z) = u(ϕ, z) + δu(λ, ϕ, z)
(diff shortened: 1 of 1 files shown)


szy21:
bors try

jiahe23 (approved):

examples/3dsphere/baroclinic_wave.jl
```diff
@@ -138,7 +140,7 @@ function pressure(ρ, e, normuvw, z)
 end
 
 # set up 3D domain - doubly periodic box
```
valeriabarra, line 142: I know this has been merged already but I was going to through it recently to test visualization. What does "doubly periodic box" mean here (since this is a spherical shell)? Was this a copy/paste mistake?

szy21: Yes, it is. I will change it next round. Thanks!

PR #399, opened 2022-01-03
ClimaCoreVTK: rename test output files to avoid overwrite

valeriabarra:
I think the output files for the 3D hybrid case had typos. This caused overwriting the 2D test outputs

### DIFF
--- a/lib/ClimaCoreVTK/test/runtests.jl
+++ b/lib/ClimaCoreVTK/test/runtests.jl
@@ -143,12 +143,12 @@ end
     fspace = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace)
     cspace = Spaces.CenterExtrudedFiniteDifferenceSpace(fspace)
     writevtk(
-        joinpath(dir, "hybrid2d_point"),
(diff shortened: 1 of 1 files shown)


valeriabarra:
bors try

jakebolewski:
good catch, it probably would be better to test that the files exist at the given path (and later we can add plot output of the files).

valeriabarra:
bors try

lib/ClimaCoreVTK/test/runtests.jl
```diff
@@ -54,6 +56,7 @@ dir = mktempdir()
         U[t] = u
     end
     writevtk(joinpath(dir, "sphere_vector_series"), times, (U = U,))
+    @test isfile(joinpath(dir, "sphere_vector_series.pvd"))
```
simonbyrne, line 59: can also check that the individual `.vtu` files also exist

valeriabarra: True. I thought this way was more direct and did not require writing for loops for each test. And I thought that the `.pvd` is created at the end of the collection, so the collection itself was successfully created if the pvd file exists

PR #404, opened 2022-01-05
add docs for ClimaCore libs

simonbyrne:


jakebolewski:
maybe it would be a good time to move the build to the cluster

simonbyrne:
> maybe it would be a good time to move the build to the cluster

I think it's fine for now: if we add more tutorials we can re-evaluate.

simonbyrne:
Since moving the plots, the graphics aren't rendering in the tutorial. I tried adding `using ClimaCorePlots`, but that didn't seem to fix it.

simonbyrne:
oh it's not running at all!

jakebolewski:
I've also noticed in the past that nothing is highlighted correctly :-/

simonbyrne:
okay, the tutorial hadn't been running since we changed some of the vector stuff

I've hopefully fixed that, and enabled strict mode on the example blocks, so hopefully should be okay now

jakebolewski:
awesome, looks much better. I'll add the docs for the plotting / makie stuff to get started in another PR

jakebolewski (approved):

PR #407, opened 2022-01-07
Add output viz tests for vtk data & test that the generated files can be read by paraview

jakebolewski:
Output is added as a summary of each run (similar to all the other CI assets).  We'll have to play around with the camera to make it reasonable for both 2D and 3D (or have a cli arg) but this should get us started.  Tests that the output files can be read by paraview and the pvpython script is useful for debugging.

https://github.com/CliMA/ClimaCore.jl/actions/runs/1665961301

Ex:
sphere vtu

### DIFF
--- a/lib/ClimaCoreVTK/test/runtests.jl
+++ b/lib/ClimaCoreVTK/test/runtests.jl
@@ -4,8 +4,8 @@ using IntervalSets
 import ClimaCore:
     Geometry, Domains, Meshes, Topologies, Spaces, Fields, Operators
 
-
-dir = mktempdir()
+OUTPUT_DIR = mkpath(get(ENV, "CI_OUTPUT_DIR", tempname()))
+mkpath(joinpath(OUTPUT_DIR, "series"))
 
 @testset "sphere" begin
     R = 6.37122e6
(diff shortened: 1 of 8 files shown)


valeriabarra:
This is nice, but yes, as you mentioned, the camera for the 2D tests should point not in the YZ-plane as it currently is, otherwise if we also want to visually inspect the plots (other than checking if paraview can open them) you can't really see anything. Is it easy to set a default camera view?

jakebolewski:
Yes I only looked at getting the sphere to work.  See the script

lib/ClimaCoreVTK/test/runtests.jl
```diff
@@ -4,8 +4,14 @@ using IntervalSets
...
 
+running_CI() = !isempty(get(ENV, "CI", ""))
 
-dir = mktempdir()
+dir = if running_CI()
+    mkdir(joinpath(@__DIR__, "output"))
+else
+    mktempdir()
+end
+mkdir(joinpath(dir, "series"))
```
simonbyrne, line 14: It is sometimes useful to save the results when running the test locally as well. Maybe we could do something like
```suggestion
dir = get(ENV, "CI_OUTPUT_DIR", mktempdir())
```
and then just set the environment variable in the .yml file.

jakebolewski: Good suggestion, I'll also these for the other artifacts.

jakebolewski:
changed the default view to XZ as it is the most useful.

Issue #346, opened 2021-12-10
Redefining operators in the same scope ressults in allocations


charleskawczynski:
See [this comment](https://github.com/CliMA/ClimaCore.jl/blob/main/test/fdspaces_opt_examples.jl#L142).

jakebolewski:
I think the issue is that what you linked to is an alias for an abstract type (unionall vector for any FT), if it was a concrete bitstype the allocation should not happen I'm guessing.

This is why it is probably good to be wary of non-concrete aliases like this

charleskawczynski:
Ah, whoops, well, I think [this test](https://github.com/CliMA/ClimaCore.jl/blob/main/test/fdspaces_opt_examples.jl#L180-L214) doesn't use the union all type

jakebolewski:
I guess i can see why one would potentially need to re-define operators with boundary conditions, but ones like these might be better to define once as a module const and use the same operator throughout the code.

I guess my point is maybe we should prioritize where re-defining the operator is not possible.

charleskawczynski:
I completely agree, I don't think this is a long-term issue for TurbulenceConvection.jl-- we should be able to easily fix anywhere that we hit this. I figure it's just helpful to keep the issue open for now so that people are aware.

PR #408, opened 2022-01-08
Add a FD perf test

charleskawczynski:
This PR adds a (broken) allocation test to the FD opt examples test suite. I'll see if I can Cthulhu in to find what's going wrong but may need help. cc @jakebolewski.

### DIFF
--- a/test/fdspaces_opt_examples.jl
+++ b/test/fdspaces_opt_examples.jl
@@ -16,6 +16,9 @@ Base.@propagate_inbounds Base.getindex(
     i::Integer,
 ) = Base.getproperty(field, i)
 
+a_bcs(::Type{FT}, i::Int) where {FT} =
+    (; bottom = Operators.SetValue(FT(0)), top = Operators.Extrapolate())
+
 if VERSION >= v"1.7.0"
     @testset "FD operator allocation tests" begin
         FT = Float64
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Looks like all we needed was to not limit recursion on `axes_args`.

jakebolewski:
Is this just another form of https://github.com/CliMA/ClimaCore.jl/issues/346?

charleskawczynski:
> Is this just another form of #346?

I don't think so, because AFAICT, this is the same as what's done in `alloc_test_operators_in_loops`, which "redefines" operators inside a loop.

jakebolewski:
https://github.com/CliMA/ClimaCore.jl/pull/408/files#diff-ee589bf4d9c0274df366b1b03a2116d98c8ef9e6689c94f25cda1240d3fbbe79R486?

charleskawczynski:
> https://github.com/CliMA/ClimaCore.jl/pull/408/files#diff-ee589bf4d9c0274df366b1b03a2116d98c8ef9e6689c94f25cda1240d3fbbe79R486?

Ahhhh, maybe that's it. I'll try changing that

jakebolewski (approved):

jakebolewski:
The determining cause of this after going through it with @charleskawczynski is that the expression is too complex for the optimizer when combining the broadcasted expression axes (`Base.Broadcast.combine_axes()` in `_axes`).  We will have to probably re-write combine_axes carefully to prevent this from happening over a given complexity limit in the expression (number of recursive broadcast calls in the call tree) which will be a bit of work.  A workaround is to break up the above expression roughly in two (adding in the second expression) which doesn't trigger any allocations (hack but it is an easy heuristic to follow in model code for the time being).

PR #413, opened 2022-01-11
Add some perf scripts

charleskawczynski:
Add some performance monitoring scripts


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

examples/3dsphere/baroclinic_wave.jl
```diff
@@ -314,6 +314,9 @@ elseif test_name == "balanced_flow"
 end
 dt = 5
 prob = ODEProblem(rhs!, Y, (0.0, time_end))
+
+haskey(ENV, "PERF_SKIP_RUN") && exit() # for performance analysis
+
```
valeriabarra, line 319: Thanks Charlie. Do we want to apply this also to the other examples?

charleskawczynski: We could, but I think @jakebolewski suggested that we first try analyzing only a few representative examples (representative w.r.t. coverage of ClimaCore internals, e.g., hitting 2D, 3D, hybrid, etc.).

It looks we have overlapping allocation sites even within these three examples, and they point to ClimaCore internals that probably affect many examples.

charleskawczynski:
I think this PR is about ready to go, unless we want to add the analysis to more examples?

examples/3dsphere/baroclinic_wave.jl
valeriabarra: Ah, gotcha. If helpful, the one I was working on that I saw could take quite a long time (when run with a resolution comparable to the one found in the literature) was the `examples/sphere/shallow-water.jl`, with the `barotropic_instability` test case =, even if it is a 2D manifold case and not a hybrid one

charleskawczynski: Is this a 2D sphere case?

szy21: I may add bubble2d_invariant and bubble3d_invariant, as they use different equations sets from those in bubble2d and bubble 3d. (Or maybe we can just replace bubble 2d and 3d with those if you prefer that?)

valeriabarra: > Is this a 2D sphere case?

The shallow-water one is a 2D manifold embedded in 3D space. Equation set has two components for velocity plus one for the height field

valeriabarra: But I think it's ok if these few examples already showed those 5/6 internal points to tackle

charleskawczynski: @jakebolewski are you familiar if this case exercises any additional ClimaCore machinery that the other examples don't exercise?

@valeriabarra I'll see if I can add it. It looks a little bit more complicated because an `ARG` needs to be passeed.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

examples/3dsphere/baroclinic_wave.jl
jakebolewski: let's not add more, one or two representative examples is enough

examples/3dsphere/baroclinic_wave.jl
```diff
@@ -314,6 +314,9 @@ elseif test_name == "balanced_flow"
 end
 dt = 5
 prob = ODEProblem(rhs!, Y, (0.0, time_end))
+
+haskey(ENV, "PERF_SKIP_RUN") && exit() # for performance analysis
```
jakebolewski, line 318: could we use `CI_PERF_SKIP`?   all others env switches start with `CI`

charleskawczynski: Sure, np

examples/3dsphere/baroclinic_wave.jl
jakebolewski: we also don't need to test the 2D cases, they are not going to matter for the final model

charleskawczynski:
bors try

jakebolewski:
LGTM

PR #419, opened 2022-01-13
init cond: discrete hydrostatic balance

jiahe23:
- add one column example with discretely balanced initial condition; `w < 1e-10` at the end of the simulation as compared to `~0.1` without the discrete balance correction;
- add the discrete balance correction to SBR; `w < 1e-11` after 1 hour of simulation as compared to `~0.8` without the discrete balance correction.


jiahe23:
bors try

jiahe23:
bors try

PR #423, opened 2022-01-13
Add conversion from LatLongZPoint to Cartesian

simonbyrne:
Fixes 3D spherical shell VTK output.

### DIFF
--- a/lib/ClimaCoreVTK/test/runtests.jl
+++ b/lib/ClimaCoreVTK/test/runtests.jl
@@ -186,7 +186,7 @@ end
 end
 
 @testset "hybrid 3d sphere" begin
-    R = 6.37122e6
+    R = 100.0
 
     hdomain = Domains.SphereDomain(R)
     hmesh = Meshes.EquiangularCubedSphere(hdomain, 4)
--- a/lib/ClimaCoreVTK/test/runtests.jl
+++ b/lib/ClimaCoreVTK/test/runtests.jl
@@ -184,3 +184,31 @@ end
     )
     @test isfile(joinpath(OUTPUT_DIR, "hybrid3d_cell.vtu"))
 end
+
+@testset "hybrid 3d sphere" begin
+    R = 6.37122e6
+
+    hdomain = Domains.SphereDomain(R)
+    hmesh = Meshes.EquiangularCubedSphere(hdomain, 4)
+    htopology = Topologies.Topology2D(hmesh)
+    quad = Spaces.Quadratures.GLL{5}()
+    hspace = Spaces.SpectralElementSpace2D(htopology, quad)
+
+
+    vdomain = Domains.IntervalDomain(
+        Geometry.ZPoint(0),
+        Geometry.ZPoint(20.0),
+        boundary_names = (:bottom, :top),
+    )
+    vmesh = Meshes.IntervalMesh(vdomain, nelems = 20)
+    vtopology = Topologies.IntervalTopology(vmesh)
+    vspace = Spaces.FaceFiniteDifferenceSpace(vtopology)
+
+    fspace = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace)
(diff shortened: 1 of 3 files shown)


simonbyrne:
cc: @jiahe23

simonbyrne:
Note that when exporting fields to VTK, you might want to export `z` to make it easier to construct contours.

lib/ClimaCoreVTK/test/runtests.jl
```diff
@@ -184,3 +184,31 @@ end
...
+    )
+    vmesh = Meshes.IntervalMesh(vdomain, nelems = 20)
+    vtopology = Topologies.IntervalTopology(vmesh)
+    vspace = Spaces.FaceFiniteDifferenceSpace(vtopology)
+
+    fspace = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace)
+    cspace = Spaces.CenterExtrudedFiniteDifferenceSpace(fspace)
+    writevtk(
+        joinpath(OUTPUT_DIR, "hybrid_sphere_point"),
+        Fields.coordinate_field(fspace);
```
valeriabarra, line 211: Is this supposed to show all points in this `fdspace` in the z direction? (i.e., showing the thickness from 0 to 20?) 
I tested this but I only see the 2D manifold
vtk-test-hybrid_sphere_point-output

simonbyrne: I think because it's really thin?

If you add an isosurface for `z` it gives you the different levels

jakebolewski: 30 km vs 6400 km is a big difference

valeriabarra: That's fair.. I was also zooming in and I could see through :eyes: 
I'll try a different value

valeriabarra: I wonder if for these viz tests would make sense to go dimensionless instead of dimensional ...

lib/ClimaCoreVTK/test/runtests.jl
```diff
@@ -184,3 +184,31 @@ end
...
+    hdomain = Domains.SphereDomain(R)
+    hmesh = Meshes.EquiangularCubedSphere(hdomain, 4)
+    htopology = Topologies.Topology2D(hmesh)
+    quad = Spaces.Quadratures.GLL{5}()
+    hspace = Spaces.SpectralElementSpace2D(htopology, quad)
+
+
+    vdomain = Domains.IntervalDomain(
+        Geometry.ZPoint(0),
+        Geometry.ZPoint(20.0),
```
valeriabarra, line 200:
```suggestion
        Geometry.ZPoint(R/3),
```

valeriabarra: There you go... if we used this z value, it'd be much more visible (tested -- images below)
vtk-test-hybrid_sphere_point-output


vtk-test-hybrid_sphere_point-output_wireframe

simonbyrne: Ah, I just pushed a change which does a similar thing.

valeriabarra: That's fine as well! Thanks

lib/ClimaCoreVTK/test/runtests.jl
simonbyrne: @OsKnoth suggested an option to exaggerate the z coordinate.

jakebolewski: I feel that would be better to have as a filter-pipeline in paraview (post-process)

valeriabarra (approved):

jiahe23:
LGTM! Thanks!

jiahe23:
bors try

Issue #341, opened 2021-12-09
`getproperty` calls inside broadcasted incur allocations


charleskawczynski:
See our broken tests, e.g., [here](https://github.com/CliMA/ClimaCore.jl/blob/main/test/fdspaces_opt_examples.jl#L44-L52)

simonbyrne:
Fixed by #755

Issue #11, opened 2021-04-16
Add size to parameter space of data layout


simonbyrne:
Some experiments by @vchuravy suggest there are some performance benefits, particularly on the GPU, to having the fast indices (i,j,k and field size) known at compile time. We could do this by adding these sizes to the parameters of the data layout objects.

simonbyrne:
Example: https://gist.github.com/vchuravy/7798cf3902168469777273cf2effeca6

jakebolewski:
We might need to add the 5 tuple size `(Ni, Nj, Nk, Nv, Nh)` to be able to better optimize field access.

jakebolewski:
A benefit of that would be that the packing data / array wouldn't have to actually be an array giving a bit more flexibility there

charleskawczynski:
Just to note: `Nf` is probably not needed in the type space, as it can be statically determined from `S` and `eltype(parent(data::AbstractData))`, using `DataLayouts.typesize`

charleskawczynski:
Closed by #1757

charleskawczynski:
Going to reopen this, as we do still have some type parameters that are only runtime known. This turns out to be important, based on recent benchmarks (#1893).

PR #424, opened 2022-01-13
Hoist some getproperty calls inside broadcast expr

charleskawczynski:
This PR should improve performance a bit via manual code hoisting, see #341.

I can confirm that this reduces the number of JET complaints.

### DIFF
--- a/examples/3dsphere/baroclinic_wave.jl
+++ b/examples/3dsphere/baroclinic_wave.jl
@@ -185,6 +185,7 @@ function rhs!(dY, Y, _, t)
     dw = dY.w
     duₕ = dY.uₕ
     dρe = dY.Yc.ρe
+    z = coords.z
 
     # 0) update w at the bottom
     # fw = -g^31 cuₕ/ g^33
@@ -275,7 +276,7 @@ function rhs!(dY, Y, _, t)
         Geometry.Contravariant12Vector(Geometry.Covariant123Vector(cuₕ))
 
     ce = @. cρe / cρ
-    cp = @. pressure(cρ, ce, norm(cuvw), coords.z)
+    cp = @. pressure(cρ, ce, norm(cuvw), z)
 
     @. duₕ -= hgrad(cp) / cρ
     vgradc2f = Operators.GradientC2F(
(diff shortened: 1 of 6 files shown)


charleskawczynski:
bors try

jakebolewski:
It is annoying that we have to do this but LGTM

jakebolewski (approved):

jakebolewski:
I wonder if we construct the property views directly (with Subarray) if that helps....

charleskawczynski:
> It is annoying that we have to do this but LGTM

Yeah, at least the fix is easy, though.

> I wonder if we construct the property views directly (with Subarray bypassing all the checks) if that would help....

Maybe, I'm still looking through the JET results..

jakebolewski:
We should just do it consistently in any case, I'll go through and make that change as it would he

jakebolewski:
Could you try?

```diff
diff --git a/src/DataLayouts/DataLayouts.jl b/src/DataLayouts/DataLayouts.jl
index 0ddeba0..20eb00d 100644
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -869,13 +869,16 @@ end
     data::VIFH{S, Ni, A},
     ::Val{Idx},
 ) where {S, Ni, A, Idx}
+    array = parent(data)
+    Nv = size(array, 1)
+    Nh = size(array, 4)
     SS = fieldtype(S, Idx)
     T = eltype(A)
     offset = fieldtypeoffset(T, S, Idx)
     nbytes = typesize(T, SS)
     field_byterange = (offset + 1):(offset + nbytes)
     return :(VIFH{$SS, $Ni}(
-        @inbounds view(parent(data), :, :, $field_byterange, :)
+        @inbounds SubArray(parent(data), Base.Slice(Base.OneTo($Nv)), Base.Slice(Base.OneTo($Ni)), $field_byterange, Base.Slice(Base.OneTo($Nh)))
     ))
 end
 ```

charleskawczynski:
> Could you try? ...

I'll try this in a follow up PR, so that we know what perf changes are attributed to.

jakebolewski:
actually with a generated function, that doesn't work if we don't have the extents in the type :-/ We could try it with the getproperty call that isn't a generated function by switching the impl.

jakebolewski:
Releated: https://github.com/CliMA/ClimaCore.jl/issues/11#issue-860112968

PR #425, opened 2022-01-13
Update lib READMEs

valeriabarra:
There was a missing keyword and apostrophe


jakebolewski (approved):
looks good, fixes like these can just be merged in

PR #427, opened 2022-01-14
Just use dispatch for `inferred_stencil_spaces_error`

charleskawczynski:
JET seems to like this better

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -2073,12 +2073,10 @@ Base.Broadcast._broadcast_getindex_eltype(
 ) = eltype(bc)
 
 # check that inferred output field space is equal to dest field space
-@noinline inferred_stencil_spaces_error(
-    dest_space_type::Type,
-    result_space_type::Type,
-) = error(
-    "dest space `$dest_space_type` is not the same instance as the inferred broadcasted result space `$result_space_type`",
+inferred_stencil_spaces_error(::Type{A}, ::Type{B}) where {A, B} = error(
+    "dest space `$A` is not the same instance as the inferred broadcasted result space `$B`",
 )
+inferred_stencil_spaces_error(::T, ::T) where {T} = nothing
 
 function Base.Broadcast.materialize!(
     ::Base.Broadcast.BroadcastStyle,
@@ -2086,11 +2084,9 @@ function Base.Broadcast.materialize!(
     bc::Base.Broadcast.Broadcasted{Style},
 ) where {Style <: AbstractStencilStyle}
     dest_space, result_space = axes(dest), axes(bc)
-    if result_space !== dest_space
-        # TODO: we pass the types here to avoid stack copying data
-        # but this could lead to a confusing error message (same space type but different instances)
-        inferred_stencil_spaces_error(typeof(dest_space), typeof(result_space))
-    end
+    # TODO: we pass the types here to avoid stack copying data
+    # but this could lead to a confusing error message (same space type but different instances)
+    inferred_stencil_spaces_error(dest_space, result_space)
     # the default Base behavior is to instantiate a Broadcasted object with the same axes as the dest
     return copyto!(
         dest,


charleskawczynski:
bors try

src/Operators/finitedifference.jl
```diff
@@ -2073,24 +2073,20 @@ Base.Broadcast._broadcast_getindex_eltype(
...
 # check that inferred output field space is equal to dest field space
-@noinline inferred_stencil_spaces_error(
-    dest_space_type::Type,
-    result_space_type::Type,
-) = error(
-    "dest space `$dest_space_type` is not the same instance as the inferred broadcasted result space `$result_space_type`",
+inferred_stencil_spaces_error(::Type{A}, ::Type{B}) where {A, B} = error(
+    "dest space `$A` is not the same instance as the inferred broadcasted result space `$B`",
 )
+inferred_stencil_spaces_error(::T, ::T) where {T} = nothing
```
jakebolewski, line 2079: `check_inferred_stencil_spaces()`?

src/Operators/finitedifference.jl
```diff
@@ -2073,24 +2073,20 @@ Base.Broadcast._broadcast_getindex_eltype(
 ) = eltype(bc)
 
 # check that inferred output field space is equal to dest field space
-@noinline inferred_stencil_spaces_error(
-    dest_space_type::Type,
-    result_space_type::Type,
-) = error(
-    "dest space `$dest_space_type` is not the same instance as the inferred broadcasted result space `$result_space_type`",
+inferred_stencil_spaces_error(::Type{A}, ::Type{B}) where {A, B} = error(
```
jakebolewski, line 2076: can we mark this as `@noinline` to outline the error call?

src/Operators/finitedifference.jl
```diff
@@ -2073,24 +2073,20 @@ Base.Broadcast._broadcast_getindex_eltype(
...
 ) where {Style <: AbstractStencilStyle}
     dest_space, result_space = axes(dest), axes(bc)
-    if result_space !== dest_space
-        # TODO: we pass the types here to avoid stack copying data
-        # but this could lead to a confusing error message (same space type but different instances)
-        inferred_stencil_spaces_error(typeof(dest_space), typeof(result_space))
-    end
+    # TODO: we pass the types here to avoid stack copying data
+    # but this could lead to a confusing error message (same space type but different instances)
+    inferred_stencil_spaces_error(dest_space, result_space)
```
jakebolewski, line 2089: we don't want to pass the spaces in that was largely the point of above, otherwise it can lead to large stack copies.

jakebolewski:
Are you sure this is real?  This would have come up all the time in looking at the TC profiles but maybe there is something different with the hybrid codepaths.

charleskawczynski:
> Are you sure this is real? This would have come up all the time in looking at the TC profiles but maybe there is something different with the hybrid codepaths.

I’ll double check today

jakebolewski:
when interpreting the profiles remember that it is a meet over paths so if the space cannot be inferred correctly statically both paths will show up in the profile.

charleskawczynski:
Yeah, JET seems to like the version in this PR better

jakebolewski:
But this is not checking the same thing, before we're checking if the spaces were the same instance and here you are not doing that

charleskawczynski:
Ahh, right. I forget, why do they need to be the same instance?

src/Operators/finitedifference.jl
charleskawczynski: I’ll try this today

jakebolewski:
it's how we check (cheaply) that the spaces have equal values

charleskawczynski:
Note to self: try removing args from error (in addition/combination with `@nospecialize`)

jakebolewski:
I would not say that is universally true but probably something to weigh for hot codepaths.

jakebolewski:
Can this be closed?

charleskawczynski:
Yeah, I did confirm that JET doesn't like this, but I don't think it's a performance issue.

jakebolewski:
I think the JET issue is a meet over error paths (it cannot prove that the spaces are equal so it shows up in the trace).

PR #428, opened 2022-01-14
Use better recursion pattern

charleskawczynski:
I'm hoping this was the big missing piece! 🤞🏻

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -224,6 +224,13 @@ end
 
 # 1D get/set node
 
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

charleskawczynski:
Well, not as big performance improvement as I hoped, but it seems to be improving allocations quite a bit.

jakebolewski:
Allocations are why we were using tuplemap and not map originally :).   Look like this one was missed, so it is good to have all cases covered now.

jakebolewski (approved):

PR #432, opened 2022-01-14
Arbitrary order 1D spectral element remapping 

jb-mackay:
Generalizes FV <-> Spectral Element remapping to handle arbitrary polynomial orders.

### DIFF
--- a/src/Operators/remapping.jl
+++ b/src/Operators/remapping.jl
@@ -53,16 +53,16 @@ by both element index and nodal order within the element.
 """
 function linear_remap_op(target::AbstractSpace, source::AbstractSpace)
     J = 1.0 ./ local_weights(target) # workaround for julia #26561
-    W = overlap_weights(target, source)
+    W = overlap(target, source)
     return W .* J
 end
 
 """
-    overlap_weights(target, source)
+    overlap(target, source)
 
 Computes local weights of the overlap mesh for `source` to `target` spaces.
 """
-function overlap_weights(
+function overlap(
     target::T,
     source::S,
 ) where {
@@ -81,7 +81,7 @@ function overlap_weights(
     source::S,
 ) where {
     T <: SpectralElementSpace1D{<:IntervalTopology},
-    S <: SpectralElementSpace1D{<:IntervalTopology, Quadratures.GL{1}},
+    S <: SpectralElementSpace1D{<:IntervalTopology},
 }
     FT = Spaces.undertype(source)
     target_topo = Spaces.topology(target)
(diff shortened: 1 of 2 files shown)


src/Operators/remapping.jl
```diff
@@ -109,35 +109,55 @@ function overlap_weights(
...
     return J_ov
 end
 
-function overlap_weights(
+function overlap(
     target::T,
     source::S,
 ) where {
-    T <: SpectralElementSpace1D{<:IntervalTopology, Quadratures.GL{1}},
+    T <: SpectralElementSpace1D{<:IntervalTopology},
```
jb-mackay, line 145: Is there only a way to have an `overlap_weights` function, and using dispatch to check the quadrature orders?

test/remapping.jl
```diff
@@ -412,53 +391,104 @@ end
...
+            R = LinearRemap(space1, space2)
             @test nnz(R.map) == 12
             @test conservative(R)
             @test consistent(R)
             @test !monotone(R)
         end
     end
 end
+
+@testset "Arbitrary Order Spectral Elements Remapping" begin
```
jb-mackay, line 430: If you have additional ideas for how to test these operators please let me know. It may be worth calculating `R_true` by hand for one or two cases.

test/remapping.jl
```diff
@@ -70,24 +100,15 @@ end
 end
 
 @testset "Finite Volume Remapping" begin
-    quad = Spaces.Quadratures.GL{1}() # FV specification
```
valeriabarra, line 73: Should we call the above test set `Finite Volume <-> Finite Volume Remapping` for clarity?

test/remapping.jl
valeriabarra: I would call this testset "Spectral Elements <-> Spectral Elements Remapping", so that we have all 3 testset names consistent. 
I am going to think about this a bit more

src/Operators/remapping.jl
simonbyrne: You can't really do arithmetic in the type domain. What you have is probably the simplest solution

simonbyrne (approved):

simonbyrne:
Maybe squash before merging?

jb-mackay:
Blue: original function on source field
Red: remapped function on target field
Green: original function represented on target field

Screen Shot 2022-01-25 at 11 33 06 AM

Screen Shot 2022-01-25 at 11 53 15 AM

Screen Shot 2022-01-25 at 11 53 36 AM

valeriabarra:
Nice job, Ben! Is the red curve in the second and third plot completely matching the other ones?

valeriabarra:
Oh, I can see a tiny red dot in the third plot

PR #439, opened 2022-01-17
Fix reference to global variable

charleskawczynski:
It looks like this may happen in many examples, I'll apply to others, I just want to see what CI looks like with these changes, first.

### DIFF
--- a/examples/3dsphere/baroclinic_wave.jl
+++ b/examples/3dsphere/baroclinic_wave.jl
@@ -175,7 +175,7 @@ uₕ = map(
 w = map(_ -> Geometry.Covariant3Vector(0.0), face_coords)
 Y = Fields.FieldVector(Yc = Yc, uₕ = uₕ, w = w)
 
-function rhs!(dY, Y, _, t)
+function rhs!(dY, Y, params, t)
     cρ = Y.Yc.ρ # scalar on centers
     fw = Y.w # Covariant3Vector on faces
     cuₕ = Y.uₕ # Covariant12Vector on centers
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

valeriabarra (approved):

simonbyrne:
@charleskawczynski what's happening with this?

charleskawczynski:
I need to apply this fix to more examples. I'll try to come back around on it tomorrow.

jakebolewski:
Again, I would focus in on a 2 representative examples (a 3D sphere case and a 3D LES case).  The model we need to be fast is not the dycore example projects but the atmos model.  Things will get lost in translation so we need only a good baseline perf. indicator(s) to do measurements on.

charleskawczynski:
Okay, I didn't get to this the other day, and I agree about where to focus efforts, but I'd still like to update this just to mitigate the spread of bad patterns.

szy21:
Just want to add to the discussion if you are not sure which examples to focus on:

Explicit:
```
examples/hybrid/bubble3d_invariant_totalenergy.jl
examples/3dsphere/baroclinic_wave.jl
```

Implicit (still waiting for the correction in the implicit solver):
```
examples/3dsphere/baroclinic_wave_rho_etot.jl
examples/3dsphere/held_suarez_rho_etot.jl
```

charleskawczynski:
This was basically superseded by #493. I'll reopen after adding perf analysis to held suarez and the 3d hybrid bubble

PR #440, opened 2022-01-17
restrict curl / weak curl to work on only covariant vector field inputs, this enables us to optimize through the hyperdiffusion operation

### LINKED ISSUE #264: Coordinate and vector conversions
There are a couple of related issues (following on from #81).

1. Convert coordinate types. For example, to get the Cartesian coordinates of the nodal points. I added a function here to do the conversion from lat long to Cartesian:
https://github.com/CliMA/ClimaCore.jl/blob/853af9c3bdbc488a78eb76e82e00053ceb248809/src/Geometry/coordinates.jl#L150-L155
but it will give a point on the unit sphere. To give the correct value, it needs the radius. Similarly we should convert `XYPoint`s to `Cartesian12Point`s.

2. Convert vector types. For example, converting a UV vector to a Cartesian vector requires some sort of global knowledge (e.g. if we're in a box, you just convert them directly, if on a sphere you need to do the rotation).

3. Handling of WVector/Covarian3Vector/Contravariant3Vector in 2D models (e.g. when using curls or cross products). Currently we have a few hacks in like this:
https://github.com/CliMA/ClimaCore.jl/blob/853af9c3bdbc488a78eb76e82e00053ceb248809/src/Geometry/conversions.jl#L73-L79
but we need to make it formalized and work properly in 3D

For 1 & 2, the two options I can think of are:
a) store the conversions to/from Cartesian coordinates/vectors at every point. Since we don't need this generally (at the moment its really only for visualization), it seems rather wasteful.
b) In addition to the `LocalGeometry` object, pass through an extra `GlobalGeometry` object that is sufficient to determine these. This should be a lightweight object.

For 3, the solution here is probably to exploit the type parameters of `LocalGeometry` and dispatch accordingly.

### LINKED ISSUE #334: Better optimize FD Curl / Specal Curl operations
Currently the curl operators do runtime dispatch on the input types, we need to re-write the implementation so that is more friendly to the optimizer.   Can lead to runtime dispatch in some inner RHS loops that kind of kills performance.

* One suggestion was to enforce that all Curl operator inputs are covariant vector types, but that requires changing the implementation + examples and is API breaking.

jakebolewski:
Curl on a covariant vector field does not depend on the local metric tensor, this enables the compiler to be able to statically infer the result type and optimize through `curl(curl())` forms found in computing the hyper viscosity terms.

Closes #334 
Closes #264

### DIFF
--- a/docs/tutorials/introduction.jl
+++ b/docs/tutorials/introduction.jl
@@ -511,8 +511,11 @@ function shallow_water_tendency!(dydt, y, _, t)
     ClimaCore.Spaces.weighted_dss!(dydt)
 
     @. dydt.u =
-        -D₄ *
-        (wgrad(sdiv(dydt.u)) - Geometry.Covariant12Vector(wcurl(curl(dydt.u))))
+        -D₄ * (
+            wgrad(sdiv(dydt.u)) - Geometry.Covariant12Vector(
+                wcurl(Geometry.Covariant3Vector(curl(dydt.u))),
+            )
+        )
     @. dydt.ρθ = -D₄ * wdiv(grad(dydt.ρθ))
 
     ## comute rest of tendency
--- a/examples/3dsphere/baroclinic_wave.jl
+++ b/examples/3dsphere/baroclinic_wave.jl
@@ -204,15 +204,20 @@ function rhs!(dY, Y, _, t)
 
     χe = @. dρe = hwdiv(hgrad(cρe / cρ))
     χuₕ = @. duₕ =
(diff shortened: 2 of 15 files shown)


jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
This looks to have made improvements for the 3D cases but the 2D case is showing a large performance regression :-/

jakebolewski:
bors try

simonbyrne (approved):

jakebolewski:
Ok it looks like tweaking the inlining heuristics recovered performance.  This is an API breaking change.

PR #442, opened 2022-01-17
Increase 2d bubble timestep, reduce io

charleskawczynski:
There aren't many inference failures left in the 2D bubble, most of the time spent is in runtime. Comparing with the 3D bubble driver, we run for 500x longer, and with 10x the output frequency. This PR increases the timestep, and decreases the output frequency.


charleskawczynski:
bors try

PR #426, opened 2022-01-14
Optimization-based limiters for advection/solidbody equation

valeriabarra:
~This is still a WIP.~ I'm opening the PR so that others can provide early feedback and/or contribute to it. 

I added a Limiters module, with an abstract limiter type, but then decided to not introduce any concrete type (structs) because they were not needed.

PR #446, opened 2022-01-18
Add neighboring_elements

### LINKED ISSUE #445: Query neighboring element indices
For the development of optimization-based limiters (#403), for a given element we need to query its neighboring elements.

valeriabarra:
This PR adds a function that given an element returns its neighboring elements (in a 2D topology). 
This is needed in #426.
While at it, I fixed a bug in `IntervalTopology`'s `opposing_face` (and realized that we do not have any tests for `IntervalTopology`!!)

Will close #445 .

### DIFF
--- a/src/Topologies/topology2d.jl
+++ b/src/Topologies/topology2d.jl
@@ -132,3 +132,62 @@ boundary_tag(topology::Topology2D, boundary_name::Symbol) =
     findfirst(==(boundary_name), boundary_names(topology))
 
 boundary_faces(topology::Topology2D, boundary) = topology.boundaries[boundary]
+
+function neighboring_elements(topology::Topology2D, elem)
+
+    # Each elem in a Topology2D will have at most 8 neighboring elements
+    #
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #
+    #
+    # Except for corner elements of the cube (they will have 7)
+    #
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #    |      |      |   ^
+    #    |      |      | < these two faces are the same at a cube corner
+    #    o------o------o
+
+    neigh_elems = Array{Int}(undef, 8)
+    op_faces = Array{Int}(undef, 8)
+    # First find the cross elements
+    for f in 1:4
+        (opelem, opface, _) = opposing_face(topology, elem, f)
+        # We look to the right of opface. Hence, subtract -1 from opface
+        opface = mod1(opface - 1, 4)
+        op_faces[f] = opface
+        neigh_elems[f] = opelem
+    end
+
+    for e in 1:4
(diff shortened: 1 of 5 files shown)


valeriabarra:
bors try

src/Topologies/topology2d.jl
```diff
@@ -132,3 +132,65 @@ boundary_tag(topology::Topology2D, boundary_name::Symbol) =
...
+    #    o------o------o------o
+    #    |      |      |      |
+    #    |      |      |      |
+    #    o------o------o------o
+    #    |      |      |   ^
+    #    |      |      | < these two faces are the same at a cube corner
+    #    o------o------o
+
+    neigh_elems = []
+    op_faces = []
```
simonbyrne, line 166: Best to make these typed:
```suggestion
    neigh_elems = Int[]
    op_faces = Int[]
```

src/Topologies/topology2d.jl
```diff
@@ -132,3 +132,65 @@ boundary_tag(topology::Topology2D, boundary_name::Symbol) =
...
+    # First find the cross elements
+    for f in 1:4
+        (opelem, opface, _) = opposing_face(topology, elem, f)
+        # We look to the right of opface. Hence, subtract -1 from opface
+        opface -= 1
+        if opface == 0
+            opface = 4
+        end
```
simonbyrne, line 174:
```suggestion
        opface = mod1(opface-1, 4)
```

src/Topologies/topology2d.jl
```diff
@@ -132,3 +132,65 @@ boundary_tag(topology::Topology2D, boundary_name::Symbol) =
...
+                opposing_face(topology, neigh_elems[e], op_faces[e])
+            # If the opposing element is already in the list of neighboring elements, we are at a cube corner and we return a 0 elem index
+            if opelem ∈ neigh_elems[1:4]
+                opelem = 0
+            end
+        end
+        push!(neigh_elems, opelem)
+    end
+
+    return Tuple(neigh_elems)
```
simonbyrne, line 195: Returning a `Vector` is probably fine here, especially since the `Tuple` will be type-unstable
```suggestion
    return neigh_elems
```

simonbyrne (commented):
This seems like a good interface: I think the only question is should we drop the boundary elements?

Internally, it may make sense to actually store the neighboring elements as a ragged array (similar to how we handle the shared vertices), but it would probably make sense to do that after the DistributedTopology stuff has been merged.  In any case, we can keep the same interface so that shouldn't require modifying downstream code.

src/Topologies/topology2d.jl
jakebolewski: I think in general it would be better to move gradually to a fixed size + seintinal value (ex. -1) for missing values for performance.

valeriabarra: I switched to a fixed-size Array. I thought that 0 was an ok sentinel value in this case (given that `opposing_face` returns 0 in case of non-existent face). Isn't it?

jakebolewski: that sounds good, you might want to switch to returning an SVector here (other parts of the mesh api are starting to do that).

valeriabarra: Thanks. I would leave that for another PR, if we are applying that change consistently

jakebolewski: The general form is:
```
mvec = MVector(...)
... fill in entries
return SVector(mvec)```

valeriabarra:
bors try

valeriabarra:
I squashed commits. Ready to merge

simonbyrne (approved):

PR #451, opened 2022-01-18
Monitor implicit baroclinic wave allocations

charleskawczynski:


charleskawczynski:
Actually, we need to call some sort of `perform_step!` for this, since there are multiple `rhs!` functions. I'll take a look at OrdinaryDiffEq.jl to figure out what's best to use here.

PR #452, opened 2022-01-19
add rho theta version of implicit baroclinic wave 

szy21:
co-authored-by: @Zhengyu-Huang @dennisYatunin 
still needs fix from @dennisYatunin


szy21:
bors try

szy21:
bors try

PR #420, opened 2022-01-13
check in Manifest files

simonbyrne:
Since it is a bit annoying to instantiate these directly. 

Alternatively, we could check in Manifest.toml files for these (which may not be a bad idea, as then they should be always reproducible?)

PR #453, opened 2022-01-19
Check in Manifest for docs, update build scripts and README

simonbyrne:
Follow on from #420.


jakebolewski:
I guess we will have to update these every version bump?

simonbyrne:
I was hoping CompatHelper would take care of that?

simonbyrne:
Oh, you mean version bump of ClimaCore? I guess we should just have that part of the release process.

jakebolewski:
It seems like it checks in the version of the 'dev' subpackages which I didn't realize (I thought it would just be at the current commit).  Maybe this info is ignored  though.

simonbyrne:
Yeah, I don't think it makes a difference, but we can make sure to update these when we bump the versions.

examples/README.md
```diff
@@ -9,17 +9,10 @@
...
-Activate the Pkg repl mode and dev the ClimaCore repo and ClimaCorePlots packages into the examples project environment:
-
-    (examples) pkg> dev . lib/ClimaCorePlots
-    (examples) pkg> instantiate
+Instantiate the examples environment (this will download and install the necessary packages)
 
-Or alternatively:
-
-    julia --project=examples -e 'using Pkg; Pkg.develop([Pkg.PackageSpec(path="."), Pkg.PackageSpec(path="lib/ClimaCorePlots")])'
+    cd ClimaCore
```
jakebolewski, line 14: I would leave a note in here that the manifest `dev` the lib/ folder sub-projects, otherwise it could be confusing

jakebolewski (approved):

PR #455, opened 2022-01-19
Fix monitoring, fix some global refs

charleskawczynski:
I've looked at #451, and I'm actually not sure how it merged because it looks like `rhs!` is called without the parameters in the perf monitoring. This PR should "fix" that, as well as do some code hoisting in the baroclinic wave files.


charleskawczynski:
bors try

charleskawczynski:
Closing in favor of #458. I'll take a second look at hoisting & global refs once the new allocation tables are up.

PR #456, opened 2022-01-20
Add 1D field plotting.

jb-mackay:
There may be a cleaner way to do this, but this uses `ClimaCore` tools I'm familiar with.

### DIFF
--- a/lib/ClimaCorePlots/src/ClimaCorePlots.jl
+++ b/lib/ClimaCorePlots/src/ClimaCorePlots.jl
@@ -13,6 +13,53 @@ import ClimaCore:
     Topologies,
     Spaces
 
+using StaticArrays
+
+RecipesBase.@recipe function f(
+    field::Fields.SpectralElementField1D;
+    interpolate = 10,
+)
+    @assert interpolate ≥ 1 "number of element quadrature points for uniform interpolation must be ≥ 1"
+
+    # compute the interpolated data to plot
+    space = axes(field)
+    topology = Spaces.topology(space)
+    nelems = Topologies.nlocalelems(space)
+    QS = Spaces.quadrature_style(space)
+    quad_name = Base.typename(typeof(QS)).name
+    Nq = Spaces.Quadratures.degrees_of_freedom(QS)
+    Nu = max(interpolate, Nq)
+    coord_field = Fields.coordinate_field(space)
+
+    lagrange_quad = Spaces.Quadratures.ClosedUniform{Nu}()
+    dataspace = Spaces.SpectralElementSpace1D(space.topology, lagrange_quad)
+    interp = Operators.Interpolate(dataspace)
+    ifield = interp.(field)
+
+    coords = vec(parent(Fields.coordinate_field(ifield).x))
+    data = vec(parent(Fields.field_values(ifield)))
+
+    # NaN injection element boundaries
+    r = range(1, nelems + 1, length(coords) + 1)
+    xcoords = [coords[1]]
+    xdata = [data[1]]
+    for k in 2:(length(r) - 1)
+        if floor(r[k - 1]) != floor(r[k]) # check this.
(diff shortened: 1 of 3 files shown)


jakebolewski:
LGTM, maybe it would be good to add a default tittle and axis labels  (ex. with the quadrature as we do for the other spectral spaces).

Test plot:
image

jakebolewski (approved):

jb-mackay:
> LGTM, maybe it would be good to add a default tittle (ex. with the quadrature as we do for the other spectral spaces).

Good idea; going to add the quadrature as a label rather than a title so that we can do things like `plot(field1); plot!(field2)` and have it be interpretable.

valeriabarra (approved):

valeriabarra:
Nice job, Ben!

simonbyrne:
It might be helpful if we make the curve discontinuous at the element boundaries: it looks like the easiest  way is to insert NaNs at the discontinuities:
https://discourse.julialang.org/t/how-to-plot-correctly-a-discontinuous-function/26931/2

jb-mackay:
@simonbyrne is this what you are envisioning? Do you just want to see the different elements distinctly? Perhaps it is possible to color them differently, but that could get confusing in the legend... 

Screen Shot 2022-01-19 at 11 03 52 PM

valeriabarra:
If the test case is plotting a CG field, I'd actually prefer without discontinuities

valeriabarra:
I see now no DSS

simonbyrne:
> @simonbyrne is this what you are envisioning? Do you just want to see the different elements distinctly? Perhaps it is possible to color them differently, but that could get confusing in the legend...

There should only be a vertical gap, not a horizontal one (since we have colocated nodes at the element boundaries), so it should look continuous if the values are the same, or discontinuous if the values are different.

lib/ClimaCorePlots/src/ClimaCorePlots.jl
```diff
@@ -13,6 +13,55 @@ import ClimaCore:
...
+        weight_idx = (Nq * (i - 1) + 1):(Nq * i)
+        vertices = Topologies.vertex_coordinates(topology, i)
+        xmin, xmax = Geometry.component(vertices[1], 1),
+        Geometry.component(vertices[2], 1)
+        coords = SVector{interpolate}(range(xmin, xmax, interpolate))
+        xcoords[interp_idx] .= coords
+        ξ, _ = Spaces.Quadratures.quadrature_points(FT, QS)
+        x = FT(0.5) * (xmin + xmax) .+ FT(0.5) * (xmax - xmin) * ξ
+        xdata[interp_idx] .=
+            Spaces.Quadratures.interpolation_matrix(coords, x) * w[weight_idx]
```
simonbyrne, line 51: If you want to get equispaced points, an easier option is probably the whole field to `ClosedUniform` quadrature rule, then just pull out the coordinates and values

jakebolewski:
You can see how to do that here: https://docs.juliaplots.org/stable/generated/gr/#gr-ref39

It could be toggled with setting `interpolate=0` (which is do in other placest)

jb-mackay:
> You can see how to do that here: https://docs.juliaplots.org/stable/generated/gr/#gr-ref39
> 
> It could be toggled with setting `interpolate=0` (which is do in other places)

Ah I see that I needed to add a NaN in the coordinate list as well

lib/ClimaCorePlots/src/ClimaCorePlots.jl
jakebolewski: see the slice model plot

jb-mackay:
Added `NaN`s at element boundaries; here's a FV example:
Screen Shot 2022-01-20 at 10 07 43 PM

simonbyrne:
That looks great!

PR #457, opened 2022-01-20
Add latlong option to VTK output

simonbyrne:


valeriabarra:
Cool. Is there a reason why the bounding box is not a rectangle and it looks like it's bent in the middle?
Test
image

simonbyrne:
It's because that's where the poles are: the top and bottom of the rectangle are each a single point on the sphere, and at that point we take define the longitude coordinate to be zero.

valeriabarra:
Ok but this way we don't seem to visualize the (-180,90) and (180,90) corner points. It's not how it is usually presented in the literature

simonbyrne:
That's true. It's not clear to me the best way to do that using VTK: perhaps switch to using polygons at the top and add extra vertices? We also need a way to handle the case when there are an odd-number of elements across the face (again, we probably need to insert extra elements).

jakebolewski:
Most publication plots are re-gridded to an RLL grid (which paraview can also read with some niceties like being able to apply coastlines) so if we are worried about that it makes more sense to focus efforts there.

simonbyrne:
On further thought, I think we should leave the extrapolation to the full rectangle until we have a proper regridding solution.

jakebolewski (approved):

valeriabarra:
Hi @simonbyrne , would the work you've done with Tempest for the regridding help in having a full lat-long vtk output that includes the (±180,±90) corner points?

simonbyrne:
Possibly, but I don't think it's a priority item at the moment, as it would be quite a bit of work.

jakebolewski:
You can read regular lat / long / level netcdf files directly into paraview as well.

PR #458, opened 2022-01-21
Fix allocation monitoring

charleskawczynski:
This explains why things were working when I thought #455 would fix the missing parameter sent into `rhs!`-- we are currently exiting the julia session, so we only run this script to the noted line

```julia
import Pkg
Pkg.develop(path = ".")

import Profile

case_name = ENV["ALLOCATION_CASE_NAME"]
ENV["CI_PERF_SKIP_RUN"] = true # we only need haskey(ENV, "CI_PERF_SKIP_RUN") == true


include(case_name)      #  <------------------- We exit in here and accidentally skip the rest of the script


rhs!(dYdt, Y, nothing, 0.0) # compile first
Profile.clear_malloc_data()
rhs!(dYdt, Y, nothing, 0.0)

# Quit julia (which generates .mem files), then call
#=
import Coverage
allocs = Coverage.analyze_malloc("src")
=#
```

This PR should reveal the true allocations occurring in the ClimaCore `rhs!` functions.

@jakebolewski / @simonbyrne, is it alright to use `try` / `catch` here like this? I was hoping to keep these performance parts as minimally invasive as possible

### DIFF
--- a/examples/sphere/shallow_water.jl
+++ b/examples/sphere/shallow_water.jl
@@ -178,7 +178,7 @@ end
 
 # Set initial condition
 if test_name == rossby_haurwitz_test_name
-    y0 = map(Fields.local_geometry_field(space)) do local_geometry
+    Y = map(Fields.local_geometry_field(space)) do local_geometry
         coord = local_geometry.coordinates
         ϕ = coord.lat
         λ = coord.long
@@ -213,7 +213,7 @@ if test_name == rossby_haurwitz_test_name
         return (h = h, u = u)
     end
 elseif test_name == steady_state_compact_test_name
-    y0 = map(Fields.local_geometry_field(space)) do local_geometry
+    Y = map(Fields.local_geometry_field(space)) do local_geometry
         coord = local_geometry.coordinates
 
         ϕ = coord.lat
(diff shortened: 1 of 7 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try-

jakebolewski (approved):

jakebolewski:
I think it is fine when used this way, if we were not eval things at the toplevel we could clean it up

charleskawczynski:
bors try

charleskawczynski:
@valeriabarra, I had to change a bunch of names in the `shallow_water.jl` file to be consistent with the other examples, in order to continue running the allocation monitoring on it, since we kind of need a unified configuration

valeriabarra:
> @valeriabarra, I had to change a bunch of names in the `shallow_water.jl` file to be consistent with the other examples, in order to continue running the allocation monitoring on it, since we kind of need a unified configuration

Thanks Charlie for the heads-up. Yeah, not a biggie changing `y` -> `Y`. Thanks for doing that and following this allocation monitoring!

examples/sphere/shallow_water.jl
```diff
@@ -354,7 +354,7 @@ else # steady-state and mountain test cases share the same form of fields
     end
 end
 
-function rhs!(dydt, y, parameters, t)
+function rhs!(dYdt, y, parameters, t)
```
valeriabarra, line 357: Wouldn't we also want to change the second argument, `y` -> `Y` to be consisten?

charleskawczynski: Ah, I thought I did. We can change this too. It's technically okay because of the scope, it's the global variable names that need to be consistent for the perf monitoring.

PR #459, opened 2022-01-21
add generalized exp stretching and interval mesh tests

### LINKED ISSUE #447: IntervalTopology not tested
While working on #446, I realized that we have no tests for IntervalTopology

jakebolewski:


jakebolewski:
I changed the implementation because with Roots there are large number of valid inputs were the default solver failed to converge and returned invalid values (this will at least raise an error).

PR #462, opened 2022-01-21
Preallocate some fields in implicit baroclinic wave

charleskawczynski:
I'm going to try to look at this a bit more offline


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
Hmm, it actually looks like the results are different from this PR.

PR #464, opened 2022-01-22
exodus format export function

simonbyrne:
(with @LenkaNovak)

This adds a function to export ClimaCore meshes in Exodus II format. I've added it into a ClimaCoreTempestRemap subpackage, as that is currently the main reason we need it, so we can add further functionality there as well.


.github/workflows/ClimaCoreTempestRemap.yml
```diff
@@ -0,0 +1,50 @@
...
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - run: sudo apt-get update && sudo apt-get install -y xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev
```
jakebolewski, line 31: don't need these

jakebolewski:
Paraview can read exodus mesh files as well so you could copy / paste the VTK CI code for viz

lib/ClimaCoreTempestRemap/src/ClimaCoreTempestRemap.jl
```diff
@@ -0,0 +1,153 @@
...
+        var_coord = defVar(dts, "coord", Float64, ("num_nodes", "num_dim"))
+
+        # variable attributes
+        var_connect1.attrib["elem_type"] = "SHELL4"
+        var_eb_prop1.attrib["name"] = "ID"
+
+        # variable values
+        dt = now()
+        qa_records = (
+            "ClimaCoreExodus.jl",
```
LenkaNovak, line 131: `ClimaCoreTempestRemap.jl`

lib/ClimaCoreTempestRemap/test/runtests.jl
```diff
@@ -0,0 +1,53 @@
...
+    # fractions should be close to 1
+    frac_a = NCDataset(nc -> Array(nc["frac_a"]), weightfile)
+    @test all(x -> x ≈ 1, frac_a)
+
+    @test_broken begin
+        run(
+            `$(TempestRemap_jll.GenerateOfflineMap_exe()) --in_mesh $meshfile_cc --out_mesh $meshfile_tr --ov_mesh $meshfile_overlap --in_np 2 --out_map $weightfile`,
+        )
+        true
+    end
```
simonbyrne, line 52: @jakebolewski when we run this one, it terminates with an error return code. However when I built it locally it seemed to run fine. Any suggestions on how we can debug this?

charleskawczynski: Don’t we still need to use LOAD_PATH for thing in runtests.jl?

charleskawczynski: How are you running locally?

simonbyrne: > Don’t we still need to use LOAD_PATH for thing in runtests.jl?

No, we set up an environment for this.

> How are you running locally?

if you `]dev ../..` you can load the package locally.

jakebolewski: hmm if it works locally but not on CI it might be an undefined behavior bug in tempest that differs between the gcc / clang toolchains.  I can build it locally and see

jakebolewski: maybe it isn't clear but the process from the above comment but the process returns a 11 exit code which means it is segfaulting, I can reproduce this locally on linux with the latest JLL release
.

simonbyrne: Yes, that's what I meant: the JLL segfaults on both Linux and OS X, but the non-JLL binaries work fine.

jakebolewski: by non-jll you mean the version of Clang on your laptop?

jakebolewski: I'm wondering if the bug is with libblastrampoline.  I had to switch to using MKL in the otther build because of bugs encountered.

simonbyrne: Thanks, that would be my first guess as well.

jakebolewski: https://github.com/JuliaPackaging/Yggdrasil/pull/4315 for the workaround (tested locally).

jakebolewski: the rebuilt package is now in the registry so the min-bounds should be set to `v2.1.2+1`

simonbyrne: Unfortunately it doesn't look like you can use builds numbers in bounds patches.

jakebolewski (approved):
Looks good, when I tested this yesterday all the test exodus files could be read by paraview.

PR #467, opened 2022-01-23
Fix baroclinic wave

charleskawczynski:
#462 somehow changed the results, so I'm going to re-apply them incrementally and fixed whatever has changed the results. Also, I'll try preallocating another field.

I misunderstood some of the broadcast assignments, it turns out that some lines were just pointer assignments. Results are back to normal.


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #470, opened 2022-01-25
Implicit Solver Experimentation

dennisYatunin:
This PR includes the following changes:

- The addition of "pointwise stencil" operations. These are a generalized form of finite difference operators, where the stencil used to determine the output values can be different at every point. To use a pointwise stencil, one needs a `Field` that stores the stencil; this is currently implemented as a `Field` of `StencilCoefs`, where each `StencilCoefs` object stores the stencil coefficients for a particular output point (in other words, it stores the row of the stencil matrix corresponding to that point). To simplify the creation of stencil fields, I have defined the `Operator2Stencil` operator, which wraps any operator defined in `finitedifference.jl` and outputs the stencil field that corresponds to that operator. Currently supported operations on stencils include `ApplyStencil` (which is effectively a matrix-vector multiplication) and `ComposeStencils` (which is a matrix-matrix multiplication). `ComposeStencils` allows one to specify the Jacobians of arbitrary finite difference broadcasts, since the chain rule can be used to express any Jacobian as a chain of matrix-matrix multiplications, where each matrix corresponds to a particular finite difference operator.

- The modification of `boundary_width` for all finite difference operators. While it previously only accepted the operator and the boundary location, it now also accepts the inputs of the operator. This allows `boundary_width` to be computed for `ApplyStencil` and `ComposeStencils` based on their input stencil fields.

- The simplification of `examples/3dsphere`. The various driver files are replaced with a single `driver.jl`. A flag at the top of this file specifies which initial conditions and explicit tendency function to use. Another file called `utilities.jl` specifies the implicit tendency function and Jacobian function, which are the same for all our tests. Another file called `schur_comlement_W.jl` defines the actual data structure for storing the Jacobian and the linear solver.

- Automatic tests for the implicit solver. The data structure for storing the Jacobian now includes a `test` flag. When this flag is true, a test is run on every Jacobian computation that compares the Jacobian of the first column of the domain to its exact value (determined using automatic differentiation), and another test is run on every linear solve that compares the result of the linear solve to its exact value (determined by using `ldiv!` on regular arrays).

- Some minor bug fixes: 1) A method was added for `RecursiveApply.radd(x) = x`. This was to simplify the use of `radd` in `ApplyStencil` and `ComposeStencils`. 2) Additional methods were added for determining the `BroadcastStyle`s of `DataLayout` broadcasts that have identical layouts but different base types, along with a method for determining the `BroadcastStyle`s of `Field` broadcasts that have different `DataLayout`s. This was to allow broadcasts over similar `Field`s with different base types; e.g., a `Field` of `Float64`s and a `Field` of `Dual`s. This made it easier to implement the Jacobian test.

It should be noted that the "storing-stencils-as-Fields" strategy is only a temporary solution. Specifically, broadcasts that act on stencil fields can be rather counterintuitive. Although `stencil_field1.^2 .+ stencil_field2 .* scalar_field` currently works as expected, that is only because I have overloaded basic arithmetic operations over `StencilCoefs` objects to automatically map over the objects. Trying to do more general operations like `f.(stencil_field)` will not work as expected, since `f` will be passed `StencilCoefs` objects, rather than the individual coefficients. In order to fix this problem, we need an alternative to a `Field` that is specifically designed for storing matrix-like objects, rather than vector-like objects. Such a data structure would also probably be useful for the coupler. Alternatively, we could define a new `BroadcastStyle` for stencil fields that automatically maps any function over `StencilCoefs` objects, though this would not be a very generalizable solution.

In the near future I will also add the stencil operation `stencil_solve!(x, stencil, b)`, which determines the `Field` `x` for which `ApplyStencil().(stencil, x) = b`. This will get used in the linear solver, which is currently the last place where I access parent arrays of `Field`s for non-debugging purposes, and will allow the solver to handle non-tridiagonal Schur complement matrices. Note that `stencil_solve!` will not be an `Operator`, since it must scan the entirety of `stencil` and `b` to determine the value of `x` at a particular point; it will be more akin to functions like `weighted_dss!`.


jakebolewski:
Thank you for squashing and rebasing :pray:

charleskawczynski:
Is there an issue with the rebase? It looks like, maybe, a bunch of files were renamed without `git mv`?

dennisYatunin:
Oh, I thought those should be new files, since they are no longer "drivers", but rather templates for the driver file.

charleskawczynski:
Ah, ok. I like that some things have been removed from global scope :slightly_smiling_face:. I think the buildkite pipeline will need to be updated.

src/Operators/pointwisestencil.jl
```diff
@@ -0,0 +1,425 @@
+#=
```
charleskawczynski, line 1: Can we add some docs here briefly describing the strategy/process?

src/Operators/finitedifference.jl
```diff
@@ -229,7 +229,7 @@ return_eltype(::FiniteDifferenceOperator, arg) = eltype(arg)
 @noinline invalid_boundary_condition_error(op_type::Type, bc_type::Type) =
     error("Boundary `$bc_type` is not supported for operator `$op_type`")
 
-boundary_width(op::FiniteDifferenceOperator, bc::BoundaryCondition) =
+boundary_width(op::FiniteDifferenceOperator, bc::BoundaryCondition, args...) =
```
jakebolewski, line 232: besides adding an argument to boundary width, most of the changes here seem unnecessary?  it makes it hard to review if we don't have minimal diffs.

Is there anything else materially that changed?

dennisYatunin: Everything else was just formatting. Sorry, I found it a bit hard to read with the inconsistent formatting.

src/Operators/pointwisestencil.jl
dennisYatunin: Definitely. I didn't make documentation or tests because Simon wants this available for others asap, but I'll definitely add those soon.

jakebolewski:
We don't have to do it in this PR, but we'll need optimizer tests to make sure we can infer these stencil expressions.

src/Operators/finitedifference.jl
dennisYatunin: Oh, I should note that stecil_interior_width was modified the same way as boundary_width, for the same reason.

src/Operators/pointwisestencil.jl
```diff
@@ -0,0 +1,425 @@
...
+    coefs_expr = :(coefs = getidx(stencil, loc, idx))
+    lbw = bandwidths(eltype(stencil_type))[1]
+    i_func = i -> :(coefs[$(i - lbw + 1)] ⊠ getidx(arg, loc, idx + $i))
+    return :(coefs_expr; return ⊞($(map(i_func, i_range)...)))
+end
+
+function stencil_interior(::ApplyStencil, loc, idx, stencil, arg)
+    lbw, ubw = bandwidths(eltype(stencil))
+    i_range = lbw:ubw
+    if @generated # TODO: Is this any better than the non-generated version?
```
jakebolewski, line 277: let's try to not use generated if it is not absolutely needed, can we test this again?

src/Operators/stencilcoefs.jl
```diff
@@ -0,0 +1,118 @@
...
+
+@inline Base.:(==)(
+    a::StencilCoefs{lbw, ubw},
+    b::StencilCoefs{lbw, ubw},
+) where {lbw, ubw} = a.coefs == b.coefs
+
+# Automatically map arithmetic functions over stencil coefficients. Makes it so
+# that broadcasting over stencil fields is just like broadcasting over matrices;
+# e.g., `stencil_field1.^2 .+ stencil_field2 .* scalar_field` should just work.
+# TODO: Give stencil fields their own broadcast style, so that this happens for
```
jakebolewski, line 90: materializing into a matrix does seem simpler, all scalar column fields will return contiguous views which should map to a blas matvec

src/Operators/pointwisestencil.jl
dennisYatunin: I figured that these sorts of loops could be easily sped up, since the bandwidths are available at compile time. The one for ApplyStencil is pretty simple, but the one for ComposeStencils almost certainly can't be optimized away without generated functions. I didn't have time to do a proper comparison, though, so I could be wrong.

dennisYatunin: Turns out the original implementation had a bug (eltype was not defined for field types, only for field instances). I just fixed the bug and ran a proper comparison. Without @generated:
```
julia> @timev compose.(stencil1, stencil2)
  0.054764 seconds (811.21 k allocations: 46.676 MiB)
elapsed time (ns): 54764000
bytes allocated:   48943552
pool allocs:       811210
malloc() calls:    1
```
With @generated:
```
julia> @timev compose.(stencil1, stencil2)
  0.008844 seconds (98.41 k allocations: 6.796 MiB)
elapsed time (ns): 8844500
bytes allocated:   7125952
pool allocs:       98410
malloc() calls:    1
```
I'd say an order-of-magnitude improvement is worth it, right?

jakebolewski: Yes absolutely.  Generated function have a cost so it is always good to check.  I don't think the way the non-generated function is written though is specializing on the bandwidth parameters, so that might be something to try.

Do you have to code used to generate the benchmark stencils?

dennisYatunin: The test above was done by running `driver.jl`, then setting `stencil1 = W.∂𝕄ₜ∂ρ` and `stencil2 = W.∂ρₜ∂𝕄`.

When you say that it isn't specializing on the bandwidth parameters, do you mean that the compiler is treating the outputs of `bandwidths(eltype(stencil1))` and `bandwidths(eltype(stencil2))` as runtime values, rather than inferring them from the types of `stencil1` and `stencil2`? What's the best way to verify/fix this?

jakebolewski: you can specialize through there where clause, ex: 

```foo(stencil1::Setencil{S1BW1, S1B2}, stencil2::Stencil{S2B1, S2B2}) where {S1B1, S2B2, ...}```

see: https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

you can check with `@code_lowered optimize=true ...`  you might need to stub out the broacast function into a toplevel function to do the later as broadcast get's lowered by the compiler into a series of calls.

Comment (dennisYatunin) [reply src/Operators/pointwisestencil.jl]: I tried your suggestion, but it did not improve the allocation/time of stencil composition at all. Unfortunately, `stencil_interior` is allocating a bunch of memory (which I checked using `perf/allocs.jl`) for stencil operations, which is definitely slowing down the implicit solver. Making `ComposeStencils` a generated function helps, but it does not eliminate the problem. If you have some time later, it might be useful for you to help me optimize it. I feel like I'm still not very certain about how to go about this sort of fine-tuning.

Comment (jakebolewski) [reply src/Operators/pointwisestencil.jl]: Sounds good, we can look at it sometime early next week when you have some time.  I don't think it is a blocker but good to track down the issues.

Comment (dennisYatunin) [reply src/Operators/pointwisestencil.jl]: I tracked down the primary issue. I had a function that needed to find a tuple for each integer in some range. I implemented this by passing in a function that returned the tuple corresponding to each integer. However, the tuples returned by this function can all have different lengths, so the compiler was unable to infer the concrete output type of the passed function (it was just inferring `Tuple`). I have changed my code so that I now pass in a tuple of tuples instead of a function, and I index the tuple at a particular integer to get the tuple corresponding to that integer. Here are my new results:
Without `@generated`:
```
julia> @timev compose.(stencil1, stencil2)
  0.005454 seconds (84.00 k allocations: 6.135 MiB)
elapsed time (ns): 5454186
bytes allocated:   6433104
pool allocs:       84004
malloc() calls:    1
```
With `@generated`:
```
julia> @timev compose.(stencil1, stencil2)
  0.004084 seconds (84.00 k allocations: 6.135 MiB)
elapsed time (ns): 4084302
bytes allocated:   6433104
pool allocs:       84004
malloc() calls:    1
```
So, the generated version is still ~30% faster, but at least the non-generated version is not allocating unnecessary memory due to type inference failure. My stencil operations are still one of the dominant sources of allocations according to `perf/allocs.jl`, but now they are roughly as bad as other finite difference operations, rather than an order of magnitude worse.

Comment (jakebolewski) [reply src/Operators/pointwisestencil.jl]: nice.  for measuring benchmarks with 1/1000th of a second difference you really need a statistical bench-marking test (ex. `@benchmark`) to make any diff conclusions, you cannot run one timing and give a conclusive answer.  It seems like they are pretty much eqv. so I would suggest dropping the `@generated` version.

Comment (dennisYatunin) [reply src/Operators/pointwisestencil.jl]: Here's a more rigorous test that does in-place operations and uses `@btime` (the first call is with `@generated`, and the second is without it):
```
julia> stencil1 = W.∂𝕄ₜ∂ρ; stencil2 = W.∂ρₜ∂𝕄; result = Operators.ComposeStencils().(stencil1, stencil2); compose!(result, stencil1, stencil2) = (result .= Operators.ComposeStencils().(stencil1, stencil2)); @btime compose!($result, $stencil1, $stencil2)
  3.034 ms (84000 allocations: 5.53 MiB)
ClimaCore.Operators.StencilCoefs{-1, 1, Tuple{Float64, Float64, Float64}}-valued Field:
  coefs: 
    1: [NaN, 0.0, -0.00223207, -0.00229962, -0.00239082, -0.0025035, -0.00262102, -0.00271167, -0.00278562, -0.00287097  …  0.0, -0.00223207, -0.00229962, -0.00239082, -0.0025035, -0.00262102, -0.00271167, -0.00278562, -0.00287097, 0.0]
    2: [-0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.0]
    3: [-0.0, 0.00119547, 0.00116286, 0.0011239, 0.00108114, 0.00104194, 0.00100713, 0.000970804, 0.00093571, 0.0  …  0.00119547, 0.00116286, 0.0011239, 0.00108114, 0.00104194, 0.00100713, 0.000970804, 0.00093571, 0.0, NaN]

julia> stencil1 = W.∂𝕄ₜ∂ρ; stencil2 = W.∂ρₜ∂𝕄; result = Operators.ComposeStencils().(stencil1, stencil2); compose!(result, stencil1, stencil2) = (result .= Operators.ComposeStencils().(stencil1, stencil2)); @btime compose!($result, $stencil1, $stencil2)
  4.309 ms (84000 allocations: 5.53 MiB)
ClimaCore.Operators.StencilCoefs{-1, 1, Tuple{Float64, Float64, Float64}}-valued Field:
  coefs: 
    1: [NaN, 0.0, -0.00223207, -0.00229962, -0.00239082, -0.0025035, -0.00262102, -0.00271167, -0.00278562, -0.00287097  …  0.0, -0.00223207, -0.00229962, -0.00239082, -0.0025035, -0.00262102, -0.00271167, -0.00278562, -0.00287097, 0.0]
    2: [-0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.0]
    3: [-0.0, 0.00119547, 0.00116286, 0.0011239, 0.00108114, 0.00104194, 0.00100713, 0.000970804, 0.00093571, 0.0  …  0.00119547, 0.00116286, 0.0011239, 0.00108114, 0.00104194, 0.00100713, 0.000970804, 0.00093571, 0.0, NaN]
```
So, the 30% difference seems to hold with better testing. Would you still say that 30% is not significant enough to merit a generated function?

Also, I've found that the other allocations are due to the same type instability problem, but at the boundaries rather than in the interior. Specifically, the tuple lengths at the boundaries depend on the values of `left_idx` and `right_idx` for the input spaces, and it seems like the compiler is unable to infer those values. It should be possible to make the inference succeed, since the number of levels of a space is a type parameter of the DataLayout, but I think I'll need your help to do that.

**Comment by dennisYatunin** [comment]
bors try

**Comment by dennisYatunin** [comment]
bors try

**Comment by dennisYatunin** [comment]
bors try

File: examples/3dsphere/baroclinic_wave_ρθ.jl [hunk 5]
```diff
@@ -0,0 +1,51 @@
...
+initial_condition_velocity(local_geometry) =
+    initial_condition_velocity(local_geometry; is_balanced_flow = false)
+
+remaining_cache_values(Y, dt) = merge(
+    baroclinic_wave_cache_values(Y, dt),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+)
+
+function remaining_tendency!(dY, Y, p, t)
+    fill!(dY, zero(eltype(dY)))
```
charleskawczynski, line 29:
```suggestion
    dY .= 0
```
This seems to be allocating ~341 MiB at every `step!`.

examples/3dsphere/baroclinic_wave_ρe.jl
```diff
@@ -0,0 +1,52 @@
...
+initial_condition_velocity(local_geometry) =
+    initial_condition_velocity(local_geometry; is_balanced_flow = false)
+
+remaining_cache_values(Y, dt) = merge(
+    baroclinic_wave_cache_values(Y, dt),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+)
+
+function remaining_tendency!(dY, Y, p, t)
+    fill!(dY, zero(eltype(dY)))
```
charleskawczynski, line 30:
```suggestion
    dY .= 0
```
This seems to be allocating ~341 MiB at every `step!`.

examples/3dsphere/driver.jl
```diff
@@ -0,0 +1,97 @@
...
+
+use_transform = !(ode_algorithm in (Rosenbrock23, Rosenbrock32))
+W_kwarg = use_transform ? (; Wfact_t = Wfact!) : (; Wfact = Wfact!)
+W = SchurComplementW(Y, use_transform, jacobian_flags, test_implicit_solver)
+problem = SplitODEProblem(
+    ODEFunction(
+        implicit_tendency!;
+        W_kwarg...,
+        jac_prototype = W,
+        tgrad = (dY, Y, p, t) -> fill!(dY, FT(0.0)),
```
charleskawczynski, line 54:
```suggestion
        tgrad = (dY, Y, p, t) -> fill!(dY, FT(0.0)),
```
Let's avoid `fill!` here since it seems to be allocating a lot (it may still be falling back to `getindex` for FieldVectors).

examples/3dsphere/held_suarez_ρe.jl
```diff
@@ -0,0 +1,54 @@
...
+    initial_condition_velocity(local_geometry; is_balanced_flow = false)
+
+remaining_cache_values(Y, dt) = merge(
+    baroclinic_wave_cache_values(Y, dt),
+    held_suarez_cache_values(Y, dt),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+)
+
+function remaining_tendency!(dY, Y, p, t)
+    fill!(dY, zero(eltype(dY)))
```
charleskawczynski, line 31:
```suggestion
    dY .= 0
```

examples/3dsphere/held_suarez_ρθ.jl
```diff
@@ -0,0 +1,53 @@
...
+    initial_condition_velocity(local_geometry; is_balanced_flow = false)
+
+remaining_cache_values(Y, dt) = merge(
+    baroclinic_wave_cache_values(Y, dt),
+    held_suarez_cache_values(Y, dt),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+)
+
+function remaining_tendency!(dY, Y, p, t)
+    fill!(dY, zero(eltype(dY)))
```
charleskawczynski, line 30:
```suggestion
    dY .= 0
```

examples/3dsphere/balanced_flow_ρe.jl
```diff
@@ -0,0 +1,66 @@
...
+initial_condition_velocity(local_geometry) =
+    initial_condition_velocity(local_geometry; is_balanced_flow = true)
+
+remaining_cache_values(Y, dt) = merge(
+    baroclinic_wave_cache_values(Y, dt),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+)
+
+function remaining_tendency!(dY, Y, p, t)
+    fill!(dY, zero(eltype(dY)))
```
charleskawczynski, line 31:
```suggestion
    dY .= 0
```

.gitignore
```diff
@@ -34,7 +34,7 @@ docs/src/tutorials/
 # committed for packages, but should be committed for applications that require a static
 # environment.
 
-
+*.DS_Store
```
simonbyrne, line 37: This is a mac specifc thing, and should go in your own excludes file (typically `$HOME/.config/git/ignore`)

dennisYatunin:
Just did another rebase+squash after addressing the new comments.

charleskawczynski:
bors try

charleskawczynski (approved):
Let's merge this! 🚀

PR #471, opened 2022-01-25
Widen allocation monitoring for implicit solve case

charleskawczynski:
This PR widens the allocation monitoring for the implicit solve to include OrginaryDiffEq.jl's `step!` routine. This should reveal if where more allocations are coming from.

Also, I changed from `$(Base.julia_cmd()) --project=examples/ --track-allocation=all perf/allocs_per_case.jl $args`  to `$(Base.julia_cmd()) --project=perf/ --track-allocation=all perf/allocs_per_case.jl $args` to ensure that I grab the correct package directories, otherwise we could silently miss allocations if the deps are different.


charleskawczynski:
bors try

charleskawczynski:
@valeriabarra, I think I'd like to remove the shallow water case from the allocation monitoring since it now depends on Plots, which I've been trying to avoid in the `perf/` environment. It seems that most of the allocations are the same as other tests, so I don't think that there are many improvements to make aside from hoisting `getproperty` calls outside of broadcast expressions.

PR #475, opened 2022-01-26
Try no io in baro wave

charleskawczynski:


charleskawczynski:
bors try

charleskawczynski:
#480 was the real fix I was looking for.

PR #480, opened 2022-01-27
Use better recursion pattern for instantiate, avoid getindex

charleskawczynski:
After looking at more flame graphs, I saw a call chain of `fast_materialize!` -> `materialize!` -> bad stuff with allocs and `getindex`. I took a look at our `materialize!` and that basically just calls `copyto!` and `instantiate`. Since we recently fixed issues with `copyto!`, I took a look at `instantiate` and recognized that we were using the same old, recursion pattern as before. This PR changes this recursion pattern to follow `column_args` and alike.

This will only affect the spectral code, since the call to `initialize` in FD's `materialize!` isn't recursive.

This PR also moves @dennisYatunin's `copyto!(::FieldVector,::Base.Broadcast.Broadcasted{FieldVectorStyle})` into `src/` with some minor adjustments (using a better recursion pattern, and supporting `VectorField`s with scalars.). This change seems to account for most of the performance benefit, as it elides calling `getindex` (which was the main culprit in the flame graph).

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -101,6 +101,8 @@ function Base.getindex(fv::FieldVector, bidx::BlockArrays.BlockIndex{1})
     X = fv[BlockArrays.block(bidx)]
     X[bidx.α...]
 end
+
+# TODO: drop support for this
 Base.getindex(fv::FieldVector, i::Integer) =
     getindex(fv, BlockArrays.findblockindex(axes(fv, 1), i))
 
@@ -108,9 +110,9 @@ function Base.setindex!(fv::FieldVector, val, bidx::BlockArrays.BlockIndex{1})
     X = fv[BlockArrays.block(bidx)]
     X[bidx.α...] = val
 end
+# TODO: drop support for this
 Base.setindex!(fv::FieldVector, val, i::Integer) =
     setindex!(fv, val, BlockArrays.findblockindex(axes(fv, 1), i))
-
 
 Base.similar(fv::FieldVector{T}) where {T} =
     FieldVector{T}(map(similar, _values(fv)))
@@ -167,14 +169,46 @@ function Base.similar(
     error("Cannot construct FieldVector")
 end
 
-@inline function Base.copyto!(
-    dest::FV,
-    src::FV,
-) where {FV <: Fields.FieldVector}
+@inline function Base.copyto!(dest::FV, src::FV) where {FV <: FieldVector}
     for symb in propertynames(dest)
         pd = parent(getproperty(dest, symb))
         ps = parent(getproperty(src, symb))
         copyto!(pd, ps)
+    end
+    return dest
+end
+
+# Recursively call transform_bc_args() on broadcast arguments in a way that is statically reducible by the optimizer
+# see Base.Broadcast.preprocess_args
+@inline transform_bc_args(args::Tuple, inds...) = (
+    transform_broadcasted(args[1], inds...),
+    transform_bc_args(Base.tail(args), inds...)...,
+)
+@inline transform_bc_args(args::Tuple{Any}, inds...) =
+    (transform_broadcasted(args[1], inds...),)
+@inline transform_bc_args(args::Tuple{}, inds...) = ()
+
+function transform_broadcasted(
+    bc::Base.Broadcast.Broadcasted{FieldVectorStyle},
+    symb,
+    axes,
+)
+    Base.Broadcast.Broadcasted(
+        bc.f,
+        transform_bc_args(bc.args, symb, axes),
+        axes,
+    )
+end
+transform_broadcasted(fv::FieldVector, symb, axes) =
+    parent(getfield(_values(fv), symb))
+transform_broadcasted(x, symb, axes) = x
+Base.@propagate_inbounds function Base.copyto!(
+    dest::FieldVector,
+    bc::Base.Broadcast.Broadcasted{FieldVectorStyle},
+)
+    for symb in propertynames(dest)
+        p = parent(getfield(_values(dest), symb))
+        copyto!(p, transform_broadcasted(bc, symb, axes(p)))
     end
     return dest
 end
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -83,12 +83,18 @@ Base.Broadcast.broadcasted(
 Base.eltype(sbc::SpectralBroadcasted) =
     operator_return_eltype(sbc.op, tuplemap(eltype, sbc.args)...)
 
+@inline instantiate_args(args::Tuple) =
+    (Base.Broadcast.instantiate(args[1]), instantiate_args(Base.tail(args))...)
+@inline instantiate_args(args::Tuple{Any}) =
+    (Base.Broadcast.instantiate(args[1]),)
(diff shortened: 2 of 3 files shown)


charleskawczynski:
bors try

jakebolewski:
This is a good change but I'm guessing it will not impact performance too much as it is called at the toplevel

charleskawczynski:
> This is a good change but I'm guessing it will not impact performance too much as it is called at the toplevel and recursive inference still works (up to a lower limit).

Yeah, this isn't fixing the problem I'm seeing..

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Fields/fieldvector.jl
```diff
@@ -179,6 +178,37 @@ end
     return dest
 end
 
+# Recursively call column() on broadcast arguments in a way that is statically reducible by the optimizer
+# see Base.Broadcast.preprocess_args
+@inline tsfmb_args(args::Tuple, inds...) = (
```
charleskawczynski, line 183:
```suggestion
# Recursively call trasnform_bc_args() on broadcast arguments in a way that is statically reducible by the optimizer
# see Base.Broadcast.preprocess_args
@inline trasnform_bc_args(args::Tuple, inds...) = (
```

jakebolewski:
If you are chasing field vector allocations for the implicit solver, isn't a potential issue the fact that we have to densify the Field vector  in calling ldiv! for QRfact?

charleskawczynski:
> If you are chasing field vector allocations for the implicit solver, isn't a potential issue the fact that we have to densify the Field vector in calling ldiv! for QRfact?

Last time I looked at the flame graph for the implicit solve, the linear solve was a relatively small fraction of the overall `step!`. Are you saying these changes may make that worse?

jakebolewski:
No just another source of potential allocations (I think that is why get / setindex is required)

jakebolewski:
Yes for the land model

charleskawczynski:
FieldVectors seem like an awkward place to store single scalars. Are these scalars updated via time stepping? cc @kmdeck

jakebolewski:
why is it an issue?  probably you are just missing an unwrap call

we do the same thing in the FD code with Refs for recursive broadcasts

kmdeck:
> FieldVectors seem like an awkward place to store single scalars. Are these scalars updated via time stepping? cc @kmdeck

yes, they would be timestepped. this arises because some of variables satisfy ODEs. For example, in a column simulation, the amount of snow on the surface is a scalar. There are are also quantities that we want to compute and store (in a field vector, with fields, and arrays) that are scalars ("aux" variables). 

However, speaking with @jb-mackay about this recently, he pointed out that in a 3d land simulation, the snow will be a 2d array of values (no flow between them, still ODEs). If all of our scalars are tied to a point on the surface, it may be cleaner overall to have it be a vector in each case (single element for column simulations vs many elements in a 3d simulation) and use broadcasting. and maybe then we dont need to use scalars. 

Im very open to discussing this more! Ben and I are hoping to demo some of our interface to the team next week as well, and that would be a good forum to bring up other cases where time-stepped scalars might arise in the land model.

charleskawczynski:
Ok, good to know, I'll take a look at `unwrap`.

src/Fields/fieldvector.jl
```diff
@@ -179,6 +178,41 @@ end
...
+end
+transform_broadcasted(fv::FieldVector, symb, axes) =
+    parent(getproperty(fv, symb))
+transform_broadcasted(x, symb, axes) = x
+@inline function Base.copyto!(
+    dest::FieldVector,
+    bc::Base.Broadcast.Broadcasted{FieldVectorStyle},
+)
+    for symb in propertynames(dest)
+        p = parent(getproperty(dest, symb))
```
jakebolewski, line 210: I think here you can call the internal field vector function `backing_array`

src/Fields/fieldvector.jl
```diff
@@ -179,6 +178,41 @@ end
...
+    axes,
+)
+    Base.Broadcast.Broadcasted(
+        bc.f,
+        transform_bc_args(bc.args, symb, axes),
+        axes,
+    )
+end
+transform_broadcasted(fv::FieldVector, symb, axes) =
+    parent(getproperty(fv, symb))
```
jakebolewski, line 203: here as well

src/Fields/fieldvector.jl
charleskawczynski: Yeah, I just changed this to `p = parent(getfield(_values(dest), symb))` and that seems to work.

charleskawczynski: I think calling `backing_array` would still be problematic, because even though you avoid calling `parent`, the `getproperty` call `unwrap`s the data and we get a method error with `copyto!(::Float64, ...)`

charleskawczynski:
bors try

jakebolewski:
Nice :rocket:

charleskawczynski:
I think that's enough changes for this PR. Okay to merge? cc @jakebolewski, @simonbyrne

src/Operators/spectralelement.jl
```diff
@@ -83,12 +83,27 @@ Base.Broadcast.broadcasted(
 Base.eltype(sbc::SpectralBroadcasted) =
     operator_return_eltype(sbc.op, tuplemap(eltype, sbc.args)...)
 
+@inline instantiate_args(args::Tuple) =
+    (Base.Broadcast.instantiate(args[1]), instantiate_args(Base.tail(args))...)
+@inline instantiate_args(args::Tuple{Any}) =
+    (Base.Broadcast.instantiate(args[1]),)
+@inline instantiate_args(::Tuple{}) = ()
+
+if VERSION >= v"1.7.0"
```
jakebolewski, line 92: I think it would be good to check if this is really needed.

jakebolewski:
Unrolling the field accesses seems to solve the bubble performance issue

src/Operators/spectralelement.jl
jakebolewski: Does it help with a particular test case?

charleskawczynski: I'm fairly certain that the main gains were from defining `Base.copyto!(::FieldVector,::Base.Broadcast.Broadcasted)`. I'll remove this.

src/Fields/fieldvector.jl
```diff
@@ -179,6 +178,41 @@ end
...
+    Base.Broadcast.Broadcasted(
+        bc.f,
+        transform_bc_args(bc.args, symb, axes),
+        axes,
+    )
+end
+transform_broadcasted(fv::FieldVector, symb, axes) =
+    parent(getfield(_values(fv), symb))
+transform_broadcasted(x, symb, axes) = x
+@inline function Base.copyto!(
```
jakebolewski, line 205: we should add an `@inbounds` here (or at the very least `@propagate_inbounds`) because we know the recursive copies are safe.

src/Operators/spectralelement.jl
simonbyrne: We require 1.7 as of #473.

charleskawczynski: Ah, yeah, lol

charleskawczynski:
Related issue: #480.

PR #482, opened 2022-01-27
Use integrator interface, add flame script

charleskawczynski:
This PR adds a flame script to the repo, and changes the monitored drivers to use OrdinaryDiffEq.jl's integrator interface. This should help make the analysis a bit more streamlined.

### DIFF
--- a/examples/hybrid/bubble_3d.jl
+++ b/examples/hybrid/bubble_3d.jl
@@ -303,12 +303,7 @@ rhs!(dYdt, Y, nothing, 0.0);
 using OrdinaryDiffEq
 Δt = 0.05
 prob = ODEProblem(rhs!, Y, (0.0, 1.0))
-
-if haskey(ENV, "CI_PERF_SKIP_RUN") # for performance analysis
-    throw(:exit_profile)
-end
-
-sol = @timev solve(
+integrator = OrdinaryDiffEq.init(
     prob,
     SSPRK33(),
     dt = Δt,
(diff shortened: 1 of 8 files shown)


simonbyrne:
Is the pprof-generated site static? Could we generate it on buildkite, and post it to a webserver?

charleskawczynski:
> Is the pprof-generated site static? Could we generate it on buildkite, and post it to a webserver?

I suspect it is, and that sounds great, but I'm not sure how to do that. Could we try to add that later?

simonbyrne:
Hmm, doesn't look like it is officially supported: https://github.com/google/pprof/issues/670

Though apparently just saving the html mostly works: https://github.com/google/pprof/issues/401#issuecomment-739576424

examples/hybrid/bubble_3d.jl
```diff
@@ -317,6 +312,12 @@ sol = @timev solve(
     progress_message = (dt, u, p, t) -> t,
 );
 
+if haskey(ENV, "CI_PERF_SKIP_RUN") # for performance analysis
+    throw(:exit_profile)
```
simonbyrne, line 316: should this just be `exit()`?

simonbyrne: oh, you catch it in the other script: in that case, maybe put the rest in a block?

charleskawczynski: I'm not sure I follow. FWIW, these lines were only moved

simonbyrne (approved):

jakebolewski:
In addition to or instead of PProf could we generate a cpuprofile file assest? 

https://github.com/julia-vscode/ChromeProfileFormat.jl

You can view these directly in vscode after downloading the file.

charleskawczynski:
> In addition to or instead of PProf could we generate a cpuprofile file assest?

Opened #483.

PR #484, opened 2022-01-27
fix slab/column broadcasting

jakebolewski:
Fixes issues uncovered in #426


jakebolewski:
I forgot we are reserving indexing for Field type values (large ntuples, static arrays):  updated

valeriabarra (approved):

PR #487, opened 2022-01-28
save cpuprofile file for buildkite testcase flamegraph

jakebolewski:
Part of $483

Can be visualized by google trace:// viewer (part of chrome)  or through the https://github.com/microsoft/vscode-js-profile-visualizer extension in vscode.


jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try

perf/flame.jl
```diff
@@ -28,6 +28,15 @@ prof = Profile.@profile begin
...
 
-import PProf
-PProf.pprof()
-# http://localhost:57599/ui/flamegraph?tf
+if !isempty(get(ENV, "CI_PERF_CPUPROFILE", ""))
+    import ChormeProfileFormat
+    output_path =
+        mkpath(get(ENV, "CI_OUTPUT_DIR", joinpath(@__DIR__, "output")))
+    testname = splitext(splitdir(filename)[end])[1]
+    cpufile = joinpath(output_path, testname * ".cpuprofile")
```
charleskawczynski, line 36: I think the profile is usually saved as `profile.pb.gz`

charleskawczynski:
Cool, I didn't realize it'd be this easy 🙂

perf/flame.jl
jakebolewski: this is for the chrometrace format  extension: https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit

the writer just grabs the latest profile data from the system

jakebolewski:
Anything that can read the chrome trace format should be able to read this (I haven't tried with speedscope but it should work?).

jakebolewski:
But yes if you install the extension and click on the file it should load in VSCode (kind of slow for larger traces so we'll see).

jakebolewski:
Screenshot from 2022-01-28 10-27-53
Screenshot from 2022-01-28 10-27-38

jakebolewski:
bors try

not dong so great with typos this morning

jakebolewski:
we are going to have to unroll all this tuple access internally to have any hope for performance lol

charleskawczynski:
> we are going to have to unroll all this ntuple field access internally to have any hope for performance lol

I think this is fixed in #480

jakebolewski:
Screenshot from 2022-01-28 10-36-44

to get the flamegraph click on the flame icon next to the funciton filter query input

jakebolewski:
Calling Pkg.develop() internal to the scripts seems to be breaking incremental compilation on the cluster

jakebolewski:
bors try

jakebolewski:
seems to work, to make it nicer we'll probably have to upstream changes so we can save / set the relative prefix paths.

charleskawczynski:
This is triggering a lot of `incremental compilation may be fatally broken for this module` in the allocations script

perf/allocs_per_case.jl
```diff
@@ -1,6 +1,5 @@
 # Launch with `julia --project --track-allocation=user`
 import Pkg
-Pkg.develop(path = ".")
```
charleskawczynski, line 3: Are we sure that this is safe? Won't this lead to analyzing the latest release and not the PR?

perf/output/bubble_2d.cpuprofile
```diff
@@ -0,0 +1 @@
...000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000]}
```
charleskawczynski: Shouldn't we not check this in?

perf/allocs_per_case.jl
jakebolewski: This is where in incremental compilation issue is coming from.  I don't think we can run this concurrently with the other test cases so we'll have to run sequentially at the end.

jakebolewski: We already `dev` the ClimaCore package at the toplevel

jakebolewski:
bors try

perf/allocs_per_case.jl
charleskawczynski: Yeah, but ReportMetrics.jl calls
```julia
run(`julia --project=perf perf/allocations_per_case.jl`)
```
And I think it's that is the session that needs the `Pkg.develop`

jakebolewski: I think the real fix for the incremental compilation issue if running in the same project do these analysis sequentially unfortunately , so I added back in the develop because it is nicer for running things locally.

charleskawczynski (approved):

PR #489, opened 2022-01-28
Bump version

charleskawczynski:
Bump version for new release


simonbyrne:
I think we need to bump all the subpackages see #490

PR #490, opened 2022-01-28
tag ClimaCore 0.7

simonbyrne:


charleskawczynski (approved):
Much better than mine, thank you 😂

simonbyrne:
bors try

simonbyrne:
bors try

perf/allocs_per_case.jl
```diff
@@ -1,6 +1,6 @@
 # Launch with `julia --project --track-allocation=user`
-import Pkg
-Pkg.develop(path = abspath(joinpath(@__DIR__, "..")))
+EXAMPLE_DIR = joinpath(dirname(@__DIR__), "examples")
+EXAMPLE_DIR in LOAD_PATH || pushfirst!(LOAD_PATH, EXAMPLE_DIR)
```
charleskawczynski, line 3: I think there's an issue with using `LOAD_PATH` when we have ClimaCore as an env dep. IIUC, we need to use `Pkg.develop` in this case. Is there a reason we can't stick with what we had?

simonbyrne:
bors try

perf/allocs_per_case.jl
simonbyrne: The problem was it couldn't resolve the ClimaCorePlots correctly. So if this doesn't work, we will have to check in the Manifest.toml

simonbyrne:
bors try

perf/allocs_per_case.jl
simonbyrne: @charleskawczynski that seemed to work: can you check the output and see how it looks?

charleskawczynski: Ah, I see. Yeah, this config looks fine, and the allocations monitoring looks fine, too.

PR #497, opened 2022-01-31
Add wrapper functions from tempestremap

simonbyrne:


simonbyrne:
Included as part of #510

PR #498, opened 2022-01-31
remove scalar index access to FieldVectors

jakebolewski:
Still need to sort through the implicit solver fallbacks


jakebolewski:
It is looking like to make this work we'll need to remove BlockArrays and define the necessary ArrayInterfaces ourselves as there are too many cases where it falls back to using abstract array / scalar indexing

jakebolewski:
Ok I have a better idea about what Implict solver interfaces we are relying on, we should discuss a better design so we can make progress on the distributed stuff next month.

PR #500, opened 2022-02-01
define mean on a Field, normalize norms by default

### LINKED ISSUE #474: Should norms be non-dimensional?
We're a bit inconsistent on how our norms on fields work: currently we use:
- l1 norm = ∑_i |x_i| * w_i * J_i ≈ ∫ |x| dΩ
- l2 norm = sqrt(∑_i |x_i|^2 * w_i * J_i) ≈ sqrt(∫ |x|^2 dΩ)
- linf norm = maximum_i |x_i| ≈ sup |x|

This implies that the dimensions are inconsistent across norms:
- l1 has units of x multiplied by the spatial unit (e.g. m^2 for 2D)
- l2 has units of x multiplied by the sqrt(spatial unit) (e.g. m for 2D)
- linf has units of x

To make these consistent, we could non-dimensionalize these by dividing by the same norm of a unit scalar, i.e.
- l1 norm = ∑_i |x_i| * w_i * J_i  /  ∑_i w_i * J_i ≈ ∫ |x| dΩ / ∫ 1 dΩ
- l2 norm = sqrt(∑_i |x_i|^2 * w_i * J_i /  ∑_i w_i * J_i ) ≈ sqrt(∫ |x|^2 dΩ / ∫ 1 dΩ)
- linf norm = maximum_i |x_i| ≈ sup |x|

simonbyrne:
Fixes #474.


src/Fields/mapreduce.jl
```diff
@@ -32,16 +32,33 @@ Base.minimum(field::Field) = minimum(identity, field)
 Base.extrema(fn, field::Field) = (minimum(fn, field), maximum(fn, field))
 Base.extrema(field::Field) = extrema(identity, field)
 
-function LinearAlgebra.norm(field::Field, p::Real = 2)
+function Statistics.mean(fn, field::Field)
+    Base.sum(fn, field) / Base.sum(weighted_jacobian(field))
+end
+
+
+function LinearAlgebra.norm(field::Field, p::Real = 2; normalize = true)
```
valeriabarra, line 40: Thanks for adding this. It would be good to add a docstring to document it

src/Fields/Fields.jl
```diff
@@ -198,6 +207,14 @@ function Base.zeros(::Type{FT}, space::AbstractSpace) where {FT}
 end
 Base.zeros(space::AbstractSpace) = zeros(Spaces.undertype(space), space)
 
+"""
+    ones([FT, ]space::AbstractSpace)
```
jakebolewski, line 211: FT other than the space undertype FT currently doesn't work

simonbyrne: okay, maybe we should just leave that undocumented.

jakebolewski: yeah it might be confusing because it differs a bit from Base behavior

valeriabarra (approved):

PR #504, opened 2022-02-01
enable opt-in operator thread parallelism

### LINKED ISSUE #503: Add Base Threading support
Add optional threading support using `Threads.@threads`.  By default julia 1.7 will start with 1 thread so users can opt into threading with `julia --threads=N`.  Support horizontal / vertical threading for 3D and vertical only for 2D slice models to start with. 

At a later point we will want to transition to `Folds.jl` infrastructure to support reductions and tunable workload scheduling but the current examples don't benefit from the added complexity quite yet.

jakebolewski:
To address the question we need to do it this way to make our static analysis tests continue to work (the optimizer can const prop and eliminate the branch by default, so everything should still work as before until the method is redefined). 

This is just to get us going and the API will change in the future.

Closes #503

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -452,12 +452,13 @@ steps:
   - label: ":computer: baroclinic wave"
     key: "cpu_baroclinic_wave"
     command:
-      - "julia --color=yes --project=examples examples/3dsphere/baroclinic_wave.jl"
+      - "julia --threads=4 --color=yes --project=examples examples/3dsphere/baroclinic_wave.jl"
     artifact_paths:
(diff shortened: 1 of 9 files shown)


jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try-

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
I think this should be good to go.  More threading can be enabled once the implicit solver cleanup PR goes through.

PR #506, opened 2022-02-02
Define `deepcopy_internal` for Fields

charleskawczynski:
This is needed for temporal convergence tests. Thanks for the help @jakebolewski!

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -135,6 +135,14 @@ _deepcopy_internal(x::Field, stackdict::IdDict) =
 Base.deepcopy_internal(fv::FieldVector{T}, stackdict::IdDict) where {T} =
     FieldVector{T}(map(x -> _deepcopy_internal(x, stackdict), _values(fv)))
 
+#=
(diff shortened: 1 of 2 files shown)


src/Fields/fieldvector.jl
```diff
@@ -135,6 +135,9 @@ _deepcopy_internal(x::Field, stackdict::IdDict) =
 Base.deepcopy_internal(fv::FieldVector{T}, stackdict::IdDict) where {T} =
     FieldVector{T}(map(x -> _deepcopy_internal(x, stackdict), _values(fv)))
 
+Base.deepcopy_internal(x::Field, stackdict::IdDict) =
```
jakebolewski, line 138: Can we add a comment for posterity why we need this?

jakebolewski (approved):

src/Fields/fieldvector.jl
jakebolewski: Also a single test would be good, you could just test the === of spaces of deepcopied fields (and !=== their field_values)

PR #507, opened 2022-02-03
Better `norm` docstring

valeriabarra:
While working on the new tutorial, I realized the docstring for `norm` was not entirely correct and matching the definition (see for instance the [Wiki](https://en.wikipedia.org/wiki/Lp_space#Lp_spaces_and_Lebesgue_integrals) page)

### DIFF
--- a/src/Fields/mapreduce.jl
+++ b/src/Fields/mapreduce.jl
@@ -8,15 +8,18 @@ weighted_jacobian(space::Spaces.AbstractSpace) =
 """
     sum([f=identity,]v::Field)
 
-Approximate integration of `v` or `f.(v)` over the domain. This is computed by
-summation of the field values multiplied by the Jacobians and the quadrature weights:
+Approximate integration of `v` or `f.(v)` over the domain. In an `AbstractSpectralElementSpace`,
+an integral over the entire space is computed by summation over the elements of the integrand
+multiplied by the Jacobian determinants and the quadrature weights at each node within an element.
+Hence, `sum` is computed by summation of the field values multiplied by the Jacobian determinants
+and quadrature weights:
 
 ```math
 \\sum_i f(v_i) W_i J_i
 \\approx
 \\int_\\Omega f(v) \\, d \\Omega
 ```
-where ``v_i``` is the value at each node, and ``f``` is the identity function if not specified.
+where ``v_i`` is the value at each node, and ``f`` is the identity function if not specified.
 """
 function Base.sum(field::Union{Field, Base.Broadcast.Broadcasted{<:FieldStyle}})
     Base.reduce(
@@ -47,15 +50,16 @@ Base.extrema(field::Field) = extrema(identity, field)
 """
     mean([f=identity, ]v::Field)
 
-The mean value of `field` or `f.(field)` over the domain, weighted by area. This is computed by
-summation of the field values multiplied by the Jacobians and the quadrature weights:
+The mean value of `field` or `f.(field)` over the domain, weighted by area.
+Similar to `sum`, in an `AbstractSpectralElementSpace`, this is computed by
+summation of the field values multiplied by the Jacobian determinants and quadrature weights:
 
 ```math
 \\frac{\\sum_i f(v_i) W_i J_i}{\\sum_i W_i J_i}
 \\approx
 \\frac{\\int_\\Omega f(v) \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}
 ```
-where ``v_i`` is the value at each node, and ``f`` is the identity function if not specified.
+where ``v_i`` is the Field value at each node, and ``f`` is the identity function if not specified.
 """
 function Statistics.mean(field::Field)
     Base.sum(field) / Base.sum(weighted_jacobian(field))
@@ -67,25 +71,33 @@ end
 """
     norm(v::Field, p=2; normalize=true)
 
-The approximate ``L^p`` norm of `v`. This is computed by summation of the norms
-of the field values multiplied by the Jacobians and the quadrature weights. If
-`normalize=true` (the default), then it is divided by the sum of the weights:
+The approximate ``L^p`` norm of `v`, where ``L^p`` represents the space of measurable
+functions for which the p-th power of the absolute value is Lebesgue integrable, that is:
 ```math
-\\left(\\frac{\\sum_i \\|v_i\\|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
-\\approx
-\\left(\\frac{\\int_\\Omega \\|v\\|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
+\\| v \\|_p = \\left( \\int_\\Omega |v|^p d \\Omega \\right)^{1/p}
 ```
-where ``v_i`` is the value at each node. If `p=Inf`, then the norm is the
-maximum of the absolute values
+where ``|v|`` is defined to be the absolute value if ``v` is a scalar-valued Field, or the 2-norm
+if it is a vector-valued Field or composite Field (see [LinearAlgebra.norm](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm)).
+Similar to `sum` and `mean`, in an `AbstractSpectralElementSpace`, this is computed by
+summation of the field values multiplied by the Jacobian determinants and quadrature weights.
+If `normalize=true` (the default), then internally the discrete norm is divided
+by the sum of the Jacobian determinants and quadrature weights:
 ```math
-\\max_i \\|v_i\\| \\approx \\sup_{\\Omega} \\|v\\|
+\\left(\\frac{\\sum_i |v_i|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
+\\approx
+\\left(\\frac{\\int_\\Omega |v|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
+```
+If `p=Inf`, then the norm is the maximum of the absolute values
+```math
+\\max_i |v_i| \\approx \\sup_{\\Omega} |v|
 ```
 
 Consequently all norms should have the same units for all ``p`` (being the same
 as calling `norm` on a single value).
 
-If `normalize=false`, then the denominator term is omitted, and so will be
-multiplied by the length/area/volume of ``\\Omega`` to the power of ``1/p``.
+If `normalize=false`, then the denominator term is omitted, and so the result will be
+the norm as described above multiplied by the length/area/volume of ``\\Omega`` to
+the power of ``1/p``.
 """
 function LinearAlgebra.norm(field::Field, p::Real = 2; normalize = true)
     if p == 2


valeriabarra:
bors try

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,30 @@ end
...
+```
+In an `AbstractSpectralElementSpace`, an integral over the entire space is
+computed by summation over the elements of the integrand multiplied by the
+Jacobians and the quadrature weights.
+If `normalize=true` (the default), then it is divided by the sum of the weights:
+```math
+\\left(\\frac{\\sum_i |v_i|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
 \\approx
-\\left(\\frac{\\int_\\Omega \\|v\\|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
+\\left(\\frac{\\int_\\Omega |v|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
```
simonbyrne, line 82: The reason I had `\|` instead of `|` is that for non-scalar-valued fields we call `norm` at each point. This should probably be explained a bit better.

valeriabarra: It looks odd to have a norm inside of a norm definition. It looks circular.... Should we just add a subscript to denote the component then? Like 
```
\\int_\\Omega |v_i|^p \\, d \\Omega
```
In case of a scalar-valued function then `v_i \equiv v`

jakebolewski: its a recursive norm (like Base julia), what Simon had looks like it aligns more with how norm is documented in `LinearAlgebra.norm`

simonbyrne: See https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm

valeriabarra: From what I see in the [`LinearAlgebra.norm` docs](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm) they define the l^2 norm (small l, not capital L. Suitable for discrete [sequence spaces](https://en.wikipedia.org/wiki/Lp_space#General_%E2%84%93p-space), not L^2 continuous function spaces like what we are trying to define here) and they also have absolute values of each component in the definition

simonbyrne: I'm happy to switch notation so that the inner norm uses |v|, but we should explain that this will call `norm`, as per the link:
> ∣a_i∣| the [norm](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm) of a_i.

> Since the p-[norm](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm) is computed using the norms of the entries of A, the p-norm of a vector of vectors is not compatible with the interpretation of it as a block vector in general if p != 2.

valeriabarra:
bors try

src/Fields/mapreduce.jl
valeriabarra: I added a clarification about vector-valued Vs scalar Fields in [[b72dfb9](https://github.com/CliMA/ClimaCore.jl/pull/507/commits/b72dfb965707503948e78fbe0a244a02d51ba3a6)](https://github.com/CliMA/ClimaCore.jl/pull/507/commits/b72dfb965707503948e78fbe0a244a02d51ba3a6). If we want to be more precise, we should use a double index. One for the Field component (I would personally call that `c`, unlike what we have used for the `IJKFVH` DataLayout) and one for the nodes

jakebolewski: A recursive norm is calling calling `norm_p(vector_val)` it's not component wise?   Maybe I'm confused about how it is worded.

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,32 @@ end
...
-The approximate ``L^p`` norm of `v`. This is computed by summation of the norms
-of the field values multiplied by the Jacobians and the quadrature weights. If
-`normalize=true` (the default), then it is divided by the sum of the weights:
+The approximate ``L^p`` norm of `v`, where ``L^p`` represents the space of measurable
+functions for which the p-th power of the absolute value is Lebesgue integrable, that is:
 ```math
-\\left(\\frac{\\sum_i \\|v_i\\|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
+\\| v \\|_p = \\left( \\int_\\Omega |v_i|^p d \\Omega \\right)^{1/p}
+```
+where ``v_i`` is the value of each component of the `Field` ``v``, in case of vector-valued fields,
```
simonbyrne, line 75: `i` below refers to the nodes at which we have values, not components.

simonbyrne:
Also you don't need to `bors try` to rebuild the docs, they build automatically.

src/Fields/mapreduce.jl
valeriabarra: That's why I said if we wanted to be more precise we should use two indices, one for the components and one for the node values. How would you want to describe this in a way that is mathematically correct to describe the L^p function norm and reflects the code?

jakebolewski: the way it is written in the `LinearAlgebra.norm` docs

simonbyrne: I think it might make more sense to draw the distinction between the field as a function on Omega, and as a discrete set of node values. I don't think you need to talk about components, other than to say that `|v(x)|` is defined to be the absolute value if scalar-values, or the 2-norm if a vector or composite valued.

valeriabarra: The way they write it is suitable for array-like objects (l^2 discrete [sequences](https://en.wikipedia.org/wiki/Lp_space#General_%E2%84%93p-space)) but it is not technically suitable for [L^2 function spaces](https://en.wikipedia.org/wiki/Lp_space#Lp_spaces_and_Lebesgue_integrals), that we are defining here in a function space. I know we discretely represent the function as a Field, but since we use the integral in the definition (unlike the `LinearAlgebra.norm` docs) we have to be careful and consistent. 
image
image

valeriabarra: @simonbyrne I reworded this following your suggestion and squashed. I also agree that we should definitely draw the distinction between the field as a function on Omega, and as a discrete set of node values. But it should perhaps be done in a broader introduction to the (Fields?) API, not in the norm function docstring

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,32 @@ end
...
-The approximate ``L^p`` norm of `v`. This is computed by summation of the norms
-of the field values multiplied by the Jacobians and the quadrature weights. If
-`normalize=true` (the default), then it is divided by the sum of the weights:
+The approximate ``L^p`` norm of `v`, where ``L^p`` represents the space of measurable
+functions for which the p-th power of the absolute value is Lebesgue integrable, that is:
 ```math
-\\left(\\frac{\\sum_i \\|v_i\\|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
+\\| v \\|_p = \\left( \\int_\\Omega |v_i|^p d \\Omega \\right)^{1/p}
+```
+where ``|v_i|`` is defined to be the absolute value if ``v` is a scalar-valued Field, or the 2-norm
```
simonbyrne, line 75:
```suggestion
\\| v \\|_p = \\left( \\int_\\Omega |v|^p d \\Omega \\right)^{1/p}
```
where ``|v|`` is defined to be the absolute value if ``v` is a scalar-valued Field, or the 2-norm
```

File: src/Fields/mapreduce.jl [hunk 4]
```diff
@@ -67,24 +67,32 @@ end
...
     norm(v::Field, p=2; normalize=true)
 
-The approximate ``L^p`` norm of `v`. This is computed by summation of the norms
-of the field values multiplied by the Jacobians and the quadrature weights. If
-`normalize=true` (the default), then it is divided by the sum of the weights:
+The approximate ``L^p`` norm of `v`, where ``L^p`` represents the space of measurable
+functions for which the p-th power of the absolute value is Lebesgue integrable, that is:
 ```math
-\\left(\\frac{\\sum_i \\|v_i\\|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
+\\| v \\|_p = \\left( \\int_\\Omega |v_i|^p d \\Omega \\right)^{1/p}
```
simonbyrne, line 73:
```suggestion
\\| v \\|_p = \\left( \\int_\\Omega |v|^p d \\Omega \\right)^{1/p}
```

src/Fields/mapreduce.jl
simonbyrne, line 75:
```suggestion
where ``|v|`` is defined to be the absolute value if ``v` is a scalar-valued Field, or the 2-norm
```

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,32 @@ end
...
+In an `AbstractSpectralElementSpace`, an integral over the entire space is
+computed by summation over the elements of the integrand multiplied by the
+Jacobian determinant and the quadrature weights at each node within an element.
+If `normalize=true` (the default), then internally the discrete norm is divided
+by the sum of the Jacobian determinants and weights:
+```math
+\\left(\\frac{\\sum_i |v_i|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
 \\approx
-\\left(\\frac{\\int_\\Omega \\|v\\|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
+\\left(\\frac{\\int_\\Omega |v_i|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
```
simonbyrne, line 85:
```suggestion
\\left(\\frac{\\int_\\Omega |v|^p \\, d \\Omega}{\\int_\\Omega \\, d \\Omega}\\right)^{1/p}
```

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,32 @@ end
...
+functions for which the p-th power of the absolute value is Lebesgue integrable, that is:
 ```math
-\\left(\\frac{\\sum_i \\|v_i\\|^p W_i J_i}{\\sum_i W_i J_i}\\right)^{1/p}
+\\| v \\|_p = \\left( \\int_\\Omega |v_i|^p d \\Omega \\right)^{1/p}
+```
+where ``|v_i|`` is defined to be the absolute value if ``v` is a scalar-valued Field, or the 2-norm
+if it is a vector-valued Field or composite Field, at each node.
+In an `AbstractSpectralElementSpace`, an integral over the entire space is
+computed by summation over the elements of the integrand multiplied by the
+Jacobian determinant and the quadrature weights at each node within an element.
```
simonbyrne, line 79: I like this addition, but we should move it to the docs for `sum`, and then refer to it here and in `mean`.

src/Fields/mapreduce.jl
```diff
@@ -67,24 +67,32 @@ end
...
-\\max_i \\|v_i\\| \\approx \\sup_{\\Omega} \\|v\\|
+\\max_i |v_i| \\approx \\sup_{\\Omega} |v|
 ```
 
 Consequently all norms should have the same units for all ``p`` (being the same
 as calling `norm` on a single value).
 
-If `normalize=false`, then the denominator term is omitted, and so will be
+If `normalize=false`, then the denominator term is omitted, and so the actual norm value will be
 multiplied by the length/area/volume of ``\\Omega`` to the power of ``1/p``.
```
simonbyrne, line 96:
```suggestion
If `normalize=false`, then the denominator term is omitted, and so the result will be the norm as described above
multiplied by the length/area/volume of ``\\Omega`` to the power of ``1/p``.
```

valeriabarra:
Hi @simonbyrne , I applied the suggestions from the review. If/when approved, I can squash commits

simonbyrne (approved):
Looks great, thanks!

simonbyrne:
The cluster is down for maintenance today, so no buildkite

valeriabarra:
> The cluster is down for maintenance today, so no buildkite

ah yeah, I've just seen your reminder on Slack now. I had forgotten about the scheduled maintenance for today. I'll `bors r+` when available again

Issue #282, opened 2021-11-04
General Remapping Operator Construction


jb-mackay:
For a remapping operation from `space1` (`n1` degrees of freedom/nodes) to`space2` (`n2` degrees of freedom), a `n2 x n1` sparse matrix must be constructed.

Pseudocode from @simonbyrne: 
```
using SparseArrays
n1 = length(space1.local_geometry) # number of nodes
n2 = length(space2.local_geometry) 

M = spzeros(n2,n1)

# fill in M; this gets tricky:
for elem1 in 1:nlocalelems(space1), elem2 in 1:nlocalelems(space2)
    # in FV, compute the overlapping area
    M[elem2, elem1] = area

    # in SE
    for n1 in nodes(elem1), n2 in nodes(elem2)
        # compute the integral of the product Lagrange polynomials associated with n1 and n2
        # space 1           |elem1       |
        # node 1                        x 
        # space 2    |elem2       |
        # node 2         x
        slab(data1, elem1)[n1]

        l1 = linearindex(elem1, n1)
        l2 = linearindex(elem2, n2)
        M[l2,l1] = integral
    end
end
# M should now be the "overlap matrix"
# divide M through by WJ of the target to get the full remapping matrix
```

This is complete for FV -> FV remapping (#251) and needs to be generalized.

PR #508, opened 2022-02-03
2D Spectral Element <-> Spectral Element Remapping

jb-mackay:
Generalizes the 2D remapping infrastructure to handle any polynomial order finite element space.

Bug fix for `nyelems`. Adds assertion checking so that remapping operators can only be applied to fields on the associated space.

### DIFF
--- a/src/Operators/remapping.jl
+++ b/src/Operators/remapping.jl
@@ -1,3 +1,4 @@
+using ..Meshes
 using ..Spaces:
     AbstractSpace, SpectralElementSpace1D, SpectralElementSpace2D, Quadratures
 using ..Topologies: Topology2D, IntervalTopology
@@ -15,6 +16,8 @@ end
     LinearRemap(target::AbstractSpace, source::AbstractSpace)
 
 A remapping operator from the `source` space to the `target` space.
+
+See [Ullrich2015](@cite) eqs. 3 and 4.
 """
 function LinearRemap(target::AbstractSpace, source::AbstractSpace)
     R = linear_remap_op(target, source)
@@ -24,7 +27,7 @@ end
 """
     remap(R::LinearRemap, source_field::Field)
 
-Applies the remapping operator `R` to `source_field`. Outputs a new field on the target space mapped to by `R`.
+Applies `R` to `source_field` and outputs a new field on the target space.
 """
 function remap(R::LinearRemap, source_field::Field)
     target_space = R.target
@@ -38,6 +41,7 @@ end
 Applies the remapping operator `R` to `source_field` and stores the solution in `target_field`.
 """
 function remap!(target_field::Field, R::LinearRemap, source_field::Field)
+    @assert (R.source == axes(source_field) && R.target == axes(target_field)) "Remap operator and field space dimensions do not match."
     mul!(vec(parent(target_field)), R.map, vec(parent(source_field)))
     return target_field
 end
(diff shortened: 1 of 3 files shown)


jb-mackay:
Issue #282 can be closed with this PR; more specific issues can be created for future in-house development of spherical remapping.

src/Operators/remapping.jl
```diff
@@ -66,56 +67,96 @@ function overlap(
...
+                kron(Y_ov[tgt_y, src_y], X_ov[tgt_x, src_x])
+        end
+    end
+    return W
 end
 
-function overlap_weights(
+function ov_idx(space, i)
+    mod(i - 1, nelems_x) + 1
+end
```
jb-mackay, line 117: Need to remove.

src/Operators/remapping.jl
```diff
@@ -138,110 +179,82 @@ function overlap_weights(
...
-    source::S,
-) where {
-    T <: SpectralElementSpace1D{<:IntervalTopology},
-    S <: SpectralElementSpace1D{<:IntervalTopology},
-}
-    QS_t = Spaces.quadrature_style(target)
-    QS_s = Spaces.quadrature_style(source)
-    Nq_t = Quadratures.degrees_of_freedom(QS_t)
-    Nq_s = Quadratures.degrees_of_freedom(QS_s)
-    if Nq_t >= Nq_s
```
jb-mackay, line 152: Moved this logic to within `x_overlap` and `y_overlap` - feels more transparent than having it here and then doing a matrix transposition. We may want to add a method that constructs the "inverse" map (target -> source) in practice, in which case transposing (and re-weighting) an existing (source -> target) remapping operator would be more efficient than building a new operator from scratch).

src/Operators/remapping.jl
```diff
@@ -138,110 +179,82 @@ function overlap_weights(
...
-}
-    # Calculate element overlap pattern in x-dimension
-    # X_ov[i,j] = overlap length along x-dimension between target elem i and source elem j
-    X_ov = fv_x_overlap(target, source)
-
-    # Calculate element overlap pattern in y-dimension
-    Y_ov = fv_y_overlap(target, source)
-
-    return kron(Y_ov, X_ov)
-end
```
jb-mackay, line 175: Is there any reason to retain special cases for remapping with FV spaces?

src/Operators/remapping.jl
```diff
@@ -138,110 +179,82 @@ function overlap_weights(
...
         min_i, max_i = ycomponent(vertices_i[v1]), ycomponent(vertices_i[v2])
-        for j in 1:nsource
+        for j in 1:nelems_s
             vertices_j = Topologies.vertex_coordinates(
                 source_topo,
                 elem_idx(j, nx_source),
             )
             min_j, max_j =
                 ycomponent(vertices_j[v1]), ycomponent(vertices_j[v2])
             min_ov, max_ov = max(min_i, min_j), min(max_i, max_j)
```
jb-mackay, line 213: Everything from this line to the end of the function is shared with `x_overlap` - can box in another method or maybe two(?), one for the calculation of the interpolation matrices with basis function values at the overlap nodes, and one to fill in `W_ov` via the integration rule.

test/Operators/remapping.jl
```diff
@@ -64,6 +68,10 @@ function consistent(R::LinearRemap)
     return row_sums ≈ v
 end
 
+function locally_consistent(R::LinearRemap)
+
+end
```
jb-mackay, line 73: Will probably split `locally_conservative` and `locally_consistent` into a separate PR that adds other test enhancements. I want to add some tests of non-square domains for existing 2D cases, some tests that remapping from a space to itself builds the identity, and perhaps tests that non-overlapping meshes return a matrix of zeros.

Additions to the test suite for this PR will focus solely on some basic 2D SE <-> SE cases checking conservation, consistency, and sparsity pattern (which I think is a good stand in for checking against `R_true` in non-trivial cases).

src/Operators/remapping.jl
valeriabarra: Wouldn't we want to retain that for the coupling with the ocean or am I not recalling correctly what was the plan?

src/Operators/remapping.jl
valeriabarra: Yes, sounds like a good idea to have a map inversion without having to build a new one from scratch

src/Operators/remapping.jl
jb-mackay: Right; a question is, would we (and can we) define the ClimaCore space that corresponds to the Oceananigans one and feed that into the ClimaCore remapping functionality? As it stands, of course we need to have some adapter to talk to them anyways, so the question is where does that connect to this code that actually computes the remap operator.

src/Operators/remapping.jl
jb-mackay: Awesome. Can do that in a future PR.

test/Operators/remapping.jl
```diff
@@ -491,4 +556,164 @@ end
...
+                Xov = Operators.x_overlap(space2, space1)
+                Yov = Operators.y_overlap(space2, space1)
+                @test Xov == Yov
+                I_true = [1, 2, 3, 2, 3, 4, 3, 4, 5, 4, 5, 6]
+                J_true = [1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4]
+                @test sparsity_pattern(Xov) == (I_true, J_true)
+
+                # SE1 -> SE2
+                R = LinearRemap(space2, space1)
+                I_true = [1, 2, 3, 4, 5, 7, 13, 14, 17,
```
jb-mackay, line 650: The formatter does a bad job here and turns this into a new line for each entry in the vector. Is there a better way @charleskawczynski ?

src/Operators/remapping.jl
```diff
@@ -38,6 +38,7 @@ end
 Applies the remapping operator `R` to `source_field` and stores the solution in `target_field`.
 """
 function remap!(target_field::Field, R::LinearRemap, source_field::Field)
+    @assert (R.source == axes(source_field) && R.target == axes(target_field)) "Remap operator and field spaces do not match."
```
valeriabarra, line 41:
```suggestion
    @assert (R.source == axes(source_field) && R.target == axes(target_field)) "Remap operator and field space dimensions do not match."
```

src/Operators/remapping.jl
```diff
@@ -65,116 +66,50 @@ Computes local weights of the overlap mesh for `source` to `target` spaces.
...
-                    # (integral of src_basis * tgt_basis on overlap) / (reference elem length * overlap elem length)
-                    J_ov[targ_idx, src_idx] =
-                        (w' * (I_mat_t[:, k] .* I_mat_s[:, l])) ./ 2 *
-                        (max_ov - min_ov)
-                end
-            end
-        end
-    end
-    return J_ov
+    return x_overlap(target, source)
```
valeriabarra, line 73: I think the convention is to use `!` if the function modifies its argument. Is this what this function does to `target`?

valeriabarra: Please document these functions

jb-mackay: > I think the convention is to use `!` if the function modifies its argument. Is this what this function does to `target`?

No, the `target` space is not modified. `x_overlap` generates a sparse matrix. I'll be sure to add some docs!

test/Operators/remapping.jl
charleskawczynski: Sorry, just saw this. We could turn the formatter off. See, e.g., [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/post_processing/compute_mse.jl#L22-L60)

src/Operators/remapping.jl
```diff
@@ -65,116 +68,58 @@ Computes local weights of the overlap mesh for `source` to `target` spaces.
...
+    for i in 1:Topologies.nlocalelems(target)
+        global_tgt = (Nq_t^2 * (i - 1) + 1):(Nq_t^2 * i) # global indices of target dof
+        ix = mod(i - 1, tgt_elems_x) + 1
+        iy = Int(ceil(i / tgt_elems_x))
+        tgt_x = (Nq_t * (ix - 1) + 1):(Nq_t * ix) # indices of target dof in Xov
+        tgt_y = (Nq_t * (iy - 1) + 1):(Nq_t * iy) # indices of target dof in Yov
+        for j in 1:Topologies.nlocalelems(source)
+            global_src = (Nq_s^2 * (j - 1) + 1):(Nq_s^2 * j)
+            jx = mod(j - 1, src_elems_x) + 1
+            jy = Int(ceil(j / src_elems_x))
```
simonbyrne, line 104: similar to aboce

src/Operators/remapping.jl
```diff
@@ -65,116 +68,58 @@ Computes local weights of the overlap mesh for `source` to `target` spaces.
...
 
 function overlap(
     target::T,
     source::S,
 ) where {
-    T <: SpectralElementSpace1D{<:IntervalTopology},
-    S <: SpectralElementSpace1D{<:IntervalTopology},
+    T <: SpectralElementSpace2D{<:Topology2D},
+    S <: SpectralElementSpace2D{<:Topology2D},
 }
```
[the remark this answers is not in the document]
simonbyrne: ah ok, that makes sense

simonbyrne (approved):

PR #509, opened 2022-02-03
Update pipeline.yml

jakebolewski:
Test new MPICH 4.0 build


jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
We'll now defer to MPICH for the CPU builds as that is what Julia MPI.jl uses by default.

PR #510, opened 2022-02-03
Add NetCDF support

simonbyrne:
I've added this to ClimaCoreTempestRemap, since the format we use is specifically aimed to be used with it.


simonbyrne:
Now working!

I'll get some docs written up soon.

lib/ClimaCoreTempestRemap/src/wrappers.jl
```diff
@@ -0,0 +1,30 @@
+function rll_mesh(filename::String; nlat = 90, nlon = round(Int, nlat * 1.6))
+    run(
```
jakebolewski, line 2: I think we should use `pipeline` here so we can redirect `stdout`, `stderr` to `devnull` if `verbose=false` keyword is set (default).

lib/ClimaCoreTempestRemap/test/exodus.jl
```diff
@@ -0,0 +1,51 @@
+import ClimaCore
+using ClimaCore: Geometry, Meshes, Domains, Topologies, Spaces
+using NCDatasets
+using TempestRemap_jll
+using Test
+using ClimaCoreTempestRemap
+
+OUTPUT_DIR = mkpath(get(ENV, "CI_OUTPUT_DIR", tempname()))
+
+@testset "mesh coordinates ClimaCore > TempestRemap" begin
```
jakebolewski, line 10: are these tests needed given the updated netcdf tests?

lib/ClimaCoreTempestRemap/test/netcdf.jl
```diff
@@ -0,0 +1,172 @@
...
+
+    datafile_rll = joinpath(OUTPUT_DIR, "data_rll_3d.nc")
+    ClimaCoreTempestRemap.apply_remap(
+        datafile_rll,
+        datafile_cc,
+        weightfile,
+        ["xlat", "xz"],
+    )
+
+    nc_rll = NCDataset(datafile_rll)
```
jakebolewski, line 161: use the `NcDataset() do ... end` form here to ensure that the file is closed consistently?

lib/ClimaCoreTempestRemap/src/netcdf.jl
```diff
@@ -0,0 +1,213 @@
...
+function write_cgll_data!(
+    var,
+    field,
+    space::Spaces.SpectralElementSpace2D,
+    extraidx...,
+)
+    nodes = Spaces.unique_nodes(space)
+    data = Fields.field_values(field)
+    for (col, ((i, j), e)) in enumerate(nodes)
+        var[col, extraidx...] = slab(data, e)[i, j]
```
jakebolewski, line 196: do we need to check if these are scalars?

jakebolewski:
Looks good, people can later can add wrapper functions to compose the sequence of operations they want.

lib/ClimaCoreTempestRemap/test/exodus.jl
simonbyrne: yes, they check that our maps align.

lib/ClimaCoreTempestRemap/src/netcdf.jl
```diff
@@ -0,0 +1,213 @@
...
+    nc::NCDataset,
+    space::Spaces.CenterFiniteDifferenceSpace,
+)
+    nlevels = Spaces.nlevels(space)
+
+    if haskey(nc, "z")
+        if size(nc["z"]) != (nlevels,)
+            error("incompatible vertical dimension already exists")
+        end
+        return nothing
```
charleskawczynski, line 61: Why do we escape here if `haskey(nc, "z")`?

simonbyrne: the dimension is already defined.

I was thinking of making this function part of `write_var_cgll!`, so it could be called multiple times.

lib/ClimaCoreTempestRemap/src/netcdf.jl
simonbyrne: I've added an assert

lib/ClimaCoreTempestRemap/src/netcdf.jl
charleskawczynski: Hmm, ok. The way we do it in TC.jl is partitioned into 2 parts: "adding fields" (writing meta info), and writing fields. At first glance it looks like we could partition things the same here. The reason I suggest this is because of [this issue](https://github.com/Alexander-Barth/NCDatasets.jl/issues/135). At least for EDMF, repeatedly writing meta info was dominating the cost of the entire simulation because of the quadratic behavior with the number of variables. Obviously this is fine for only a few variables, but it may be something to consider.

simonbyrne: That seems like a good idea.

simonbyrne: I've updated the interface to be more explicit.

lib/ClimaCoreTempestRemap/src/netcdf.jl
```diff
@@ -0,0 +1,257 @@
...
+        lon[col] = coord.long
+        lat[col] = coord.lat
+    end
+    nc.attrib["node_type"] = type
+    return lat, lon
+end
+
+function def_space_coord(
+    nc::NCDataset,
+    space::Spaces.CenterFiniteDifferenceSpace,
```
charleskawczynski, line 103: I don't mean to slow things down but, should we also support `FaceFiniteDifferenceSpace `? If so, we'll need to have `zf` / `zc` defined variables, and maybe a helper to get the appropriate coordinate name

charleskawczynski: I suppose vertical velocity is one variable that someone might want

charleskawczynski: Or we could write a wrapper that interpolates to a cell center field.

charleskawczynski: Ah, this may not be needed if TempestRemap is collocated at cell centers.

docs/src/lib/ClimaCoreTempestRemap.md
```diff
@@ -4,12 +4,32 @@
 CurrentModule = ClimaCoreTempestRemap
```
valeriabarra, line 4: What does `CurrentModule` refer to in this document? 
What would be the workflow if one wanted to test/use this to output lat-long grids? Can we document it?

charleskawczynski: It's for [Documenter.jl](https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Adding-Some-Docstrings)

valeriabarra: Thanks Charlie!

lib/ClimaCoreTempestRemap/src/netcdf.jl
simonbyrne: TempestRemqp doesn't do anything to the vertical.

I'm not sure what we should call the face locations (`z_half`?). @szy21 is there a standard name that is used?

charleskawczynski (approved):
Overall this LGTM

lib/ClimaCoreTempestRemap/src/netcdf.jl
szy21: `z_half` seems to be a common name (I'm not sure if it's standard). I think as long as we all agree that half levels are on cell faces then it's fine.

szy21: `z_half` seems to be a common name (I'm not sure if it's standard). I think as long as we all agree that half levels are on cell faces then it's fine.

jiahe23:
bors try

jiahe23 (approved):

lib/ClimaCoreTempestRemap/src/netcdf.jl
simonbyrne: okay, i went with that.

docs/src/lib/ClimaCoreTempestRemap.md
```diff
@@ -33,3 +33,82 @@ overlap_mesh
...
+# Example
+
+The following example converts an OrdinaryDiffEq solution object `sol` to a netcdf file, and remaps it to an regular latitude-longitude (RLL) grid.
+
+```julia
+using ClimaCore: Geometry, Meshes, Domains, Topologies, Spaces
+using NCDatasets, ClimaCoreTempestRemap
+
+# sol is the integrator solution
+# cspace is the center extrduded space
```
valeriabarra, line 46: Thanks @simonbyrne for adding this example! 
Is there a similar approach for our spherical 2D cases?

simonbyrne: yes, it should be basically the same, except that you don't need to worry about the staggering.

valeriabarra: Thanks. So in that case in the `def_space_coord` call below, one would just use the (unique) horizontal `space`, as 
```
def_space_coord(nc, space)
```

PR #511, opened 2022-02-03
opt_limiters_solidbody.jl: Update norms and gaussian bells IC

valeriabarra:
After #500, I needed to update the norms computation in this example. 
I also homogenized the gaussian bells definition with the single bell example (`sphere/solidbody.jl`) and it gives better errors


valeriabarra:
bors try

PR #512, opened 2022-02-03
Add temporal convergence test

charleskawczynski:
This PR adds a temporal convergence test for the explicit baroclinic wave. We need to wait for [ODEConvergenceTester.jl](https://github.com/CliMA/ODEConvergenceTester.jl)'s registration to complete.

We do seem to get the correct convergence rate for SSPRK33:

```
[ Info: ––– Convergence parameters –––
[ Info: nsteps : [216.0, 432.0, 864.0, 1728.0, 3456.0]
[ Info: refinement factors : [2.0, 2.0, 2.0, 2.0]
[ Info: dts : [4.0, 2.0, 1.0, 0.5, 0.25]
[ Info: tfinal (rounded) : [864.0, 864.0, 864.0, 864.0, 864.0]
[ Info: –- Running convergence study... –-
@timing iteration 1: 35.272038 seconds (121.24 k allocations: 13.743 MiB, 22.60% compilation time)
@timing iteration 2: 55.694182 seconds (55.73 k allocations: 18.675 MiB)
@timing iteration 3:112.194313 seconds (111.46 k allocations: 36.414 MiB)
@timing iteration 4:211.563645 seconds (222.91 k allocations: 71.891 MiB, 0.02% gc time)
@timing iteration 5:490.880064 seconds (445.82 k allocations: 142.845 MiB)
[ Info: ––- Convergence study output ––-
[ Info: Errors : [10.543135914720834, 2.3307444518740135, 0.35735670941838943, 0.048074174273188885]
[ Info: Convergence orders : [2.1774413032595343, 2.705354052344195, 2.894030892296653]
[ Info: Expected convergence order: 3.0
```

Also, there's some decisions that must be made about how to normalize the error norm so as to not allow temperature errors (on order of 100) to dominate the norm over, say, `q_tot` (on order < 1). `compute_err_norm`, which calls `non_dimensionalize!` then `LinearAlgebra.norm` is passed into ODEConvergenceTester.jl to address this. The way we normalize each individual field is as follows (`u` and `v` are solutions computed from two different Δts):

```julia
        ∑u = sum(abs.(parent(u))) / length(parent(u))
        ∑v = sum(abs.(parent(v))) / length(parent(v))
        ∑uv = (∑u + ∑v) / 2
        if !(∑uv ≈ 0)
            u .= u ./ ∑uv
            v .= v ./ ∑uv
        end
```

It might be nice if we change our drivers to be a bit more unified to duplicate less code. This basically duplicates the explicit baroclinic wave until the integrator is created.

### DIFF
--- a/test/TimeStepping/temporal_convergence.jl
+++ b/test/TimeStepping/temporal_convergence.jl
@@ -1,0 +1,394 @@
+import Pkg
+# Pkg.add(Pkg.PackageSpec(; url = "https://github.com/CliMA/ODEConvergenceTester.jl"))
+Pkg.develop(Pkg.PackageSpec(; path = "../ODEConvergenceTester.jl"))
+Pkg.develop(Pkg.PackageSpec(; path = "."))
+
+import Logging
+import TerminalLoggers
+Logging.global_logger(TerminalLoggers.TerminalLogger())
+
+import LinearAlgebra: norm, ×
+import UnPack
+import OrdinaryDiffEq
+const ODE = OrdinaryDiffEq
+
+import ClimaCore:
+    ClimaCore, Spaces, Domains, Meshes, Geometry, Topologies, Fields, Operators
+const CC = ClimaCore
+
+non_dimensionalize!(u, v) = nothing
+function non_dimensionalize!(u::FV, v::FV) where {FV <: CC.Fields.Field}
+    if isempty(propertynames(u))
+        ∑u = sum(abs.(parent(u))) / length(parent(u))
+        ∑v = sum(abs.(parent(v))) / length(parent(v))
+        ∑uv = (∑u + ∑v) / 2
+        if !(∑uv ≈ 0)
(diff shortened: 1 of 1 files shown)


jakebolewski:
Maybe this would be better in ClimaAtmos?

charleskawczynski:
> Maybe it would be better to add this to ClimaAtmos?

I think it's probably better suited here, actually. Once the implicit solver pieces are in, this will be a good test that they're configured correctly. I think it's more numerics-focused integration test than physics-focused test. We can use a different example, or share more code when it's easier.

charleskawczynski:
I'm personally interested in this PR, so I'd like to not forget about it for when I have time to look into this more.

charleskawczynski:
I've thought about this a bit more, and perhaps it's best that we just keep unit tests in ClimaCore for the operators and split this out somehow-- The main issue I think is that there are many operators and many problems that we could configure for testing

PR #515, opened 2022-02-07
Add Cent types--complement to PlusHalf

charleskawczynski:
This PR defines a new type, `Cent`, to support `ntuple` fields, because we must be able to differentiate between
 - `getindex` to get the i-th n-tuple field (using `::Int`)
 - `getindex` to get the k-th vertical cell center (using `::Cent`)

This was pulled out from TC.jl

I still need to add the tests... I'll probably just mirror the tests from `PlusHalf`.


charleskawczynski:
bors try

charleskawczynski:
I guess we'll just be type pirates in TC. 🦜

PR #520, opened 2022-02-08
Add euclidean_distance support

### LINKED ISSUE #519: Euclidean distance
It would be nice to define interfaces for 2D and 3D Euclidean distances between `XYPoint`s and `XYZPoint`s, similar to the [`great_circle_distance`](https://github.com/CliMA/ClimaCore.jl/blob/51c2f070a0a3496fe75f8a44a8af9752c9c390fe/src/Geometry/globalgeometry.jl#L125) we have for spherical domains. 

As explained, in #517 , initially I was trying to use `hypot` between two XYPoints, as in 

```
hypot(point_1 .- point_2...)
```
but currently we do not support `.-` between XY/XYZPoints so we have to do it compoenent-wise

valeriabarra:
This will close #519

### DIFF
--- a/src/Geometry/globalgeometry.jl
+++ b/src/Geometry/globalgeometry.jl
@@ -146,6 +146,18 @@ function great_circle_distance(
     )
 end
 
+"""
+    euclidean_distance(pt1::XYPoint, pt2::XYPoint)
+
+Compute the 2D or 3D Euclidean distance between `pt1` and `pt2`.
+"""
+function euclidean_distance(
+    pt1::T,
+    pt2::T,
+) where {T <: Union{XPoint, YPoint, ZPoint, XYPoint, XZPoint, XYZPoint}}
+    return hypot((components(pt1) .- components(pt2))...)
+end
+
 # vectors
 CartesianVector(
     u::CartesianVector,
--- a/test/Geometry/geometry.jl
+++ b/test/Geometry/geometry.jl
@@ -575,6 +575,66 @@ end
 
 end
 
+@testset "1D XPoint Euclidean distance" begin
+    for FT in (Float32, Float64, BigFloat)
+        pt_1 = Geometry.XPoint(one(FT))
+        pt_2 = Geometry.XPoint(zero(FT))
+        @test Geometry.euclidean_distance(pt_1, pt_2) ≈
+              hypot((Geometry.components(pt_1) .- Geometry.components(pt_2))...) rtol =
+            2eps()
+    end
+end
+
+@testset "1D YPoint Euclidean distance" begin
+    for FT in (Float32, Float64, BigFloat)
+        pt_1 = Geometry.YPoint(one(FT))
+        pt_2 = Geometry.YPoint(zero(FT))
+        @test Geometry.euclidean_distance(pt_1, pt_2) ≈
+              hypot((Geometry.components(pt_1) .- Geometry.components(pt_2))...) rtol =
+            2eps()
+    end
+end
+
+@testset "1D ZPoint Euclidean distance" begin
+    for FT in (Float32, Float64, BigFloat)
+        pt_1 = Geometry.ZPoint(one(FT))
+        pt_2 = Geometry.ZPoint(zero(FT))
+        @test Geometry.euclidean_distance(pt_1, pt_2) ≈
+              hypot((Geometry.components(pt_1) .- Geometry.components(pt_2))...) rtol =
+            2eps()
+    end
+end
+
+@testset "2D XYPoint Euclidean distance" begin
+    for FT in (Float32, Float64, BigFloat)
+        pt_1 = Geometry.XYPoint(one(FT), one(FT))
+        pt_2 = Geometry.XYPoint(zero(FT), zero(FT))
+        @test Geometry.euclidean_distance(pt_1, pt_2) ≈
+              hypot((Geometry.components(pt_1) .- Geometry.components(pt_2))...) rtol =
+            2eps()
+    end
+end
+
+@testset "2D XZPoint Euclidean distance" begin
+    for FT in (Float32, Float64, BigFloat)
+        pt_1 = Geometry.XZPoint(one(FT), one(FT))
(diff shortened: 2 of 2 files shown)


src/Geometry/globalgeometry.jl
```diff
@@ -146,6 +146,66 @@ function great_circle_distance(
...
+end
+
+"""
+    euclidean_distance(pt1::XYPoint, pt2::XZPoint)
+
+Compute the 2D Euclidean distance between `pt1` and `pt2`.
+"""
+function euclidean_distance(
+    pt1::XYPoint{FT},
+    pt2::XZPoint{FT},
```
simonbyrne, line 168: Shouldn't these be the same type?

src/Geometry/globalgeometry.jl
```diff
@@ -146,6 +146,66 @@ function great_circle_distance(
...
+    euclidean_distance(pt1::XYPoint, pt2::XYPoint)
+
+Compute the 2D Euclidean distance between `pt1` and `pt2`.
+"""
+function euclidean_distance(
+    pt1::XYPoint{FT},
+    pt2::XYPoint{FT},
+) where {FT}
+    return hypot((components(pt1) .- components(pt2))...)
+end
```
simonbyrne, line 156: You can avoid duplicate definitions by defining a single method and use "diagonal" dispatch:
```suggestion
function euclidean_distance(
    pt1::T,
    pt2::T,
) where {T<:Union{XYPoint,XZPoint,XYZPoint}
    return hypot((components(pt1) .- components(pt2))...)
end
```

src/Geometry/globalgeometry.jl
valeriabarra: Do we want to allow for distances between non-planar points? (i.e., one point lying on the XY plane and one that is non on the XY plane?) Am I seeing this wrong?

simonbyrne: Sure, but then this definition doesn't make sense: you're effectively doing:
```
sqrt((pt1.x - pt2.x)^2 + (pt1.y - pt2.z)^2)
```

src/Geometry/globalgeometry.jl
valeriabarra: This way, it would only allow for same type. For hybrid spaces, wouldn't we want to allow a distance between two different types? Am I seeing this wrong?

src/Geometry/globalgeometry.jl
valeriabarra: Right, I was considering the missing component as an omitted zero

valeriabarra: I will have to promote them then

valeriabarra: Would the promotion I added in [6ee763b](https://github.com/CliMA/ClimaCore.jl/pull/520/commits/6ee763b2a67fe551746c7867b57f14e07b9c5893) be an ok way to take care of that?

simonbyrne: Honestly I don't think we should do promotion: is there a particular case where this is needed?

valeriabarra: Yes, it would be needed in 2D hybrid spaces, to compute radial distance for gaussian/cosine bells types of initial conditions for advection/transport cases, similar to the 2D non-hybrid cases

valeriabarra: What's wrong with the promotion?

simonbyrne: I guess I don't quite see why you need promotion for those cases? Can you give an example of how it would be used?

My main concern is that unlike vectors, for positions there is nothing "special" about the 0 plane, so its not clear that `XPoint(2)` should be the same as `XYPoint(2,0)`.

It also adds potential for unexpected behavior: e.g. accidentally computing the distance from `XPoint(1)` to `YPoint(3)`.

src/Geometry/globalgeometry.jl
simonbyrne: why would a hybrid space be different?

src/Geometry/globalgeometry.jl
valeriabarra: Ok probably I was seeing the hybrid case as needing two different types, but it may not be the case. In the case of, say a 2D hybrid space (1D CG x 1D FD) the coordinates field would be of XZPoint type, right? Then the radial distance would just be defined between two XZPoints and in that case we don't need a promotion like I was imaging. Similarly, for a 3D hybrid space (2D CG x 1D FD), the coordinates field would be XYZPoints and the distances can be computed among them. Sorry, I think the ides of the hybrid spaces and these different coordinate types as algebraic tuples and not geometric vectors (position vectors) confused me

valeriabarra: I removed it in [ab7216d](https://github.com/CliMA/ClimaCore.jl/pull/520/commits/ab7216d1dc2a07426d94845060a904fe8d4097a1). Please let me know if this is looking good and can be squashed/merged. Thanks a lot!

src/Geometry/globalgeometry.jl
```diff
@@ -146,6 +146,18 @@ function great_circle_distance(
...
 
+"""
+    euclidean_distance(pt1::XYPoint, pt2::XYPoint)
+
+Compute the 2D or 3D Euclidean distance between `pt1` and `pt2`.
+"""
+function euclidean_distance(
+    pt1::T,
+    pt2::T,
+) where {T <: Union{XYPoint, XZPoint, XYZPoint}}
```
simonbyrne, line 157: we might as well support single dimensions as well:
```suggestion
) where {T <: Union{XPoint, YPoint, ZPoint, XYPoint, XZPoint, XYZPoint}}
```

valeriabarra: Yes, I had considered that honestly

valeriabarra: I added that, related tests and squashed commits. Please let me know if it looks good to merge. Thanks!

simonbyrne (approved):

PR #527, opened 2022-02-11
Define fill! for field vectors

charleskawczynski:
The allocation monitoring shows that the fallback implementation to `fill!` is allocating a lot. This PR implements `fill!` for `FieldVector`s similar to `copyto!`.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -186,6 +186,14 @@ end
     return dest
 end
 
+@inline function Base.fill!(fv::FV, val::Real) where {FV <: FieldVector}
+    for symb in propertynames(fv)
+        fvp = parent(getproperty(fv, symb))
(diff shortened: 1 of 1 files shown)


charleskawczynski:
> Does `fill!(FV, val) = FV .= val` work instead? it should do the right thing I think falling back on the datalayout impl and we can keep it generic.

I don't think so, but I'll give that a try.

jakebolewski:
it looks like it is already falling back to the broadcasted `copyto!` so I wonder why this is slow?

The original version was fine, just don't specialize the `val` argument to Real

jakebolewski (approved):

jakebolewski:
I think Dennis did the fix I had in mind so this can be closed.

charleskawczynski:
Closing in favor of #528.

Issue #524, opened 2022-02-10
Third-order upwind FD scheme


valeriabarra:
Since a 1st-order upwind FD scheme is too dissipative (edit: typo), we want to add a 3rd-order upwind FD scheme. 

This can then be used with limiters in hybrid domains (both 2D hybrid domain, 1D CG x 1D FD, and 3D hybrid domain, 2D CG x 1D FD).

Here are a couple of references:
* Book: Hundsdorfer and Verwer - 2003 - Numerical Solution of Time-Dependent Advection-Diffusion-Reaction Equations (equations (3.27) & (3.28) for regular unstaggered upwind third-order finite differences stencil)
* Paper on staggered reconstruction: Pressel et al. - 2015 - Large-eddy simulation in an anelastic framework with closed
water and entropy balances [doi: 10.1002/2015MS000496](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015MS000496) (equations (60)-(63))

valeriabarra:
@simonbyrne who can look into this?

valeriabarra:
This was merged, but it currently works only with periodic meshes. If anyone wants to work on adding the boundary support, they can feel free to re-open this issue or create a new one.

PR #533, opened 2022-02-15
Advection & Divergence convergence unit tests

valeriabarra:
Since I was confused about the order of convergence of our current `AdvectionC2C` operator from its docs and stencil formula, I added unit tests that check this, both in a 1D FD space and 2d hybrid, i.e., 1D SE x 1D FD. 

I also wanted to add the same for `AdvectionF2F`, but I need some clarifications about this operator as well

### DIFF
--- a/test/Operators/hybrid/2d.jl
+++ b/test/Operators/hybrid/2d.jl
@@ -16,6 +16,9 @@ import ClimaCore:
 import ClimaCore.Domains.Geometry: ⊗
 import ClimaCore.Utilities: half
 
+convergence_rate(err, Δh) =
+    [log(err[i] / err[i - 1]) / log(Δh[i] / Δh[i - 1]) for i in 2:length(Δh)]
+
 function hvspace_2D(;
     xlim = (-π, π),
     zlim = (0, 4π),
@@ -107,13 +110,7 @@ end
 
 @testset "1D SE, 1D FD Extruded Domain vertical advection operator" begin
 
-    # Advection Operator
-    # c ∂_z f
-    # for this test, we use f(z) = sin(z) and c = 1, a WVector
-    # => c ∂_z f = cos(z)
-    hv_center_space, hv_face_space = hvspace_2D()
-
-    function advection(c, f)
+    function advection(c, f, hv_center_space)
         adv = zeros(eltype(f), hv_center_space)
         A = Operators.AdvectionC2C(
             bottom = Operators.SetValue(0.0),
@@ -122,15 +119,34 @@ end
         return @. adv = A(c, f)
     end
 
-    # advective velocity
-    c = Geometry.WVector.(ones(Float64, hv_face_space),)
-    # scalar-valued field to be advected
-    f = sin.(Fields.coordinate_field(hv_center_space).z)
+    n_elems_seq = 2 .^ (5, 6, 7, 8)
+    err, Δh = zeros(length(n_elems_seq)), zeros(length(n_elems_seq))
 
-    # Call the advection operator
-    adv = advection(c, f)
+    for (k, n) in enumerate(n_elems_seq)
+        # Advection Operator
+        # c ∂_z f
+        # for this test, we use f(z) = sin(z) and c = 1, a WVector
+        # => c ∂_z f = cos(z)
+        hv_center_space, hv_face_space = hvspace_2D(helem = n, velem = n)
 
-    @test norm(adv .- cos.(Fields.coordinate_field(hv_center_space).z)) ≤ 5e-2
+        Δh[k] = 1.0 / n
+
+        # advective velocity
+        c = Geometry.WVector.(ones(Float64, hv_face_space),)
+        # scalar-valued field to be advected
+        f = sin.(Fields.coordinate_field(hv_center_space).z)
+
+        # Call the advection operator
+        adv = advection(c, f, hv_center_space)
+
+        err[k] = norm(adv .- cos.(Fields.coordinate_field(hv_center_space).z))
+    end
+    # AdvectionC2C convergence rate
+    conv_adv_c2c = convergence_rate(err, Δh)
+    @test err[3] ≤ err[2] ≤ err[1] ≤ 0.1
+    @test conv_adv_c2c[1] ≈ 2 atol = 0.1
+    @test conv_adv_c2c[2] ≈ 2 atol = 0.1
+    @test conv_adv_c2c[3] ≈ 2 atol = 0.1
 end
 
 @testset "1D SE, 1D FD Extruded Domain horizontal divergence operator" begin
@@ -163,7 +179,7 @@ end
     ) ≤ 5e-5
 end
 
-@testset "1D SE, 1D FD Extruded Domain horz & vert divergence operator" begin
+@testset "1D SE, 1D FD Extruded Domain horz & vert divergence operator, with Extrapolate BCs" begin
 
     # Divergence operator in 2D Cartesian domain with
     # Cₕ = (cₕ, 0), Cᵥ = (0, cᵥ)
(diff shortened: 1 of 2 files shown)


valeriabarra:
In [e88fb0f](https://github.com/CliMA/ClimaCore.jl/pull/533/commits/e88fb0fd01279547a1a81ab9526d75d4688d6b17), I confirmed that we do have the desired 2nd order convergence, when using the Dirichelet BCs for the composite vertical operator. 
Hence, the degraded order was indeed induced by the `Extruded` BC (as we suspected)
cc: @charleskawczynski , @simonbyrne

valeriabarra:
@simonbyrne , @charleskawczynski , can anyone give some feedback on this? 

I heard we might deprecate the Advection operators in favor of the Interpolate + Divergence composite ones in the future, but I feel some of the convergence/unit tests in this PR are still valuable. 

Just to summarize: this PR adds convergence/unit tests to both the AdvectionC2C and Interpolate + Divergence composite operator. This was needed to establish and verify the convergence rate of the current operator before extending it to high-order (as detailed in #524 , needed for the limiters)

test/Operators/hybrid/2d.jl
```diff
@@ -172,63 +188,154 @@ end
...
 
-    @test norm(exact_divuf .- divuf.h) ≤ 5e-2
+        err[k] = norm(exact_divuf .- divuf.h)
+    end
+    # Divergence convergence rate
+    conv_comp_div = convergence_rate(err, Δh)
+    @test err[3] ≤ err[2] ≤ err[1] ≤ 0.1
+    @test conv_comp_div[1] ≈ 0.5 atol = 0.1
+    @test conv_comp_div[2] ≈ 0.5 atol = 0.1
+    @test conv_comp_div[3] ≈ 0.5 atol = 0.1
```
charleskawczynski, line 258: Why are these still at 0.5 order?

valeriabarra: Because this test uses the Extruded BC. The one that uses Dirichelet (SetValue) are 2nd order

test/Operators/hybrid/2d.jl
```diff
@@ -172,63 +188,154 @@ end
...
+        # A horizontal Gaussian blob, centered at (-x_0, -z_0),
+        # with `a` the height of the blob and σ the standard deviation
+        function h_blob(x_0, z_0, a = 1.0, σ = 0.2)
+            coords = Fields.coordinate_field(hv_center_space)
+            f = map(coords) do coord
+                a * exp(-((coord.x + x_0)^2 + (coord.z + z_0)^2) / (2 * σ^2))
+            end
+
+            return f
+        end
```
charleskawczynski, line 312: Can we remove the closure here?

charleskawczynski (commented):
Overall the PR looks good. I'd prefer we remove some of the closures, and there's still one test with 0.5 convergence rate, shouldn't it be higher?

charleskawczynski:
> Thanks. Unfortunately, with the current Extruded (0th-order BC) that's what we get at the moment. We may want to consider different ways (higher-order) of specifying Extruded BCs, but that's not in the scope of this PR

Ah, ok, I thought we changed the `Extrapolate` bc to a `SetValue` BC. Can we just use a (different, higher order) BC? Do we need to add the test that confirms that we get degraded order of convergence with a low order BC and a higher order interior stencil? It's just not clear to me that order 0.5 is theoretically expected for a mixed order stencil/bc, in general.

I think we should (maybe in a separate PR) document that the extrapolate bc is 0th order accurate.

valeriabarra:
> Ah, ok, I thought we changed the `Extrapolate` bc to a `SetValue` BC. 

I kept both tests.

valeriabarra:
> Do we need to add the test that confirms that we get degraded order of convergence with a low order BC and a higher order interior stencil?
We will do when we add the higher-order interior stencil

test/Operators/hybrid/2d.jl
```diff
@@ -172,63 +188,154 @@ end
...
+        exact_divuf = div_h_blob(0.0, 0.0)
+
+        err[k] = norm(exact_divuf .- divuf.h)
+    end
+    # Divergence convergence rate
+    conv_comp_div = convergence_rate(err, Δh)
+    @test err[3] ≤ err[2] ≤ err[1] ≤ 0.1
+    @test conv_comp_div[1] ≈ 2 atol = 0.1
+    @test conv_comp_div[2] ≈ 2 atol = 0.1
+    @test conv_comp_div[3] ≈ 2 atol = 0.1
```
valeriabarra, line 337: @charleskawczynski 
Here is the Interpolate + Divergence test with Dirichelet SetValue BC that indeed is 2nd order

charleskawczynski:
> We will do when we add the higher-order interior stencil

Ah, ok. I figured that we'd add higher order BCs so that we can globally preserve the order of the interior stencil. I guess this is fine if we plan to use mixed order stencils.

charleskawczynski (approved):

valeriabarra:
> > We will do when we add the higher-order interior stencil
> 
> Ah, ok. I figured that we'd add higher order BCs so that we can globally preserve the order of the interior stencil. 

That's my hope for the future as well, but it was not the aim/scope of this PR. This was just to assess the order of our current setup (advection/divergence interior stencil operators + boundary conditions) before we attempt to modify them to achieve higher-order. 
Thanks :) 
I'm going to apply the suggested changes to the closures soon

valeriabarra:
I applied Charlie's suggestions and squashed commits. 
I'm going to `bors r+` if that's ok. Thanks!

test/Operators/hybrid/2d.jl
```diff
@@ -163,7 +179,7 @@ end
     ) ≤ 5e-5
 end
 
-@testset "1D SE, 1D FD Extruded Domain horz & vert divergence operator" begin
+@testset "1D SE, 1D FD Extruded Domain horz & vert divergence operator, with Extruded BCs" begin
```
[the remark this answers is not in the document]
valeriabarra: Oh, absolutely, this was a typo! Thanks for catching that!

test/Operators/hybrid/2d.jl
```diff
@@ -172,63 +188,152 @@ end
...
+        hv_center_space, hv_face_space =
+            hvspace_2D(xlim = (-1, 1), zlim = (-1, 1), helem = n, velem = n)
+
+        Δh[k] = 1.0 / n
+
+        function divergence(f)
+
+            divuf = Fields.FieldVector(h = zeros(eltype(f), hv_center_space))
+            h = f.h
+            dh = divuf.h
```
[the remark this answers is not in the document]
valeriabarra: Sorry, perhaps I'm missing something. How do you define the FieldVector of the right size if you don't have the space here?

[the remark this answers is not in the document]
charleskawczynski: We can change the function signature to
```julia
        function divergence(f, hv_center_space)

            divuf = Fields.FieldVector(h = zeros(eltype(f), hv_center_space))
            h = f.h
            dh = divuf.h
            ...
```
And then move the function outside of the loop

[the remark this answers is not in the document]
valeriabarra: I was referring to the call, not the definition. I see now what you mean, thanks

test/Operators/hybrid/2d.jl
```diff
@@ -107,30 +110,43 @@ end
...
+    for (k, n) in enumerate(n_elems_seq)
+        # Advection Operator
+        # c ∂_z f
+        # for this test, we use f(z) = sin(z) and c = 1, a WVector
+        # => c ∂_z f = cos(z)
+        hv_center_space, hv_face_space = hvspace_2D(helem = n, velem = n)
+
+        Δh[k] = 1.0 / n
+
+        function advection(c, f)
```
charleskawczynski, line 125: One last one here! cc @valeriabarra

charleskawczynski (approved):

valeriabarra:
this closes #534

Issue #537, opened 2022-02-16
writevtk with latlong for time series output 


valeriabarra:
I think in #457 we missed the time series output. And now we have an error when attempting this. For instance, 

```
    times = 0:5*dt:end_time
    Sol_q =  Array{Fields.Field}(undef, length(times))
    for t in 1:length(times)
        Sol_q[t] = sol.u[t].ρq ./ sol.u[t].ρ
    end
    
    using ClimaCoreVTK
    writevtk(
        joinpath(path, "q_latlong"),
        (q = Sol_q,);
        latlong = true,
        basis = :point,
    )
```

valeriabarra:
Currently gives a Method error:
```
ERROR: LoadError: MethodError: no method matching vtk_grid_space(::Vector{ClimaCore.Fields.Field})
Closest candidates are:
  vtk_grid_space(!Matched::ClimaCore.Spaces.SpectralElementSpace1D) at /central/home/vbarra/ClimaCore.jl/lib/ClimaCoreVTK/src/space.jl:184
  vtk_grid_space(!Matched::ClimaCore.Spaces.SpectralElementSpace2D) at /central/home/vbarra/ClimaCore.jl/lib/ClimaCoreVTK/src/space.jl:192
  vtk_grid_space(!Matched::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{ClimaCore.Spaces.CellFace}) at /central/home/vbarra/ClimaCore.jl/lib/ClimaCoreVTK/src/space.jl:200
```

simonbyrne:
I think you need to add `times` as well?

valeriabarra:
that might have been it! 
I'll try again. I suggest adding a unit test for this case, as we often refer to the unit tests for the right syntax, and this particular case is missing.
Thanks

valeriabarra:
@simonbyrne , I tried again passing `times`, in the same order as we have in the t[ests](https://github.com/CliMA/ClimaCore.jl/blob/6ebb06e2a411af95c9ead0020c21e6b1f6acc9d7/lib/ClimaCoreVTK/test/runtests.jl#L79) for the regular not latlong version

```
writevtk(
        joinpath(path, "sphere_lat_long"),
        times,
        (q = Sol_q,);
        latlong = true,
        basis = :point,
    )
``` 

but it still gives the same error:
```
ERROR: LoadError: MethodError: no method matching vtk_grid_space(::Vector{ClimaCore.Fields.Field})
[...]

PR #539, opened 2022-02-17
add writepvd

simonbyrne:
To help avoid problems like #527, I propose splitting the paraview collection functionality into a new function `writepvd`: now if you use the wrong arguments you should get a better error message.

### DIFF
--- a/docs/src/lib/ClimaCoreVTK.md
+++ b/docs/src/lib/ClimaCoreVTK.md
@@ -10,6 +10,7 @@ ClimaCoreVTK.jl provides functionality for writing ClimaCore fields to VTK files
 
 ```@docs
 writevtk
+writepvd
 ```
 
 # Internal functions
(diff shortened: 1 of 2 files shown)


jakebolewski:
agree, it would be nice also to replace with the `series` format IMO as that will be the way forward in paraview / VTK,  also it is a more descriptive name

simonbyrne:
> agree, it would be nice also to replace with the `series` format IMO as that will be the way forward in paraview / VTK, also it is a more descriptive output name.

what's the series format?

jakebolewski:
https://kitware.github.io/paraview-docs/latest/cxx/classvtkFileSeriesReader.html

It's the JSON version in the description

valeriabarra:
Thanks @simonbyrne ! 
I think you meant #537 . 

Can we add the corresponding unit tests for this and the latlong writevtk version for completeness?

PR #541, opened 2022-02-17
add simple restart to examples/3dsphere

simonbyrne:
Adds a very crude restart mechanism to the 3d sphere examples driver.


dennisYatunin (approved):
Looks good! It would be nice to allow additional callbacks to be specified in the test files, but I can add that when the need arises.

simonbyrne:
bors try

PR #548, opened 2022-02-21
add support for periodic finite difference spaces

### LINKED ISSUE #545: Allow for periodic finite difference meshes.
This would be very helpful for testing our stencils as we don't have to then worry about boundary conditions.

cc: @akshaysridhar @valeriabarra

simonbyrne:
Fixes #545 

Specify `periodic=true` when constructing the `IntervalDomain`.

### DIFF
--- a/test/Operators/finitedifference/column.jl
+++ b/test/Operators/finitedifference/column.jl
@@ -87,6 +87,54 @@ end
     end
 end
 
+
+@testset "Scalar Field FiniteDifferenceSpaces - periodic" begin
+    for FT in (Float32, Float64)
+        domain = Domains.IntervalDomain(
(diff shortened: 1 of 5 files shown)


test/Operators/finitedifference/column.jl
```diff
@@ -87,6 +87,54 @@ end
     end
 end
 
+
+@testset "Scalar Field FiniteDifferenceSpaces - periodic" begin
```
valeriabarra, line 91: Thanks for adding this. Can we also add Mesh and Topology tests, in addition to Operators test?

simonbyrne: I've added Mesh tests. We don't really use any of the Topology properties.

valeriabarra (approved):

PR #557, opened 2022-02-25
Reorganize drivers, add inertial gravity wave, add rhoe_int

dennisYatunin:
This PR puts a `driver.jl` in the main examples folder, along with a `staggered_nonhydrostatic_model.jl`. The latter defines a split IMEX tendency function and a Jacobian function for the implicit part of the tendency function. It can handle both 2D and 3D spaces, and it can handle three different energy variables (ρθ, ρe, and ρe_int). This setup makes it easy to create a wide range of test files that are all based on the same model.

Also, this PR adds the `inertial_gravity_wave.jl` example to `examples/hybrid`, which is based on [this paper](https://rmets.onlinelibrary.wiley.com/doi/epdf/10.1002/qj.2105).


examples/driver.jl
```diff
@@ -0,0 +1,90 @@
+if !haskey(ENV, "BUILDKITE")
```
valeriabarra, line 1: As mentioned in the software meeting, it's better to move (and rename?) this file in the subdirectory(ies) for the hybrid/3D spaces. Please use `git mv` when moving/renaming files so that the git history and contributions are preserved :) 
Thanks!!

dennisYatunin:
The analytical solution for inertial gravity wave is still not working, but Akshay and I have decided to merge in the current version anyway. Also, git decided to separate some of the moves into deletes+adds, I guess because the changes were too significant? Apologies for the unnecessarily large PR.

jiahe23:
I do not want to appear harsh, but this PR made too many changes and made it impossible to review. Also, better to have a clear instruction on how to set up a slurm pipeline job.

dennisYatunin:
@jiahe23 Yes, I apologize about the size of the PR. I was trying to juggle too many things at once.

In terms of setting up a pipeline, the old driver functionality is unchanged. The driver now just exists in examples/hybrid instead of examples/3dsphere (as was decided in an earlier software meeting). The environment variables that can be used for the driver (restart files, output directories) are basically the same as in main.

szy21:
Do you know why some files are deleted and added rather than being renamed? If we can change that it will be easier to review.

szy21:
@simonbyrne Is there anything we can do about the git history here? Although as @dennisYatunin mentioned this PR is mostly moving things around, we think it might be risky to merge it without knowing what the changes are exactly.

valeriabarra:
> Do you know why some files are deleted and added rather than being renamed? If we can change that it will be easier to review.

@dennisYatunin , I wonder if you used `git mv` when you renamed those files? And did you commit the renaming only in an atomic commit, or in the same commit there were other changes? It's always good to keep commits as atomic as possible, especially when renaming/moving of files happens.

[git bisect](https://git-scm.com/docs/git-bisect), [git reflog](https://git-scm.com/docs/git-reflog) and [git rerere](https://git-scm.com/docs/git-rerere) are all good tools worth checking, if you don't know those already, for identifying and rewinding some hazardous or dangerous commits.

szy21:
After talking with a few people I think it would be good to break this down to smaller PRs, even if it takes some time. @dennisYatunin Please try to modify it so that the git diff is reasonable. The easiest thing might be to start with a new PR, and commit the renaming first, then commit other changes. @charleskawczynski could help with that if you need help. @valeriabarra may know the git tricks as well:)

dennisYatunin:
Ok, I will break this into 2 PRs. The first PR will only move files into hybrid/box and hybrid/sphere (modifying the driver and pipeline appropriately). The second PR will contain the remainder of this PR's modifications.

szy21:
Should we close this one?

PR #559, opened 2022-02-25
Add unit tests for stencil operations

dennisYatunin:
This PR adds the test script `test/Operators/finitedifference/stencil_operations.jl`, which checks for the correctness of `src/Operators/operator2stencil.jl` and `src/Operators/pointwisestencil.jl`. This should make it easier to fix the issue of memory allocation during stencil application/composition at the right boundaries of Fields. This PR also extends `Operator2Stencil` to work for all F2C and C2F Operators, including flux correction (which needs to get added to the implicit tendency function); it also fixes some minor bugs in stencil operations.


dennisYatunin:
bors try

src/Operators/operator2stencil.jl
```diff
@@ -33,7 +75,7 @@ get_boundary(op::Operator2Stencil, bw::RightBoundaryWindow{name}) where {name} =
     get_boundary(op.op, bw)
 
 function return_eltype(op::Operator2Stencil, args...)
-    lbw, ubw = stencil_interior_width(op.op, args...)[end]
+    lbw, ubw = stencil_interior_width(op.op, args...)[1]
```
charleskawczynski, line 78: Is this one of the bug fixes?

test/Operators/finitedifference/stencil_operations.jl
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+        # We can't use non-zero non-extrapolation BCs because Operator2Stencil
+        # does not account for them (it only handles linear transformations).
+        zero_scalar = Operators.SetValue(zero(FT))
+        zero_vector = Operators.SetValue(Geometry.Covariant3Vector(zero(FT)))
+        zero_grad = Operators.SetGradient(Geometry.Covariant3Vector(zero(FT)))
+        zero_div = Operators.SetDivergence(zero(FT))
+        zero_curl = Operators.SetCurl(Geometry.Contravariant3Vector(zero(FT)))
+        extrap = Operators.Extrapolate()
```
charleskawczynski, line 76: Can we please add tests with non-zero BCs for `SetValue`, `SetGradient` and `SetDivergence`?

src/Operators/operator2stencil.jl
dennisYatunin: This is actually the extension of `Operator2Stencil` to advection and flux-correction operators. I initially linearized every operator with respect to its last argument (because that is what determines the operator's name; i.e., C2F, C2C, etc.), but these operators need to be linearized with respect to velocity, which is their first argument. So, I am now linearizing every operator with respect to its first argument.

test/Operators/finitedifference/stencil_operations.jl
dennisYatunin: That would correspond to an affine transformation (rather than a linear transformation), which I cannot currently represent with my stencil data structure. I do plan to add that capability in the future, though.

dennisYatunin: To clarify, I could add non-zero BCs to the unit tests, and they would run, but the tests would not pass because the "linearized" expressions would not be equal to the "exact" expressions at the boundaries.

charleskawczynski (approved):

PR #567, opened 2022-03-01
ClimaCoreVTK: writevtk for latlong time series

### LINKED ISSUE #537: writevtk with latlong for time series output

valeriabarra:
This PR adds unit tests for writevtk with latlong=true for time series outputs (currently working!) for both scalar-valued series and vector-valued ones. 

This will close #537

### DIFF
--- a/lib/ClimaCoreVTK/test/runtests.jl
+++ b/lib/ClimaCoreVTK/test/runtests.jl
@@ -34,6 +34,7 @@ mkpath(joinpath(OUTPUT_DIR, "series"))
         basis = :point,
     )
     @test isfile(joinpath(OUTPUT_DIR, "sphere.vtu"))
+
     writevtk(
         joinpath(OUTPUT_DIR, "sphere_latlong"),
         (coords = coords,);
@@ -76,6 +77,7 @@ mkpath(joinpath(OUTPUT_DIR, "series"))
         end
         U[t] = u
     end
+
     writevtk(
         joinpath(OUTPUT_DIR, "series", "sphere_vector_series"),
         times,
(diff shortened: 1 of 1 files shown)


lib/ClimaCoreVTK/test/runtests.jl
```diff
@@ -88,6 +90,30 @@ mkpath(joinpath(OUTPUT_DIR, "series"))
         )
     end
     @test isfile(joinpath(OUTPUT_DIR, "series", "sphere_vector_series.pvd"))
+
+    writevtk(
+        joinpath(OUTPUT_DIR, "sphere_latlong_scalar_series"),
+        times,
+        (coords = coords,);
```
simonbyrne, line 97: `coords` here is a single field, not an array of fields

valeriabarra: That's right! Thanks for catching that. I fixed it in [0ddb2e9](https://github.com/CliMA/ClimaCore.jl/pull/567/commits/0ddb2e92006b71c7a4b44dae438f8187fc0c2658). Tests are passing now

valeriabarra: Rebased onto `main` and squashed

simonbyrne (approved):

PR #568, opened 2022-03-02
Reduce allocations in implicit operators

charleskawczynski:
This PR significantly reduces allocations in the held suarez examples. The `@time` of `OrdinaryDiffEq.solve!()` cost breakdown is:

```
                      main                 |         This PR
ρe:      94.001 GiB, 3.44% gc time         |  10.646 GiB, 0.46% gc time
ρθ:      91.321 GiB, 3.14% gc time         |   7.967 GiB, 0.42% gc time
```
I think the big gain here was from using `ntuple` instead of `map(, (1:N...,))`.

### DIFF
--- a/src/Operators/pointwisestencil.jl
+++ b/src/Operators/pointwisestencil.jl
@@ -345,12 +345,12 @@ function apply_stencil_at_idx(i_vals, stencil, arg, loc, idx)
     lbw = bandwidths(eltype(stencil))[1]
     i_func = i -> coefs[i - lbw + 1] ⊠ getidx(arg, loc, idx + i)
     return length(i_vals) == 0 ? zero(eltype(eltype(stencil))) :
-           ⊞(map(i_func, i_vals)...)
+           mapreduce(i_func, ⊞, i_vals)
 end
 
 function stencil_interior(::ApplyStencil, loc, idx, stencil, arg)
     lbw, ubw = bandwidths(eltype(stencil))
-    i_vals = (lbw:ubw...,)
+    i_vals = lbw:ubw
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

src/Operators/pointwisestencil.jl
```diff
@@ -422,22 +421,20 @@ function compose_stencils_at_idx(i_vals_tuple, stencil1, stencil2, loc, idx)
...
     i_func_at_j(j) =
         i -> coefs1[i - lbw1 + 1] ⊠ getidx(stencil2, loc, idx + i)[j - i + lbw1]
     function j_func(j)
         i_vals = i_vals_tuple[j]
         return length(i_vals) == 0 ? zero(eltype(eltype(stencil1))) :
-               return ⊞(map(i_func_at_j(j), i_vals)...)
+               return mapreduce(i_func_at_j(j), ⊞, i_vals)
     end
-    return StencilCoefs{lbw, ubw}(map(j_func, j_vals))
+    return StencilCoefs{lbw, ubw}(ntuple(j -> j_func(j), (ubw - lbw + 1)))
```
jakebolewski, line 431: @charleskawczynski shouldn't we use the same form as we do in the previous stencil code?  this is a `tuplemap` like we had previously and the backing `map` was able to be optimized through in 1.7+.  The real gain was getting rid of the splatting like we discussed.

charleskawczynski: Hmm, maybe you're right, I thought I had tried on just the splatting and didn't see much difference, but I don't remember exactly when the difference in allocations came through.

I'd be happy to continue working on this with you if you'd like. It'd be nice if we can eliminate the remaining allocations, for good measure.

PR #569, opened 2022-03-02
Parallelize unit tests in CI

charleskawczynski:
I wanted to try out buildkite's group feature.


charleskawczynski:
bors try

charleskawczynski:
bors try

simonbyrne:
We don't actually need to run the unit tests on the cluster, since they run on GitHub Actions already.

charleskawczynski:
> We don't actually need to run the unit tests on the cluster, since they run on GitHub Actions already.

That's a good point, the only advantages I saw was to keep an eye on coverage and testing on other operating systems. What we've decided to do in TC.jl is still run unit tests on GHA, but make bors not depend on it. That way we can keep an eye on both without slowing down development.

PR #593, opened 2022-03-09
Add up_deps.jl script to dev folder

charleskawczynski:


.dev/up_deps.jl
```diff
@@ -0,0 +1,27 @@
+#=
+A simple script for updating the manifest
+files in all of our environments.
+=#
+
+root = dirname(@__DIR__)
```
jakebolewski, line 6: this is not the correct root?

jakebolewski: ah nvm I see that you are taking the dirname of the dir which is a bit confusing

simonbyrne (approved):

PR #595, opened 2022-03-09
Interpolate before computing products

charleskawczynski:
This PR makes the BCs for `cuₕ` consistent, with
```julia
@. fu¹² =
        Geometry.Contravariant12Vector(Geometry.Covariant123Vector(Ic2f(cuₕ)))
```
, by interpolating before computing products with other variables inside vertical divergence operators. @jiahe23, could you try running held suarez with this PR?

### DIFF
--- a/examples/3dsphere/baroclinic_wave_utilities.jl
+++ b/examples/3dsphere/baroclinic_wave_utilities.jl
@@ -223,7 +223,7 @@ function baroclinic_wave_ρθ_remaining_tendency!(dY, Y, p, t; κ₄)
         Geometry.Covariant123Vector(cuₕ) + Geometry.Covariant123Vector(If2c(fw))
 
     @. dY.Yc.ρ -= hdiv(cρ * cuvw)
-    @. dY.Yc.ρ -= vdivf2c(Ic2f(cρ * cuₕ))
+    @. dY.Yc.ρ -= vdivf2c(Ic2f(cρ) * Ic2f(cuₕ))
 
     # Momentum conservation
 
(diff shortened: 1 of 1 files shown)


jiahe23:
BTW, can you also open an issue in ClimaAtmos so that this propagates there as well.

charleskawczynski:
Hmm, I'm realizing now that I think this won't have any impact for the `Extrapolate` BC, since the first interior values are simply set to the boundary faces, so the order won't matter.

jiahe23:
> Hmm, I'm realizing now that I think this won't have any impact for the `Extrapolate` BC, since the first interior values are simply set to the boundary faces, so the order won't matter.

ok

charleskawczynski:
That said, this _will_ have an impact on the tendencies in the interior, just not at the boundaries because our BCs are doing constant extrapolation. The proposed form in this PR is consistent for higher order BCs, and will have an impact on the first interior cell center for higher order BCs.

So it may still be worth trying out held suarez with this PR.

jiahe23:
> That said, this _will_ have an impact on the tendencies in the interior, just not at the boundaries because our BCs are doing constant extrapolation. The proposed form in this PR is consistent for higher order BCs, and will have an impact on the first interior cell center for higher order BCs.
> 
> So it may still be worth trying out held suarez with this PR.

Ok, I am running a test on it. Could we check how Tempest handle the order for interpolation? @OsKnoth

szy21:
Could you post the question on the dycore channel so that Oswald and Paul can see it?

PR #600, opened 2022-03-15
Move hybrid and 3dsphere to hybrid/box and hybrid/sphere

dennisYatunin:
This PR moves all the files in examples/hybrid to examples/hybrid/box, and it moves all the files in examples/3dsphere to examples/hybrid/box. The driver file (which currently only works for some files in hybrid/sphere) and the pipeline are updated appropriately.


examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -473,14 +473,10 @@ end
 function output_writer(integrator)
     savefile = joinpath(
         output_dir,
-        test_name *
-        "_Day" *
-        string(Int(floor((integrator.t + tend) / 3600 / 24))) *
-        ".jld2",
+        "day" * string(Int(floor((integrator.t + tend) / 3600 / 24))) * ".jld2",
```
szy21, line 476: Any reason why `test_name` is removed?

szy21 (approved):

valeriabarra:
Please allow some time for others to review and give feedback before triggering the merge

valeriabarra:
I see that both 2d and 3d Cartesian examples where moved in the same `examples/hybrid/box`. While for the non-hybrid cases, we have separate `plane` and `sphere` directories. 
So I'm thinking this could (should?) have followed the same pattern and used `examples/hybrid/plane` , `examples/hybrid/box`,  and `examples/hybrid/sphere` (for the hybrid sphere cases)

.buildkite/pipeline.yml
```diff
@@ -312,9 +312,9 @@ steps:
   - label: ":computer: Rising Bubble 2D hybrid"
     key: "cpu_rising_bubble_2d_hybrid"
     command:
-      - "julia --color=yes --project=examples examples/hybrid/bubble_2d.jl"
+      - "julia --color=yes --project=examples examples/hybrid/box/bubble_2d.jl"
```
valeriabarra, line 315: Please consider moving the 2d examples in a separate `examples/hybrid/plane/` directory for consistency with the non-hybrid case.
```suggestion
      - "julia --color=yes --project=examples examples/hybrid/plane/bubble_2d.jl"
```

valeriabarra (changes requested):

.buildkite/pipeline.yml
jiahe23: I thought 2d plane is usually a 2d horizontal cg plane, which is different from here

dennisYatunin:
I'm pretty neutral on the subfolder naming. I guess it could be somewhat helpful to keep all the tests that include `baroclinic_wave_utilities.jl` (baroclinic wave and Held-Suarez) in a single subfolder, but other than that it doesn't really matter. After my next PR (which unifies the tendency functions for all 2D and 3D hybrid examples), there will be no behavioral differences between any of the examples. Is there a reason to separate things into subfolders?

.buildkite/pipeline.yml
valeriabarra: It's just a category based on geometry/domain type.  There was a discussion on [Slack](https://climate-machine.slack.com/archives/C019DHELDA9/p1643931125682369) where we talked about characterizing them by physical behavior or by domain type. Since this PR makes a move in the latter direction, I think at least it would be good to follow a consistent pattern

examples/hybrid/sphere/baroclinic_wave_utilities.jl
jiahe23: same question from me

.buildkite/pipeline.yml
valeriabarra: So we would have the non-hybrid (CG only examples) in `examples/plane`, and the corresponding hybrid ones in `examples/hybrid/plane` and `examples/hybrid/box`, respectively (for 2D and 3D). It's just clearer and more consistent IMO.

dennisYatunin: Ok, so if we're categorizing things based on domains, perhaps we could have hybrid/line (or hybrid/interval), hybrid/plane (or hybrid/rectangle), and hybrid/sphere (or hybrid/cubed_sphere).

examples/hybrid/sphere/baroclinic_wave_utilities.jl
dennisYatunin: The test_name global variable contains both the test file directory (e.g., sphere) and the test file name (e.g., baroclinic_wave_rhoe). I could have replaced test_name with the test file name, but that is already included in the output directory. For example, day5.jld2 will be stored in hybrid/sphere/output/baroclinic_wave_rhoe, so there is no need to call it baroclinic_wave_rhoe_day5.jld2. If you do want to call it that, then perhaps we should also add the test names to the plots; e.g., baroclinic_wave_rhoe_v.jld2?

dennisYatunin: Ok, how about this. Vote on what you would most rather see:

1. hybrid/sphere/output/baroclinic_wave_rhoe/baroclinic_wave_rhoe_day5.jld2
2. hybrid/sphere/output/baroclinic_wave_rhoe/day5.jld2
3. hybrid/output/sphere/baroclinic_wave_rhoe/baroclinic_wave_rhoe_day5.jld2
4. hybrid/output/sphere/baroclinic_wave_rhoe/day5.jld2
5. hybrid/output/sphere/baroclinic_wave_rhoe_day5.jld2
6. hybrid/output/sphere_baroclinic_wave_rhoe_day5.jld2
7. hybrid/output/baroclinic_wave_rhoe_day5.jld2

szy21:
I think it's still good to keep box and sphere separate for now, but I'm pretty neutral on box or plane. It's your call @dennisYatunin:)

.buildkite/pipeline.yml
valeriabarra: In the absence of a broader discussion with the others, I personally think it's ok to leave the `hybrid/box` for the hybrid 3d ones as you did. And just change the `hybrid/plane` ones as I was asking, to make them at least consistent with the planar non-hybrid case. Some of these examples will be moved anyways, so not worth doing a major reorganization of the examples folder right now.

examples/hybrid/sphere/baroclinic_wave_utilities.jl
dennisYatunin: To clarify, the current PR will result in option 2.

.buildkite/pipeline.yml
dennisYatunin: Alright, will do

valeriabarra: Thank you

examples/hybrid/sphere/baroclinic_wave_utilities.jl
jiahe23: I see. I do not care much about the folder because it will be overwritten by the environment var set in the slurm script (at least that's how this is being done in the current main), this is because we want to save it to scratch or storage on central so that the simulation won't explode one's home did. But I do think we should have it inthe folder name and the file name.

jiahe23: so 3

dennisYatunin: Should I also add the test name to the generated plots? Then I could use option 5 for everything.

dennisYatunin: Nevermind, option 5 might make the CI more convoluted. I'll just add the test file names everywhere, but stick with option 3.

jiahe23: 👍

dennisYatunin:
bors try

examples/hybrid/sphere/baroclinic_wave_utilities.jl
jiahe23: Yea. Let's stay with 3… and make sure if people specify 'OUTPUT_DIR' then it will be OUTPUT_DIR'/baroclinic_wave_rhoe_day5.jld2'

jiahe23 (approved):

valeriabarra (approved):

PR #602, opened 2022-03-16
Docs: add some introductory material, vector bases and SEM descriptions

### LINKED ISSUE #601: Add some introductory docs
Since we produced some useful material for the second ClimaWorkshop on ClimaCore.jl's API fundamentals, it would be good to port over some of that material and visual aid to our docs

valeriabarra:
This PR is an attempt at enriching our docs, with some of the material produced for the second workshop in the ClimaWorkshops series. 

It fixes some docstrings. It adds an Introduction section. And it adds more verbose descriptions of the different vector bases ClimaCore supports and different discretizations, especially SEM. Since these two latter topics were a bit more lengthy, I decided to include them in separate, dedicated `.md` files. But we can trim, move, or re-organize things as desired. 

This will close #601


docs/src/api.md
```diff
@@ -25,6 +25,29 @@ Geometry.AbstractPoint
 Geometry.float_type
 ```
 
+Points represent _locations_ in space (not to be confused with _positions_,
+_position vectors_ or _location vectors_ in geometry). Vectors, on the other hand,
```
simonbyrne, line 29: I still don't understand the distinction between a coordinate and a position.

docs/src/api.md
```diff
@@ -25,6 +25,29 @@ Geometry.AbstractPoint
...
+represent displacements in space. An analogy with time works well: times,
+(also called instants or datetimes) are _locations_ in time. While, durations
+are displacements in time.
+
+**Note 1**: In ClimaCore.jl, we use angles (and, therefore, trigonometric functions:
+`cosd`, `sind`, `acosd`, `asind`, `tand`,...) in degrees, not in radians.
```
simonbyrne, line 35:
```suggestion
**Note 1**: Latitude and longitude are specified via angles (and, therefore, trigonometric functions:
`cosd`, `sind`, `acosd`, `asind`, `tand`,...) in degrees, not in radians.
```

docs/src/api.md
```diff
@@ -107,10 +130,32 @@ Topologies.vertices
...
 ```
+### Finite Difference Spaces
+ClimaCore.jl supports staggered Finite Difference discretizations. Finite Differences
+discretize an interval domain by approximating the function by a value at either
+the center of each element (also referred to as _cell_) (`CenterFiniteDifferenceSpace`),
+or the interfaces (faces in 3D, edges in 2D or points in 1D) between elements
+(`FaceFiniteDifferenceSpace`).
+
+Users can construct either the center or face space from the mesh, then construct
+the dual space from the original one (this is to avoid allocating additional memory).
```
simonbyrne, line 154:
```suggestion
Users should construct either the center or face space from the mesh, then construct
the other space from the original one: this internally reuses the same datastructures, and avoids allocating additional memory.
```

docs/src/api.md
valeriabarra: This comment was here only for the people with a geometry background (like me) that might be confused because in geometry the terminology _position_ can be used for [_position vector_ or _location vector_](https://en.wikipedia.org/wiki/Position_(geometry)#:~:text=The%20term%20%22position%20vector%22%20is,affine%20spaces%20of%20any%20dimension.), i.e., a Euclidean vector that represents the position of a point P in space in relation to an arbitrary reference origin O. But if not useful for this community, I can certainly remove it.

simonbyrne: Having thought about this a bit more, I think the more useful distinction is that a "point' describes a position via a particular set of coordinates, but that that the coordinates need not be Cartesian.

valeriabarra: Yes. So do we prefer to rephrase the sentence above or is it good as is?

simonbyrne: I would prefer to rephrase it if possible, as I think the current wording is confusing.

valeriabarra: Sure, what do you have in mind?

valeriabarra: @simonbyrne , I applied your other suggestions. For this point in particular I am not sure what you had in mind. As I said, we can just remove the part in parentheses. I don't mind.

docs/src/api.md
```diff
@@ -25,6 +25,29 @@ Geometry.AbstractPoint
 Geometry.float_type
 ```
 
+Points represent _locations_ in space (not to be confused with _positions_,
+_position vectors_ or _location vectors_ in geometry). Vectors, on the other hand,
+represent displacements in space. An analogy with time works well: times,
+(also called instants or datetimes) are _locations_ in time. While, durations
+are displacements in time.
```
simonbyrne, line 32:
```suggestion
Points represent _locations_ in space, specified by a coordinates in a given coordinate system (Cartesian, spherical, etc), whereas vectors, on the other hand, represent displacements in space. 
An analogy with time works well: times (also called instants or datetimes) are _locations_ in time, while, durations are displacements in time.
```

valeriabarra:
@simonbyrne I applied all sugeestions. 

What do you think of the two separate files for the vectors and FEM theory? Do they break the narration? Would a user follow the links? Or is better to put all in one file?

simonbyrne (approved):

simonbyrne:
I think the split makes sense

PR #603, opened 2022-03-16
Reorganize driver

dennisYatunin:
This PR moves everything related to restart files into `driver.jl`. It also cleans up some variable names and clarifies exactly what needs to be defined in each test case file for `driver.jl` to work.

### DIFF
--- a/examples/common_spaces.jl
+++ b/examples/common_spaces.jl
@@ -1,0 +1,87 @@
+using ClimaCore: Geometry, Domains, Meshes, Topologies, Spaces
+
+abstract type Space{FT} end
+
+Base.@kwdef struct PeriodicLine{FT} <: Space{FT}
+    xmax::FT
+    xelem::Int
+    npoly::Int
+end
+
+Base.@kwdef struct PeriodicRectangle{FT} <: Space{FT}
+    xmax::FT
+    ymax::FT
+    xelem::Int
+    yelem::Int
+    npoly::Int
+end
+
+Base.@kwdef struct CubedSphere{FT} <: Space{FT}
+    radius::FT
+    helem::Int # number of elements along side of each panel (6 panels in total)
+    npoly::Int
+end
+
+Base.@kwdef struct ExtrudedSpace{FT, S <: Space{FT}} <: Space{FT}
+    zmax::FT
+    zelem::Int
+    hspace::S
+end
+
+function make_space((; xmax, xelem, npoly)::PeriodicLine)
+    domain = Domains.IntervalDomain(
+        Geometry.XPoint(zero(xmax)),
+        Geometry.XPoint(xmax);
+        periodic = true,
+    )
+    mesh = Meshes.IntervalMesh(domain; nelems = xelem)
+    topology = Topologies.IntervalTopology(mesh)
+    quad = Spaces.Quadratures.GLL{npoly + 1}()
+    return Spaces.SpectralElementSpace1D(topology, quad)
+end
+
+function make_space((; xmax, ymax, xelem, yelem, npoly)::PeriodicRectangle)
+    xdomain = Domains.IntervalDomain(
+        Geometry.XPoint(zero(xmax)),
+        Geometry.XPoint(xmax);
+        periodic = true,
+    )
+    ydomain = Domains.IntervalDomain(
+        Geometry.YPoint(zero(ymax)),
+        Geometry.YPoint(ymax);
+        periodic = true,
+    )
+    domain = Domains.RectangleDomain(xdomain, ydomain)
+    mesh = Meshes.RectilinearMesh(domain, xelem, yelem)
+    topology = Topologies.Topology2D(mesh)
+    quad = Spaces.Quadratures.GLL{npoly + 1}()
+    return Spaces.SpectralElementSpace2D(topology, quad)
+end
+
+function make_space((; radius, helem, npoly)::CubedSphere)
+    domain = Domains.SphereDomain(radius)
+    mesh = Meshes.EquiangularCubedSphere(domain, helem)
+    topology = Topologies.Topology2D(mesh)
+    quad = Spaces.Quadratures.GLL{npoly + 1}()
+    return Spaces.SpectralElementSpace2D(topology, quad)
+end
+
+function make_space((; zmax, zelem, hspace)::ExtrudedSpace)
+    vdomain = Domains.IntervalDomain(
+        Geometry.ZPoint(zero(zmax)),
+        Geometry.ZPoint(zmax);
+        boundary_tags = (:bottom, :top),
+    )
+    vmesh = Meshes.IntervalMesh(vdomain, nelems = zelem)
+    vspace = Spaces.CenterFiniteDifferenceSpace(vmesh)
+    return Spaces.ExtrudedFiniteDifferenceSpace(make_space(hspace), vspace)
+end
+
+function local_geometry_fields(space::ExtrudedSpace)
+    center_space = make_space(space)
+    face_space = Spaces.FaceExtrudedFiniteDifferenceSpace(center_space)
+    return (
+        Fields.local_geometry_field(center_space),
+        Fields.local_geometry_field(face_space),
+    )
+end
--- a/examples/hybrid/sphere/balanced_flow_rhoe.jl
+++ b/examples/hybrid/sphere/balanced_flow_rhoe.jl
@@ -1,25 +1,21 @@
 using Test
 using ClimaCorePlots, Plots
-using DiffEqCallbacks
-using JLD2
 
 include("baroclinic_wave_utilities.jl")
 
-jld2_callback = PeriodicCallback(output_writer, 864000; initial_affect = true)
+const sponge = false
 
-driver_values(FT) = (;
-    zmax = FT(30.0e3),
-    velem = 10,
-    helem = 4,
-    npoly = 4,
-    tmax = FT(60 * 60),
-    dt = FT(5.0),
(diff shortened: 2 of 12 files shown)


examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -1,25 +1,20 @@
...
-    dt = FT(5.0),
-    ode_algorithm = OrdinaryDiffEq.SSPRK33,
-    jacobian_flags = (; ∂𝔼ₜ∂𝕄_mode = :no_∂P∂K, ∂𝕄ₜ∂ρ_mode = :exact),
-    max_newton_iters = 2,
-    save_every_n_steps = 10,
-    additional_solver_kwargs = (;), # callback = jld2_callback), 
+space = ExtrudedSpace(;
+    zmax = FT(30e3),
+    zelem = 10,
+    hspace = CubedSphere(; radius = R, pelem = 4, npoly = 4),
```
jiahe23, line 11: should `pelem` be `helem`?

examples/common_spaces.jl
```diff
@@ -0,0 +1,87 @@
...
+    xmax::FT
+    ymax::FT
+    xelem::Int
+    yelem::Int
+    npoly::Int
+end
+
+Base.@kwdef struct CubedSphere{FT} <: Space{FT}
+    radius::FT
+    pelem::Int # number of elements along side of each panel (6 panels in total)
```
jiahe23, line 21: I think `pelem` is confusing. I prefer `helem` or `ne`

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -28,7 +23,7 @@ initial_condition_velocity(local_geometry) =
 
 remaining_cache_values(Y, dt) = merge(
     baroclinic_wave_cache_values(Y, dt),
-    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = false),
+    final_adjustments_cache_values(Y, dt; use_rayleigh_sponge = sponge),
```
jiahe23, line 26: I prefer `use_rayleigh_sponge = false` here. If later on we have an option for e.g., viscous sponge, then we could do something like `use_sponge = NoSponge/RayleighDamping/DiffusiveSponge` and make `NoSponge` the default, but for now let's not consider this design in ClimaCore.

jiahe23: Oh, I just realize `const sponge = false` is set up above. Then never mind.

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -1,25 +1,20 @@
...
-    additional_solver_kwargs = (;), # callback = jld2_callback), 
+space = ExtrudedSpace(;
+    zmax = FT(30e3),
+    zelem = 10,
+    hspace = CubedSphere(; radius = R, pelem = 4, npoly = 4),
 )
+tend = FT(60 * 60)
+dt = FT(5)
+dt_save_to_sol = FT(50)
+# dt_save_to_disk = FT(60 * 60 * 24 * 10)
```
jiahe23, line 16: If I understand correctly, `dt_save_to_disk = FT(0)` means no jld2 output? Then, can we add a `const dt_save_to_disk = FT(0)` above together with the `const sponge` and use `dt_save_to_disk=dt_save_to_disk` here.

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: I wanted it to mean "panel elements", but I now realize that `helem` is less confusing than `pelem`. I'll change it to `helem`.

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -1,25 +1,20 @@
 using Test
 using ClimaCorePlots, Plots
-using DiffEqCallbacks
-using JLD2
 
 include("baroclinic_wave_utilities.jl")
 
-jld2_callback = PeriodicCallback(output_writer, 864000; initial_affect = true)
+const sponge = false
```
jiahe23, line 6: Let's add 
```
#########################
# set your experiment specs here
#########################
```
so that people have a clear idea of where to specify their prefered experiment setups

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: `dt_save_to_disk = FT(0)` is already specified as the default in `driver.jl`. I also did not want to make it a const since is is not used in any functions and is subject to change during testing.

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: Sure. I can add that. Though, I'm unclear how it clarifies things? What else can you do in a test case file other than set experiment specs?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -4,7 +4,9 @@ using LinearAlgebra: norm_sqr
 using ClimaCore:
     Geometry, Domains, Meshes, Topologies, Operators, Spaces, Fields
 
+include("../implicit_solver_debugging_tools.jl")
 include("../ordinary_diff_eq_bug_fixes.jl")
+include("../common_spaces.jl")
 include("schur_complement_W.jl")
 
 const R = 6.371229e6    # Earth's radius
```
jiahe23, line 12: Can we move this to each of the experiment files, e.g., `baroclinic_wave_rhoe.jl`? There will be chances that people want to run a different planet radius (not sure whether in ClimaCore or not but better to give this flexibility.

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: The commented-out line was only included to match your comment `# callback = jld2_callback),`

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Common spaces does not specify the planet radius `R`. That is included in `baroclinic_wave_utilites.jl`. Any test that does not include `baroclinic_wave_utilities.jl` must specify its own radius.

jiahe23 (changes requested):
Overall, it looks great! I left a few comments. The comments in experiment files I only left in `balanced_flow_rhoe.jl` and should propagate to other experiments.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: I only added `common_spaces.jl` so that I could specify the 2D/3D box test cases using the same format as the sphere test cases (both can just specify `space`, rather than `velem`, `helem`, etc.)

examples/hybrid/sphere/balanced_flow_rhoe.jl
jiahe23: For example, I would rather not touch the anything below this section (intial condition, tendency set up, etc.) and make it clear that everything you want to specify for this experiment is in this section.

examples/hybrid/sphere/balanced_flow_rhoe.jl
jiahe23: How about we move it in the experiment files so that people who want to save jld2's know where to change. I was doing the comment only because I did not figure out a simple way without the comment.

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: Okie dokie, I'll add a comment.

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -1,25 +1,20 @@
...
-    jacobian_flags = (; ∂𝔼ₜ∂𝕄_mode = :no_∂P∂K, ∂𝕄ₜ∂ρ_mode = :exact),
-    max_newton_iters = 2,
-    save_every_n_steps = 10,
-    additional_solver_kwargs = (;), # callback = jld2_callback), 
+space = ExtrudedSpace(;
+    zmax = FT(30e3),
+    zelem = 10,
+    hspace = CubedSphere(; radius = R, pelem = 4, npoly = 4),
 )
+tend = FT(60 * 60)
```
szy21, line 13: I have a slight preference of `tmax` or `t_end` because `tend` sounds like tendency to me...but up to you

dennisYatunin: Okay yeah, I also prefer `t_start` and `t_end` to `tstart` and `tend`.

dennisYatunin:
bors try

jiahe23 (approved):
Thanks!

examples/common_spaces.jl
```diff
@@ -0,0 +1,87 @@
...
+
+Base.@kwdef struct PeriodicRectangle{FT} <: Space{FT}
+    xmax::FT
+    ymax::FT
+    xelem::Int
+    yelem::Int
+    npoly::Int
+end
+
+Base.@kwdef struct CubedSphere{FT} <: Space{FT}
```
valeriabarra, line 19: @dennisYatunin , I know this has been merged already. But I did not have a chance to go through this before and I have questions. Why the need to redefine the CubedSphere struct here?

PR #606, opened 2022-03-18
Unify hybrid model tendencies and add rhoe_int

dennisYatunin:
This PR moves the explicit tendencies defined in `baroclinic_wave_utilities.jl` into `staggered_nonhydrostatic_model.jl`, unifying them into a single function and generalizing them to work with both 2D and 3D domains. It also separates the default tendencies from the non-default tendencies (e.g., hyperdiffusion and Held-Suarez forcing), so that test case files only need to specify non-default tendencies (if there are any), which can be done by defining methods for `additional_cache()` and `additional_tendency!()`.

In addition, this PR introduces a new naming convention for Fields defined on hybrid domains---Fields on cell centers are prefixed with a superscript "c", and Fields on cell faces are prefixed with a superscript "f". Previously, this was only done for some fields, and with non-superscript "c" and "f", which caused confusion when such fields were used alongside variables whose names start with "c" and "f" (e.g., heat capacity and Coriolis frequency). Similarly, this PR introduces a new data structure convention---Fields on cell centers are stored in `Y.c`, and Fields on cell faces are stored in `Y.f`. Previously, the prognostic state was stored as `Y.Yc.ρ`, `Y.Yc.ρe` (or equivalent), `Y.uₕ`, and `Y.w`, which was arbitrary and inefficient (for example, dss needed to be called on both `Y.Yc` and `Y.uₕ`, whereas it can now be called just once on `Y.c`). These new conventions account for most of the lines changed in this PR, but hopefully they will make the code easier to understand.

This PR also adds support for using internal energy as an prognostic variable, and it adds a new Held-Suarez test case that demonstrates this capability ("held_suarez_rhoe_int.jl"). It also adds the "inertial_gravity_wave.jl" test case to "hybrid/plane"; the analytic solution in this test case is still not working, so it can't be used as a proper integration test, but that can be fixed later.

Finally, this PR fixes a minor buildkite bug I introduced in my previous PR that was causing some artifacts from the rising bubble tests to be missing.

### DIFF
--- a/examples/Project.toml
+++ b/examples/Project.toml
@@ -1,6 +1,3 @@
-[compat]
-OrdinaryDiffEq = "5.64.0"
-
 [deps]
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
@@ -18,12 +15,16 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TerminalLoggers = "5d786b92-1e48-4d6f-9151-6b4477ca9bed"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
+
+[compat]
+OrdinaryDiffEq = "5.64.0"
 
 [extras]
 CPUSummary = "2a0fbf3d-bb9c-48f3-b0a9-814d99fd7ab9"
--- a/examples/hybrid/hyperdiffusion.jl
+++ b/examples/hybrid/hyperdiffusion.jl
@@ -1,0 +1,72 @@
+hyperdiffusion_cache(
+    ᶜlocal_geometry,
+    ᶠlocal_geometry;
+    κ₄ = FT(0),
+    use_tempest_mode = false,
+) = merge(
+    (;
+        ᶜχ = similar(ᶜlocal_geometry, FT),
+        ᶜχuₕ = similar(ᶜlocal_geometry, Geometry.Covariant12Vector{FT}),
+        κ₄,
+        use_tempest_mode,
+    ),
+    use_tempest_mode ? (; ᶠχw_data = similar(ᶠlocal_geometry, FT)) : (;),
+)
+
+function hyperdiffusion_tendency!(Yₜ, Y, p, t)
+    ᶜρ = Y.c.ρ
+    ᶜuₕ = Y.c.uₕ
+    (; ᶜp, ᶜχ, ᶜχuₕ, κ₄, use_tempest_mode) = p # assume that ᶜp has been updated
+    point_type = eltype(Fields.local_geometry_field(axes(Y.c)).coordinates)
+
+    if use_tempest_mode
+        @. ᶜχ = wdivₕ(gradₕ(ᶜρ)) # ᶜχρ
+        Spaces.weighted_dss!(ᶜχ)
+        @. Yₜ.c.ρ -= κ₄ * wdivₕ(gradₕ(ᶜχ))
+
+        if :ρθ in propertynames(Y.c)
+            @. ᶜχ = wdivₕ(gradₕ(Y.c.ρθ)) # ᶜχρθ
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρθ -= κ₄ * wdivₕ(gradₕ(ᶜχ))
+        else
+            error("use_tempest_mode must be false when not using ρθ")
+        end
+
+        (; ᶠχw_data) = p
+        @. ᶠχw_data = wdivₕ(gradₕ(Y.f.w.components.data.:1))
+        Spaces.weighted_dss!(ᶠχw_data)
+        @. Yₜ.f.w.components.data.:1 -= κ₄ * wdivₕ(gradₕ(ᶠχw_data))
+    else
+        if :ρθ in propertynames(Y.c)
+            @. ᶜχ = wdivₕ(gradₕ(Y.c.ρθ / ᶜρ)) # ᶜχθ
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρθ -= κ₄ * wdivₕ(ᶜρ * gradₕ(ᶜχ))
+        elseif :ρe in propertynames(Y.c)
+            @. ᶜχ = wdivₕ(gradₕ((Y.c.ρe + ᶜp) / ᶜρ)) # ᶜχe
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρe -= κ₄ * wdivₕ(ᶜρ * gradₕ(ᶜχ))
+        elseif :ρe_int in propertynames(Y.c)
+            @. ᶜχ = wdivₕ(gradₕ((Y.c.ρe_int + ᶜp) / ᶜρ)) # ᶜχe_int
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρe_int -= κ₄ * wdivₕ(ᶜρ * gradₕ(ᶜχ))
+        end
+    end
+
+    if point_type <: Geometry.Abstract3DPoint
+        @. ᶜχuₕ =
+            wgradₕ(divₕ(ᶜuₕ)) - Geometry.Covariant12Vector(
+                wcurlₕ(Geometry.Covariant3Vector(curlₕ(ᶜuₕ))),
+            )
+        Spaces.weighted_dss!(ᶜχuₕ)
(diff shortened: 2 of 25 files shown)


dennisYatunin:
bors try

szy21 (approved):

examples/hybrid/box/bubble_3d_invariant.jl
```diff
@@ -321,7 +321,7 @@ function linkfig(figpath, alt = "")
     end
 end
 
-dir = "bubble3d_invariant"
+dir = "bubble_3d_invariant"
```
jiahe23, line 324:
```suggestion
dir = "bubble_3d_invariant_potential_temperature"
```

jiahe23: can you also change the name of the file adding 'potential_temperature' to be consistent with the total energy version.

examples/hybrid/hyperdiffusion.jl
```diff
@@ -0,0 +1,78 @@
...
+) = merge(
+    (;
+        ᶜχ = similar(ᶜlocal_geometry, FT),
+        ᶜχuₕ = similar(ᶜlocal_geometry, Geometry.Covariant12Vector{FT}),
+        κ₄,
+        use_tempest_mode,
+    ),
+    use_tempest_mode ? (; ᶠχw_data = similar(ᶠlocal_geometry, FT)) : (;),
+)
+
```
jiahe23, line 15: The unicode is over-used here. `div`, `grad`, `curl`, and their weak forms by adding a `w` in front is explanatory enough and very easy to read. I think we should be careful and not abuse unicode

examples/hybrid/plane/bubble_2d_invariant.jl
```diff
@@ -265,7 +265,7 @@ ENV["GKSwstype"] = "nul"
 using ClimaCorePlots, Plots
 Plots.GRBackend()
 
-dir = "bubble_invariant"
+dir = "bubble_2d_invariant"
```
jiahe23, line 268:
```suggestion
dir = "bubble_2d_invariant_potential_temperature"
```

jiahe23: and the file name

examples/hybrid/plane/inertial_gravity_wave.jl
```diff
@@ -0,0 +1,364 @@
...
+    # p(z + Δz) + grav * Δz * ρ(z + Δz) / 2 = p(z) - grav * Δz * ρ(z) / 2    ==>
+    # p(z + Δz) * (1 + δ * Δz / 2) = p(z) * (1 - δ * Δz / 2)                 ==>
+    # p(z + Δz) / p(z) = (1 - δ * Δz / 2) / (1 + δ * Δz / 2)                 ==>
+    # p(z) = p(0) * ((1 - δ * Δz / 2) / (1 + δ * Δz / 2))^(z / Δz)
+    p₀(z) = p_0 * ((1 - δ * Δz / 2) / (1 + δ * Δz / 2))^(z / Δz)
+else
+    p₀(z) = p_0 * exp(-δ * z)
+end
+Tb_init(x, z) = ΔT * exp(-(x - xmid)^2 / d^2) * sin(π * z / zmax)
+T′_init(x, z) = Tb_init(x, z) * exp(δ * z / 2)
```
jiahe23, line 75: I am opposed to use `'` in var name

examples/hybrid/plane/inertial_gravity_wave.jl
```diff
@@ -0,0 +1,364 @@
...
+    ᶜub .= FT(0)
+    ᶜvb .= FT(0)
+    ᶠwb .= FT(0)
+    max_ikx, max_ikz = (size(ρ̂b_init_array) .- 1) .÷ 2
+    for ikx in (-max_ikx):max_ikx, ikz in (-max_ikz):max_ikz
+        kx = 2 * π / xmax * ikx
+        kz = 2 * π / (2 * zmax) * ikz
+
+        # Fourier coefficient of ᶜρb_init (for current kx and kz)
+        ρ̂b_init = ρ̂b_init_array[ikx + max_ikx + 1, ikz + max_ikz + 1]
```
jiahe23, line 297: Please do not use `ρ̂` such notation

examples/hybrid/hyperdiffusion.jl
```diff
@@ -0,0 +1,78 @@
...
+function hyperdiffusion_tendency!(Yₜ, Y, p, t)
+    ᶜρ = Y.c.ρ
+    ᶜuₕ = Y.c.uₕ
+    (; ᶜp, ᶜχ, ᶜχuₕ, κ₄, use_tempest_mode) = p # assume that ᶜp has been updated
+    point_type = eltype(Fields.local_geometry_field(axes(Y.c)).coordinates)
+
+    if use_tempest_mode
```
jiahe23, line 22: tempest mode does contain the total energy and internal energy set up.

examples/hybrid/hyperdiffusion.jl
```diff
@@ -0,0 +1,78 @@
...
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρθ -= κ₄ * W∇◦ₕ(∇ₕ(ᶜχ))
+        elseif :ρe in propertynames(Y.c)
+            @. ᶜχ = W∇◦ₕ(∇ₕ(Y.c.ρe + ᶜp)) # ᶜχρe
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρe -= κ₄ * W∇◦ₕ(∇ₕ(ᶜχ))
+        elseif :ρe_int in propertynames(Y.c)
+            @. ᶜχ = W∇◦ₕ(∇ₕ(Y.c.ρe_int + ᶜp)) # ᶜχρe_int
+            Spaces.weighted_dss!(ᶜχ)
+            @. Yₜ.c.ρe_int -= κ₄ * W∇◦ₕ(∇ₕ(ᶜχ))
```
jiahe23, line 38: no need for these two and they do not seem correct.

examples/Project.toml
```diff
@@ -18,12 +15,16 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
```
jiahe23, line 18: where do we need this?

jiahe23 (changes requested):
Overall looks good. I like the idea of make it clear between center and face. 

Here are some issues:

hyperdiffusion file needs to revise a bit, @szy21 can you also take a close look at this file just in case I missed something.

unicode is over-used. Here are two examples:
- `ρ` is good, but `ρ̂ ` is too much; and we used to have reports on  superscripts/subscripts/overbars not appearing properly in some browser;
- `hdiv` and `hwdiv` is self-explanatory enough and the symbols only make it hard to read.

szy21:
I agree with Jia that unicode is overused. I’m fine with doing this in climacore for now but let’s not do it in climaatmos. It would be good to talk about guidelines on unicodes before we make changes in climaatmos.

jiahe23:
> I agree with Jia that unicode is overused. I’m fine with doing this in climacore for now but let’s not do it in climaatmos. It would be good to talk about guidelines on unicodes before we make changes in climaatmos.

Yep, let's make minimal changes here and have it in for implicit development; but bear in mind the unicode issue.

dennisYatunin:
I've addressed the above comments as follows:
1. I renamed all the rising bubble filenames to follow the same convention as the sphere filenames, so that they now end with `_rhoe` or `_rhotheta`. I also updated the output directories and buildkite pipeline appropriately.
2. I made the Unicode replacements `∇◦ → div`, `W∇◦ → wdiv`, `∇⨉ → curl`, `W∇⨉ → wcurl`, `∇ → grad`, `W∇ → wgrad`, and `ρ̂ → ρf` (and likewise for all the other variables with hats). Also, although this was not requested, I made the replacement `I → interp`, since `I` almost always means "identity". I did not replace the `′` symbol, since it is already in use in ClimaCore's source code and is a very common mathematical symbol. Please let me know if you feel that is a bad decision; I can replace `′` with `prime`, but I don't feel that this is necessary.
3. I replaced the ρe and ρe_int cases in tempest-mode hyperdiffusion with `error("use_tempest_mode must be false when not using ρθ")`.

In general, how does the following work as a Unicode rule of thumb:
"Only common mathematical symbols, non-English alphabets, English/numerical superscripts, and English/numerical subscripts may be used."

examples/Project.toml
dennisYatunin: This is used for progress bars in `inertial_gravity_wave.jl`. Generating the analytical solution takes a very long time, so I added `@progress` to some for loops.

dennisYatunin:
bors try

jiahe23:
> I've addressed the above comments as follows:
> 
> 1. I renamed all the rising bubble filenames to follow the same convention as the sphere filenames, so that they now end with `_rhoe` or `_rhotheta`. I also updated the output directories and buildkite pipeline appropriately.
> 2. I made the Unicode replacements `∇◦ → div`, `W∇◦ → wdiv`, `∇⨉ → curl`, `W∇⨉ → wcurl`, `∇ → grad`, `W∇ → wgrad`, and `ρ̂ → ρf` (and likewise for all the other variables with hats). Also, although this was not requested, I made the replacement `I → interp`, since `I` almost always means "identity". I did not replace the `′` symbol, since it is already in use in ClimaCore's source code and is a very common mathematical symbol. Please let me know if you feel that is a bad decision; I can replace `′` with `prime`, but I don't feel that this is necessary.
> 3. I replaced the ρe and ρe_int cases in tempest-mode hyperdiffusion with `error("use_tempest_mode must be false when not using ρθ")`.
> 
> In general, how does the following work as a Unicode rule of thumb: "Only common mathematical symbols, non-English alphabets, English/numerical superscripts, and English/numerical subscripts may be used."

Thanks! It looks good.

We could leave the prime for later decision. I find it confusing with its common use in matrix transpose.

jiahe23 (approved):

PR #611, opened 2022-03-21
Add ClimaCoreMakie to examples Project and Manifest

valeriabarra:
This is needed for this week's ClimaWorkshop to be able to have a reproducible environment for people to run examples and test the ClimaCoreMakie support


jakebolewski:
The reason I was hesitant to add these as normal dependencies in the package itself and in the examples is that each require specific system libraries to be installed correctly on Linux (outside of BinaryBuilder deps).  This is probably not an issue on OSX and Win where the OpenGL driver stack is already installed.   I don't know how you want to handle this but that was the original rationale.

valeriabarra:
> The reason I was hesitant to add these as normal dependencies in the package itself and in the examples is that each require specific system libraries to be installed correctly on Linux (outside of BinaryBuilder deps). This is probably not an issue on OSX and Win where the OpenGL driver stack is already installed. I don't know how you want to handle this but that was the original rationale.

Thanks for explaining.

valeriabarra:
@simonbyrne it seems like as is, there would be a circular dependency issue. How do we want to solve this for the workshop?

simonbyrne:
Unfortunately it looks like GLMakie won't work on the cluster. For the workshop it is probably best to use a separate Project.toml file.

PR #610, opened 2022-03-21
Hybrid 3D box limiters advection example

valeriabarra:
This PR adds a test case for 3D limiters in a Cartesian (box) domain. 
The flow is ~non-uniform~ constant in the vertical, rotational in the horizontal and reversed halfway the period.

This is the first example/integration test case in ClimaCore.jl that showcases the usage of `ClimaTimeSteppers`, and that also demonstrates the usage of a third-order upwinding `Upwind3rdOrderBiasedProductC2F` FD operator in the vertical. 

Dependencies for this PR: PR #620, #643, and #683.

Sub-tasks:

- [x] Modify the ODE solve to work separately in the horizontal and in the vertical  (i.e., remove the stage callback mechanism we had when using the `OrdinaryDiffEq` package). Adapted, based in what was done in 2D in #620.
- [x] use third-order upwinding `Upwind3rdOrderBiasedProductC2F` FD operator in the vertical
- [x] Use slotted spheres IC: a slight modification of the 2D slotted cylinders IC  
- [x] Added this example to buildkite's pipeline and check artifacts
- [x] Added documentation 

All done. 

This will close #575 

cc: @simonbyrne

PR #620, opened 2022-03-23
use ClimaTimeSteppers.jl for limiters

### LINKED ISSUE #612: Custom timestepper interface
To use the horizontal limiters in 3D, we need to apply them after the horizontal transport, but before the vertical transport and source terms. In a forward Euler scheme, this would look like:
```
Y_{t+1} = L(Y_t + dt * f_h(Y_t)) + dt*f_v(Y_t)
```

One way to do this is to provide an interface which takes in a forward Euler update, e.g.  `f(yn, y, p, t, dt)` would give you `yn .= y .+ dt .* f(y, p, t)`.

For SSPRK methods, we can use this for via the Shu-Osher form (which allows us to rewrite it as a complex combination of forward Euler updates).

Questions:
 - how much of the OrdinaryDiffEq interface do we want to reuse? Initially it may be easier to just define our own integrator object, and pull in the pieces we want.
 - how do we make this work with IMEX solvers?

cc: @charleskawczynski @OsKnoth

simonbyrne:
This switches the ODE solver used in the plane limiter example to use TimeMachine.jl and the ForwardEulerODEProblem.

Fixes #612


examples/plane/limiters_advection.jl
```diff
@@ -212,48 +215,49 @@ for (k, ne) in enumerate(ne_seq)
...
     ystar = similar(y0)
     parameters = (space = space, min_Q = min_Q, max_Q = max_Q)
-    f!(ystar, y0, parameters, 0.0)
+    f!(ystar, y0, parameters, 0.0, dt)
 
     # Solve the ODE
-    prob = ODEProblem(f!, y0, (0.0, end_time), parameters)
-    sol = solve(
+    prob = ODEProblem(ForwardEulerODEFunction(f!), copy(y0), (0.0, end_time), parameters)
+    yend = solve(
```
simonbyrne, line 251: At the moment, TimeMachine doesn't store the full solution. We should probably have an option for that.

examples/plane/limiters_advection.jl
```diff
@@ -175,7 +176,9 @@ for (k, ne) in enumerate(ne_seq)
         return (ρ = ρ, ρq = Q)
     end
 
-    function f!(ystar, y, parameters, t)
+    y0 = Fields.FieldVector(ρ=y0.ρ, ρq=y0.ρq)
```
simonbyrne, line 179: TimeMachine only works with FieldVectors at the moment, which seems fine.

simonbyrne:
bors try

valeriabarra:
> This switches the ODE solver used in the plane limiter example to use TimeMachine.jl and the ForwardEulerODEProblem.
> 
> Fixes #612

Thanks Simon. I wonder why you used the plane example, since it doesn't need the horizontal/vertical splitting, rather than the hybrid 3D one that I was developing in #610

simonbyrne:
I wanted to see if it worked first. You could merge #610 with this one to see if it works?

examples/plane/limiters_advection.jl
valeriabarra: Yes, that'd be great, so that we can do post-processing analysis easier rather than on-the-fly per each mesh refinement run

valeriabarra:
bors try

valeriabarra:
I tired to download the raw log file from buildkite last successful merged PR to diff'it with the results from this branch, but it was rendering with weird characters, so here is a crude screenshot comparison. 

Results from this branch:
limiters2Dbranch

Results from current `main`:
limiters2Dmain

The coarsest (4x4 elem) run is the one that exhibits the largest differences in the L_inf error. The other runs are on the same order of magnitude. 

I think it would be important to see all steps so that we can inspect the transient behavior in an animation.

valeriabarra:
Without limiters comparison.

From this branch:
[log-nolim-branch.log](https://github.com/CliMA/ClimaCore.jl/files/8366797/log-nolim-branch.log)

From `main`:
[log-nolim-main.log](https://github.com/CliMA/ClimaCore.jl/files/8366798/log-nolim-main.log)

simonbyrne:
bors try

valeriabarra:
@simonbyrne there is a `MethodError: no method matching (::var"#f!#8"{ClimaCore.Fields.Field{ClimaCore.DataLayouts.IJFH` error now

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

examples/plane/limiters_advection.jl
```diff
@@ -211,49 +215,56 @@ for (k, ne) in enumerate(ne_seq)
...
-        Spaces.weighted_dss!(ystar)
+        Spaces.weighted_dss!(ystar.ρq)
         @. ystar.ρq = -D₄ * wdiv(y.ρ * grad(ystar.ρq))
 
         # Add advective flux divergence
-        @. ystar.ρ = -wdiv(y.ρ * u)         # contintuity equation
-        @. ystar.ρq += -wdiv(y.ρq * u)      # adevtion of tracers equation
-    end
+        @. dy.ρ = beta * dy.ρ - alpha * wdiv(y.ρ * u)         # contintuity equation
+        @. dy.ρq = beta * dy.ρq - alpha * wdiv(y.ρq * u) + alpha * ystar.ρq   # adevtion of tracers equation
```
valeriabarra, line 225:
```suggestion
        @. dy.ρq = beta * dy.ρq - alpha * wdiv(y.ρq * u) + alpha * ystar.ρq   # advection of tracers equation
```

valeriabarra:
Comparisons of all test cases (`cosine_bells`, `gaussian_bells`, and `cylinders`) with and without limiters (all without hyperdiffusion though). On the left, always the `main` branch and on the right always the feature branch. 

Cosine bells, no limiter:
image

Cosine bells, with limiter:
image

Gaussian bells, no limiter:
image

Gaussian bells, with limiter:
image

Cylinders, no limiter:
image

Cylinders, with limiter:
image

The most striking differences are on the L∞ error on the finest mesh (which actually could be related to the time step not being optimal for that fine mesh)

valeriabarra:
But they are much more pronounced with the limiter, rather than without

simonbyrne:
wait which case is different?

valeriabarra:
> wait which case is different?

All test cases. You can see from the diffs reported above that without limiters, the main branch and this feature branch errors are on the same order of magnitude (this branch actually slightly better). But with the limiters, this feature branch shows almost consistently across all test cases and resolutions an order of magnitude worse for the erros.

examples/Manifest.toml
```diff
@@ -159,6 +159,14 @@ path = "../lib/ClimaCoreVTK"
 uuid = "c8b6d40d-e815-466f-95ae-c48aefa668fa"
 version = "0.6.0"
 
+[[deps.ClimaTimeSteppers]]
+deps = ["CUDA", "DiffEqBase", "KernelAbstractions", "LinearAlgebra", "MPI", "SciMLBase", "StaticArrays"]
+git-tree-sha1 = "0e9e7c00554eb564e56aac460f7b4cdda8b57367"
+repo-rev = "main"
+repo-url = "https://github.com/CliMA/ClimaTimeSteppers.jl"
+uuid = "595c0a79-7f3d-439a-bc5a-b232dc3bde79"
+version = "0.1.0"
```
simonbyrne, line 168: switch to released version

simonbyrne:
bors try

examples/plane/limiters_advection.jl
valeriabarra: I guess the typo re-appeared in the merge/rebase

valeriabarra: @simonbyrne is there anywhere that this interface with the `alpha` and `beta` is documented? I can't find it anywhere in the ClimaTimeSteppers.jl docs and I'd like to understand what we are doing here.

simonbyrne:
bors try

valeriabarra:
I was going to test this again with and without limiters and compare with the `main` branch today

simonbyrne:
didn't you do that above?

valeriabarra:
Yes, but I thought you pushed some changes after that

simonbyrne:
No, just switched to the released version of ClimaTimeSteppers

examples/plane/limiters_advection.jl
```diff
@@ -211,49 +215,56 @@ for (k, ne) in enumerate(ne_seq)
...
         min_Q = parameters.min_Q
         max_Q = parameters.max_Q
 
         if lim_flag
             # Call quasimonotone limiter, to find optimal ρq (where ρq gets updated in place)
             Limiters.quasimonotone_limiter!(
-                ydoublestar.ρq,
-                ydoublestar.ρ,
+                dy.ρq,
+                dy.ρ,
```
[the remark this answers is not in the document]
valeriabarra: Answer to self, for future reference. No, we're not applying it to the tendency, we're still applying it to the state. The way it's written in CliMATimeSteppers interface it's just like a temporary increment, since we use `alpha = dt` and `beta = 1`.

PR #626, opened 2022-03-24
Add flattened iterators over field & fieldvector variables

charleskawczynski:
One task for debugging the dycore is to check the BCs. This PR adds an iterator over `Field` & `FieldVector`s so that we can more easily iterate over these fields (for plotting).

This PR is basically adding the ClimaCore version of ClimateMachine's `flattened_tup_chain`.

The user-facing components here are `property_chains` and `single_field`.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -252,6 +252,7 @@ include("broadcast.jl")
 include("mapreduce.jl")
 include("compat_diffeq.jl")
 include("fieldvector.jl")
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Maybe `_getproperty` -> `single_field`?

simonbyrne (approved):

charleskawczynski:
Going to merge #628 instead, which includes this commit.

charleskawczynski:
Let's do this incrementally.

PR #629, opened 2022-03-25
Add divergence damping

dennisYatunin:
This PR adds the ability to use divergence damping by setting the value of `divergence_damping_factor` when calling `hyperdiffusion_cache()`. The default value of this factor is `1`.


dennisYatunin:
bors try

simonbyrne (approved):

PR #634, opened 2022-03-26
Add performance writeup to ClimaCore

charleskawczynski:
I wasn't sure exactly where to put this performance writeup, is this okay? Or should I transfer to overleaf?

### DIFF
--- a/docs/src/performance_tips.md
+++ b/docs/src/performance_tips.md
@@ -1,0 +1,103 @@
+# Performance tips and monitoring
+
+This document outlines some performance tips and performance monitoring strategies.
+
+For the most part, this document focuses on some of the common performance gotcha's that we've observed in the CliMA codebase.
+
+There is a very good and thorough overview of performance tips in [Julia's docs](https://docs.julialang.org/en/v1/manual/performance-tips/).
+
+## Avoiding global variables
+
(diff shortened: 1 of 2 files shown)


valeriabarra:
Hi Charlie, thanks for this write up! This is great! Do you think we could use some of this material for a workshop? In that case, I would think we could later port some of it to the CliMAWorkshop repo, together with a non-optimized example to show how to apply these recommendations. 

Let me know what you think. 
Thanks!

charleskawczynski:
> Hi Charlie, thanks for this write up! ...

I'm not sure if I'll have the bandwidth for the CliMA workshop, let's play it by ear. I'm fine if people would like to port this to the workshop repo, though.

simonbyrne (approved):

docs/src/performance_tips.md
```diff
@@ -0,0 +1,105 @@
...
+CliMA's [ReportMetrics.jl](https://github.com/CliMA/ReportMetrics.jl) applies the strategy in the above section and provides a re-useable interface for reporting the top-most important allocations. Here is an example of it in use:
+
+ - rep_workload.jl
+ - perf.jl
+
+```julia
+# File: rep_workload.jl
+import Profile
+
+import SpecialFunctions # for triggering invalidations
```
valeriabarra, line 50: Thanks Charlie for adding this! 
Just a minor question. What do you need this package for here?

charleskawczynski: Ah, yeah, I can remove this, it was for something else

valeriabarra: That'd be nice. Thanks!

valeriabarra (approved):

PR #635, opened 2022-03-26
Add a cleanup script for deleting unwanted files

charleskawczynski:
I'm finding as many as 6 `.DS_Store` files are generated when running a single example. This PR adds a script to help make deleting them easier.


simonbyrne:
Why not just add them to your global exclude file?

simonbyrne:
I think the default is `~/.config/git/ignore`, see https://git-scm.com/docs/gitignore

PR #636, opened 2022-03-28
Flux form equations in staggered non-hydrostatic model

akshaysridhar:
@szy21  Please see changes to include flux form towards box HS tests. 
cc @dennisYatunin
TODO: flags for implicit solver,  cleanup / re-factor driver tendency terms. Currently `flux_form` is a constant flag at the HS experiment level


szy21:
@akshaysridhar Should we close this?

PR #613, opened 2022-03-22
Distribute hybrid driver.

sriharshakandala:
Distribute hybrid driver.

PR #638, opened 2022-03-29
Update driver for distributed runs

dennisYatunin:
This PR reverts all the hybrid examples that use `driver.jl` to how they looked before #613; that is, it isolates all distributed computations to `driver.jl`. The interface for using `driver.jl` is unchanged, except now one must specify `horizontal_mesh_description`, `npoly`, `zmax`, and `zelem` instead of specifying `space` (this is to allow the driver to construct a distributed horizontal topology before constructing the entire hybrid space). Since the inertial gravity wave test case works with `driver.jl` again, I've added it back to the Buildkite pipeline.

In addition, this PR allows restart files to work in distributed mode (any restart file that needs to get read by or written to from multiple processors gets `_pid$pid` appended to its name). Also, while testing this PR, I found it useful to modify `t_end` and `dt_save_to_disk` from the run script (by setting `ENV[T_END]` and `ENV[DT_SAVE_TO_DISK]`), instead of repeatedly modifying the test case files. I've left this change in the PR, in case others also find it helpful.

### DIFF
--- a/examples/common_spaces.jl
+++ b/examples/common_spaces.jl
@@ -1,87 +1,81 @@
 using ClimaCore: Geometry, Domains, Meshes, Topologies, Spaces
 
-abstract type Space{FT} end
-
-Base.@kwdef struct PeriodicLine{FT} <: Space{FT}
-    xmax::FT
-    xelem::Int
-    npoly::Int
-end
-
-Base.@kwdef struct PeriodicRectangle{FT} <: Space{FT}
-    xmax::FT
-    ymax::FT
-    xelem::Int
-    yelem::Int
-    npoly::Int
-end
-
-Base.@kwdef struct CubedSphere{FT} <: Space{FT}
-    radius::FT
-    helem::Int # number of elements along side of each panel (6 panels in total)
-    npoly::Int
-end
-
-Base.@kwdef struct ExtrudedSpace{FT, S <: Space{FT}} <: Space{FT}
-    zmax::FT
-    zelem::Int
-    hspace::S
-end
-
-function make_space((; xmax, xelem, npoly)::PeriodicLine)
+function periodic_line_mesh(; x_max, x_elem)
     domain = Domains.IntervalDomain(
-        Geometry.XPoint(zero(xmax)),
-        Geometry.XPoint(xmax);
+        Geometry.XPoint(zero(x_max)),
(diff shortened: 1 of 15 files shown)


dennisYatunin:
Okay, I've removed the option to have them as environment variables.

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

sriharshakandala:
It is unclear if each process writing its own file is currently supported. It is ideal to write a single file using MPI IO. Its design is currently in the works.
Gathering data onto root process and writing the file may be a feasible temporary solution!

dennisYatunin:
@sriharshakandala The version currently on main does not support restart files when using MPI. I'm sure having each process write to its own file is not optimal, but it does work (I tested it with several hybrid examples). Perhaps we can use this way for now, and then you can add a better mechanism that uses MPI later?

dennisYatunin:
bors try

examples/common_spaces.jl
```diff
@@ -1,87 +1,101 @@
...
-    npoly::Int
+Base.@kwdef struct PeriodicRectangle{FT} <: MeshDescription{FT}
+    x_max::FT
+    y_max::FT
+    x_elem::Int
+    y_elem::Int
 end
 
-Base.@kwdef struct CubedSphere{FT} <: Space{FT}
+Base.@kwdef struct CubedSphere{FT} <: MeshDescription{FT}
```
sriharshakandala, line 17: Can we not use struct names like CubedSphere, etc. There is potential for confusion with structs defined in src/Meshes. Can we replace them with default arguments to make_space / make_mesh functions!

dennisYatunin: How about I replace each struct with function, so `h_mesh = periodic_line_mesh(; x_max, x_elem)`, etc.?

dennisYatunin:
bors try

dennisYatunin:
Per Zhaoyi's request, I have added a call to `weighted_dss!` after the calls to the initial condition functions, just in case the initial conditions are discontinuous across elements.

dennisYatunin:
bors try

dennisYatunin:
bors try-

dennisYatunin:
bors try

Issue #639, opened 2022-03-30
mismatched spaces


simonbyrne:
Now that we are defining more derived spaces, this error is becoming a problem:
https://github.com/CliMA/ClimaCore.jl/blob/427fcc675d86eb094b0a3c4a7047f1bdc87b0c83/src/Fields/broadcast.jl#L119-L127

The aim of this was so that people don't define the same space multiple times. There are a couple of fixes we could do here:

1. Disable it for now, and just check equality? 
  - could make it a warning?
  - or a fast hash?
2. keep track of all spaces that are created (e.g. via weak references).
  - would require more thought on how to handle to construct these

For now, 1 is probably the best option.

PR #640, opened 2022-03-30
Change same space different instance error to warning.

jb-mackay:
Addresses #639

### DIFF
--- a/src/Fields/broadcast.jl
+++ b/src/Fields/broadcast.jl
@@ -114,6 +114,20 @@ function Base.copyto!(
     return dest
 end
 
+allow_mismatched_diagonalized_spaces() = false
+
+@noinline function warn_mismatched_spaces(
+    space1::Type{S},
+    space2::Type{S},
+) where {S <: AbstractSpace}
+    @warn "Broacasted spaces are the same ClimaCore.Spaces type but not the same instance"
+    return nothing
+end
+
+is_diagonalized_spaces(::Type{S}, ::Type{S}) where {S <: AbstractSpace} = true
+
+is_diagonalized_spaces(::Type, ::Type) = false
+
 @noinline function error_mismatched_spaces(
     space1::Type{S},
     space2::Type{S},
@@ -132,7 +146,16 @@ end
     space2::AbstractSpace,
 )
     if space1 !== space2
-        error_mismatched_spaces(typeof(space1), typeof(space2))
+        if is_diagonalized_spaces(typeof(space1), typeof(space2)) &&
+           allow_mismatched_diagonalized_spaces() &&
+           (
+               parent(Spaces.local_geometry_data(space1)) ==
+               parent(Spaces.local_geometry_data(space2))
+           )
+            warn_mismatched_spaces(typeof(space1), typeof(space2))
+        else
+            error_mismatched_spaces(typeof(space1), typeof(space2))
+        end
     end
     return space1
 end
@@ -145,7 +168,16 @@ end
     space2::AbstractSpace,
 )
     if space1 !== space2
-        error_mismatched_spaces(typeof(space1), typeof(space2))
+        if is_diagonalized_spaces(typeof(space1), typeof(space2)) &&
+           allow_mismatched_diagonalized_spaces() &&
+           (
+               parent(Spaces.local_geometry_data(space1)) ==
+               parent(Spaces.local_geometry_data(space2))
+           )
+            warn_mismatched_spaces(typeof(space1), typeof(space2))
+        else
+            error_mismatched_spaces(typeof(space1), typeof(space2))
+        end
     end
     return nothing
 end
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -292,3 +292,21 @@ end
     @test isapprox(Fields.field_values(sin_field)[], FT(0.0); atol = √eps(FT))
     @test isapprox(Fields.field_values(add_field)[], FT(2π))
 end
+
+@testset "Broadcasting same spaces different instances" begin
+    space1 = spectral_space_2D()
+    space2 = spectral_space_2D()
+    field1 = ones(space1)
+    field2 = 2 .* ones(space2)
+    @test Fields.is_diagonalized_spaces(typeof(space1), typeof(space2))
+    @test_throws ErrorException(
+        "Broacasted spaces are the same ClimaCore.Spaces type but not the same instance",
+    ) field1 .= field2
+
+    # turn warning on
+    Fields.allow_mismatched_diagonalized_spaces() = true
+    @test_warn "Broacasted spaces are the same ClimaCore.Spaces type but not the same instance" field1 .=
+        field2
+    @test parent(field1) == parent(field2)
+    Fields.allow_mismatched_diagonalized_spaces() = false
+end


simonbyrne:
Looks like we need to tweak the JET tests to make exclude this: look at `function_filter` in https://aviatesk.github.io/JET.jl/stable/optanalysis/#JET.OptAnalyzer

simonbyrne:
Or you might just be able to stick a `@nospecialized` tag on the function, as they are excluded by default

test/Operators/spectralelement/opt.jl
```diff
@@ -2,6 +2,7 @@ using Test
 using JET
 using LinearAlgebra, IntervalSets
 
+using ClimaCore
```
jb-mackay, line 5:
```suggestion
```

test/Operators/finitedifference/opt.jl
```diff
@@ -224,6 +224,9 @@ end
             center_values = ones(FT, center_space)
             center_velocities = Geometry.WVector.(center_values)
 
+            function_filter(@nospecialize(ft)) =
+                ft !== typeof(Base.CoreLogging.handle_message)
+
```
jb-mackay, line 229: This doesn't seem to be a great solution, as each `@test_opt` that fails this needs to have this function filter added. Plus, other tests still fail.

I'm a little surprised that we are seeing this `runtime dispatch detected: Base.CoreLogging.handle_message` since [`handle_message` uses `@nospecialize`](https://github.com/JuliaLang/julia/blob/244ada361432462012835c93d3bac031e8046793/base/logging.jl#L661-L663). At least given my current understanding of runtime dispatch and `@nospecialize`... 

@simonbyrne any tips?

simonbyrne: @aviatesk any idea what is causing the problem here?

src/Fields/broadcast.jl
```diff
@@ -107,9 +107,8 @@ end
     space1::Type{S},
     space2::Type{S},
 ) where {S <: AbstractSpace}
-    error(
-        "Broacasted spaces are the same ClimaCore.Spaces type but not the same instance",
-    )
+    @warn "Broacasted spaces are the same ClimaCore.Spaces type but not the same instance"
+    return nothing
```
charleskawczynski, line 111: Wouldn't we need to modify `error_mismatched_spaces` below, too? I think JET looks at all possible dispatch options when analyzing this.

charleskawczynski: Also, we should probably change the function to `warn_mismatched_spaces` if this is the new behavior.

jb-mackay: > Also, we should probably change the function to `warn_mismatched_spaces` if this is the new behavior.

Yes, agreed!

> Wouldn't we need to modify error_mismatched_spaces below, too? I think JET looks at all possible dispatch options when analyzing this.

Yeah I think that is the problem we are trying to circumvent as we still want an error if the spaces aren't the same `ClimaCore.Spaces` type.

jb-mackay:
Here is a minimum working example of the issue:

```
using JET

function dummy_warn()
    @warn "Something not quite right"
end

function dummy_error()
    error("Something is broken")
end

function mwe()
    dummy_warn()
    dummy_error()
end

@test_opt mwe() # fails
@code_typed mwe()
```

JET complains about runtime dispatch for [handle_message](https://github.com/JuliaLang/julia/blob/6f8662b2c29f5d642de8f452ec58323ab411e6f4/base/logging.jl#L661)

src/Fields/broadcast.jl
charleskawczynski: Ah, I see. Yeah, I think we need a different (and still cheap) way of comparing spaces if we want this to pass the JET test.. Or for something in JET to change but I know less about that.

test/Operators/finitedifference/opt.jl
aviatesk: That `@nospecialize` doesn't eliminate dynamic dispatch when `logger` is abstract as `handle_message` has multiple methods for different loggers.

I'd say that usage of `function_filter` is a solution until we make the logging module type stable, which I'm not sure is worthwhile as I guess logging is designed to be very dynamic for some good reasons.

aviatesk:
> Is there any other way to throw a warning?

Use `function_filter`. Maybe you want something like:
```julia
macro test_climaopt(ex0...)
    return var"@test_opt"(__source__, __module__, :(function_filter=$clima_function_filter), ex0...)
end
```

> JET does fine with `error`, though not `@error`.

On code path leading to `error` (or, `throw` internally), Julia compiler intentionally disables most optimizations for latency reasons (as error path in general is not really supposed to be performance sensitive), so JET doesn't anything on it neither. If you put `@error` instead, it is just a type unstable code and JET currently doesn't special case it because, e.g. it's certainly very bad for performance if you put it in a heavy loop.

jb-mackay:
> On code path leading to error (or, throw internally), Julia compiler intentionally disables most optimizations for latency reasons (as error path in general is not really supposed to be performance sensitive), so JET doesn't anything on it neither. If you put @error instead, it is just a type unstable code and JET currently doesn't special case it because, e.g. it's certainly very bad for performance if you put it in a heavy loop.

That makes a lot of sense! 

> Use function_filter. Maybe you want something like...

Thank you for taking a look at this, this macro seems like a good solution.

jb-mackay:
@simonbyrne would we want to use `@test_climaopt` to replace all instances of `test_opt`?

simonbyrne:
> @simonbyrne would we want to use `@test_climaopt` to replace all instances of `test_opt`?

that might be the easiest option?

jb-mackay:
@charleskawczynski this change to a warning creates additional allocations that are caught by your tests - what would you advise as the best way to proceed?

charleskawczynski:
> @charleskawczynski this change to a warning creates additional allocations that are caught by your tests - what would you advise as the best way to proceed?

We should find a way to not incur allocations. I'm not sure of the best way to do this. One option is to just disable it all-together (that would be type-stable). That said, it would be nice to continue having some sort of sanity check. Maybe we can define a `debug_spaces() = false` method, and only apply the warning when `debug_spaces() = true`? The only issue I see with this is that it'd be good for us to have tests that typical drivers we're using somehow apply this test.

A long term solution might be: generate a hash from the grid points and add this to the type parameter space, then this check can be applied at compile time.

jb-mackay:
Thank you @charleskawczynski for working on this with me earlier. We are almost ready to go, I would like to add a test that checks that the error is thrown when we want it to be. 

@simonbyrne let us know if you have any suggestions, but I think this is a good way to move forward.

jb-mackay:
@LenkaNovak when this is merged, we will simply need to turn on the warning (and thereby turn off the error) via:

```
import ClimaCore.Fields
Fields.allow_mismatched_diagonalized_spaces() = true
```

This changes the error to a warning when the spaces are the same but different instances. I think we can simply put this in the ClimaCoupler.jl module/file so that this is automatically enabled when the Coupler is imported. We will also probably want to turn off warning logs so that the warning is not printed a million times during a coupled simulation:

```
import Logging
Logging.disable_logging(Logging.Warn)
```

Not sure if we can target this specific warning to be filtered out.

LenkaNovak:
Thanks so much for adding this, Ben! That's really helpful! And yeah, def want to avoid printing out lots of warning messages. If it's just that one, would it be possible to just do an if statement for now?

LenkaNovak:
And are the stack allocations still a problem, with the same instance warning turned off?

jb-mackay:
@LenkaNovak 
> Thanks so much for adding this, Ben! That's really helpful! And yeah, def want to avoid printing out lots of warning messages. If it's just that one, would it be possible to just do an if statement for now?

I don't think an `if` would work here. I really don't know if there is a good way to target specific warnings.

> And are the stack allocations still a problem, with the same instance warning turned off?

With the warning turned off (and error turned on), the allocations are not an issue, due to the way Charlie rewrote this.

src/Fields/broadcast.jl
```diff
@@ -132,7 +146,12 @@ end
     space2::AbstractSpace,
 )
     if space1 !== space2
-        error_mismatched_spaces(typeof(space1), typeof(space2))
+        if is_diagonalized_spaces(typeof(space1), typeof(space2)) &&
+           allow_mismatched_diagonalized_spaces()
+            warn_mismatched_spaces(typeof(space1), typeof(space2))
```
simonbyrne, line 151: should we still check if they are `==` equal?

charleskawczynski: This is inside a `space1 !== space2` branch?

simonbyrne: `!==` is not `===`. We should check if they are `==`.

charleskawczynski: Ah, right, agreed.

jb-mackay: Can you explain this further @simonbyrne or @charleskawczynski, given that the `if space1 !== space2` has not been changed by this PR? I thought we wanted this error to check that memory was not being used wastefully, so `===` for identical memory address makes sense.

simonbyrne: So there are 3 possible cases we should consider:
- spaces are `===`: we're good
- spaces are not `===`, but are `==` (i.e. they are both constructed from the same way). In this case, we should warn, but let it go
- spaces are not `==`: should be an error.

The extra question is how to define `==`: in this case, the easiest is probably to just check if the `local_geometry` fields are `==`?

jb-mackay: - [x]  spaces are `===`: we're good

- [ ] spaces are not `===`, but are `==` (i.e. they are both constructed from the same way). In this case, we should warn, but let it go

- [x]  spaces are not `==`: should be an error.

> The extra question is how to define `==`: in this case, the easiest is probably to just check if the `local_geometry` fields are `==`?

Right, currently our "`==`" is `is_diagonalized_spaces`:
```
is_diagonalized_spaces(::Type{S}, ::Type{S}) where {S <: AbstractSpace} = true

is_diagonalized_spaces(::Type, ::Type) = false
```

Is it insufficient to check at this type level? Or overkill?

jb-mackay:
@simonbyrne I think we are ready, feel free to merge if you are happy or suggest any further changes.

PR #643, opened 2022-03-30
High-order BCs for Upwind 3rd-Order composite operator

valeriabarra:
This is my attempt at adding BCs for the `Upwind3rdOrderBiasedProductC2F` operator that so far was supported only on periodic meshes.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -783,6 +783,7 @@ end
 
 stencil_interior_width(::UpwindBiasedProductC2F, velocity, arg) =
     ((0, 0), (-half, half))
+
 function stencil_interior(::UpwindBiasedProductC2F, loc, idx, velocity, arg)
     space = axes(arg)
     a⁻ = stencil_interior(LeftBiasedC2F(), loc, idx, arg)
@@ -795,6 +796,7 @@ function stencil_interior(::UpwindBiasedProductC2F, loc, idx, velocity, arg)
 end
 
 boundary_width(::UpwindBiasedProductC2F, ::SetValue, velocity, arg) = 1
+
 function stencil_left_boundary(
     ::UpwindBiasedProductC2F,
     bc::SetValue,
@@ -813,6 +815,7 @@ function stencil_left_boundary(
     )
     return Geometry.Contravariant3Vector(upwind_biased_product(vᶠ, aᴸᴮ, a⁺))
 end
+
 function stencil_right_boundary(
     ::UpwindBiasedProductC2F,
     bc::SetValue,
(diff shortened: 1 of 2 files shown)


src/Operators/finitedifference.jl
```diff
@@ -873,7 +878,20 @@ U(v,x)[i] = \\begin{cases}
 This stencil is based on [WickerSkamarock2002](@cite), eq. 4(a).
 
 Supported boundary conditions are:
-- At the moment this is defined only for periodic meshes.
+- [`SetValue(x₀)`](@ref): set the value of `x` to be `x₀` on the boundary. On
+  the left boundary the stencil is
+  ```math
+  U(v,x)[\\tfrac{1}{2}] = \\begin{cases}
+   v[\\tfrac{1}{2}] \\left(8 x_0 + 4 x[1+\\tfrac{1}{2}] \\right) / 12  \\textrm{, if } v[i] > 0 \\\\
+   v[\\tfrac{1}{2}] \\left(4 x_0 + 10 x[1+\\tfrac{1}{2}] -2 x[1+\\tfrac{3}{2}]  \\right) / 12  \\textrm{, if } v[i] < 0
```
valeriabarra, line 886: Not sure if the docs here are a good representation of the code or if this is more aspirational

test/Operators/finitedifference/column.jl
```diff
@@ -675,14 +675,118 @@ end
...
+    end
+
+    # Check convergence rate
+    conv_adv_wc = convergence_rate(err_adv_wc, Δh)
+
+    # Upwind3rdOrderBiasedProductC2F conv, with f(z) = sin(z)
+    @test err_adv_wc[3] ≤ err_adv_wc[2] ≤ err_adv_wc[1] ≤ 5e-2
+    @test conv_adv_wc[1] ≈ 0.5 atol = 0.1
+    @test conv_adv_wc[2] ≈ 0.5 atol = 0.1
+    @test conv_adv_wc[3] ≈ 0.5 atol = 0.1
```
valeriabarra, line 735: @simonbyrne , unfortunately, both with `SetValue` and `Extrapolate` BCs, we now have only 0.5 convergence rate. So I am afraid we have to revisit the way we handled the boundary cases, as this is supposed to be higher-order  :/

simonbyrne:
For the boundaries, instead of defining it for `SetValue`, we want to have options that allow setting the 1st interior face (level 1+half) as either:
- 1st order upwind (e.g. what you would get by `UpwindBiasedProductC2F`
- 3rd order one-sided (e.g. what you would get by using the right-biased reconstruction on the right boundary, and vice-versa)

valeriabarra:
@simonbyrne I tried to apply your suggestions to the best of my understanding but obviously I am missing something. 
I tried to use `stencil_interior(RightBiasedC2F()....)` in the `stencil_left_boundary` function both at the boundary point and the first interior one. If we use it in the first interior one, I am not sure I understand how to set the other ones then

valeriabarra:
@simonbyrne do you know what is left to do here, according to what @OsKnoth suggested?
It would be good if this could be merged by tomorrow's meeting.

simonbyrne (approved):

PR #646, opened 2022-03-31
Apply dss to state in hybrid driver

dennisYatunin:
This PR applies dss to every field in the state Y for all hybrid examples that use `driver.jl`.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -120,9 +120,6 @@ else
         c = center_initial_condition.(ᶜlocal_geometry),
         f = face_initial_condition.(ᶠlocal_geometry),
     )
-    # apply dss in case the initial conditions are discontinuous across elements
-    Spaces.weighted_dss!(Y.c, comms_ctx)
-    Spaces.weighted_dss!(Y.f, comms_ctx)
 end
 p = get_cache(ᶜlocal_geometry, ᶠlocal_geometry, comms_ctx, dt)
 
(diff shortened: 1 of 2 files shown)


examples/hybrid/driver.jl
```diff
@@ -120,9 +120,6 @@ else
         c = center_initial_condition.(ᶜlocal_geometry),
         f = face_initial_condition.(ᶠlocal_geometry),
     )
-    # apply dss in case the initial conditions are discontinuous across elements
-    Spaces.weighted_dss!(Y.c, comms_ctx)
-    Spaces.weighted_dss!(Y.f, comms_ctx)
```
simonbyrne, line 125: I think we may still need this, as it needs to get called every stage

dennisYatunin: `func_start = true` in the callback ensures that it gets called on the initial conditions

simonbyrne: no, i meant it needs to get called every RK stage (as opposed to every step)

dennisYatunin: I don't think there's a way to supply stage callbacks to OrdinaryDiffEq. We can only make a callback that gets called once per tilmestep.

dennisYatunin: In theory, if we dss the tendency on every RK stage, then the state will remain continuous across elements. This is apparently not true after many timestep, but applying dss at the end of every timestep should fix that.

simonbyrne (approved):

PR #650, opened 2022-04-05
Enable multi-setup driver

dennisYatunin:
This PR modifies the hybrid examples driver so that it can handle more than one setup (model specification + domain/initial condition specification + timestepping specification). Instead of each test case file passing several variables to the driver, it now passes a list of setups, and the driver loops over those setups. Post-processing can then be done on a list of `sols`, each of which corresponds to a given setup.

This PR also modifies the inertial gravity wave test case to loop over multiple discretizations (though not all the ones used in the paper because that would be too slow).

### DIFF
--- a/examples/hybrid/sphere/balanced_flow_rhoe.jl
+++ b/examples/hybrid/sphere/balanced_flow_rhoe.jl
@@ -19,6 +19,7 @@ setups = [
         t_end = FT(60 * 60),
         dt = FT(5),
         dt_save_to_sol = FT(50),
+        dt_save_to_disk = FT(0), # 0 means don't save to disk
         ode_algorithm = SSPRK33,
     ),
 ]
--- a/examples/hybrid/sphere/balanced_flow_rhoe.jl
+++ b/examples/hybrid/sphere/balanced_flow_rhoe.jl
@@ -4,32 +4,27 @@ using ClimaCore.DataLayouts
 
 include("baroclinic_wave_utilities.jl")
 
-const sponge = false
+sponge = false
 
-# Variables required for driver.jl (modify as needed)
-horizontal_mesh = cubed_sphere_mesh(; radius = R, h_elem = 4)
-npoly = 4
-z_max = FT(30e3)
-z_elem = 10
-t_end = FT(60 * 60)
-dt = FT(5)
-dt_save_to_sol = FT(50)
-dt_save_to_disk = FT(0) # 0 means don't save to disk
-ode_algorithm = OrdinaryDiffEq.SSPRK33
+setups = [
+    HybridDriverSetup(;
+        additional_cache = make_additional_cache(sponge; κ₄ = FT(2e17)),
+        additional_tendency! = make_additional_tendency(sponge),
+        center_initial_condition = make_center_initial_condition(:ρe, true),
+        face_initial_condition = make_face_initial_condition(),
+        horizontal_mesh = cubed_sphere_mesh(; radius = R, h_elem = 4),
+        npoly = 4,
+        z_max = FT(30e3),
+        z_elem = 10,
+        t_end = FT(60 * 60),
+        dt = FT(5),
+        dt_save_to_sol = FT(50),
+        ode_algorithm = SSPRK33,
+    ),
+]
 
-additional_cache(ᶜlocal_geometry, ᶠlocal_geometry, dt) = merge(
-    hyperdiffusion_cache(ᶜlocal_geometry, ᶠlocal_geometry; κ₄ = FT(2e17)),
-    sponge ? rayleigh_sponge_cache(ᶜlocal_geometry, ᶠlocal_geometry, dt) : (;),
-)
-function additional_tendency!(Yₜ, Y, p, t)
-    hyperdiffusion_tendency!(Yₜ, Y, p, t)
-    sponge && rayleigh_sponge_tendency!(Yₜ, Y, p, t)
-end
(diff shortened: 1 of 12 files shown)


charleskawczynski:
Let's apply these changes after we've moved Held Suarez to ClimaAtmos.

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -4,32 +4,27 @@ using ClimaCore.DataLayouts
...
-
-# Variables required for driver.jl (modify as needed)
-horizontal_mesh = cubed_sphere_mesh(; radius = R, h_elem = 4)
-npoly = 4
-z_max = FT(30e3)
-z_elem = 10
-t_end = FT(60 * 60)
-dt = FT(5)
-dt_save_to_sol = FT(50)
-dt_save_to_disk = FT(0) # 0 means don't save to disk
```
jiahe23, line 17: where does the `dt_save_to_disk` go now?

jiahe23, line 17: where does the `dt_save_to_disk` go now?

examples/hybrid/sphere/balanced_flow_rhoe.jl
```diff
@@ -4,32 +4,27 @@ using ClimaCore.DataLayouts
...
-ode_algorithm = OrdinaryDiffEq.SSPRK33
-
-additional_cache(ᶜlocal_geometry, ᶠlocal_geometry, dt) = merge(
-    hyperdiffusion_cache(ᶜlocal_geometry, ᶠlocal_geometry; κ₄ = FT(2e17)),
```
jiahe23, line 21: and `κ₄`?  people will need to change `κ₄` if they change the resolution

jiahe23, line 21: and `κ₄`?  people will need to change `κ₄` if they change the resolution

jiahe23: oh never mind, just saw it at the very top

jiahe23: oh never mind, just saw it at the very top

jiahe23:
I don't think there is any significant change that may interrupt the pipeline, but just to make sure nothing goes wrong, could you double check the pipeline written [here](https://github.com/CliMA/ClimaCore.jl/tree/main/examples/hybrid/sphere) still works? (run a short simulation, restart with jld2, postprocess to lat/lon and save in nc)

examples/hybrid/sphere/balanced_flow_rhoe.jl
dennisYatunin: Oh, it still goes in the same place as `dt_save_to_sol`. I'll add the `dt_save_to_sol = FT(0)` back into every test case file to clarify that.

valeriabarra: Also, does `save_to_disk` only refers to JLD2 outputs? Is the interface compatible with the other output/visualization options we have? 
I'm honestly a bit lost in the way these files have been modified to account for pre-and-post-processing options

dennisYatunin: `dt_save_to_disk` toggles how frequently restart files are generated. Currently, we only support JLD2 restart files, so that's the only option. All other outputs must be generated in the `postprocessing` function, which is defined in every test-case file.

dennisYatunin: In general, a "driver setup" contains all information needed before/during a simulation run, and "postprocessing" is everything that occurs after a simulation has been run.

examples/hybrid/sphere/held_suarez_rhoe.jl
```diff
@@ -3,35 +3,28 @@ using ClimaCore.DataLayouts
...
-    hyperdiffusion_tendency!(Yₜ, Y, p, t)
-    sponge && rayleigh_sponge_tendency!(Yₜ, Y, p, t)
-    held_suarez_tendency!(Yₜ, Y, p, t)
-end
-
-center_initial_condition(local_geometry) =
-    center_initial_condition(local_geometry, Val(:ρe))
-
-function postprocessing(sol, output_dir)
+function postprocessing(sols, output_dir)
```
valeriabarra, line 26: sorry, I'm trying to follow. How many `sols` there are? Where are they defined?

dennisYatunin: There used to be a single `sol`, which corresponded to a single "setup", though there did not used to be any proper data structure for storing a setup. Now, the driver gets a list of `setups`, and it generates a list of corresponding `sols`.

examples/hybrid/sphere/balanced_flow_rhoe.jl
valeriabarra: @dennisYatunin as we discussed today, it would be nice if these setups and file organization could be documented. Let's add a README.md in this `example/hybrid` directory to explain where things are or need to be modified for pre-and-postprocessing. Thanks!

valeriabarra: Otherwise it is currently a bit hard to follow the flow of the program (what is common to all test cases, what needs to be setup for each test  case, etc...)

PR #651, opened 2022-04-05
update example, docs manifests

simonbyrne:


charleskawczynski:
Shouldn't `julia .dev/up_deps.jl` catch these?

PR #657, opened 2022-04-08
Define set! on fields

charleskawczynski:
This PR adds a function, `set!`:
```julia
"""
    set!(f::Function, field::Field, args = ())
Apply function `f` to populate
values in field `field`. `f` must
have a function signature with signature
`f(::LocalGeometry[, args...])`.
Additional arguments may be passed to
`f` with `args`.
"""
```
This is used in Oceananigans, ClimaAtmos, and it would be useful in TC. I think it makes sense to make it a ClimaCore feature.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -332,4 +332,22 @@ function level(field::FaceExtrudedFiniteDifferenceField, v::PlusHalf)
     Field(data, hspace)
 end
 
+"""
+    set!(f::Function, field::Field, args = ())
+
+Apply function `f` to populate
+values in field `field`. `f` must
+have a function signature with signature
+`f(::LocalGeometry[, args...])`.
+Additional arguments may be passed to
+`f` with `args`.
+"""
+function set!(f::Function, field::Field, args = ())
+    space = axes(field)
+    local_geometry = local_geometry_field(space)
+    field .= f.(local_geometry, args...)
+    return nothing
+end
+
+
 end # module
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -196,6 +196,7 @@ Base.ones(::Spaces.AbstractSpace)
 Base.sum(::Fields.Field)
 Fields.Statistics.mean(::Fields.Field)
 Fields.LinearAlgebra.norm(::Fields.Field)
+Fields.set!
 ```
 
 ## Limiters
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -257,3 +257,18 @@ end
         @test var === Fields.single_field(Y, prop_chain)
     end
 end
+
+@testset "Set!" begin
+    space = spectral_space_2D()
+    function FieldFromNamedTuple(space, nt::NamedTuple)
+        cmv(z) = nt
+        return cmv.(Fields.coordinate_field(space))
+    end
+    FT = Float64
+    nt = (; x = FT(0), y = FT(0))
+    Y = FieldFromNamedTuple(space, nt)
+    foo(local_geom) =
+        sin(local_geom.coordinates.x * local_geom.coordinates.y) + 3
+    Fields.set!(foo, Y.x)
+    @test all((parent(Y.x) .> 1))
+end


src/Fields/Fields.jl
```diff
@@ -332,4 +332,20 @@ function level(field::FaceExtrudedFiniteDifferenceField, v::PlusHalf)
     Field(data, hspace)
 end
 
+"""
+    set!(field::Field, f::Field)
```
valeriabarra, line 336: Thanks Charlie. Does it matter in this signature which one is the input/output Field? If so, do you mind slightly changing the docstring so that this is clearly documented?

src/Fields/Fields.jl
```diff
@@ -332,4 +332,20 @@ function level(field::FaceExtrudedFiniteDifferenceField, v::PlusHalf)
...
+    set!(field::Field, f::Function)
+
+Assign field `field` to values in
+`f`, which may be a field or a
+function.
```
charleskawczynski, line 341: @valeriabarra, the inputs / outputs are explained here-- I'm following the julia convention that the first argument is mutated.

valeriabarra: Yes, thanks, I had seen that. I was just wondering if it was clearer to mention in the docstring that indeed the first argument was going to be the output, but it's ok if you think the convention is clear enough and commonly known

charleskawczynski: Sure, can add a suggestion?

valeriabarra: If the first argument is the input/output, how does something like this sound?
 
```suggestion
Assign values in `f`, 
which may be a field or a
function, to field `field`.
```

valeriabarra: Or did I get what you wanted to say backward?

charleskawczynski: This looks backwards to me, but maybe I've re-read this too many times. @simonbyrne, what do you think?

valeriabarra: Then I must have misunderstood what this does. I thought we said that the first argument (`field`) is both an input and the output (i.e., mutated by the function). Hence I thought the assignment was of the values of `f` to `field`. As in:
```
field <-- f
```

charleskawczynski: I guess I was interpreting
```
Assign values in `f`
```
to mean that `f` is mutating, and that's not the case. I'm fine with this doc string if people find it more clear.

valeriabarra: Oh, how about 

```suggestion
Assign values of `f`, 
which may be a field or a
function, to field `field`.
```
?

valeriabarra (approved):

charleskawczynski:
Hmm, that's a bit weird. The failure doesn't seem related to these changes.

valeriabarra:
Unfortunately, it wouldn't be the first time that we have these rundom CI failures. It happened to some of my PRs in the past too. Just try again, and if not, @simonbyrne might have a better clue

charleskawczynski:
Ah, I see what's going on.. CI is kind of broken because our OrdinaryDiffEq is auto-updating to the latest version when tested via `Pkg.test()`.

src/Fields/Fields.jl
```diff
@@ -332,4 +332,20 @@ function level(field::FaceExtrudedFiniteDifferenceField, v::PlusHalf)
     Field(data, hspace)
 end
 
+"""
+    set!(field::Field, f::Field)
+    set!(field::Field, f::Function)
+
+Assign values of `f`, 
+which may be a field or a
+function, to field `field`.
```
simonbyrne, line 341: Docs should explain what arguments `f` should take.

Also should `f` be the first arg, so we can use `do` notation?

Docstring needs to be added to documenter as well

charleskawczynski:
@simonbyrne pointed out some good points. Maybe we can just define `set!(f::Function, field::Field)` and users can just use `field .= f` if `f isa Field` themselves. I think this is a bit more explicit, and we'll be able to support the `do`-syntax, which seems like a good idea in the first place.

@simonbyrne, how does this sound?

simonbyrne:
That sounds good to me!

valeriabarra (approved):

PR #659, opened 2022-04-08
Generalize remap2latlon

charleskawczynski:
This PR generalizes the `remap2latlon` function in `examples/hybrid/sphere/remap_pipeline.jl`. AFAIU, this function was used to create the animations for the recent Held Suarez runs. This function accepts a FieldVector and exports an NC file. Exporting this NC file during IO callbacks will be useful for both visualization and regression tests. This interface no longer relies on an existing JLD2 file, and is independent of the structure of `Y` (e.g., `Y.c` doesn't need to exist).

### DIFF
--- a/examples/hybrid/sphere/remap_pipeline.jl
+++ b/examples/hybrid/sphere/remap_pipeline.jl
@@ -42,6 +42,21 @@ jld2_files = filter(x -> endswith(x, ".jld2"), readdir(jld2_dir, join = true))
 
 function remap2latlon(filein, nc_dir, nlat, nlon)
     datain = jldopen(filein)
+    t_now = datain["t"]
+    Y = datain["Y"]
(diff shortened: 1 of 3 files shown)


jiahe23:
It looks good to me. Have you tried to run it over some jld2 outputs and confirm it works fine to generate the expected netcdf? @charleskawczynski

charleskawczynski:
> It looks good to me. Have you tried to run it over some jld2 outputs and confirm it works fine to generate the expected netcdf? @charleskawczynski

Working on it!

charleskawczynski:
Still working on this and thinking about design here.. I think a few more changes are needed.

charleskawczynski:
I think the upstream boiler plate has been reduced, and the new workflow has been improved, closing.

Issue #664, opened 2022-04-14
`DataLayout` Backing Array Dimension Checks


jb-mackay:
Some `DataLayout` objects have constructors that allow for backing arrays with dimensions that don't match what one would expect given the stored types.

e.g.
```
FT = Float64
S = Complex{FT}
vf1 = VF{S}(Matrix{FT}, 3)
vf2 = VF{S}(ones(3, 7))
```

Here, `vf1` generates a `VF` struct with a backing array of the expected size (3 x 2 as v = 3 and complex components need two elements). However, `vf2` also works and creates a backing array with size 3 x 7; in this case, more than half of this stored data is inaccessible and unused.

We can add a `typesize` check in the bugged constructor to prevent this. The other `DataLayout` constructors should be reviewed for similar issues.

PR #665, opened 2022-04-14
`DataLayouts` F Dimension Constructor Checks

jb-mackay:
Adds `F` dimension checks on backing data arrays for `DataLayouts`. See #664

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -191,6 +191,7 @@ function IJKFVH{S, Nij, Nk}(array::AbstractArray{T, 6}) where {S, Nij, Nk, T}
     @assert size(array, 2) == Nij
     @assert size(array, 3) == Nk
     check_basetype(T, S)
+    @assert size(array, 4) == typesize(T, S)
     IJKFVH{S, Nij, Nk, typeof(array)}(array)
 end
 
@@ -257,6 +258,7 @@ function IJFH{S, Nij}(array::AbstractArray{T, 4}) where {S, Nij, T}
     @assert size(array, 1) == Nij
     @assert size(array, 2) == Nij
     check_basetype(T, S)
+    @assert size(array, 3) == typesize(T, S)
     IJFH{S, Nij, typeof(array)}(array)
 end
 
@@ -412,6 +414,7 @@ end
 function IFH{S, Ni}(array::AbstractArray{T, 3}) where {S, Ni, T}
     @assert size(array, 1) == Ni
     check_basetype(T, S)
+    @assert size(array, 2) == typesize(T, S)
     IFH{S, Ni, typeof(array)}(array)
 end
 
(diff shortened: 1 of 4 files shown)


jb-mackay:
- [x] `VF`
- [ ] `IF`
- [ ] `IFH`
- [ ] `IJF`
- [ ] `IJFH`
- [ ] `VIFH`
- [ ] `VIJFH`
- [ ] `IJKFVH`

jb-mackay:
@simonbyrne is this all sensible?

jb-mackay:
https://github.com/CliMA/ClimaCore.jl/blob/5540fc43fc00819d0c21fdbb102d9986191dd2a4/test/DataLayouts/data1d.jl#L114-L117

This does not make sense - `data1` and `data2` are arrays, not `VF` structs which we are testing. The code without `@test_prop` errors. They should be replaced with `dl1, dl2`(?), which then fail the JET tests and also error on the mapping functions.

dennisYatunin (approved):

src/DataLayouts/DataLayouts.jl
```diff
@@ -633,16 +633,19 @@ end
...
     VF{S, typeof(array)}(array)
 end
 
 function VF{S}(array::AbstractVector{T}) where {S, T}
+    check_basetype(T, S)
     @assert typesize(T, S) == 1
     VF{S}(reshape(array, (:, 1)))
 end
 
 function VF{S}(ArrayType, nelements) where {S}
```
dennisYatunin, line 646: Might be better to use `::Type{A}` instead of `ArrayType` to ensure that the compiler specializes on `A`.

jb-mackay: You mean `function VF{S}(::Type{A}, nelements) where {S, A}`?

jb-mackay:
@sriharshakandala would you be able to help me understand/debug the errors now in `dss_2d!`?

src/DataLayouts/DataLayouts.jl
simonbyrne: it should still specialize in either case

simonbyrne (approved):

simonbyrne:
Can you figure out why the test is failing?

sriharshakandala:
> @sriharshakandala would you be able to help me understand/debug the errors now in `dss_2d!`?

Its seems to do with the version of ClimaComms being used. Rebasing should fix the issue.

sriharshakandala:
> @sriharshakandala would you be able to help me understand/debug the errors now in `dss_2d!`?

DSS buffer sizing can be different from the number of fields. The sizing is conservative and typically determined based on maximum number of fields being dss'd in a given simulation.

simonbyrne:
Maybe change the assertions to inequalities?

jb-mackay:
> Maybe change the assertions to inequalities?

It seems like the info necessary to compute a conservative upper bound aren't really accessible at the `DataLayout` level. I'm inclined to close this PR and leave things as is.

PR #670, opened 2022-04-15
Remove ClimaCore from own test dependencies.

jb-mackay:


charleskawczynski (approved):
Sorry for the delay. This was a gripe of mine from way back!

PR #671, opened 2022-04-18
add some docs to DSS

simonbyrne:


src/Spaces/dss.jl
```diff
@@ -255,6 +269,14 @@ function fill_send_buffer!(
     return nothing
 end
 
+"""
+    dss_interior_faces!(topology, data [, local_geometry_data=nothing, local_weights=nothing])
+
+Perform DSS on the local interior faces of the topology.
```
valeriabarra, line 275: Thanks for adding these! 
I think it would be better to use ref, as in 
```suggestion
Perform [`DSS`](@ref) on the local interior faces of the topology.
```

simonbyrne: what should that link to?

simonbyrne: The link text needs the function name: https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@ref-link

valeriabarra: Yes, I meant the `weighted_dss` docstring, indeed, but if not helpful or not the correct reference, never mind

PR #672, opened 2022-04-18
Add scaling test

sriharshakandala:
Add scaling test configuration for Held Suarez \rho theta simulation, to buildkite long run pipeline.


simonbyrne (approved):
Would be good to add a README on how to run it.

PR #678, opened 2022-04-20
Add `PointSpace`

jb-mackay:
Adds a 0D `PointSpace` to `ClimaCore.Spaces`. The `PointSpace` allows for the creation of a space existing at a single point. This makes `Spaces` consistent. Calling `level` on a column field and calling `column` on a 2D horizontal field will return a field at that point. Enables #325 to be completed.

### DIFF
--- a/src/Spaces/pointspace.jl
+++ b/src/Spaces/pointspace.jl
@@ -1,0 +1,34 @@
+abstract type AbstractPointSpace <: AbstractSpace end
+
+local_geometry_data(space::AbstractPointSpace) = space.local_geometry
+
+"""
+    PointSpace <: AbstractSpace
+
+A zero-dimensional space. 
+"""
+struct PointSpace{LG} <: AbstractPointSpace
+    local_geometry::LG
+end
+
+function PointSpace(local_geometry::LG) where {LG <: Geometry.LocalGeometry}
+    FT = Geometry.undertype(LG)
+    local_geometry_data = DataLayouts.DataF{LG}(Matrix{FT})
+    local_geometry_data[] = local_geometry
+    return PointSpace(local_geometry_data)
+end
+
+function PointSpace(coord::Geometry.Abstract1DPoint{FT}) where {FT}
+    CoordType = typeof(coord)
+    AIdx = Geometry.coordinate_axis(CoordType)
+    local_geometry = Geometry.LocalGeometry(
+        coord,
+        FT(1.0),
+        FT(1.0),
+        Geometry.AxisTensor(
+            (Geometry.LocalAxis{AIdx}(), Geometry.CovariantAxis{AIdx}()),
+            FT(1.0),
+        ),
+    )
+    return PointSpace(local_geometry)
+end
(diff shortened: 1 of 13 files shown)


src/Spaces/pointspace.jl
```diff
@@ -0,0 +1,17 @@
...
+
+struct PointSpace{LG} <: AbstractPointSpace
+    local_geometry::LG
+end
+
+function PointSpace(coord::Geometry.Abstract1DPoint{FT}) where {FT}
+    CoordType = typeof(coord)
+    AIdx = Geometry.coordinate_axis(CoordType)
+    LG = Geometry.LocalGeometry{AIdx, CoordType, FT, SMatrix{1, 1, FT, 1}}
+    local_geometry = DataLayouts.DataF{LG}([Geometry.component(coord, 1) 1.0 1.0 1.0 1.0])
```
jb-mackay, line 14: Does this seem like a fair way to initialize?

jb-mackay:
To do:
- define `level` for `VF` to return `DataF`
- define `column` for `IJFH, IJF, IFH, IF` to return `DataF`

src/Spaces/pointspace.jl
```diff
@@ -0,0 +1,17 @@
+abstract type AbstractPointSpace <: AbstractSpace end
+
+local_geometry_data(space::AbstractPointSpace) = space.local_geometry
+
+struct PointSpace{LG} <: AbstractPointSpace
```
simonbyrne, line 5: Add a docstring?

src/Spaces/pointspace.jl
simonbyrne: I would suggest defining a constuctor:
```
PointSpace(lg::LocalGeometry)
```
and then defining other constructors in terms of that.

jb-mackay: Makes sense!

jb-mackay: Is there any reason not to simply make that kind of type assertion for the default constructor? e.g.
```
struct PointSpace{LG <: Geometry.LocalGeometry} <: AbstractPointSpace
    local_geometry::LG
end
```

jb-mackay:
@simonbyrne I will get to the above comments tomorrow and will ask your advice on them as well. Please let me know if you think there are additions we should definitely make to the unit tests.

simonbyrne:
> @simonbyrne I will get to the above comments tomorrow and will ask your advice on them as well. Please let me know if you think there are additions we should definitely make to the unit tests.

Tests seem fine. I think the next step `PointField` will require more tests

src/Spaces/pointspace.jl
simonbyrne: e.g. something like
```julia
function PointSpace(local_geometry::LocalGeometry)
    local_geometry_data = DataF{typeof(local_geometry(}()
    local_geometry_data[] = local_geometry
    PointSpace(local_geometry_data)
end
```

jb-mackay:
@simonbyrne take a look and let me know if there are further tests to hit this with or if we are ready to squash & merge.

simonbyrne (approved):
Looks good! When you're ready please squash and merge

jb-mackay:
bors try

PR #679, opened 2022-04-20
fix upwind docs

simonbyrne:


src/Operators/finitedifference.jl
```diff
@@ -937,28 +937,32 @@ return_eltype(::AdvectionOperator, velocity, arg) = eltype(arg)
...
+U(\\boldsymbol{v},x)[i] = \\begin{cases}
+  v^3[i] x[i-\\tfrac{1}{2}]\\boldsymbol{e}_3 \\textrm{, if } v^3[i] > 0 \\\\
+  v^3[i] x[i+\\tfrac{1}{2}]\\boldsymbol{e}_3 \\textrm{, if } v^3[i] < 0
   \\end{cases}
 ```
+where ``\\boldsymbol{e}_3`` is the 3rd covariant basis vector.
 
 Supported boundary conditions are:
-- [`SetValue(x₀)`](@ref): set the value of `x` to be `x₀` on the boundary. On
+- [`SetValue(x₀)`](@ref): set the value of `x` to be `x₀` on the opposing side of the boundary. On
```
valeriabarra, line 954: What does "opposing side of the boundary" means here? It would sound like it would be the opposing face at index 3/2, but it shouldn't, as  `x₀` lives at cell center.

simonbyrne: I clarified that this means a ghost cell.

PR #683, opened 2022-04-26
Fix  ThirdOrderOneSided stencil right boundary assert

valeriabarra:
While reviewing the usage of the `Upwind3rdOrderBiasedProductC2F` operator, I think I stumbled in a small bug that was not caught by the unit test, since it's an assertion for a corner/limit case. 

I also attempted at describing the materialization issue, if one tries to use this operator with a temporary assignment. 

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -1229,7 +1229,7 @@ function stencil_right_boundary(
     arg,
 )
     space = axes(arg)
-    @assert idx <= right_face_boundary_idx(space) + 1
+    @assert idx <= right_face_boundary_idx(space) - 1
 
     vᶠ = Geometry.contravariant3(
         getidx(velocity, loc, idx),
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -1094,6 +1094,9 @@ Supported boundary conditions are:
   and the first-order upwind scheme to compute `x` on the right boundary.
 - [`ThirdOrderOneSided(x₀)`](@ref): uses the third-order downwind reconstruction to compute `x` on the left boundary,
 and the third-order upwind reconstruction to compute `x` on the right boundary.
+
+!!! note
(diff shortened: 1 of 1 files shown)


src/Operators/finitedifference.jl
```diff
@@ -1229,7 +1232,7 @@ function stencil_right_boundary(
     arg,
 )
     space = axes(arg)
-    @assert idx <= right_face_boundary_idx(space) + 1
+    @assert idx <= right_face_boundary_idx(space) - 1
```
valeriabarra, line 1236: @simonbyrne I think I spotted a bug. Or at least, here we were not using the same index as we are in the `FirstOrderOneSided` case, on line 1188 above

src/Operators/finitedifference.jl
```diff
@@ -1094,6 +1094,9 @@ Supported boundary conditions are:
   and the first-order upwind scheme to compute `x` on the right boundary.
 - [`ThirdOrderOneSided(x₀)`](@ref): uses the third-order downwind reconstruction to compute `x` on the left boundary,
 and the third-order upwind reconstruction to compute `x` on the right boundary.
+
+When this operator is composed with a [`DivergenceF2C`](@ref) operator, the user
+should not materialize the result of the `Upwind3rdOrderBiasedProductC2F` operator, but apply the composition directly.
```
simonbyrne, line 1099:
```suggestion
!!! note
    These boundary conditions do not define the value at the actual boundary faces, and so this operator cannot be materialized directly: it needs to be composed with another operator that is does not make use of this value, e.g. a [`DivergenceF2C`](@ref) operator, with a [`SetValue`](@ref) boundary.
```

src/Operators/finitedifference.jl
simonbyrne: Good catch!

valeriabarra:
@simonbyrne if this looks good, can you please approve and merge? Thanks!

Issue #687, opened 2022-04-27
limiters should act on q, not rhoq


simonbyrne:
I noticed the example:
https://github.com/CliMA/ClimaCore.jl/blob/6e2d89f47f7f78f2850e28f6f1875f204114ee21/examples/plane/limiters_advection.jl#L197-L214
we compute the min/max of rhoq, not q. This and the limiter code should be updated.

valeriabarra:
Thanks for pointing this out!
I'll get to this after #674 is merged

PR #689, opened 2022-04-27
Specializing DSS versions

sriharshakandala:
Specialize DSS for various Spaces.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #690, opened 2022-04-29
Limiters: use q_min/q_max instead of Q_min/Q_max

valeriabarra:
This PR addresses #687 . 

It modifies the limiters function and existing examples to use the q_min/q_max instead of Q_min/Q_max. As suspected, since we were always using a ρ starting = 1, we don't notice any difference up to round off precision.

### DIFF
--- a/examples/plane/limiters_advection.jl
+++ b/examples/plane/limiters_advection.jl
@@ -100,7 +100,7 @@ domain = Domains.RectangleDomain(
 )
 
 # Set up spatial discretization
-ne_seq = 2 .^ (2, 3, 4, 5)
+ne_seq = 2 .^ (2, 3, 4)
 Δh = zeros(FT, length(ne_seq))
 L1err, L2err, Linferr = zeros(FT, length(ne_seq)),
 zeros(FT, length(ne_seq)),
@@ -116,8 +116,8 @@ for (k, ne) in enumerate(ne_seq)
 
     # Initialize variables needed for limiters
     n_elems = Topologies.nlocalelems(space.topology)
-    min_Q = zeros(n_elems)
-    max_Q = zeros(n_elems)
+    min_q = zeros(n_elems)
+    max_q = zeros(n_elems)
 
     coords = Fields.coordinate_field(space)
     Δh[k] = (xmax - xmin) / ne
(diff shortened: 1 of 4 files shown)


valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

simonbyrne (approved):

valeriabarra:
bors try

examples/plane/limiters_advection.jl
```diff
@@ -189,33 +189,33 @@ for (k, ne) in enumerate(ne_seq)
...
+                    neigh_elems_q_max[i] = -Inf
                 else
-                    neigh_elems_Q_min[i] =
-                        Fields.minimum(Fields.slab(y.ρq, neigh_elems[i]))
-                    neigh_elems_Q_max[i] =
-                        Fields.maximum(Fields.slab(y.ρq, neigh_elems[i]))
+                    neigh_elems_q_min[i] =
+                        Fields.minimum(Fields.slab(y.ρq ./ y.ρ, neigh_elems[i]))
+                    neigh_elems_q_max[i] =
+                        Fields.maximum(Fields.slab(y.ρq ./ y.ρ, neigh_elems[i]))
```
simonbyrne, line 214: Looking at this more closely, should this be:
```suggestion
                    neigh_elems_q_min[e] =
                        Fields.minimum(Fields.slab(y.ρq ./ y.ρ, neigh_elems[i]))
                    neigh_elems_q_max[e] =
                        Fields.maximum(Fields.slab(y.ρq ./ y.ρ, neigh_elems[i]))
```
?

valeriabarra: Thanks for this review. I am puzzled to understand what is going on here in this PR now with CI. It used to work fine before being rebased onto `main` after the confusion about this file being accidentally edited/reverted by #689 . Could you please shed some light on what is going on? It seems like it just stalls after the `@ info` print-outs

.buildkite/pipeline.yml
```diff
@@ -293,7 +293,7 @@ steps:
       queue: central
       slurm_ntasks: 1
 
-  - label: ":computer: Advection CG plane limiter cosine bells"
+  - label: ":computer: Plane limiters advection cosine bells"
```
simonbyrne, line 296: I don't have a problem with renaming these, but I would prefer it to be a different PR

valeriabarra: Ok, sure. I opened #700 and dropped the commit here.

valeriabarra:
bors try

valeriabarra:
Although this PR did not change the resolution of the integration tests, it seems like bors times out now on these tests (?) I don't understand why. They work locally fine when tested on my machine. 
@simonbyrne do you understand what is going on here with CI?

examples/plane/limiters_advection.jl
```diff
@@ -189,33 +189,33 @@ for (k, ne) in enumerate(ne_seq)
...
-                    neigh_elems_Q_max[i] = -Inf
+                    neigh_elems_q_min[i] = +Inf
+                    neigh_elems_q_max[i] = -Inf
                 else
-                    neigh_elems_Q_min[i] =
-                        Fields.minimum(Fields.slab(y.ρq, neigh_elems[i]))
-                    neigh_elems_Q_max[i] =
-                        Fields.maximum(Fields.slab(y.ρq, neigh_elems[i]))
+                    neigh_elems_q_min[i] =
+                        Fields.minimum(Fields.slab(y.ρq ./ y.ρ, neigh_elems[i]))
```
simonbyrne, line 212: You're doing `y.ρq ./ y.ρ` at every iteration of the loop

valeriabarra: What would you suggest we should do? We need this on each element slab

valeriabarra: One thing I can think of is returning `q` only instead of `ρq` from the limiters routine. But that will require some more changes

valeriabarra:
bors try

charleskawczynski:
It looks like this is significantly increasing the CI time.

valeriabarra:
> It looks like this is significantly increasing the CI time.

It was, but actually with the latest edits, it decreased it significantly

valeriabarra:
But if that's still too slow, I will reduce the number of outputs produced.

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

Issue #686, opened 2022-04-27
Add distributed support for limiters


simonbyrne:
- [ ] `neighboring_elements` should be split into two functions (local and ghost elements), and should be allocation free (i.e. it shouldn't create an `Array` inside).
- [ ] Create a function which computes the min/max values of a given field:
https://github.com/CliMA/ClimaCore.jl/blob/55761cd7a6cd1ee149b48303653b36f8f98d28d4/examples/sphere/limiters_advection.jl#L183-L196

PR #697, opened 2022-04-30
define neighbor iterator

simonbyrne:
This is the first part of  #686.

Note that the `ghost_neighboring_elements` may seem a bit odd (it's the local indices which neighbor a ghost element): the rationale here is that it is probably more efficient to use a different iteration pattern. Currently we do
```
for e in elements
  # compute min/max in element e
  for n in neighbors
    # compute min/max in neighbors e
  end
  out_min[e] = min(min_e, minimum(nbrs_min_e))
end
```
which requires multiple passes over all the nodes

Instead, it is probably better to do
```
for e in elements
   # compute extrema in e
   out_min[e] = min(out_min[e], min_e)
   for n in neighbors
         out_min[n] = min(out_min[n], min_e)
   end
end
```


simonbyrne:
Note that the 2nd one would be difficult to add threading, since there are multiple writes to `out_min[n]`. We can cross that bridge when we get to it.

sriharshakandala:
May be something like
```
for e in local+ghost_elements
    min_e[e] = minimum(e_internal)....
end
for e in local_elements
    out_e = min_e[e]
    for n in neighbors_e
         out_e = min(out_e, min_e[n])....
     end
end
```
Part 1 is the most expensive and can be multithreaded. Part 2 is very cheap.

sriharshakandala (approved):

valeriabarra:
Just to let you know (to avoid reverting changes as it happened last week), I don't know when you last rebased onto `main` this branch, but a bunch of changes have gone through already

src/Topologies/topology2d.jl
```diff
@@ -119,61 +137,10 @@ boundary_tag(topology::Topology2D, boundary_name::Symbol) =
...
-    #    |      |      |      |
-    #    |      |      |      |
-    #    o------o------o------o
-    #    |      |      |      |
-    #    |      |      |      |
-    #    o------o------o------o
-    #    |      |      |      |
-    #    |      |      |      |
-    #    o------o------o------o
-    #
```
valeriabarra, line 136: I thought the ASCII art could have been useful, but it's ok if not. 

@charleskawczynski today when we were chatting about the number of neighboring elements, pulling this up could have helped picturing that

simonbyrne: I'm not sure it makes much sense to keep: the code no longer iterates over the various connections, instead it precomputes the neighbors for each element and iterates directly over those.

simonbyrne: If you think it would be suitable to move elsewhere, then do feel free.

valeriabarra: Absolutely, not a problem if that is not reflecting/helping explaining the current code. I was mentioning that because today Charlie and I were chatting about this neighboring elements search

PR #702, opened 2022-05-04
3D sphere limiters example: deformation flow

valeriabarra:
This PR will close #701 . 
It will add a modified deformation flow example, in which we exhibit how to apply limiters in a 3D sphere example, with a 3rd-order upwinding scheme in the vertical and ClimaTimeSteppers.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #703, opened 2022-05-04
Add great_circle_distance between LatLongZPoints

valeriabarra:
While working on PR #702 , I realized that this small utility function could have been handy (and would have made the 3d sphere example consistent with the 2d sphere one). 

It basically ignores the Z coordinate and calls the existing method for LatLongPoints. 
Since basically re-uses the same method, I did not repeat all of the cases in the unit tests, but only one.


valeriabarra:
bors try

PR #710, opened 2022-05-06
3D Box limiters example: small edit in velocity for 3rd-order upwinding

valeriabarra:
@OsKnoth suggested that we use the velocity interpolated at faces (both horizontal and vertical velocity components) in the third-order upwinding operator. 

### DIFF
--- a/examples/hybrid/box/limiters_advection.jl
+++ b/examples/hybrid/box/limiters_advection.jl
@@ -81,7 +81,7 @@ end
 
 # Advection problem on a 3D Cartesian domain with bounds-preserving quasimonotone horizontal limiter.
 # The initial condition can be set via a command line argument.
-# Possible test cases are: cosine_bells (default), gaussian_bells, and cylinders
+# Possible test cases are: cosine_bells (default), gaussian_bells, and slotted_spheres
 
 FT = Float64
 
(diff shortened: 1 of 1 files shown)


valeriabarra:
bors try

examples/hybrid/box/limiters_advection.jl
```diff
@@ -262,11 +262,15 @@ for (k, horz_ne) in enumerate(horz_ne_seq)
...
+        fuₕ = first_order_Ic2f.(cuₕ)
+        fuvw =
+            Geometry.Contravariant123Vector.(fuₕ) .+
+            Geometry.Contravariant123Vector.(fw)
+        vert_flux_wρ = vdivf2c.(fw .* first_order_Ic2f.(y.ρ))
         vert_flux_wρq =
             vdivf2c.(
                 first_order_Ic2f.(y.ρ) .*
-                third_order_upwind_c2f.(w, y.ρq ./ y.ρ),
+                third_order_upwind_c2f.(fuvw, y.ρq ./ y.ρ),
```
valeriabarra, line 273: @OsKnoth here is where we use `fuvw`

simonbyrne:
Does this make a difference? It will in the presence of topography, but shouldn't otherwise?

OsKnoth:
without orography the first term do nothing, but with orography its different and that is our goal.

valeriabarra:
> Does this make a difference? It will in the presence of topography, but shouldn't otherwise?

The results here in this branch compared to `main` are indeed identical. Do we still want to merge these changes in preparation of what we will need with topography? 

Also, do you know by any chance why buildkite says it cannot find the artifacts? In the error message, I see that the prefix `examples/` from the path is missing, but the file exists and has the full file name, including the `examples/` bit, so I'm not sure what is going on. Thanks!

valeriabarra:
@simonbyrne the results here are identical as expected, since there is no topography for now. Do we want to go ahead and merge this and then I can propagate this same change both in #714  (if we want to update that example without limiters) and #702 ? 

Did you have a chance to see my question above? 

> Also, do you know by any chance why buildkite says it cannot find the artifacts? In the error message, I see that the prefix `examples/` from the path is missing, but the file exists and has the full file name, including the `examples/` bit, so I'm not sure what is going on. Thanks!

PR #714, opened 2022-05-09
Deformation flow: 3rd-order upwinding

valeriabarra:
I'm opening this WIP PR (we don't need to merge) because I am investigating the effects of 3rd-order upwinding alone, since #702 introduces more changes: usage of ClimaTimeStepper, 3rd-order-upwinding, addition of continuity equation. I am investigating why the need for a smaller time step there, why the simulation goes unstable, and decided to do one incremental change only here.

By only switching to 3rd-order upwinding in this test case, keeping resolution and time step the same, we incur in errors. 

### DIFF
--- a/examples/hybrid/sphere/deformation_flow.jl
+++ b/examples/hybrid/sphere/deformation_flow.jl
@@ -47,7 +47,7 @@ function sphere_3D(
     R = 6.37122e6,
     zlim = (0, 12.0e3),
     helem = 4,
-    zelem = 12,
+    zelem = 36,
     npoly = 4,
 )
     FT = Float64
@@ -171,6 +171,10 @@ function rhs!(dydt, y, (coords, face_coords), t)
         top = Operators.SetValue(Geometry.Contravariant3Vector(0.0)),
         bottom = Operators.SetValue(Geometry.Contravariant3Vector(0.0)),
     )
+    third_order_upwind_c2f = Operators.Upwind3rdOrderBiasedProductC2F(
+        bottom = Operators.ThirdOrderOneSided(),
+        top = Operators.ThirdOrderOneSided(),
+    )
     hdiv = Operators.Divergence()
     hwdiv = Operators.WeakDivergence()
     hgrad = Operators.Gradient()
@@ -197,19 +201,19 @@ function rhs!(dydt, y, (coords, face_coords), t)
     cuvw = Geometry.Covariant123Vector.(uₕ) .+ Geometry.Covariant123Vector.(cw)
 
     @. dρq1 -= hdiv(cuvw * ρq1)
-    @. dρq1 -= vdivf2c(w * Ic2f(ρq1))
+    @. dρq1 -= vdivf2c(Ic2f(ρ) * third_order_upwind_c2f.(w, ρq1 ./ ρ))
     @. dρq1 -= vdivf2c(Ic2f(uₕ * ρq1))
 
     @. dρq2 -= hdiv(cuvw * ρq2)
-    @. dρq2 -= vdivf2c(w * Ic2f(ρq2))
+    @. dρq2 -= vdivf2c(Ic2f(ρ) * third_order_upwind_c2f.(w, ρq2 ./ ρ))
     @. dρq2 -= vdivf2c(Ic2f(uₕ * ρq2))
 
     @. dρq3 -= hdiv(cuvw * ρq3)
-    @. dρq3 -= vdivf2c(w * Ic2f(ρq3))
+    @. dρq3 -= vdivf2c(Ic2f(ρ) * third_order_upwind_c2f.(w, ρq3 ./ ρ))
     @. dρq3 -= vdivf2c(Ic2f(uₕ * ρq3))
 
     @. dρq4 -= hdiv(cuvw * ρq4)
-    @. dρq4 -= vdivf2c(w * Ic2f(ρq4))
+    @. dρq4 -= vdivf2c(Ic2f(ρ) * third_order_upwind_c2f.(w, ρq4 ./ ρ))
     @. dρq4 -= vdivf2c(Ic2f(uₕ * ρq4))
 
     Spaces.weighted_dss!(dydt)
--- a/examples/hybrid/sphere/deformation_flow.jl
+++ b/examples/hybrid/sphere/deformation_flow.jl
@@ -14,7 +14,9 @@ import ClimaCore:
     Operators
 import ClimaCore.Utilities: half
 
-using OrdinaryDiffEq: ODEProblem, solve, SSPRK33
+using OrdinaryDiffEq: ODEProblem, solve
+using DiffEqBase
+using ClimaTimeSteppers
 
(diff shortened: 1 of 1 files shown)


examples/hybrid/sphere/deformation_flow.jl
```diff
@@ -197,19 +201,19 @@ function rhs!(dydt, y, (coords, face_coords), t)
     cuvw = Geometry.Covariant123Vector.(uₕ) .+ Geometry.Covariant123Vector.(cw)
 
     @. dρq1 -= hdiv(cuvw * ρq1)
-    @. dρq1 -= vdivf2c(w * Ic2f(ρq1))
+    @. dρq1 -= vdivf2c(third_order_upwind_c2f.(w, ρq1 ./ ρ))
```
simonbyrne, line 204: I think you need to multiply by ρ:
```suggestion
    @. dρq1 -= Ic2f(ρ) * vdivf2c(third_order_upwind_c2f.(w, ρq1 ./ ρ))
```

valeriabarra: Oh, ok. Thanks. I had tried both versions in the two commit. One with `ρ` inside and the other one outside, but I mistakenly did not add it outside. Will do now. Thanks

valeriabarra: @simonbyrne wait, actually, that would give a `LoadError: Broacasted spaces are not the same ClimaCore.Spaces type` error. `ρ` is already at the center and we want the operation to land at centers

valeriabarra: So it will have to be just a simple multiplication by `ρ`, without the interpolation operator

valeriabarra: @simonbyrne I did that in commit [514aa26](https://github.com/CliMA/ClimaCore.jl/pull/714/commits/514aa261cc0f122cb808968d2f8c898d9a248d06) and I still see the same error with roughly the same magnitude, for same `dt` and resolution. 

```
Test Failed at /home/valeria/CliMA/ClimaCore.jl/examples/hybrid/sphere/deformation_flow.jl:244
  Expression: ≈(q1_error, 0.0, atol = 0.7)
   Evaluated: 0.7360619443393059 ≈ 0.0 (atol=0.7)
```

examples/hybrid/sphere/deformation_flow.jl
```diff
@@ -197,19 +201,19 @@ function rhs!(dydt, y, (coords, face_coords), t)
     cuvw = Geometry.Covariant123Vector.(uₕ) .+ Geometry.Covariant123Vector.(cw)
 
     @. dρq1 -= hdiv(cuvw * ρq1)
-    @. dρq1 -= vdivf2c(w * Ic2f(ρq1))
+    @. dρq1 -= ρ * vdivf2c(third_order_upwind_c2f.(w, ρq1 ./ ρ))
```
simonbyrne, line 204: Ah, it should go inside the divergence
```suggestion
    @. dρq1 -= vdivf2c(Ic2f(ρ) * third_order_upwind_c2f.(w, ρq1 ./ ρ))
```

valeriabarra: of course!

valeriabarra: That results in 
```
ERROR: LoadError: MethodError: no method matching Jcontravariant3(::Float64, ::ClimaCore.Geometry.LocalGeometry{(1, 2, 3), ClimaCore.Geometry.LatLongZPoint{Float64}, Float64, StaticArrays.SMatrix{3, 3, Float64, 9}})
Closest candidates are:
  Jcontravariant3(::ClimaCore.Geometry.AxisTensor, ::ClimaCore.Geometry.LocalGeometry) at ~/CliMA/ClimaCore.jl/src/Geometry/conversions.jl:157
Stacktrace:
  [1] stencil_left_boundary
```
I will now try to just put the `Ic2f(ρ) ` bit inside the `vdivf2c` operator

valeriabarra:
@simonbyrne I think I applied the suggestion properly now, but sitll, we cannot use the same time step as in the original test case.

simonbyrne:
How big of a difference are you seeing?

OsKnoth:
Higher order spatial discretization results in tighter time step restrictions for the same explicit time integration method.
https://www.sciencedirect.com/science/article/pii/S0021999108001769. I will test the code tomorrow.

valeriabarra:
Just reporting here on my findings so far: 
Reducing dt from `dt = 60 * 60` six times to `dt = 10 * 60`, but keeping the resolution the same results in:
```
Expression: ≈(q1_error, 0.0, atol = 0.7)
Evaluated: 0.7348440901642557 
```

Keeping the dt the same , i.e., `dt = 60 * 60` but doubling the vertical resolution (`zelem = 24`), results in `q1` passing and `q2`  with a much smaller error:
```
Expression: ≈(q2_error, 0.0, atol = 0.03)
Evaluated: 0.03223662743377384 ≈ 0.0 (atol=0.03)
```

valeriabarra:
Also, I am not sure how these tolerances were chosen when the problem was set up. Some of them are very large (those for `q1` and `q3`)

valeriabarra:
Last update on this investigation: using a 3x finer resolution in the vertical `zelem = 36` instead of `zelem = 12`, is sufficient to stabilize this without any change in the `dt`.

valeriabarra:
bors try

valeriabarra:
We don't have to merge this PR if we don't want to introduce the 3rd-order upwinding operator in this test case. I will continue my development of the 3D sphere limiters example in #702

valeriabarra:
bors try

valeriabarra:
I tried to use ClimaTimeSteppers here in a separate commit and now I see NaNs. We didn't have this with the original time stepper. I might have done something wrong, of course. I am investigating.

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
Latest update on this investigation: Switching to ClimaTimeSteppers.jl alone resulted in a reduction of a factor of two of the time step needed to have a stable simulation.

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
@simonbyrne if we want to merge these changes (i.e., introducing the 3rd-order-upwinding and ClimaTimeSteppers to the current deformation_flow.jl example so that people have an example of how to use these operators), this PR is ready. 

The only thing is that I had to reduce the resolution of a 3x factor. But it can still run in a reasonable time for CI and produces the desired output (actually even sharper ones, given the higher res).

PR #719, opened 2022-05-12
Reformulate limiters

simonbyrne:
Adds a limiter object which stores the necessary buffers. Applying the limiter is now
split across two operations: one to compute the bounds, another to apply the limiter.

Builds on #697.


simonbyrne:
bors try

examples/hybrid/box/limiters_advection.jl
```diff
@@ -373,8 +317,7 @@ for (k, horz_ne) in enumerate(horz_ne_seq)
     ystar = copy(y0)
     parameters = (
         horzspace = horzspace,
-        min_q = min_q,
-        max_q = max_q,
+        limiter = Limiters.QuasiMonotoneLimiter(y0.ρq, y0.ρ),
```
valeriabarra, line 320: Thank you for this much nicer interface! Just a question. I know here `y0` is copied into `ystar` anyway, but to avoid any confusion on the variable names, shouldn't we pass `(ystar.ρq, ystar.ρ)` to the limiter function so that it doesn't look like we would be limiting the initial condition only?

valeriabarra: Also, can you advice/show how this interface would be used in a problem that has multiple tracers, e.g., in #702?

simonbyrne: we could, this is just used to set up the state, it doesn't matter which one we use.

valeriabarra: I know, but given the confusion on the state Vs tendency that was brought up the other day, I'd try to avoid these types of things. 

Also, can you please show/advise on how to use this for an arbitrary number of tracers? Via a variable tuple?

src/Limiters/quasimonotone.jl
```diff
@@ -0,0 +1,348 @@
...
+            end
+            slab_q_bounds = slab(q_bounds, v, h)
+            slab_q_bounds[1] = q_min
+            slab_q_bounds[2] = q_max
+        end
+    end
+    return nothing
+end
+
+function compute_neighbor_bounds_local!(
```
valeriabarra, line 118: Please add documentation for this function.

src/Limiters/quasimonotone.jl
```diff
@@ -0,0 +1,348 @@
...
+            end
+            slab_q_bounds_nbr = slab(q_bounds_nbr, v, h)
+            slab_q_bounds_nbr[1] = q_min
+            slab_q_bounds_nbr[2] = q_max
+        end
+    end
+    return nothing
+end
+
+function compute_neighbor_bounds_ghost!(
```
valeriabarra, line 143: Please add documentation for this function.

src/Limiters/quasimonotone.jl
```diff
@@ -0,0 +1,348 @@
...
+            end
+        end
+    end
+    return nothing
+end
+
+"""
+    compute_bounds!(limiter::QuasiMonotoneLimiter, ρq::Field, ρ::Field)
+
+Compute the tracer bounds for the tracer concentration `q` based on the fields `ρq` and `ρ`.
```
valeriabarra, line 174:
```suggestion
Compute the desired bounds for the tracer concentration `q`, based on the fields `ρq` and `ρ`.
```

src/Limiters/quasimonotone.jl
```diff
@@ -0,0 +1,348 @@
...
+    end
+end
+
+
+
+
+"""
+    apply_limiter!(ρq, ρ, limiter::QuasiMonotoneLimiter)
+
+Apply the computed bounds of the tracer concentration (`q`) in the limiter to `ρq`.
```
valeriabarra, line 217: Does the following sound more what we are doing?
```suggestion
Apply the limiter on the tracer density  `ρq`, using the computed desired bounds and density, `ρ`, as an optimal weight.
```

src/Limiters/quasimonotone.jl
```diff
@@ -0,0 +1,348 @@
...
+            slab_ρq = slab(ρq_data, v, h)
+            slab_wj = slab(WJ_data, v, h)
+            slab_q_bounds = slab(limiter.q_bounds_nbr, v, h)
+            apply_limit_slab!(slab_ρq, slab_ρ, slab_wj, slab_q_bounds)
+        end # end of horz elem loop
+    end # end of vert level loop
+    return ρq
+end
+
+function apply_limit_slab!(
```
valeriabarra, line 246: Please add documentation to this function. I see that here is where the bulk of the original implementation was. 
By any chance, did you use `git mv` when renaming this file? It helps track the history of the changes. Especially on GitHub, which is very finicky about showing history of files after rebases already (but I use a Chrome extension to see those). So please consider using `git mv` if you haven't done so. Thank you!

simonbyrne: I did, but `git mv` doesn't actually do anything fancy internally: because git is content-addressed, it is the same as deleting and adding a new file: https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F

valeriabarra: Yes, I know. I resorted using this Chrome extension that sometimes helps 
https://chrome.google.com/webstore/detail/git-history-browser-exten/laghnmifffncfonaoffcndocllegejnf?hl=en

PR #729, opened 2022-05-17
3D box limiters example: apply limiters between horz/vert tendency

valeriabarra:
This PR implements the suggestion of applying limiters after the horizontal tendency specification, but before the vertical one in the 3D box limiters example. The results are comparable with what we had already. So the change is not significant.  


valeriabarra:
bors try

PR #734, opened 2022-05-20
Add profiling.

sriharshakandala:
Add Nsight profiling for distributed simulations.


sriharshakandala:
bors try

simonbyrne:
Do you mind if we leave this open for a bit? I think we want to register NVTX first.

sriharshakandala:
> NVTX.jl has now been registered: can you switch to using the registered version?

Done.

sriharshakandala:
bors try

charleskawczynski:
This looks like a pretty invasive application of monitoring, can't we just apply this to `OrdinaryDiffEq.step!`? Or will `NVTX` not provide sufficient granularity?

sriharshakandala:
> This looks like a pretty invasive application of monitoring, can't we just apply this to `OrdinaryDiffEq.step!`? Or will `NVTX` not provide sufficient granularity? Or is there some other reason / issue?

One objective was to measure the relative time being spent in the DSS portions of the code per timestep, which is what we did here.
I did try to expose the step of the ODE solver, but it was very entangled. Also, given that we will be moving to CliMATimesteppers anyways, I did not think it was worth the effort at this point.

PR #752, opened 2022-06-12
Obtain elemorder from space filling curve.

sriharshakandala:
Obtain `elemorder` from space filling curve.
cs_sfc


simonbyrne:
Looks good:
- you will need to `Pkg.resolve` the `docs` Project
- can we add a couple of tests? e.g. that every element is connected to its successor
- should we make this the default?

sriharshakandala:
bors try

test/Project.toml
```diff
@@ -10,6 +10,7 @@ CubedSphere = "7445602f-e544-4518-8976-18f8e8ae6cdb"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GaussQuadrature = "d54b0c1a-921d-58e0-8e36-89d8069c0969"
+GilbertCurves = "88fa7841-ef32-4516-bb70-c6ec135699d9"
```
simonbyrne, line 13: You're not actually using it directly in the tests, so it doesn't need to be a test dependency

src/Topologies/dtopology2d.jl
```diff
@@ -85,6 +86,41 @@ struct DistributedTopology2D{
...
+    elemorder = CartesianIndex{3}[]
+    for j in 1:6
+        push!(
+            elemorder,
+            [
+                CartesianIndex(i, j) for
+                i in gilbertindices((ne, ne); majdim = majordim[j])
+            ]...,
+        )
+    end
```
simonbyrne, line 108: Use slightly more descriptive names, and `append!` rather than splatting into `push!`
```suggestion
    for panel in 1:6
        append!(
            elemorder,
            [
                CartesianIndex(cartidx, panel) for
                cartidx in gilbertindices((ne, ne); majdim = majordim[panel])
            ],
        )
    end
```

test/runtests.jl
```diff
@@ -19,7 +19,9 @@ using Base: operator_associativity
 @time include("Meshes/cubedsphere.jl")
 
 @time include("Topologies/rectangle.jl")
+@time include("Topologies/rectangle_sfc.jl")
 @time include("Topologies/cubedsphere.jl")
+@time include("Topologies/cubedsphere_sfc.jl")
```
simonbyrne, line 24: Could you add a couple of unit tests that checks the necessary properties of `spacefillingcurve` and `linearindices` for both rectangle and cubed sphere meshes:

- that `spacefillingcurve(mesh)` actually includes all elements (sort and compare with `elements(mesh)`)
- that `linearindices` is actually the inverse (iterate through all elements, and check that it gives the correct values)

(you shouldn't need to construct a topology for these)

src/Topologies/dtopology2d.jl
```diff
@@ -85,6 +86,41 @@ struct DistributedTopology2D{
     boundaries::BF
 end
 
+"""
+    spacefillingcurve(mesh::Meshes.AbstractCubedSphere)
+
+Generate element ordering, `elemorder`, based on a space filling curve
+for a `CubedSphere` mesh.
+
+"""
```
simonbyrne, line 95: can you embed the image you used in the PR into the docs? add it to `docs/src/assets` and then add a link via
```
![descriptive caption for the image](assets/image.png)
```

src/Topologies/dtopology2d.jl
sriharshakandala: append!  is not working correctly with CartesianIndices here! All other revisions are incorporated

sriharshakandala:
bors try

PR #739, opened 2022-05-29
add lengthscale functions to mesh and spaces

simonbyrne:
This will make https://github.com/CliMA/ClimaAtmos.jl/issues/491 easier

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #757, opened 2022-06-17
2D rising thermal bubble: increase dt and final time

### LINKED ISSUE #756: Examples: Verify largest Δt for 2D bubble invariant rhoe problem
For the requested investigation about time-stepping limitations (if any) in using high-order upwinding operators in our applications, it is first useful to determine what is the largest Δt we can run with in a simple, explicit problem set up, such as the 2D rising thermal bubble.

Try increasing the current Δt in the experiment

valeriabarra:
This closes #756 

I tried to push Δt as large as I could, with this configuration, but it could only be doubled. Also running for longer time to see the interesting stuff happening in the output videos.

cc: @akshaysridhar 

### DIFF
--- a/examples/hybrid/plane/bubble_2d_invariant_rhoe.jl
+++ b/examples/hybrid/plane/bubble_2d_invariant_rhoe.jl
@@ -67,7 +67,7 @@ const T_0 = 273.16 # triple point temperature
 Φ(z) = grav * z
 
 # Reference: https://journals.ametsoc.org/view/journals/mwre/140/4/mwr-d-10-05073.1.xml, Section 5a
-# Prognostic thermodynamic variable: Total Energy 
+# Prognostic thermodynamic variable: Total Energy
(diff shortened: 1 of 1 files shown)


valeriabarra:
bors try

akshaysridhar (approved):

valeriabarra:
> @valeriabarra Since we are updating these, it would be useful to include the forms of grid-dependent kappa_4 defined in (77)-(79) in https://gmd.copernicus.org/articles/9/2007/2016/. It seems in this instance we have 25m uniform resolution (minimal nodal distance in the horizontal is then smaller) so dt ≈ 0.07 results in Courant number = 1 (acoustic courant)

Thanks Akshay, there is a separate open PR by @simonbyrne that adds that support. So I was thinking we could update the examples once that is in.

valeriabarra:
That is #739 I was referring to

PR #760, opened 2022-06-17
add bycolumn function

simonbyrne:
This should make it easier to apply multiple column operations in a single pass, and leverage threading.

Example:
```
bycolumn(axes(f)) do colidx
    @. dg[colidx] = ∇(f[colidx])
    @. df[colidx] = div(g[colidx])
end
```

The reason to use `getindex` here is that it doesn't get broadcasted by `@.`, so makes writing these operations easier.

### DIFF
--- a/src/Fields/indices.jl
+++ b/src/Fields/indices.jl
@@ -43,8 +43,8 @@ multiple column-wise operations in a single pass, making use of multiple threads
 div = DivergenceC2F()
 
 bycolumn(axes(f)) do colidx
-    @. dg[colidx] = ∇(f[colidx])
-    @. df[colidx] = div(g[colidx])
+    @. ∇f[colidx] = ∇(f[colidx])
+    @. df[colidx] = div(∇f[colidx])
 end
 ```
 """
--- a/src/Fields/indices.jl
+++ b/src/Fields/indices.jl
@@ -1,0 +1,75 @@
+"""
+    ColumnIndex(ij,h)
+
+An index into a column of a field. This can be used as an argument to `getindex`
+of a `Field`, to return a field on that column.
+
+# Example
+```julia
+colidx = ColumnIndex((1,1),1)
+field[colidx]
+```
+"""
+struct ColumnIndex{N}
+    ij::NTuple{N, Int}
+    h::Int
+end
+
+Base.getindex(field::Field, colidx::ColumnIndex) = column(field, colidx)
+
+function column(field::SpectralElementField1D, colidx::ColumnIndex{1})
+    column(field, colidx.ij[1], colidx.h)
+end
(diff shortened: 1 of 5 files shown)


charleskawczynski (commented):
I'll try testing this out with TC on Monday

charleskawczynski:
Better yet, can you add a test that tests this out with ntuple fields? ( and does the overloading/indexing like we do in TC? )

Basically, in a test, you want to test this, with ntuple fields alongside what's done in #515.

src/Fields/indices.jl
```diff
@@ -0,0 +1,75 @@
...
+
+# Example
+
+```julia
+∇ = GradientF2C()
+div = DivergenceC2F()
+
+bycolumn(axes(f)) do colidx
+    @. dg[colidx] = ∇(f[colidx])
+    @. df[colidx] = div(g[colidx])
```
charleskawczynski, line 47:
```suggestion
    @. ∇f[colidx] = ∇(f[colidx])
    @. df[colidx] = div(∇f[colidx])
```

charleskawczynski:
Actually, I think that this should be fine since we're overloading `getindex` on `ColumnIndex` (`Base.getindex(field::Field, colidx::ColumnIndex) = column(field, colidx)`).

src/Fields/indices.jl
```diff
@@ -0,0 +1,75 @@
...
+bycolumn(axes(f)) do colidx
+    @. ∇f[colidx] = ∇(f[colidx])
+    @. df[colidx] = div(∇f[colidx])
+end
+```
+"""
+function bycolumn(fn, space::Spaces.SpectralElementSpace1D)
+    Nh = Topologies.nlocalelems(space)
+    Nq = Spaces.Quadratures.degrees_of_freedom(Spaces.quadrature_style(space))
+    Threads.@threads for h in 1:Nh
```
charleskawczynski, line 54: Do we need `@inbounds`? And the `@threads` is ignored when no threading is used, correct?

src/Fields/indices.jl
```diff
@@ -0,0 +1,75 @@
...
+            fn(ColumnIndex((i,), h))
+        end
+    end
+    return nothing
+end
+function bycolumn(fn, space::Spaces.SpectralElementSpace2D)
+    Nh = Topologies.nlocalelems(space)
+    Nq = Spaces.Quadratures.degrees_of_freedom(Spaces.quadrature_style(space))
+    Threads.@threads for h in 1:Nh
+        for j in 1:Nq, i in 1:Nq
```
charleskawczynski, line 65: Do we need `@inbounds`? And the `@threads` is ignored when no threading is used, correct?

simonbyrne: I'm not sure: I think `@threads` still has some overhead, so we may want to make it optional.

Issue #747, opened 2022-06-06
Standard data format


simonbyrne:
## Purpose
We need a standard way to save data for restart and reanalysis, which is
 a. performant: there should be minimal conversion from in-memory format
 b. self-describing: it contains sufficient metadata to reload without access to the script and options used to generate it
 c. supports distributed fields: uses MPI I/O or some other distributed layer

Relevant issues:
- https://github.com/CliMA/ClimateMachine.jl/issues/114
- https://github.com/CliMA/ClimaCore.jl/issues/547

### Costs
- it will take some time to implement all possible backends, but we will initially focus on one
- although the container format is standard, the conventions are not the same as e.g. CF conventions. As a result, extra steps will be required to convert to tools which use this format.
- it is not compatible with the NetCDF CF conventions: this would require additional post-processing.

### Benefits
- saving data from distributed runs, using distributed I/O
- will allow us to read/write data in a single format for both restart and for post-processing
- we can potentially reduce memory consumption, by writing data without any intermediate copies
- make it possible to reload data after upgrading packages, as it won't serialize internals

### Risks
- C libraries which make use of parallel MPI I/O (HDF5, NetCDF, Adios2) need to be compiled for a particular MPI implementation.
- Future changes may mean that it is not possible to retain full backward compatibility. By adding the ClimaCore version to the file, we can hopefully reduce this risk.
- It may not be possible to support all features in each container format

## Producers

- Simon
- Sriharsha
- Ben

## Components

The main components will consist of
 - some sort of hierarchical container file: HDF5, NetCDF, Zarr, ADIOS2 BP: below we use the HDF5 naming convention of datasets, groups and attributes, but analogies to these exist in all of these formats.
 - a metadata and naming standard for representing the data in that file
 - code for saving/loading data in the relevant file format adhering to the standard

## Inputs
A description of the inputs to the solution (designs, references, equations, closures, discussions etc).

### Format description

Overall, the general principal of this design is to match the ClimaCore representation of spatial discretization internals wherever possible, make use of self-description features where possible (e.g. determine floating-point types from the container format instead of including it as an attribute), and only store what cannot be directly recomputed at load time.

The top level will consist of the following groups:
- `/domains`
- `/meshes`
- `/topologies`
- `/spaces`
- `/fields`
- `/timescales`

and the relevant data will be written to each group.

[...]

PR #762, opened 2022-06-21
HDF5 reader/writer

simonbyrne:
Implement #747.
Work in progress


src/IO/writers.jl
```diff
@@ -0,0 +1,204 @@
...
+
+"""
+function write_new(
+    writer::HDF5Writer,
+    mesh::Meshes.IntervalMesh,
+    name::AbstractString = "interval",
+)
+    group = create_group(writer.file, name)
+    write_attribute(group, "type", "IntervalMesh")
+    write_new(writer, mesh.domain, "$name/domain")
```
simonbyrne, line 68:
```suggestion
    write_attribute(group, "domain", HDF5.name(write(writer, mesh.domain))
```

src/IO/writers.jl
```diff
@@ -0,0 +1,256 @@
...
+function write_new(
+    writer::HDF5Writer,
+    mesh::Meshes.IntervalMesh,
+    name::AbstractString = defaultname(mesh),
+)
+    group = create_group(writer.file, name)
+    write_attribute(group, "type", "IntervalMesh")
+    write_new(writer, mesh.domain, "$name/domain")
+    write_attribute(group, "nelements", length(mesh.faces))
+    return nothing
```
simonbyrne, line 90: return `name`?

src/IO/writers.jl
```diff
@@ -0,0 +1,256 @@
...
+    writer::HDF5Writer,
+    mesh::Meshes.RectilinearMesh,
+    name::AbstractString = defaultname(mesh),
+)
+    group = create_group(writer.file, "meshes/$name")
+    write_attribute(group, "type", "RectilinearMesh")
+    write_attribute(group, "domain", typeof(mesh.domain))
+    write_attribute(group, "interval1", "interval1") # TODO
+    write_attribute(group, "inverval2", "interval2") # TODO
+    return group
```
simonbyrne, line 103: return `name`

src/IO/readers.jl
```diff
@@ -0,0 +1,193 @@
...
+            DataLayout = eval(Meta.parse(data_layout))
+            ElType = eval(
+                Meta.parse(replace(attrs(g)["value_type"], "ClimaCore." => "")),
+            )
+            values = DataLayout{ElType, Nij}(data)
+            spacename =
+                occursin("c", fieldname) ?
+                "CenterExtrudedFiniteDifferenceSpace" :
+                "FaceExtrudedFiniteDifferenceSpace"
+            cache[fieldname] = Fields.Field(values, cache[spacename])
```
simonbyrne, line 170: We probably don't need to cache fields

src/IO/readers.jl
```diff
@@ -0,0 +1,193 @@
...
+    return nothing
+end
+
+function read_fieldvector(reader::HDF5Reader)
+    read_domains(reader)
+    read_meshes(reader)
+    read_topologies(reader)
+    read_spaces(reader)
+    read_fields(reader)
+    return Fields.FieldVector(c = reader.cache["c"], f = reader.cache["f"])
```
simonbyrne, line 182: we don't want to hardcode names

sriharshakandala:
bors try

sriharshakandala:
> Does this work for with MPI runs?

The current PR is intended for use with serial runs.
We can temporarily try the case where each process writes its data to a separate file, while parallel HDF5 support is being added.  However, parallel HDF5 is recommended for parallel runs!

jiahe23 (approved):
looks good. please clean up commit history and merge

LenkaNovak:
I'm trying to run this from `ClimaAtmos` and getting an error because `IO` the module name clashes with `IO` the DataType. I then tried `restart_Y = ClimaCore.IO.read("testfile.hdf5", "Y")` and this doesn't recognize ClimaCore with
Screen Shot 2022-07-13 at 5 59 49 PM
:

LenkaNovak:
It's also strange that it's throwing the error [here](https://github.com/CliMA/ClimaCore.jl/blob/dcbedcc09a98487ba0ddf6642fd31acb27cdec59/src/IO/readers.jl#L146). I'm happy to connect tomorrow, but it would be good to see that Atmos state can be saved / read before merging. (unless Jia's already checked and I'm doing something I shouldn't ;) ). Otherwise the code looks very neat and legible.

LenkaNovak (changes requested):
(see comments above)

sriharshakandala:
I am open to a different name.  We do plan to overload `read/write` to support other formats as well (e.g.: custom binary) in the future. cc: @simonbyrne @bischtob

src/IO/IO.jl
```diff
@@ -0,0 +1,6 @@
+module IO
```
simonbyrne, line 1: Following discussion in the standup meeting, lets rename this to be `InputOutput`

src/IO/writers.jl
```diff
@@ -0,0 +1,391 @@
...
+    domain::Domains.IntervalDomain,
+    name::AbstractString = defaultname(domain),
+)
+    group = create_group(writer.file, name)
+    write_attribute(group, "type", "IntervalDomain")
+    write_attribute(group, "coord_type", string(typeof(domain.coord_min)))
+    write_attribute(group, "coord_min", string(domain.coord_min))
+    write_attribute(group, "coord_max", string(domain.coord_max))
+    !isnothing(domain.boundary_names) &&
+        write_attribute(group, "boundary_names", string(domain.boundary_names))
```
simonbyrne, line 73: write this as a collection of strings, not a single string
```suggestion
        write_attribute(group, "boundary_names", map(string, domain.boundary_names)))
```

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+function read_new!(
+    reader::HDF5Reader,
+    name::AbstractString,
+    ::Type{IntervalDomain},
+)
+    g = reader.file["domains/" * name]
+    coord_type = replace(attrs(g)["coord_type"], "ClimaCore." => "")
+    boundary_names =
+        haskey(attributes(g), "boundary_names") ?
+        eval(Meta.parse(attrs(g)["boundary_names"])) : nothing
```
simonbyrne, line 105: by writing this as a collection of strings, you should be able to do this without `eval`

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+    periodic = boundary_names === nothing ? true : false
+    CoordType = eval(Meta.parse(coord_type))
```
simonbyrne, line 107: There are only handful of coord types: can we just make this an if-else branch instead?

src/IO/writers.jl
```diff
@@ -0,0 +1,391 @@
...
+    writer::HDF5Writer,
+    mesh::Meshes.IntervalMesh,
+    name::AbstractString = defaultname(mesh),
+)
+    meshname = split(name, "/")[end]
+    group = create_group(writer.file, name)
+    write_attribute(group, "type", "IntervalMesh")
+    write!(writer, mesh.domain, "domains/$meshname")
+    write_attribute(group, "nelements", length(mesh.faces))
+    write_attribute(group, "faces", string(mesh.faces))
```
simonbyrne, line 114: if `faces` is an `Array`, it should be written as an array. If it's a range, we shouldn't need to write it?

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+function read_new!(
+    reader::HDF5Reader,
+    name::AbstractString,
+    ::Type{IntervalMesh},
+)
+    (; file, cache) = reader
+    g = file["meshes/" * name]
+    FT = attrs(g)["FT"] # Float type
+    faces_string = replace(attrs(g)["faces"], "ZPoint" => "ZPoint{$FT}")
+    faces = eval(Meta.parse(faces_string))
```
simonbyrne, line 146: Again, if we write it as an `Array` we shouldn't need to `eval` here

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+Reads an object named 'name' of type 'EquiangularCubedSphere'.
+"""
+function read_new!(
+    reader::HDF5Reader,
+    name::AbstractString,
+    ::Type{EquiangularCubedSphere},
+)
+    (; file, cache) = reader
+    g = file["meshes/" * name]
+    localelementmap = eval(Meta.parse(attrs(g)["localelementmap"]))
```
simonbyrne, line 186: use a branch instead of `eval`

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+    ::Type{SpectralElementSpace1D},
+)
+    (; file, cache) = reader
+    g = file["spaces/" * name]
+    quadrature_num_points = attrs(g)["quadrature_num_points"]
+    quadrature_style = eval(
+        Meta.parse(
+            attrs(g)["quadrature_type"] * "{$(quadrature_num_points)}()",
+        ),
+    )
```
simonbyrne, line 248: again, branch instead of `eval`

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+    ::Type{SpectralElementSpace2D},
+)
+    (; file, cache) = reader
+    g = file["spaces/" * name]
+    quadrature_num_points = attrs(g)["quadrature_num_points"]
+    quadrature_style = eval(
+        Meta.parse(
+            attrs(g)["quadrature_type"] * "{$(quadrature_num_points)}()",
+        ),
+    )
```
simonbyrne, line 274: branch instead of `eval` (can write a helper function to avoid duplicating code with above)

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+Reads an object named 'name' of type 'ExtrudedFiniteDifferenceSpace'.
+"""
+function read_new!(
+    reader::HDF5Reader,
+    name::AbstractString,
+    ::Type{T},
+) where {T <: ExtrudedFiniteDifferenceSpace}
+    (; file, cache) = reader
+    g = file["spaces/" * name]
+    staggering = eval(Meta.parse(attrs(g)["staggering"]))
```
simonbyrne, line 295: if-else branch insetad of eval

src/IO/readers.jl
```diff
@@ -0,0 +1,356 @@
...
+function read_new!(reader::HDF5Reader, name::AbstractString, ::Type{Field})
+    (; file, cache) = reader
+    @assert name ∈ keys(file["fields"])
+    g = file["fields/" * name]
+    data = read_dataset(g, "data")
+    data_layout = attrs(g)["data_layout"]
+    Nij = size(data, findfirst("I", data_layout)[1])
+    DataLayout = eval(Meta.parse(data_layout))
+    ElType =
+        eval(Meta.parse(replace(attrs(g)["value_type"], "ClimaCore." => "")))
```
simonbyrne, line 321: can we use an `if-else` branch instead of `eval`?

src/IO/writers.jl
sriharshakandala: `faces` can be range. We we like it to be written as an `Array` anyways?

src/IO/readers.jl
sriharshakandala: `eval` statement has been replaced for `DataLayout`. `ElType` is much more complex to predict!

src/IO/writers.jl
sriharshakandala: Done. `Range` is used to construct a uniform mesh.

src/IO/readers.jl
sriharshakandala: A test is added for stretched mesh. However, the last element of the stretched mesh seems to be a NaN, causing problems with the test.

sriharshakandala: Bug fix added, resolved.

src/Meshes/interval.jl
```diff
@@ -123,7 +123,7 @@ function IntervalMesh(
     cmax = Geometry.component(domain.coord_max, 1)
     R = cmax - cmin
     h = stretch.H / R
-    η(ζ) = -h * log1p((expm1(-1 / h)) * ζ)
+    η(ζ) = ζ == 1 ? ζ : -h * log1p((expm1(-1 / h)) * ζ)
```
charleskawczynski, line 126: Can these changes, if intentional, be moved to a separate PR?

sriharshakandala: This changes were add by @simonbyrne , to fix a `NaN` issue in the ExponentialStretchedGrid code!

test/InputOutput/hybrid2dbox_stretched.jl
```diff
@@ -0,0 +1,107 @@
...
+function init_dry_rising_bubble_2d(x, z)
+    x_c = 0.0
+    z_c = 350.0
+    r_c = 250.0
+    θ_b = 300.0
+    θ_c = 0.5
+    cp_d = C_p
+    cv_d = C_v
+    p_0 = MSLP
+    g = grav
```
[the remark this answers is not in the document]
sriharshakandala: Removed global variables.

simonbyrne:
Maybe we need a mechanism to close the writers and readers?

simonbyrne:
e.g. define `Base.close` for each, which will close the underlying HDF5 handles

sriharshakandala:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

.buildkite/pipeline.yml
```diff
@@ -25,6 +25,7 @@ steps:
       - "julia --project=perf -e 'using Pkg; Pkg.precompile(;strict=true)'"
 
       - echo "--- Instantiate examples"
+      - "julia --project=perf -e 'using Pkg; Pkg.develop(path=\".\")'"
```
charleskawczynski, line 28: Please remove this, our perf environment has the manifest checked in.

sriharshakandala:
bors try

charleskawczynski:
I think we just need to

```
julia --project=docs/
] add HDF5
```
?

charleskawczynski:
Also, can we rebase/squash?

charleskawczynski (approved):

docs/Manifest.toml
```diff
@@ -200,7 +200,7 @@ uuid = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 version = "0.3.0"
 
 [[deps.ClimaCore]]
-deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "Plots", "RecursiveArrayTools", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
+deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "RecursiveArrayTools", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
```
charleskawczynski, line 203: It looks like Plots was removed here?

examples/Manifest.toml
```diff
@@ -160,7 +160,7 @@ uuid = "5f86816e-8b66-43b2-912e-75384f99de49"
 version = "0.3.2"
 
 [[deps.ClimaCore]]
-deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "Plots", "RecursiveArrayTools", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
+deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "RecursiveArrayTools", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
```
charleskawczynski, line 163: It looks like Plots was removed here?

charleskawczynski: Maybe this is the issue?

charleskawczynski:
It looks like there's nothing wrong in the main branch. Something wrong is happening with this PR. Looking at the log, this seems suspect:

```julia
┌ Warning: Replacing module ClimaCore
```

in the `Column Heat Diffusion Eq` job. It looks like there's still use of `eval`, which I would recommend looking into first. I'm still looking into this, but just wanted to point this out in case it's helpful to others in finding out what's wrong.

I also recommend insuring that `Plots` remains part of the `examples` and `docs` Manifest, this PR should probably just add `HDF5`

src/InputOutput/readers.jl
```diff
@@ -0,0 +1,419 @@
...
+function read_new!(reader::HDF5Reader, name::AbstractString, ::Type{Field})
+    (; file, cache) = reader
+    @assert name ∈ keys(file["fields"])
+    g = file["fields/" * name]
+    data = read_dataset(g, "data")
+    data_layout = attrs(g)["data_layout"]
+    Nij = size(data, findfirst("I", data_layout)[1])
+    DataLayout = _scan_data_layout(data_layout)
+    ElType =
+        eval(Meta.parse(replace(attrs(g)["value_type"], "ClimaCore." => "")))
```
charleskawczynski, line 378: Can we avoid `eval` here?

test/InputOutput/hybrid3dbox.jl
```diff
@@ -0,0 +1,124 @@
+using Test
+using StaticArrays, IntervalSets, LinearAlgebra, UnPack
+
+import ClimaCore:
+    ClimaCore,
```
charleskawczynski, line 5:
```suggestion
import ClimaCore
import ClimaCore:
```
I don't think that this should matter, but I've not really used this syntax before

charleskawczynski: Also, there are a couple of these

sriharshakandala:
bors try

sriharshakandala:
bors try

PR #764, opened 2022-06-21
Update def_coord_space for TerrainWarpedIntervalTopology remapped output

akshaysridhar:
Update `def_space_coord` to handle TerrainWarpedIntervalTopology. Existing tests remain unaffected 

### DIFF
--- a/lib/ClimaCoreTempestRemap/src/netcdf.jl
+++ b/lib/ClimaCoreTempestRemap/src/netcdf.jl
@@ -228,10 +228,12 @@ function def_space_coord(
     type = "dgll",
 ) where {S <: Spaces.Staggering}
     hvar = def_space_coord(nc, space.horizontal_space; type = type)
-    vvar = def_space_coord(
-        nc,
-        Spaces.FiniteDifferenceSpace{S}(space.vertical_topology),
-    )
+    vertical_topology =
+        space.vertical_topology isa
+        ClimaCore.Hypsography.TerrainWarpedIntervalTopology ?
+        space.vertical_topology.topology : space.vertical_topology
+    vvar =
+        def_space_coord(nc, Spaces.FiniteDifferenceSpace{S}(vertical_topology))
     (hvar..., vvar...)
 end
 
--- a/lib/ClimaCoreTempestRemap/test/netcdf.jl
+++ b/lib/ClimaCoreTempestRemap/test/netcdf.jl
@@ -1,5 +1,6 @@
 import ClimaCore
-using ClimaCore: Geometry, Meshes, Domains, Topologies, Spaces, Fields
+using ClimaCore:
+    Geometry, Meshes, Domains, Topologies, Spaces, Fields, Hypsography
 using NCDatasets
 using TempestRemap_jll
 using Test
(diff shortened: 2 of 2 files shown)


akshaysridhar:
bors try

akshaysridhar:
@charleskawczynski  Could you please take a quick look at this PR ?

lib/ClimaCoreTempestRemap/src/netcdf.jl
```diff
@@ -228,10 +228,19 @@ function def_space_coord(
     type = "dgll",
 ) where {S <: Spaces.Staggering}
     hvar = def_space_coord(nc, space.horizontal_space; type = type)
-    vvar = def_space_coord(
-        nc,
-        Spaces.FiniteDifferenceSpace{S}(space.vertical_topology),
-    )
+    if space.vertical_topology isa Topologies.IntervalTopology
```
sriharshakandala, line 231: vertical_topology = space.vertical_topology isa
           ClimaCore.Hypsography.TerrainWarpedIntervalTopology ? space.vertical_topology.topology : space.vertical_topology
    vvar = def_space_coord(
        nc,
        Spaces.FiniteDifferenceSpace{S}(vertical_topology),
    )

sriharshakandala: In general, can we just use `space.vertical_topology`. Switching between `space.vertical_topology` and `space.vertical_topology.topology` does not seem to be intuitive

LenkaNovak:
@akshaysridhar , looks good, but I'm seeing that although it passes the integrations tests above, when I try to include `runtests.jl` locally, the last test breaks on line [312](https://github.com/CliMA/ClimaCore.jl/blob/9220828a513942784f66f189d6a4b88a8b3ccf4a/lib/ClimaCoreTempestRemap/test/netcdf.jl#L312) with error: `incompatible horizontal dimension already exists`. If you run the whole test suite locally, does it behave?

lib/ClimaCoreTempestRemap/test/netcdf.jl
```diff
@@ -153,6 +154,118 @@ end
...
+    ϕₘ = FT(0) # degrees (equator)
+    λₘ = FT(3 / 2 * 180)  # degrees
+    rₘ = @. FT(acos(sind(ϕₘ) * sind(ϕ) + cosd(ϕₘ) * cosd(ϕ) * cosd(λ - λₘ))) # Great circle distance (rads)
+    Rₘ = FT(3π / 4) # Moutain radius
+    ζₘ = FT(π / 16) # Mountain oscillation half-width
+    h₀ = FT(10)
+    if rₘ < Rₘ
+        zₛ = FT(h₀ / 2) * (1 + cospi(rₘ / Rₘ))
+    else
+        zₛ = @. FT(0)
```
charleskawczynski, line 169: Does this ever get executed? Looks like it should maybe be `zₛ = FT(0)`?

charleskawczynski (approved):
Code changes look fine to me.

lib/ClimaCoreTempestRemap/test/netcdf.jl
LenkaNovak: Both are used (which one depends on the coords).

LenkaNovak (approved):
If the local test issue is not a problem then let's merge. Visualizations look correct too. Also, in case you wanted a quantitative estimate of the error due to the regridding, you could apply the `apply_remap` twice and reconstruct your original field. Not necessary here, but thought it was worth mentioning.

simonbyrne:
Note that it still won't handle topography correctly according to the CF conventions, but we can fix that later.

PR #779, opened 2022-06-28
Remove unused type parameters

charleskawczynski:
This PR removes some unused type parameters. This turned out to be a performance bug in RootSolvers.jl, which was fixed in [RS#36](https://github.com/CliMA/RootSolvers.jl/pull/36).

Hopefully this will reduce the allocations in the dss operations.


charleskawczynski:
Also, opened [JuliaLang#45846](https://github.com/JuliaLang/julia/issues/45846).

PR #781, opened 2022-06-29
Enforce tuple with uniform eltype

charleskawczynski:
This will enforce that we use a uniform `eltype` in `StencilCoefs`s.


dennisYatunin (approved):
Was this causing a performance hit because `!isconcretetype(eltype(coefs))` was not a constant at compile time?

charleskawczynski:
> Was this causing a performance hit because `!isconcretetype(eltype(coefs))` was not a constant at compile time?

I'm not 100% sure, but using this restriction will at least eliminate the possibility of different eltypes, which we've seen is important for compiler optimizations in the past. I think that this s a safe move if it looks good to you, @dennisYatunin.

PR #782, opened 2022-06-29
Reuse stencil operator in Op2Stencil

charleskawczynski:
This is a tracer bullet PR (we can spread this pattern to other operators if reviewers think this is a good idea) for refactoring the implicit operator stencils by

 - Splitting the stencil function into a "parts" function and one that applies the reduce over the parts
 - Makes the implicit operator call the "parts" function


dennisYatunin (approved):
I think this is a great idea! If all the operators in `finitedifference.jl` are rewritten like this, then we can completely get rid of the code duplication in `operator2stencil.jl`.

charleskawczynski:
A lot of this duplication has been deleted.

PR #785, opened 2022-07-01
Add and fix Aqua tests

charleskawczynski:
This PR adds [Aqua](https://github.com/JuliaTesting/Aqua.jl) tests, specifically, `Aqua.test_unbound_args(ClimaCore)`, which tests that no functions defined in ClimaCore run into [this julia issue](https://github.com/JuliaLang/julia/issues/29393).

Adding this test revealed 9 instances of this issue, and the PR also includes fixes so that we pass all of the tests.

### DIFF
--- a/src/DataLayouts/broadcast.jl
+++ b/src/DataLayouts/broadcast.jl
@@ -261,7 +261,7 @@ end
 function Base.similar(
     ::Union{IF{<:Any, Ni, A}, Broadcast.Broadcasted{IFStyle{Ni, A}}},
     ::Type{Eltype},
-) where {S, Ni, A, Eltype}
+) where {Ni, A, Eltype}
     Nf = typesize(eltype(A), Eltype)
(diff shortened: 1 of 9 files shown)


charleskawczynski:
It may be that we want to include type parameters in `copyto!`, in which case, the only solution I see is to split the method based on the unions of the input arguments to ensure that we don't run into this problem.

charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
I think this is ready to go, cc @simonbyrne. Going to squash now.

charleskawczynski:
All these changes look safe to me. Merging. cc @simonbyrne.

simonbyrne:
Thanks!

PR #787, opened 2022-07-02
Add support for truncated field type printing

charleskawczynski:
This PR adds support for truncated field type printing. Users can set this (dynamically if desired) with:

```julia
import ClimaCore
ClimaCore.Fields.truncate_printing_field_types() = true
```

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -393,5 +393,64 @@ function set!(f::Function, field::Field, args = ())
     return nothing
 end
 
+#=
+This function can be used to truncate the printing
+of ClimaCore `Field` types, which can get rather
+long.
+
+# Example
+```
+import ClimaCore
+ClimaCore.Fields.truncate_printing_field_types() = true
+```
+=#
+truncate_printing_field_types() = false
+
+function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
+    if truncate_printing_field_types()
+        print(io, truncated_field_type_string(T))
+    else
+        invoke(show, Tuple{IO, Type}, io, T)
+    end
+end
+
+# Defined for testing
+function truncated_field_type_string(::Type{T}) where {T <: Fields.Field}
+    values_type(::Type{T}) where {V, T <: Fields.Field{V}} = V
+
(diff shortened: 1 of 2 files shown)


simonbyrne:
Can you give some examples of how this prints?

If it would be easier, I'd be happy to change it so that it prints just as `Field{...}`?

charleskawczynski:
> Can you give some examples of how this prints?

Yeah, that's a good idea

> If it would be easier, I'd be happy to change it so that it prints just as `Field{...}`?

That would indeed be much, much, easier.

charleskawczynski:
I've updated the PR by moving this into ClimaCore src and it's now conditionally loaded 😬, plus, I've added a test. cc @simonbyrne

src/Fields/Fields.jl
```diff
@@ -393,5 +393,64 @@ function set!(f::Function, field::Field, args = ())
...
+ClimaCore.Fields.truncate_printing_field_types() = true
+```
+=#
+truncate_printing_field_types() = false
+
+function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
+    if truncate_printing_field_types()
+        print(io, print_truncated_field_type(T))
+    else
+        invoke(show, io, print_truncated_field_type(T))
```
simonbyrne, line 413: shouldn't this be
```suggestion
        invoke(show, Tuple{IO, Type}, io, T)
```

src/Fields/Fields.jl
```diff
@@ -393,5 +393,64 @@ function set!(f::Function, field::Field, args = ())
...
+    end
+end
+
+# Defined for testing
+function print_truncated_field_type(::Type{T}) where {T <: Fields.Field}
```
simonbyrne, line 418: don't call this `print_truncated_field_type` if it doesn't print anything

src/Fields/Fields.jl
charleskawczynski: I've added a test to make sure the standard printing isn't broken, too.

src/Fields/Fields.jl
charleskawczynski: Changed to `truncated_field_type_string`.

PR #791, opened 2022-07-05
Attempt to optimize implicit stencils

charleskawczynski:
This is a "future variant" of #790.

This is the best I've done so far on cracking down on inference. It's still failing JET tests, which are, at this point, not very informative. And now, this is breaking Cthulhu. Using this with

```julia
using Test

import ClimaCore
ClimaCore.Operators.allow_mismatched_fd_spaces() = true

using ClimaCore: Geometry, Domains, Meshes, Topologies, Spaces, Fields, Operators

import ClimaCore.Utilities: half
import LinearAlgebra: norm_sqr

const FT = Float32
radius = FT(1e7)
zmax = FT(1e4)
helem = npoly = 2
velem = 4

hdomain = Domains.SphereDomain(radius)
hmesh = Meshes.EquiangularCubedSphere(hdomain, helem)
htopology = Topologies.Topology2D(hmesh)
quad = Spaces.Quadratures.GLL{npoly + 1}()
hspace = Spaces.SpectralElementSpace2D(htopology, quad)

vdomain = Domains.IntervalDomain(
    Geometry.ZPoint{FT}(zero(FT)),
    Geometry.ZPoint{FT}(zmax);
    boundary_tags = (:bottom, :top),
)
vmesh = Meshes.IntervalMesh(vdomain, nelems = velem)
center_space = Spaces.CenterFiniteDifferenceSpace(vmesh)

#=
# TODO: Replace this with a space that includes topography.
center_space = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace)
center_coords = Fields.coordinate_field(center_space)
face_space = Spaces.FaceExtrudedFiniteDifferenceSpace(center_space)
=#
face_space = Spaces.FaceFiniteDifferenceSpace(center_space)

ᶜρe = ones(center_space)
ᶜρ = ones(center_space)
ᶜp = ones(center_space)
ᶠw = Geometry.Covariant3Vector.(ones(face_space))
BDT = Operators.StencilCoefs{-half, half, NTuple{2, FT}}
QDT = Operators.StencilCoefs{-(1 + half), 1 + half, NTuple{4, FT}}

# this is just for testing the optimizations, so we can just make everything 1s
∂ᶜK∂ᶠw_data = similar(ᶜρ, BDT)
∂ᶜ𝔼ₜ∂ᶠ𝕄 = similar(ᶜρ, QDT)


# Allow one() to be called on vectors.
Base.one(::T) where {T <: Geometry.AxisTensor} = one(T)
[...]

PR #783, opened 2022-06-29
add JET test case for Wfact dispatch

simonbyrne:
This is a reproducible test case for the JET failures in ClimaAtmos

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #792, opened 2022-07-05
Attempt to optimize implicit stencils

charleskawczynski:
Here's an alternative to #791.

On the one hand, this may not require creating an intermediate tuple, however, Julia docs say

> Evaluates to the same as reduce(op, map(f, A...); dims=dims, init=init), but is generally faster because the intermediate array is avoided.

### DIFF
--- a/perf/Manifest.toml
+++ b/perf/Manifest.toml
@@ -1,18 +1,72 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.2"
+julia_version = "1.7.3"
 manifest_format = "2.0"
+
+[[deps.AbstractFFTs]]
+deps = ["ChainRulesCore", "LinearAlgebra"]
+git-tree-sha1 = "8e9c3482c61d06343a6199814bf84f7df82f2b28"
+uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+version = "1.2.0"
 
 [[deps.AbstractTrees]]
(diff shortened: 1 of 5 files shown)


charleskawczynski:
Update:

At commit [5c50027](https://github.com/CliMA/ClimaCore.jl/pull/792/commits/5c50027b1cd0f0b72a9bf63f31f1a9a6696481d9), Cthulhu says that all types are inferred through the recursion of `getidx`, however, `JET` still says that the compiler fails to optimize the code. So, we need to devise a plan to circumvent this. Half baked thoughts of things that come to mind are:

 - refactor `getindex` to somehow return multiple `getidx`s in a single pass (if that even makes sense, due to recursion complexity)
 - Write a custom `getidx` for stencil coefficients (and dispatch to a custom `apply_stencil!` for composed stencils).
 - (if possible) Construct stencils ahead of time, store them in the cache, and only use them for mat-vec products.
 - Brainstorm other ideas?

I've probably not thought about this enough, and I need to review what exactly is needed to construct the stencil coefficients.

The good news is that the compiler can propagate type information all the way through with these changes, and I'm hoping that applying similar changes to the apply stencils will result in optimized code.

charleskawczynski:
Update update: it's working and passing the JET tests! The trick was to avoid `reduce` and just use loops! 🎉 Supersedes #783.

charleskawczynski:
Superseded by #797

PR #795, opened 2022-07-08
Default solving grid stretching w Float64

charleskawczynski:
This PR adjusts the `GeneralizedExponentialStretching` by adding `tol` as a kwarg and allowing users to specify the float type when solving for the grid stretch parameters. This PR changes the default to `Float64`, to avoid accumulated precision errors with Float32, then type casts to `FT` on the resulting faces.

This is technically behavior changing, we could change `FT_solve = Float64` to `FT_solve = FT` to make it non-behavior changing, however, I think this is probably what users would want / expect from this function.

### DIFF
--- a/src/Meshes/interval.jl
+++ b/src/Meshes/interval.jl
@@ -191,13 +191,16 @@ function IntervalMesh(
     domain::IntervalDomain{CT},
     stretch::GeneralizedExponentialStretching{FT};
     nelems::Int,
+    FT_solve = Float64,
+    tol = 1e-3,
     reverse_mode::Bool = false,
 ) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
     if nelems ≤ 1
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Looks good to me, thanks! I don’t know why the unit tests fail though.

charleskawczynski:
> Looks good to me, thanks! I don’t know why the unit tests fail though.

Yeah, I'm not sure it's related, though. Those same tests are failing in other PRs (e.g., #793). It very well may be from a dependency update (we don't check in the manifest for the test environment).

charleskawczynski:
I'll let that get sorted out before merging.

charleskawczynski:
bors try

charleskawczynski:
I just revived this PR, cc @szy21!

szy21 (approved):
Thanks Charlie!

charleskawczynski:
bors try

PR #800, opened 2022-07-13
Bump patch version for new release

charleskawczynski:


charleskawczynski:
Plots updates are broken... @simonbyrne, alright if we just pin Plots.jl? Or should I just not update the deps in this PR? We saw some really weird changes when TC updated (see [Plots#4273](https://github.com/JuliaPlots/Plots.jl/issues/4273)).

PR #804, opened 2022-07-16
Downgrade plots and pin version

charleskawczynski:
This PR downgrades Plots.jl to `1.30.1` and pins this version, which seems to have fewer issues. This is a followup to #800, which only set the compat entry. For this PR I've actually pinned it with `Pkg`:

```julia
import Pkg
Pkg.add(Pkg.PackageSpec(;name="Plots", version = "1.30.1"))
Pkg.pin("Plots")
```

### DIFF
--- a/.dev/up_deps.jl
+++ b/.dev/up_deps.jl
@@ -18,6 +18,7 @@ dirs = (
 
 cd(root) do
     for dir in dirs
+        @info "Updating environment `$dir`"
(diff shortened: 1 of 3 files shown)


charleskawczynski:
bors try

valeriabarra (approved):

valeriabarra:
Thanks!

PR #805, opened 2022-07-16
Try correction to exponential stretching

charleskawczynski:
This is an attempt to peel off a piece from #762.


charleskawczynski:
bors try

PR #807, opened 2022-07-18
2D bubble: Use ExponentialStretching

### LINKED ISSUE #806: Try ExponentialStretching in the rising thermal bubble example
This is a follow-up to Issue #803 

In the incremental investigation of stability of the dry HS test case, we should first try an explicit example with stretched grid and 3rd-order-upwinding in the vertical, before trying an implicit configuration (which would be more similar to the desired one in the HS case).

*Describe the solution you'd like* 
Add exponential stretching to the 2D rising bubble test case

valeriabarra:
This will close #806 

I tried `ExponentialStretching` in this example and it works fine (I had to reduce the `dt` of a factor of 4, but it works). 


valeriabarra:
bors try

PR #811, opened 2022-07-18
Extend ClimaCore Remap Interface for TempestRemap

jb-mackay:
Extends the `remap` and `remap!` methods of ClimaCore for the TempestRemap module.

Builds off #793 by introducing `LinearTempestRemap` subtype of `AbstractRemap`.


jb-mackay:
@simonbyrne let me know if you think it makes sense to keep this as a separate module or if we want to move this code to `src/Operators`.

LenkaNovak (approved):
This looks good to me. I would retain it in the `lib` directory, to keep wrappers related to third-party packages separate. Eventually we can add another `remap` that will dispatch on a ClimaCore-based conservative interpolate operator, once we get round to developing it. @simonbyrne would you agree?

PR #829, opened 2022-07-25
Improve performance of column operators

charleskawczynski:
This PR improves the performance of column operations by:
 - Adding `@inline` and `@inbounds` annotations
 - Fixing some type inference issues that prevented inlining (in `idxin`)

### DIFF
--- a/src/DataLayouts/struct.jl
+++ b/src/DataLayouts/struct.jl
@@ -119,10 +119,13 @@ promote_parent_array_type(
 Create an instance of type `T` from a tuple of field values `args`, bypassing
 possible internal constructors. `T` should be a concrete type.
 """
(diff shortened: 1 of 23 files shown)


charleskawczynski:
bors try

src/DataLayouts/struct.jl
```diff
@@ -261,5 +257,8 @@ if VERSION >= v"1.7.0-beta1"
         for m in methods(set_struct!)
             m.recursion_relation = dont_limit
         end
+        for m in methods(bypass_constructor)
+            m.recursion_relation = dont_limit
+        end
```
charleskawczynski, line 262: This can be removed

charleskawczynski:
That should do it. 🤞🏻

charleskawczynski:
bors try

PR #834, opened 2022-07-26
use project instead of transform to access contravariant components

simonbyrne:
This should help avoid some unnecessary ops and checks

### DIFF
--- a/src/Geometry/conversions.jl
+++ b/src/Geometry/conversions.jl
@@ -141,21 +141,18 @@ LocalVector(u::ContravariantVector{<:Any, (3,)}, ::LocalGeometry{(1, 2)}) =
     CovariantVector(u, local_geometry).u₃
 
 @inline contravariant1(u::AxisVector, local_geometry::LocalGeometry) =
-    transform(Contravariant123Axis(), u, local_geometry).u¹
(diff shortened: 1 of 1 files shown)


charleskawczynski:
I can go ahead and rebase this to double check if this improves anything

charleskawczynski (approved):

src/Geometry/conversions.jl
```diff
@@ -141,21 +141,18 @@ LocalVector(u::ContravariantVector{<:Any, (3,)}, ::LocalGeometry{(1, 2)}) =
     CovariantVector(u, local_geometry).u₃
 
 @inline contravariant1(u::AxisVector, local_geometry::LocalGeometry) =
-    transform(Contravariant123Axis(), u, local_geometry).u¹
+    @inbounds project(Contravariant1Axis(), u, local_geometry)[1]
```
charleskawczynski, line 144: Wait, why was this changed from `Contravariant123Axis` to `Contravariant1Axis`? I assume this is equivalent? cc @simonbyrne

simonbyrne: We only care about the 1st component

PR #843, opened 2022-07-29
Try parallelizing unit tests

charleskawczynski:
This PR parallelizes CI by binning some of the more expensive tests into separate jobs.


charleskawczynski:
bors try

charleskawczynski:
Ah, I'm realizing that this wouldn't actually have a useful rebalancing workflow, and it's just a bit too complicated

PR #866, opened 2022-08-05
IGW: workaround XZ-space Covariant12Vector issue in `linear_solution!`

### LINKED ISSUE #865: IGW example with small_scale=false issue in linear_solution
## Describe the bug
The IGW example, in its large-scale configuration, i.e., with `small_scale = false` breaks in the `linear_solution!` function that is called in the post-processing routine. 

## To Reproduce
Switch `is_small_scale = false` and run 

```julia
ENV["FLOAT_TYPE"] = "Float64"; ENV["TEST_NAME"] = "plane/inertial_gravity_wave.jl"; include("hybrid/driver.jl")
```

<details><summary>Project</summary>
<p>
Using the `examples` project:

</p>
</details>

## System details

Any relevant system information:
- Julia version 1.7.2
- operating system
- modules loaded on cluster (`module list`)

valeriabarra:
This will close #865 

Co-authored-by: Dennis Yatunin <dyatun@gmail.com>

### DIFF
--- a/examples/hybrid/plane/inertial_gravity_wave.jl
+++ b/examples/hybrid/plane/inertial_gravity_wave.jl
@@ -400,6 +400,9 @@ function linear_solution!(Y, lin_cache, t)
     elseif ᶜ𝔼_name == :ρe_int
         @. Y.c.ρe_int = ᶜρ * cv_d * (ᶜT - T_tri)
     end
-    @. Y.c.uₕ = Geometry.Covariant12Vector(Geometry.UVVector(ᶜu, ᶜv))
+    # NOTE: The following two lines are a temporary workaround b/c Covariant12Vector won't accept a non-zero second component in an XZ-space.
+    # So we temporarily set it to zero and then reassign its intended non-zero value (since in case of large-scale config ᶜv is non-zero)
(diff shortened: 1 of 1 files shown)


valeriabarra:
bors try

valeriabarra:
bors try

charleskawczynski (approved):

valeriabarra:
Thanks @charleskawczynski ! 
Do we want to open a separate issue for the uncovered unexpected behavior (i.e., restrictive check on second component to be zero in the [`_transform`](https://github.com/CliMA/ClimaCore.jl/blob/868eccf89bbc3dfe62bf04220ffc0eefa4ba0707/src/Geometry/axistensors.jl#L458) or is that already considered in Issue #867 ?

charleskawczynski:
> Thanks @charleskawczynski ! Do we want to open a separate issue for the uncovered unexpected behavior (i.e., restrictive check on second component to be zero in the [`_transform`](https://github.com/CliMA/ClimaCore.jl/blob/868eccf89bbc3dfe62bf04220ffc0eefa4ba0707/src/Geometry/axistensors.jl#L458) or is that already considered in Issue #867 ?

It's kind of part of #867. I think @simonbyrne suggested changing the `transform` calls to `project`. Checking values (which is done at runtime) is not really performant. I think the better solution is to allow an option for users to turn this on, but have some way to disable it.

PR #872, opened 2022-08-08
Add simple FCT unit tests

### LINKED ISSUE #871: Add simple FCT unit tests
#856 added the simple FCT as an integration test directly. But it's better to add unit tests as well.

Add dedicated unit tests, by replicating the existing third-order upwinding operator unit tests and verify that indeed we fall back on the existing behavior.

valeriabarra:
This will close #871 

### DIFF
--- a/test/Operators/finitedifference/column.jl
+++ b/test/Operators/finitedifference/column.jl
@@ -637,11 +637,11 @@ end
         # c = sin(z), scalar field defined at the centers
         c = sin.(centers)
 
-        fluxᶠ = Operators.Upwind3rdOrderBiasedProductC2F()
-        fluxsinᶠ = fluxᶠ.(w, c)
+        third_order_fluxᶠ = Operators.Upwind3rdOrderBiasedProductC2F()
+        third_order_fluxsinᶠ = third_order_fluxᶠ.(w, c)
 
         divf2c = Operators.DivergenceF2C()
-        adv_wc = divf2c.(fluxsinᶠ)
+        adv_wc = divf2c.(third_order_fluxsinᶠ)
 
         Δh[k] = cs.face_local_geometry.J[1]
 
@@ -689,7 +689,7 @@ end
             c = (cos.(centers .- Δz / 2) .- cos.(centers .+ Δz / 2)) ./ Δz
             s = sin.(centers)
 
-            fluxᶠ = Operators.Upwind3rdOrderBiasedProductC2F(
+            third_order_fluxᶠ = Operators.Upwind3rdOrderBiasedProductC2F(
                 bottom = Operators.FirstOrderOneSided(),
                 top = Operators.FirstOrderOneSided(),
             )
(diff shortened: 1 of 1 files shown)


test/Operators/finitedifference/column.jl
```diff
@@ -775,6 +775,196 @@ end
...
+    @test conv_adv_wc[3] ≈ 3 atol = 0.1
+    @test conv_adv_wc[1] ≤ conv_adv_wc[2] ≤ conv_adv_wc[2]
+end
+
+@testset "Simple FCT: lin combination of UpwindBiasedProductC2F + Upwind3rdOrderBiasedProductC2F on (uniform and stretched) non-periodic mesh, with FirstOrderOneSided BCs" begin
+    FT = Float64
+    n_elems_seq = 2 .^ (4, 6, 8, 10)
+    stretch_fns = (Meshes.Uniform(), Meshes.ExponentialStretching(1.0))
+
+    for (i, stretch_fn) in enumerate(stretch_fns)
```
charleskawczynski, line 839: It's probably easier to write this for only the exponential case because the convergence orders won't match up in general

test/Operators/finitedifference/column.jl
```diff
@@ -775,6 +775,196 @@ end
...
+                top = Operators.SetValue(
+                    Geometry.Contravariant3Vector(FT(0.0)),
+                ),
+            )
+
+            corrected_antidiff_flux = @. divf2c(
+                C * (third_order_fluxᶠ(w, c) - first_order_fluxᶠ(w, c)),
+            )
+            adv_wc =
+                @. @. divf2c.(first_order_fluxᶠ(w, c)) + corrected_antidiff_flux
```
charleskawczynski, line 886:
```suggestion
            adv_wc =
                @. divf2c(first_order_fluxᶠ(w, c)) + corrected_antidiff_flux
```

valeriabarra: Good eye! Sorry that slipped by accident

test/Operators/finitedifference/column.jl
```diff
@@ -775,6 +775,196 @@ end
     end
 end
 
+
+@testset "Simple FCT: lin combination of UpwindBiasedProductC2F + Upwind3rdOrderBiasedProductC2F on (uniform) periodic mesh" begin
```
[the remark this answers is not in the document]
valeriabarra: I think these are already covered in the integration tests (some of which use quantitative tests to check against expected behavior). However, it would be good to improve some of our simple unit tests as well. But I don't think this PR is the right venue for that. We should discuss perhaps in a separate issue how to improve our unit tests. These are not the only one which need some improvement and/or corner cases covered.

test/Operators/finitedifference/column.jl
valeriabarra: It was very close and some cases also better actually. I don't think it would be the right venue here to split them up and reformulate these tests

valeriabarra:
bors try

charleskawczynski (approved):

valeriabarra:
@charleskawczynski , I am not familiar with the error in the CI job:
```
🚨 Error: The command was interrupted by a signal
```
Do we just punch it again?

valeriabarra:
Might be related to time-out. Should we increase it?

PR #875, opened 2022-08-08
Upwind3rdOrderBiasedProductC2F and simple FCT: Add varying sign veloc…

### LINKED ISSUE #873: Add unit test to test 3rd-order-upwinding with varying sign velocity
Currently, our unit test for the `Upwind3rdOrderBiasedProdctC2F` uses a constant and always positive velocity. 

We could add a test with a varying sign velocity (e.g., a cosine profile) and test this, at least in the simple, periodic mesh case.

valeriabarra:
…ity unit tests

This was requested in the review of #872 

This will close #873 
I think it's expected that we see some degraded convergence rate. I think I remember Tapio mentioned this when this operator was discussed a while ago.

I added unit tests with varying sign velocity for the periodic mesh case and non-periodic mesh cases with `ThirdOrderOneSided` BC. The other one (`FirstOrderOneSided`) is already not accurate enough and also Tapio suggested we should favor using the `ThirdOrderOneSided` for this reason.

### DIFF
--- a/test/Operators/finitedifference/column.jl
+++ b/test/Operators/finitedifference/column.jl
@@ -610,7 +610,7 @@ end
     @test conv_curl_sin_f[1] ≤ conv_curl_sin_f[2] ≤ conv_curl_sin_f[3]
 end
 
-@testset "Upwind3rdOrderBiasedProductC2F + DivergenceF2C on (uniform) periodic mesh" begin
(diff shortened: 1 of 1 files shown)


valeriabarra:
bors try

valeriabarra:
Exact solution
exact
Vs computed
adv
for the stretched, non-periodic mesh case with `ThirdOrderOneSided`

Looking good

charleskawczynski (approved):
Thanks for adding this! LGTM

PR #880, opened 2022-08-10
Fix broken JET tests, with latest version, on Julia 1.8

charleskawczynski:
This PR applies similar code simplification to #797 in order to fix broken JET tests, with the latest version, on Julia 1.8.


charleskawczynski:
Absorbed into #912.

Issue #877, opened 2022-08-10
Missing support for similar / zeros for point spaces


charleskawczynski:
MWE:

```julia
import ClimaCore: Domains,Geometry,Meshes,Spaces,Fields,Utilities

FT = Float64
column = Domains.IntervalDomain(
    Geometry.ZPoint{FT}(-10.0),
    Geometry.ZPoint{FT}(0.0);
    boundary_tags = (:top, :bottom)
);
mesh = Meshes.IntervalMesh(column; nelems = 10);
center_space = Spaces.CenterFiniteDifferenceSpace(mesh);
face_space = Spaces.FaceFiniteDifferenceSpace(center_space);
surface_space = Fields.level(
    face_space,
    Utilities.PlusHalf(Spaces.nlevels(face_space)-1)
);

Fields.zeros(surface_space) # breaks
Fields.Field(FT, surface_space) # breaks
```

Ported from #869.

charleskawczynski:
I'm actually wondering if #863 will fix this, or if there's any overlap

kmdeck:
A note - the functions do work with a PointSpace when that space is created from scratch (not via `level`):
```
julia> ClimaCore.Fields.zeros(ClimaCore.Spaces.PointSpace(ClimaCore.Geometry.ZPoint{Float64}(0.0)))
Float64-valued Field:
  [0.0]
```

kmdeck:
Draft PR
https://github.com/CliMA/ClimaCore.jl/pull/882

PR #882, opened 2022-08-11
Removed reshaped arrays when making a DataF DataLayout

kmdeck:
Possible way to address Issue https://github.com/CliMA/ClimaCore.jl/issues/877

The root cause of that was that level and column, when applied to data layouts where a point `DataF` layout was returned, returned something with a `ReshapedArray` subtype. `parent_array_type` was not defined for this, so this was a bug. This addresses that issue by making the returned `DataF` have subtype `SubArray`, for which `parent_array_type` is defined.

Another option would be to define `parent_array_type` for `ReshapedArray`s.

Open to suggestions on how to improve the tests :)
- [X] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [X] Unit tests are included OR N/A.
- [X] Code is exercised in an integration test OR N/A.
- [X] Documentation has been added/updated OR N/A.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -328,7 +328,7 @@ end
 @inline function column(data::IJFH{S, Nij}, i, j, h) where {S, Nij}
     @boundscheck (1 <= j <= Nij && 1 <= i <= Nij && 1 <= h <= length(data)) ||
                  throw(BoundsError(data, (i, j, h)))
-    dataview = @inbounds view(parent(data), [i], j, :, h)
+    dataview = @inbounds view(parent(data), i:i, j, :, h)
     DataF{S}(dataview)
 end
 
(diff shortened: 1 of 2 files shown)


kmdeck:
I'm not sure why this isnt passing. On my own computer:
```
spaces = all_spaces(Float64)
space = spaces[2]
@info space
┌ Info: CenterFiniteDifferenceSpace:
└   IntervalTopology on 1-element IntervalMesh of IntervalDomain(ZPoint(0.0), ZPoint(5.0); boundary_names = (:bottom, :top))
data = space.center_local_geometry
v = 1
```

If I call:
```
using ClimaCore.DataLayouts:level
level(data, v) # has a reshaped array
ClimaCore.Fields.zeros(ClimaCore.Spaces.PointSpace(level(data,1))) # does not work
```


If I instead just copy paste the lines of ClimaCore.DataLayouts.level:
```
Nv = size(data, 4)
@boundscheck (1 <= v <= Nv) || throw(BoundsError(data, (v)))
array = parent(data)
dataview = @inbounds view(array, [v], :)
data_f = DataF{eltype(data)}(dataview) # subarray
ClimaCore.Fields.zeros(ClimaCore.Spaces.PointSpace(data_f)) # runs
```

kmdeck:
Is it possible that the test environment is using the registered ClimaCore, and not the version with the new changes?

kmdeck:
The tests fail now because 
```
julia> lg_field_space === lg_space
false
```
in https://github.com/CliMA/ClimaCore.jl/blob/1c425bb98d28dea8aa63add14232cdeb1f4e3e1c/test/Fields/field.jl#L335

So something about changing ClimaCore.DataLayouts.level has affected this? Before it was passing.

charleskawczynski:
bors try

charleskawczynski (approved):

charleskawczynski:
Hmm, looks like the unit tests timed out the first go around.

simonbyrne:
It might make more sense to just make the underlying type a vector than a matrix. let me see how that goes.

charleskawczynski:
Superseded by #892.

PR #883, opened 2022-08-12
Implement FCTBorisBook as a new operator

valeriabarra:
This adds the Boris & Book FCT as a new ClimaCore operator, and adds a unit/integration test and doc reference.

WIP: Needs debugging!

### DIFF
--- a/examples/column/bb_fct_advection.jl
+++ b/examples/column/bb_fct_advection.jl
@@ -1,6 +1,8 @@
 using Test
 using LinearAlgebra
-using OrdinaryDiffEq: ODEProblem, solve, SSPRK33
+using OrdinaryDiffEq: ODEProblem, solve
+using DiffEqBase
+using ClimaTimeSteppers
 
 import ClimaCore:
     Fields,
@@ -22,7 +24,7 @@ import ClimaCore:
 ENV["GKSwstype"] = "nul"
 using ClimaCorePlots, Plots
 Plots.GRBackend()
-dir = "fct_advection"
+dir = "bb_fct_advection"
 path = joinpath(@__DIR__, "output", dir)
 mkpath(path)
 
@@ -35,10 +37,11 @@ function linkfig(figpath, alt = "")
     end
 end
 
-function f!(dydt, y, parameters, t)
+function f!(dydt, y, parameters, t, alpha, beta)
 
-    (; w, C, corrected_antidiff_flux) = parameters
-    FT = Spaces.undertype(axes(y))
+    (; w, A, y_td) = parameters
+    y = y.y
+    dydt = dydt.y
 
     first_order_fluxᶠ = Operators.UpwindBiasedProductC2F(
         bottom = Operators.Extrapolate(),
@@ -52,24 +55,32 @@ function f!(dydt, y, parameters, t)
         bottom = Operators.SetValue(Geometry.WVector(FT(0.0))),
         top = Operators.SetValue(Geometry.WVector(FT(0.0))),
     )
+    FCTBB = Operators.FCTBorisBook(
+        bottom = Operators.ThirdOrderOneSided(),
+        top = Operators.ThirdOrderOneSided(),
+    )
 
-    @. corrected_antidiff_flux =
-        divf2c(C * (third_order_fluxᶠ(w, y) - first_order_fluxᶠ(w, y)))
-    @. dydt = -(divf2c(first_order_fluxᶠ(w, y)) + corrected_antidiff_flux)
+    @. y_td = beta * dydt - alpha * divf2c(first_order_fluxᶠ(w, y))
+    @. dydt =
+        y_td -
+        alpha * divf2c(
+            FCTBB(
+                third_order_fluxᶠ(w, y) - first_order_fluxᶠ(w, y),
+                y_td * alpha,
+            ),
+        )
 
     return dydt
 end
 
 FT = Float64
-t₀ = FT(0)
-t₁ = FT(1)
-z₀ = FT(0)
-zₕ = FT(1)
-z₁ = FT(1)
+t₀ = FT(0.0)
+t₁ = FT(1.0)
+z₀ = FT(0.0)
+zₕ = FT(1.0)
+z₁ = FT(1.0)
 speed = FT(1.0)
-C = FT(0.5) # flux-correction coefficient: ∈ [0,1] with
-#                              0 = first-order upwinding
-#                              1 = third-order upwinding
+
 n = 2 .^ 6
 
 domain = Domains.IntervalDomain(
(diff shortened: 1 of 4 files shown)


valeriabarra:
With a constant positive velocity the Boris & Book FCT eliminates the under/over-shoots observed with the simple, vanilla FCT 
exact_and_computed_advected_square_wave_BBFCT

valeriabarra:
With a varying-sign velocity (e.g., a cosine profile), it still shows a small monotone overshoot, but no undershoot
exact_and_computed_advected_square_wave_BBFCT

valeriabarra:
bors try

src/Operators/finitedifference.jl
```diff
@@ -1459,6 +1459,139 @@ end
...
+    velocity_space::Spaces.FaceExtrudedFiniteDifferenceSpace,
+    arg_space::Spaces.CenterExtrudedFiniteDifferenceSpace,
+) = velocity_space
+
+@inline function fct_boris_book(v, a⁻, a⁺, a⁺⁺, a⁺⁺⁺, step)
+    if v != zero(eltype(a⁻))
+        sign(v) ⊠ (RecursiveApply.rmap(
+            max,
+            zero(eltype(a⁻)),
+            RecursiveApply.rmap(
```
charleskawczynski, line 1509: I think we can improve this a bit-- I need a bit of time to think about this, but the compiler may have trouble with the very nested `rmap`s.

A more important concern is about the `v != zero(eltype(a⁻))` comparison. Let's consider how this comparison behaves:

```julia
julia> compare(num::FT) where {FT} = num ≠ FT(0)
compare (generic function with 1 method)

julia> compare(Float32(1e-7))
true

julia> compare(Float32(1e-15))
true

julia> compare(Float32(1e-60))
false

julia> compare(Float32(1e-30))
true

julia> compare(Float64(1e-30))
true

julia> compare(Float64(1e-300))
true

julia> compare(Float64(1e-600))
false
```

Is this what we want?

src/Operators/finitedifference.jl
```diff
@@ -1459,6 +1459,139 @@ end
...
+    space = axes(arg)
+    a⁻ = getidx(arg, loc, idx - half, hidx)
+    a⁺ = getidx(arg, loc, idx + half, hidx)
+    a⁺⁺ = getidx(arg, loc, idx + half + 1, hidx)
+    a⁺⁺⁺ = getidx(arg, loc, idx + half + 2, hidx)
+    vᶠ = Geometry.contravariant3(
+        getidx(velocity, loc, idx, hidx),
+        Geometry.LocalGeometry(space, idx, hidx),
+    )
+    step = Geometry.LocalGeometry(space, idx, hidx).∂x∂ξ[1]
```
charleskawczynski, line 1545: `∂x∂ξ[1]` is the only part of this function that I'm not certain about in terms of generality. We may need to dig into how to make sure that we're grabbing `dz` for all space types.

src/Operators/finitedifference.jl
valeriabarra: Interesting. I saw that this was used in several other places, e.g., the construction of a [zero field](https://github.com/CliMA/ClimaCore.jl/blob/b133c3ae65e40c993f46c57466563d8cd1f99faa/src/Fields/Fields.jl#L212) and I thought that this was indeed the type-safe way to do this

src/Operators/finitedifference.jl
valeriabarra: I thought about this and as you pointed out yesterday in our chat, for any space, `LocalGeometry(space, idx, hidx)` would return the info for all space types. And we use this only for vertical spaces (i.e., vertical center space or vertical face space). I checked that the syntax and indexing is the same in both cases

examples/column/bb_fct_advection.jl
```diff
@@ -0,0 +1,142 @@
...
+    # buildkite-agent upload figpath
+    # link figure in logs if we are running on CI
+    if get(ENV, "BUILDKITE", "") == "true"
+        artifact_url = "artifact://$figpath"
+        print("\033]1338;url='$(artifact_url)';alt='$(alt)'\a\n")
+    end
+end
+
+function f!(dydt, y, parameters, t, alpha, beta)
+
```
Zhengyu-Huang, line 41: A question for my learning, what are alpha, beta? Is beta 1, alpha dt/dx ?

src/Operators/finitedifference.jl
```diff
@@ -1459,6 +1459,139 @@ end
...
+) = velocity_space
+return_space(
+    ::FCTBorisBook,
+    velocity_space::Spaces.FaceExtrudedFiniteDifferenceSpace,
+    arg_space::Spaces.CenterExtrudedFiniteDifferenceSpace,
+) = velocity_space
+
+@inline function fct_boris_book(v, a⁻, a⁺, a⁺⁺, a⁺⁺⁺, step)
+    if v != zero(eltype(a⁻))
+        sign(v) ⊠ (RecursiveApply.rmap(
```
Zhengyu-Huang, line 1506: Does your v represent A in section 5.4.1, does your a represent phi^td ? Is step = dx/dt? You can change it to if v < zero(eltype(a⁻))

Zhengyu-Huang (commented):
Thank you for the great work, I have some questions for my understanding!

src/Operators/finitedifference.jl
```diff
@@ -1459,6 +1459,139 @@ end
...
+
+stencil_interior_width(::FCTBorisBook, velocity, arg) =
+    ((0, 0), (-half, half + 2))
+
+@inline function stencil_interior(::FCTBorisBook, loc, idx, hidx, velocity, arg)
+    space = axes(arg)
+    a⁻ = getidx(arg, loc, idx - half, hidx)
+    a⁺ = getidx(arg, loc, idx + half, hidx)
+    a⁺⁺ = getidx(arg, loc, idx + half + 1, hidx)
+    a⁺⁺⁺ = getidx(arg, loc, idx + half + 2, hidx)
```
Zhengyu-Huang, line 1540: When idx is the  first interior cell face from top,  will `idx + half + 2` go out of the range?

charleskawczynski: No, the infrastructure checks whether we are inside the boundary window or not, [here](https://github.com/CliMA/ClimaCore.jl/blob/b133c3ae65e40c993f46c57466563d8cd1f99faa/src/Operators/finitedifference.jl#L2622-L2644). That said, ensuring that we're not running into bounds errors is not a bad idea to verify. @valeriabarra, can you please add the `bounds-check=yes` to this unit test in the buildkite script? It will make the test run a bit slower, but it'll improve our correctness confidence.

examples/column/bb_fct_advection.jl
valeriabarra: Beta = 1, alpha =  dt

src/Operators/finitedifference.jl
valeriabarra: Yes, A in section 5.4.1 has the role of `v` here (defined at faces, like the velocity) and `a` is phi^td in Sec. 5.4.1. Step is only the spatial step size, so only `dz`

charleskawczynski: > You can change it to if v < zero(eltype(a⁻))

This should be the same as what is there, I've opened a [comment](https://github.com/CliMA/ClimaCore.jl/pull/883#discussion_r944626258) regarding the floating point comparison.

src/Operators/finitedifference.jl
```diff
@@ -1459,6 +1459,139 @@ end
...
+    end
+end
+
+stencil_interior_width(::FCTBorisBook, velocity, arg) =
+    ((0, 0), (-half, half + 2))
+
+@inline function stencil_interior(::FCTBorisBook, loc, idx, hidx, velocity, arg)
+    space = axes(arg)
+    a⁻ = getidx(arg, loc, idx - half, hidx)
+    a⁺ = getidx(arg, loc, idx + half, hidx)
```
Zhengyu-Huang, line 1538: Does `a⁻` represent `phi_{j-1}^{td}`, why's it idx - half instead of idx - half - 1 ? Is the idx  j+1/2 ?

src/Operators/finitedifference.jl
charleskawczynski: Discussed this, looks good because the only difference is based on `sign`

src/Operators/finitedifference.jl
valeriabarra: `a⁺` is `phi_{j}^{td}`, and `a⁻` is `phi_{j-1}^{td}`, similar to what we have for the 3rd-order-upwinding operator, see this [line](https://github.com/CliMA/ClimaCore.jl/blob/b133c3ae65e40c993f46c57466563d8cd1f99faa/src/Operators/finitedifference.jl#L1354)

charleskawczynski: @Zhengyu-Huang, another way to think about this is that `PlusHalf(0)` comes before `1`.

charleskawczynski (approved):

valeriabarra:
Thanks all for the review. I squashed my commits to one

src/Operators/finitedifference.jl
Zhengyu-Huang: I still need some help to understand the code here, since it is different from 3rd order upwinding stencils, so I can implement Zalesak Corrector

1. Does this function compute the flux on the `idx` face, and the faces are labeled as 1, 2, ..., 
2. a is a cell center variable, phi^{td}, and they are labeled as 1, 2, ...
3. for face j, the cells on left and right are labeled as j-1 and j
4. to compute flux on the jth face, we need, phi^{td}[j-2], phi^{td}[j-1], phi^{td}[j], phi^{td}[j+1]
5. `idx - half` in the code means `idx - 2` here?

valeriabarra: See #885

PR #887, opened 2022-08-12
Implement Zalesak Correction

Zhengyu-Huang:

The advection test with periodic boundary condition for 1 period 
exact_and_computed_advected_square_wave_ZalesakFCT

The advection test with non-penetration boundary condition
exact_and_computed_advected_square_wave_ZalesakFCT

### DIFF
--- a/docs/refs.bib
+++ b/docs/refs.bib
@@ -30,6 +30,25 @@
   author = {Jay P Boris and David L Book}
 }
 
+@book{durran2013numerical,
+  title={Numerical methods for wave equations in geophysical fluid dynamics},
+  author={Durran, Dale R},
+  volume={32},
+  year={2013},
+  publisher={Springer Science \& Business Media}
+}
+
+@article{zalesak1979fully,
+  title={Fully multidimensional flux-corrected transport algorithms for fluids},
+  author={Zalesak, Steven T},
+  journal={Journal of computational physics},
+  volume={31},
+  number={3},
+  pages={335--362},
+  year={1979},
+  publisher={Elsevier}
+}
+
 @article{Galewsky2004,
   author = {Joseph Galewsky and Richard K. Scott and Lorenzo M. Polvani},
   title = {An initial-value problem for testing numerical models of the global shallow-water equations},
--- a/examples/column/zalesak_fct_advection.jl
+++ b/examples/column/zalesak_fct_advection.jl
@@ -1,0 +1,144 @@
+using Test
+using LinearAlgebra
+using OrdinaryDiffEq: ODEProblem, solve
+using DiffEqBase
+using ClimaTimeSteppers
+
+import ClimaCore:
+    Fields,
+    Domains,
+    Topologies,
+    Meshes,
+    DataLayouts,
+    Operators,
+    Geometry,
+    Spaces
+
+
+# Advection Equation, with constant advective velocity (so advection form = flux form)
+# ∂_t y + w ∂_z y  = 0
+# the solution translates to the right at speed w,
+# so at time t, the solution is y(z - w * t)
+
+# visualization artifacts
+ENV["GKSwstype"] = "nul"
+using ClimaCorePlots, Plots
+Plots.GRBackend()
+dir = "zalesak_fct_advection"
+path = joinpath(@__DIR__, "output", dir)
+mkpath(path)
+
+function linkfig(figpath, alt = "")
+    # buildkite-agent upload figpath
+    # link figure in logs if we are running on CI
+    if get(ENV, "BUILDKITE", "") == "true"
+        artifact_url = "artifact://$figpath"
+        print("\033]1338;url='$(artifact_url)';alt='$(alt)'\a\n")
+    end
+end
+
+function f!(dydt, y, parameters, t, alpha, beta)
+
+    (; w, A, y_td) = parameters
+    y = y.y
+    dydt = dydt.y
+
+    first_order_fluxᶠ = Operators.UpwindBiasedProductC2F(
+        bottom = Operators.Extrapolate(),
+        top = Operators.Extrapolate(),
+    )
+    third_order_fluxᶠ = Operators.Upwind3rdOrderBiasedProductC2F(
+        bottom = Operators.ThirdOrderOneSided(),
+        top = Operators.ThirdOrderOneSided(),
+    )
+    divf2c = Operators.DivergenceF2C(
+        bottom = Operators.SetValue(Geometry.WVector(FT(0.0))),
+        top = Operators.SetValue(Geometry.WVector(FT(0.0))),
+    )
+    FCTZalesak = Operators.FCTZalesak(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
+    )
+
+    @. y_td = beta * dydt - alpha * divf2c(first_order_fluxᶠ(w, y))
+    @. dydt =
+        y_td -
+        alpha * divf2c(
+            FCTZalesak(
+                third_order_fluxᶠ(w, y) - first_order_fluxᶠ(w, y),
+                y / alpha,
+                y_td / alpha,
+            ),
+        )
+
+    return dydt
+end
+
+FT = Float64
+t₀ = FT(0.0)
+t₁ = FT(1.0)
+z₀ = FT(0.0)
+zₕ = FT(1.0)
+z₁ = FT(1.0)
+speed = FT(1.0)
+
+n = 2 .^ 6
+
+domain = Domains.IntervalDomain(
+    Geometry.ZPoint{FT}(-π),
+    Geometry.ZPoint{FT}(π);
+    boundary_names = (:bottom, :top),
+    # periodic = true,
+)
+mesh = Meshes.IntervalMesh(domain, nelems = n)
+cs = Spaces.CenterFiniteDifferenceSpace(mesh)
+fs = Spaces.FaceFiniteDifferenceSpace(cs)
+zc = Fields.coordinate_field(cs)
+
+# Unitary, constant advective velocity
+w = Geometry.WVector.(speed .* ones(FT, fs))
+
+# Define a pulse wave or square wave
+pulse(z, t, z₀, zₕ, z₁) = abs(z.z - speed * t) ≤ zₕ ? z₁ : z₀
+
+# Initial condition
+y0 = pulse.(zc, 0.0, z₀, zₕ, z₁)
+
+# ClimaTimeSteppers need a FieldVector
+y0 = Fields.FieldVector(y = y0)
+
+# Set up parameters needed for time-stepping
+Δt = 0.0001
+dydt = copy(y0)
+A = similar(dydt.y)
+y_td = similar(dydt.y)
+
+parameters = (; w, A, y_td)
+# Call the RHS function
+f!(dydt, y0, parameters, 0.0, Δt, 1)
+t₁ = 100Δt
+prob = ODEProblem(IncrementingODEFunction(f!), copy(y0), (t₀, t₁), parameters)
+sol = solve(
+    prob,
+    SSPRK33ShuOsher(),
+    dt = Δt,
+    saveat = Δt,
+    progress = true,
+    adaptive = false,
+    progress_message = (dt, u, p, t) -> t,
+)
+computed_result = sol.u[end].y
+analytical_result = pulse.(zc, t₁, z₀, zₕ, z₁)
+err = norm(computed_result .- analytical_result)
+initial_mass = sum(sol.u[1].y)
+mass = sum(sol.u[end].y)
+rel_mass_err = norm((mass - initial_mass) / initial_mass)
+
+@test err ≤ 0.019
+@test rel_mass_err ≤ 13eps()
+
+plot(sol.u[end].y)
+Plots.png(
+    Plots.plot!(analytical_result, title = "Zalesak FCT"),
+    joinpath(path, "exact_and_computed_advected_square_wave_ZalesakFCT.png"),
+)
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -1588,6 +1588,210 @@ end
     return Geometry.Contravariant3Vector(vᶠ)
 end
 
+
+
+#########################
+"""
+    U = FCTZalesak(;boundaries)
+    U.(A, Φ, Φᵗᵈ)
+
+Correct the flux using the flux-corrected transport formulation by Zalesak
+Input arguments:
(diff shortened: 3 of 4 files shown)


examples/Project.toml
```diff
@@ -24,6 +24,7 @@ Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SnoopPrecompile = "66db9d55-30c0-4569-8b51-7e840670fc0c"
```
valeriabarra, line 27: The `examples/Manifest.toml` and `examples/Project.toml` files shouldn't have been checked-in, unless this test adds a specific new necessary dependency

examples/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.7.1"
```
charleskawczynski, line 3: @Zhengyu-Huang, it looks like you're using an older version of Julia, can you upgrade? Let's revert these manifest file changes

examples/Project.toml
charleskawczynski, line 27: Let's remove this

src/Operators/finitedifference.jl
```diff
@@ -1592,6 +1592,214 @@ end
...
+Φ_j^{n+1} = Φ_j^{td} - \frac{dt}{dx}(C_{j+\frac{1}{2}}A_{j+\frac{1}{2}} - C_{j-\frac{1}{2}}A_{j-\frac{1}{2}})
+```
+This stencil is based on [durran2013numerical](@cite).
+
+Supported boundary conditions are:
+- [`ThirdOrderOneSided(x₀)`](@ref): uses the third-order downwind reconstruction to compute `x` on the left boundary,
+and the third-order upwind reconstruction to compute `x` on the right boundary.
+
+!!! note
+    Similar to the [`Upwind3rdOrderBiasedProductC2F`](@ref) operator, these boundary conditions do not define the value at the actual boundary faces, and so this operator cannot be materialized directly: it needs to be composed with another operator that does not make use of this value, e.g. a [`DivergenceF2C`](@ref) operator, with a [`SetValue`](@ref) boundary.
```
charleskawczynski, line 1617: Doc string needs fixed, also, can we break the line so that it doesn't go so far out?

src/Operators/finitedifference.jl
```diff
@@ -1592,6 +1592,214 @@ end
...
+    Rⱼ⁻ = (Pⱼ⁻ > 0 ? min(1, Qⱼ⁻ / Pⱼ⁻) : 0)
+
+    ϕⱼ₊₁ᵐᵃˣ = max(ϕⱼ, ϕⱼ₊₁, ϕⱼ₊₂, ϕⱼᵗᵈ, ϕⱼ₊₁ᵗᵈ, ϕⱼ₊₂ᵗᵈ)
+    ϕⱼ₊₁ᵐⁱⁿ = min(ϕⱼ, ϕⱼ₊₁, ϕⱼ₊₂, ϕⱼᵗᵈ, ϕⱼ₊₁ᵗᵈ, ϕⱼ₊₂ᵗᵈ)
+    Pⱼ₊₁⁺ = max(0, Aⱼ₊₁₂) - min(0, Aⱼ₊₃₂)
+    Qⱼ₊₁⁺ = (ϕⱼ₊₁ᵐᵃˣ - ϕⱼ₊₁ᵗᵈ) * dxⱼ₊₁
+    Rⱼ₊₁⁺ = (Pⱼ₊₁⁺ > 0 ? min(1, Qⱼ₊₁⁺ / Pⱼ₊₁⁺) : 0)
+    Pⱼ₊₁⁻ = max(0, Aⱼ₊₃₂) - min(0, Aⱼ₊₁₂)
+    Qⱼ₊₁⁻ = (ϕⱼ₊₁ᵗᵈ - ϕⱼ₊₁ᵐⁱⁿ) * dxⱼ₊₁
+    Rⱼ₊₁⁻ = (Pⱼ₊₁⁻ > 0 ? min(1, Qⱼ₊₁⁻ / Pⱼ₊₁⁻) : 0)
```
charleskawczynski, line 1682: Please fix the type instabilities here, see the julia docs on [type promotions](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/).

src/Operators/finitedifference.jl
```diff
@@ -1592,6 +1592,214 @@ end
...
+    Φᵗᵈ_space,
+)
+    face_space = axes(A_space)
+    @assert idx <= right_face_boundary_idx(face_space) - 1
+
+    Aⱼ₊₁₂ = Geometry.contravariant3(
+        getidx(A_space, loc, idx, hidx),
+        Geometry.LocalGeometry(face_space, idx, hidx),
+    )
+    return Geometry.Contravariant3Vector(Aⱼ₊₁₂ * 0.0)
```
charleskawczynski, line 1798: Why is this multiplying by zero?

src/Operators/finitedifference.jl
```diff
@@ -1592,6 +1592,214 @@ end
...
+    Φᵗᵈ_space,
+)
+    face_space = axes(A_space)
+    @assert idx <= left_face_boundary_idx(face_space) + 1
+
+    Aⱼ₊₁₂ = Geometry.contravariant3(
+        getidx(A_space, loc, idx, hidx),
+        Geometry.LocalGeometry(face_space, idx, hidx),
+    )
+    return Geometry.Contravariant3Vector(Aⱼ₊₁₂ * 0)
```
charleskawczynski, line 1778: Why is this multiplying by zero?

docs/refs.bib
```diff
@@ -30,6 +30,14 @@ @article{BorisBook1973
   author = {Jay P Boris and David L Book}
 }
 
+@book{durran2013numerical,
```
valeriabarra, line 33: I would also add the original citation, Zalesak 1979

src/Operators/finitedifference.jl
Zhengyu-Huang: I want to set to 0, but if I set it to 0, I got errors

src/Operators/finitedifference.jl
Zhengyu-Huang: I change all 0 to `stable_zero = zero(eltype(Aⱼ₊₁₂))`

src/Operators/finitedifference.jl
Zhengyu-Huang: fixed

src/Operators/finitedifference.jl
valeriabarra: Does A here represent the same thing as in the BB FCT? If so, in #888 we left only the value of A in the BCs. Why here we want this to be always zero?

src/Operators/finitedifference.jl
```diff
@@ -1592,6 +1592,217 @@ end
     return Geometry.Contravariant3Vector(vᶠ * a)
 end
 
+
+
+#########################
+"""
+    U = FCTZalesak(;boundaries)
+    U.(v, x)
```
charleskawczynski, line 1600:
```suggestion
    U.(A, Φ, Φᵗᵈ)
```
?

src/Operators/finitedifference.jl
charleskawczynski: If it should be zero, then use `zero` (first determine the correct type), and don't call `contravariant3` (this performs operations). If it's not meant to be zero, then let's correct that issue

examples/column/zalesak_fct_advection.jl
```diff
@@ -0,0 +1,144 @@
...
+mesh = Meshes.IntervalMesh(domain, nelems = n)
+cs = Spaces.CenterFiniteDifferenceSpace(mesh)
+fs = Spaces.FaceFiniteDifferenceSpace(cs)
+zc = Fields.coordinate_field(cs)
+
+# Unitary, constant advective velocity
+w = Geometry.WVector.(speed .* ones(FT, fs))
+
+# Define a pulse wave or square wave
+pulse(z, t, z₀, zₕ, z₁) = abs(z.z - speed * t) ≤ zₕ ? z₁ : z₀
```
valeriabarra, line 102: I would generalize this (and the comment above), for the case in which you want to run this for one (or more) periods

src/Operators/finitedifference.jl
Zhengyu-Huang: I think both A and 0 are correct, A->third order upwind boundary condition implementation, 0->first order upwind boundary condition implementation. I tried both, 0-> gives better results in terms of overshoting

examples/column/zalesak_fct_advection.jl
Zhengyu-Huang: I think the current setup is fine, I copy the test as your bb_fct test. I only run 1 period to understand the algorithm

docs/refs.bib
Zhengyu-Huang: just updated the citation

Zhengyu-Huang:
@charleskawczynski could you help me with the unit tests? I saw the unit test failed at JET Test, do you know what that is?

src/Operators/finitedifference.jl
```diff
@@ -1588,6 +1588,209 @@ end
...
+        Aⱼ₊₁₂ * (ϕⱼ₊₂ᵗᵈ - ϕⱼ₊₁ᵗᵈ) < stable_zero ||
+        Aⱼ₊₁₂ * (ϕⱼᵗᵈ - ϕⱼ₋₁ᵗᵈ) < stable_zero
+    )
+        Aⱼ₊₁₂ = stable_zero
+    end
+    ϕⱼᵐᵃˣ = max(ϕⱼ₋₁, ϕⱼ, ϕⱼ₊₁, ϕⱼ₋₁ᵗᵈ, ϕⱼᵗᵈ, ϕⱼ₊₁ᵗᵈ)
+    ϕⱼᵐⁱⁿ = min(ϕⱼ₋₁, ϕⱼ, ϕⱼ₊₁, ϕⱼ₋₁ᵗᵈ, ϕⱼᵗᵈ, ϕⱼ₊₁ᵗᵈ)
+    Pⱼ⁺ = max(stable_zero, Aⱼ₋₁₂) - min(stable_zero, Aⱼ₊₁₂)
+    Qⱼ⁺ = (ϕⱼᵐᵃˣ - ϕⱼᵗᵈ) * dxⱼ
+    Rⱼ⁺ = (Pⱼ⁺ > stable_zero ? min(1, Qⱼ⁺ / Pⱼ⁺) : stable_zero)
```
charleskawczynski, line 1669: Can we also use `one(::Type)` instead of `1` here?

charleskawczynski:
> @charleskawczynski could you help me with the unit tests? I saw the unit test failed at JET Test, do you know what that is?

If what you're talking about is the 1.8 unit tests, those are broken on main, so it's not related to this PR.

src/Operators/finitedifference.jl
Zhengyu-Huang: just updated

Zhengyu-Huang:
> > @charleskawczynski could you help me with the unit tests? I saw the unit test failed at JET Test, do you know what that is?
> 
> If what you're talking about is the 1.8 unit tests, those are broken on main, so it's not related to this PR.

should I merge this PR without touching it?

src/Operators/finitedifference.jl
valeriabarra: You're saying that we should use 0 also for the BB FCT. I'd like these BCs to be consistent for the two operators, unless there is a specific reason why they shouldn't be. Otherwise it gets confusing for people who might be looking into this in the future and we can't remember why we made one choice over the other.

charleskawczynski (approved):

valeriabarra:
@Zhengyu-Huang if you can, please see my question above. I think it's better to be consistent.

src/Operators/finitedifference.jl
Zhengyu-Huang: I am not sure whether we should use 0 also for the BB FCT, could you try both for more tests to see which one is better? For Zalesak, using 0 reduces undershooting

Zhengyu-Huang:
bors try

src/Operators/finitedifference.jl
```diff
@@ -1588,6 +1588,210 @@ end
...
+- a center-valued field `Φ`
+- a center-valued field `Φᵗᵈ`
+```math
+Φ_j^{n+1} = Φ_j^{td} - \frac{dt}{dx}(C_{j+\frac{1}{2}}A_{j+\frac{1}{2}} - C_{j-\frac{1}{2}}A_{j-\frac{1}{2}})
+```
+This stencil is based on [zalesak1979fully](@cite).
+
+Supported boundary conditions are:
+- [`ThirdOrderOneSided(x₀)`](@ref): uses the third-order downwind reconstruction to compute `x` on the left boundary,
+and the third-order upwind reconstruction to compute `x` on the right boundary.
```
valeriabarra, line 1610: If in this case, the boundary condition is falling back to the first-order-upwinding case, then this docstring should be updated

Zhengyu-Huang: thanks! just change it to FirstOrder

src/Operators/finitedifference.jl
```diff
@@ -1588,6 +1588,210 @@ end
...
+    idx,
+    hidx,
+    A_space,
+    Φ_space,
+    Φᵗᵈ_space,
+)
+    face_space = axes(A_space)
+    @assert idx <= left_face_boundary_idx(face_space) + 1
+
+    return Geometry.Contravariant3Vector(zero(eltype(eltype(A_space))))
```
valeriabarra, line 1774: Why two `eltype`s?

src/Operators/finitedifference.jl
```diff
@@ -1588,6 +1588,210 @@ end
...
+    idx,
+    hidx,
+    A_space,
+    Φ_space,
+    Φᵗᵈ_space,
+)
+    face_space = axes(A_space)
+    @assert idx <= right_face_boundary_idx(face_space) - 1
+
+    return Geometry.Contravariant3Vector(zero(eltype(eltype(A_space))))
```
valeriabarra, line 1790: Why two `eltype`s?

src/Operators/finitedifference.jl
Zhengyu-Huang: @simonbyrne  helps me with this, I guess the first one returns a vector, and the second returns a scalar?

Issue #884, opened 2022-08-12
Add test for retrieving `dz` in LocalGeometry for different spaces


charleskawczynski:
One thing that we do, somewhat automatically but not very transparently in our finite difference stencils is extracting `dz` from the LocalGeometry. We should add a test, for clarity, how to do this in general. #883 implements this using `step = Geometry.LocalGeometry(space, idx, hidx).∂x∂ξ[1]`, and my main question / request is to add a test demonstrating that this is a valid `dz` for all spaces.

Zhengyu-Huang:
I agree we need verify that, I changed  it to `step = Geometry.LocalGeometry(space, idx, hidx).∂x∂ξ[3, 3]`

charleskawczynski:
Closed by #962.

PR #889, opened 2022-08-13
Examples: Add BB FCT to deformation flow

valeriabarra:
Adding a WIP PR for now to gather feedback. 

This adds the BB FCT to the deformation flow example.

### DIFF
--- a/examples/hybrid/sphere/deformation_flow.jl
+++ b/examples/hybrid/sphere/deformation_flow.jl
@@ -47,6 +47,7 @@ const κ₄ = 1.0e16 # hyperviscosity
 # time constants
 T = 86400.0 * 12.0
 dt = 60.0 * 60.0
+
 FT = Float64
 
 # set up function space
@@ -368,7 +369,7 @@ q1_error =
 q2_error =
     norm(sol.u[end].ρq2 ./ ρ_ref.(coords.z) .- y0.ρq2 ./ ρ_ref.(coords.z)) /
     norm(y0.ρq2 ./ ρ_ref.(coords.z))
-@test q2_error ≈ 0.0 atol = 0.031
+@test q2_error ≈ 0.0 atol = 0.032
 
 q3_error =
     norm(sol.u[end].ρq3 ./ ρ_ref.(coords.z) .- y0.ρq3 ./ ρ_ref.(coords.z)) /
@@ -378,7 +379,7 @@ q3_error =
 q4_error =
     norm(sol.u[end].ρq4 ./ ρ_ref.(coords.z) .- y0.ρq4 ./ ρ_ref.(coords.z)) /
     norm(y0.ρq4 ./ ρ_ref.(coords.z))
-@test q4_error ≈ 0.0 atol = 0.03
+@test q4_error ≈ 0.0 atol = 0.025
 
 # Tracer mass conservation checks
 q1_initial_mass = sum(y0.ρq1)
@@ -399,7 +400,7 @@ q3_rel_mass_err = norm((q3_mass - q3_initial_mass) / q3_initial_mass)
 q4_initial_mass = sum(y0.ρq4)
 q4_mass = sum(sol.u[end].ρq4)
 q4_rel_mass_err = norm((q4_mass - q4_initial_mass) / q4_initial_mass)
-@test q4_rel_mass_err ≈ 0.0 atol = 7e1eps(FT)
+@test q4_rel_mass_err ≈ 0.0 atol = 8e1eps(FT)
 
 # visualization artifacts
 ENV["GKSwstype"] = "nul"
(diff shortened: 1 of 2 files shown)


examples/hybrid/sphere/deformation_flow.jl
```diff
@@ -221,35 +229,79 @@ function rhs!(dydt, y, parameters, t, alpha, beta)
...
-            vdivf2c(Ic2f(ρ) * first_order_upwind_c2f(w, ρq1 / ρ)) +
-            corrected_antidiff_flux
+        dρq1_td -
+        alpha * vdivf2c(
+            FCTBB(
+                Ic2f(ρ) * (
+                    third_order_upwind_c2f(w, ρq1 / ρ) -
+                    first_order_upwind_c2f(w, ρq1 / ρ)
+                ),
+                dρq1_td / alpha,
```
valeriabarra, line 245: @simonbyrne I modified this as we discussed but this test still doesn't work. I also tried to lower the dt significantly, but didn't help. It must be something else that I am missing now...

valeriabarra: @simonbyrne , @charleskawczynski , @daniel my first thought (and question in the Slack channel) was that `dρq1_td` should account for the history of the vertical transport, according to the algorithm in the book. I tried to account for the history of this, by saving it as another state variable in [0fd9830](https://github.com/CliMA/ClimaCore.jl/pull/889/commits/0fd98304e01a4e2dd1421e409730d2e6b3ad1c2d), but it still didn't work. Can we please discuss these implementation details about how to use these FCT operators in 3D problems?

Zhengyu-Huang (commented):
Can I modify directly in your branch?

valeriabarra:
> Can I modify directly in your branch?

sure

valeriabarra:
> > Can I modify directly in your branch?
> 
> sure

Since #904 was just merged, this will need to be rebase onto `main` and updated to use `FirstOrderOneSided`. I can do it, and push again to this branch. Let me know if you're not touching it already

Zhengyu-Huang (approved):
I think the current version works, there is one bug in the dx computation in the flux corrector, could you double check that？

examples/hybrid/sphere/deformation_flow.jl
```diff
@@ -197,6 +215,10 @@ function rhs!(dydt, y, parameters, t, alpha, beta)
         bottom = Operators.ThirdOrderOneSided(),
         top = Operators.ThirdOrderOneSided(),
     )
+    FCTZalesak = Operators.FCTZalesak(
```
valeriabarra, line 218: This test case was for BB. Could we do things incrementally please so that later we know how the Zalesak one compares with BB?

valeriabarra: We can Zalesak next

valeriabarra:
> I think the current version works, there is one bug in the dx computation in the flux corrector, could you double check that？

What do you mean? We had opened an issue (#884) to check the dz indeed, and it also needs to be tested also on stretched grids

examples/hybrid/sphere/deformation_flow.jl
Zhengyu-Huang: Sure, you can change Zalesak to BB, it should work, you need check dz in BB too

src/Operators/finitedifference.jl
```diff
@@ -1728,9 +1728,8 @@ stencil_interior_width(::FCTZalesak, A_space, Φ_space, Φᵗᵈ_space) =
     )
 
 
-    dxⱼ = Geometry.LocalGeometry(center_space, idx - half, hidx).∂x∂ξ[1]
-    dxⱼ₊₁ = Geometry.LocalGeometry(center_space, idx + half, hidx).∂x∂ξ[1]
-
+    dxⱼ = Geometry.LocalGeometry(center_space, idx - half, hidx).∂x∂ξ[3, 3]
+    dxⱼ₊₁ = Geometry.LocalGeometry(center_space, idx + half, hidx).∂x∂ξ[3, 3]
```
valeriabarra, line 1732: @simonbyrne what is a robust way to get `dz` across all types of domains (1D-3D), both uniform and stretched? I had tested the `∂x∂ξ[1]` for both 1D and extruded 3D spherical domains and it seemed to be working as expected

valeriabarra: #884 was a reminder for ourselves to add unit tests for this indeed

charleskawczynski: Related issue: #884.

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
@simonbyrne  and @Zhengyu-Huang , I removed the step from the BB FCT and tested this, together with the 1D test and they look fine. 
Once bors is done I think we can merge this. In the meantime, I will work on removing the dx also from Zalesak FCT and test that one on the Hadley-like circulation test case.

src/Operators/finitedifference.jl
valeriabarra: We discussed removing the dx altogether. Related PR: #907

valeriabarra:
bors try

PR #892, opened 2022-08-15
use a vector as a backing array for DataF

### LINKED ISSUE #877: Missing support for similar / zeros for point spaces

simonbyrne:
Alternative to #882, fixes #877.

### DIFF
--- a/src/Spaces/pointspace.jl
+++ b/src/Spaces/pointspace.jl
@@ -5,7 +5,7 @@ local_geometry_data(space::AbstractPointSpace) = space.local_geometry
 """
     PointSpace <: AbstractSpace
 
-A zero-dimensional space. 
+A zero-dimensional space.
 """
 struct PointSpace{LG} <: AbstractPointSpace
     local_geometry::LG
@@ -13,7 +13,7 @@ end
 
 function PointSpace(local_geometry::LG) where {LG <: Geometry.LocalGeometry}
     FT = Geometry.undertype(LG)
-    local_geometry_data = DataLayouts.DataF{LG}(Matrix{FT})
+    local_geometry_data = DataLayouts.DataF{LG}(Array{FT})
     local_geometry_data[] = local_geometry
     return PointSpace(local_geometry_data)
 end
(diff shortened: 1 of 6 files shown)


charleskawczynski (approved):

src/Spaces/pointspace.jl
```diff
@@ -5,15 +5,15 @@ local_geometry_data(space::AbstractPointSpace) = space.local_geometry
...
+A zero-dimensional space.
 """
 struct PointSpace{LG} <: AbstractPointSpace
     local_geometry::LG
 end
 
 function PointSpace(local_geometry::LG) where {LG <: Geometry.LocalGeometry}
     FT = Geometry.undertype(LG)
-    local_geometry_data = DataLayouts.DataF{LG}(Matrix{FT})
+    local_geometry_data = DataLayouts.DataF{LG}(Array{FT})
```
charleskawczynski, line 16: Shouldn't we be specifying the size? What will this do exactly?

charleskawczynski: Ever since fixing inference issues in Thermodynamics, I've tried avoiding UnionAll types

simonbyrne: just determines what type (Array, CuArray) and floating point type (Floa32, Float64). Dimensionality is handled by the constructor

kmdeck:
this looks like the better solution! Though it is failing because it is a bit slower than previous code was?

simonbyrne:
@charleskawczynski the failure looks unrelated: can we make the benchmark tests a little less strict?

charleskawczynski:
> @charleskawczynski the failure looks unrelated: can you look into this, or make the benchmark tests a little less strict?

Yeah, we can make the benchmark less strict, I'll push a fix

PR #895, opened 2022-08-15
Update remap_weights in wrappers.jl

juliasloan25:
Co-authored-by: LenkaNovak <LenkaNovak@users.noreply.github.com>

Add functionality in remap_weights to correctly parse arguments without a value, for example the flag `--mono`. This is used for the function `TempestRemap_jll.GenerateOfflineMap_exe()` by the coupler.

### DIFF
--- a/lib/ClimaCoreTempestRemap/src/wrappers.jl
+++ b/lib/ClimaCoreTempestRemap/src/wrappers.jl
@@ -77,7 +77,9 @@ Keyword arguments are passed as command-line options. These include:
   - `"cgll"`: continuous GLL finite element method (a single value for colocated nodes)
   - `"dgll"`: discontinuous GLL finite element method (duplicate values for colocated nodes)
 - 'in_np'/'out_np': Order of input and output meshes
+- 'mono': Monotonicity of remapping. Note: must be used with in_np = 1
 
+Set `mono = true` and `in_np = 1` for monotone remapping
 Set `verbose=true` to print information.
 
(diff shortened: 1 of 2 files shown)


lib/ClimaCoreTempestRemap/src/wrappers.jl
```diff
@@ -98,7 +98,11 @@ function remap_weights(
     --out_map $weightfile
     ```
     for (k, v) in kwargs
-        append!(cmd.exec, [string("--", k), string(v)])
+        if isempty(v)
```
simonbyrne, line 101: what would `v` be in this case?

simonbyrne:
Maybe add a test case as well? See existing ones in https://github.com/CliMA/ClimaCore.jl/blob/main/lib/ClimaCoreTempestRemap/test/netcdf.jl

lib/ClimaCoreTempestRemap/src/wrappers.jl
juliasloan25: `v` would be the value you want to set argument `k` to in the command we're constructing. For example, `k=in_np` and `v=1` would produce the string `--in_np 1`, which would get appended to `cmd`.

simonbyrne: I meant what would the user set it to if they wanted to use the mono flag?

juliasloan25:
bors try

juliasloan25:
bors try

simonbyrne (approved):

PR #905, opened 2022-08-19
work around PkgVersion bug

simonbyrne:
PkgVersion doesn't work when called from subdirectories of `src` (https://github.com/KlausC/PkgVersion.jl/issues/8), so we define a constant `VERSION` in the main ClimaCore file which we import to subfiles.

Also cleans up some imports, and adds a test which uses a specified element ordering (which triggers the version issue)


simonbyrne:
bors try

simonbyrne:
bors try

PR #907, opened 2022-08-22
Zalesak FCT: remove dx and add it to Hadley circ test

### LINKED ISSUE #851: Flux-corrected transport
## Purpose
Flux-corrected transport is a requested feature. 

My current understanding is that this, applied only in the vertical direction, is a combination of the first-order upwinding  ([UpwindBiasedProductC2F](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.UpwindBiasedProductC2F)) and third-order upwinding ([Upwind3rdOrderBiasedProductC2F](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.Upwind3rdOrderBiasedProductC2F)) schemes. 

Since this work builds on the third-order upwinding operator and there have been observed instabilities when using this operator in the high-resolution runs in ClimaAtmos, it is essential to first rule out any potential issue with the third-order upwinding operator first. 

There can be several options to combine these two low-and-high order operators. The simplest one is a linear combination of the two (I would call this the _vanilla case_), where one could reduce to first-order upwinding only, or third-order upwinding only, by switching on/off the correction factor coefficient. 

## Cost/benefits/risks
My understanding is that this flux-corrected transport will act similarly to the flux-limiters we currently have that act only in the horizontal direction, but in the vertical direction. Hence, the corrected flux will be ensured to be not large enough or small enough to generate new extrema in the vertically advected quantities.

## Producers
@valeriabarra with help of others if needed

## Components
This proposal shouldn't need to create/develop new operators, but should be able to reuse our existing operators (i.e., library code) and only combine them in application code, i.e., examples. 
Perhaps, when turning our attention to the non-vanilla applications, we may have to develop some fancy way of computing the correction factor and this might require new library code/utilities. 

## Inputs
Chapter 5.4 in the following reference: 
"Numerical Methods for Fluid Dynamics", by Dale Durran (2010)

## Results and deliverables
Use some of the existing examples and add the flux-corrected transport to show that some ripples/oscillations in the vertical direction can be reduced. Similarly to what was done in the limiters examples.

## Task breakdown
A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results.
- [x] Task 1: Start with the simple, vanilla case (Sec. 5.4 in reference book) Timeline less than a week.
     - [x] Task 1.1: Add spatial convergence unit tests for this case
     - [x] Task 1.2: Add temporal integration/unit test for this case
     - [x] Task 1.3: Add the vanilla case to a more complicated integration example: the deformation flow example.
- [ ] Task 2: Add more advanced cases from references (Expected timeline not so clear at the moment. Need to read more about the most advanced cases to assess their implementation challenges). Start with Sec. 5.4.1 in the reference book, i.e., "Boris and Book FCT" (1973):
     - [x] Task 2.1:  Edited: Add a new ClimaCore operator for the Boris and Book FCT
     - [x] Task 2.2: Add temporal integration/unit test for this case (<- this FCT type requires the usage of step-function specification, a la ClimaTimeSteppers, instead of tendency specification; so for now it is better to use it directly in temporal unit/integration tests). Reused part of the same unit test from Task 1.2, but adapted to use ClimaTimeSteppers.
     - [x] Task 2.3: Add the Boris and Book FCT case to a more complicated integration example: the deformation flow example.
- [ ] Task 3: Repeat with the next section, Sec. 5.4.2 in the reference book, i.e., "Zalesak corrector FCT" (1979):
     - [x] Task 3.1: Add a new ClimaCore operator for the Zalesak FCT (done by @Zhengyu-Huang  in #887 )
     - [x] Task 3.2: Add temporal integration/unit test for this case. (#887 reused the same one from Task 1.2)
     - [ ] Task 3.3: Add the Zalesak corrector FCT case to a more complicated integration example: the Hadley-like meridional circulation test case.
...

## Reviewers
@simonbyrne

valeriabarra:
This PR removes dx from the Zalesak operator and adds it to the Hadley-like meridional circulation test case. I also updated the mass conservation check in the example, since they were erroneously considering only the tracer instead of the tracer mass. 

The error looks slightly larger with Zalesak compared to only 3rd-order-upwinding. 

This will also complete Task 3.3 in the SDI #851, and therefore, close #851

### DIFF
--- a/examples/hybrid/sphere/hadley_circulation.jl
+++ b/examples/hybrid/sphere/hadley_circulation.jl
@@ -79,7 +79,7 @@ end
 # RHS tendency specification
 function rhs!(dydt, y, parameters, t, alpha, beta)
 
-    (; coords, face_coords, τ, u₀, w₀, K, ρ₀, ystar) = parameters
+    (; coords, face_coords, τ, u₀, w₀, K, ρ₀, ystar, y_td) = parameters
 
     ϕ = coords.lat
     zc = coords.z
@@ -104,6 +104,7 @@ function rhs!(dydt, y, parameters, t, alpha, beta)
     ρ = y.ρ
     ρq1 = y.ρq1
     dρq1 = dydt.ρq1
+    ρq1_td = y_td.ρq1
 
     # No change in density: divergence-free flow
     @. dydt.ρ = beta * dydt.ρ + 0 .* y.ρ
@@ -117,9 +118,17 @@ function rhs!(dydt, y, parameters, t, alpha, beta)
         top = Operators.SetValue(Geometry.Contravariant3Vector(0.0)),
         bottom = Operators.SetValue(Geometry.Contravariant3Vector(0.0)),
     )
+    first_order_upwind_c2f = Operators.UpwindBiasedProductC2F(
+        bottom = Operators.Extrapolate(),
+        top = Operators.Extrapolate(),
+    )
     third_order_upwind_c2f = Operators.Upwind3rdOrderBiasedProductC2F(
         bottom = Operators.ThirdOrderOneSided(),
         top = Operators.ThirdOrderOneSided(),
+    )
+    FCTZalesak = Operators.FCTZalesak(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
     )
     hdiv = Operators.Divergence()
     hwdiv = Operators.WeakDivergence()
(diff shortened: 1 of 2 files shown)


valeriabarra:
bors try

simonbyrne (approved):

Zhengyu-Huang (approved):
Look good for me! We can do more tests on stretching grid/ 1d advection after this is merged

PR #917, opened 2022-08-25
Do inlining and inbounds more carefully

charleskawczynski:
This PR is an attempt to partially revert (and extend/complete) #829, by:
 - More properly using `@inline`, `@inbounds` and `Base.@propagate_inbounds`
 - Removing `@inline` for "expensive" functions that do not have or uses indices which require boundscheck elision
 - Applying inlining and `@inbounds` to slab functions

The only thing that I'm not 100% sure about is the `@inbounds` around the threading loops, does what I've done look okay? cc @simonbyrne 

Supersedes #891.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -270,7 +270,7 @@ end
 
 function Base.fill!(data::IJFH, val)
     (_, _, _, _, Nh) = size(data)
-    for h in 1:Nh
+    @inbounds for h in 1:Nh
         fill!(slab(data, h), val)
     end
     return data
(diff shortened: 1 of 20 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
Alright, I think this should be fine so long we have a job that applies `check-bounds=yes`.

valeriabarra:
Maybe squash first? Or we don't care?

charleskawczynski:
Ah, I probably should have, CI is almost finished, so I'm fine with letting it go. If the implicit stencil tests hang again then I'll make sure to squash after fixing that

charleskawczynski:
Ok, canceled this, I don't want to completely tank CI for this. I'll revert the apply ops and squash

PR #920, opened 2022-08-25
Add monotonic vertical mesh check

charleskawczynski:
This PR adds a check into our vertical mesh generation, to ensure that the mesh is monotonically increasing. Now that I think of it, @kmdeck, are there meshes where we want monotonically decreasing meshes? If so, we can modify the logic.

### DIFF
--- a/src/Meshes/interval.jl
+++ b/src/Meshes/interval.jl
@@ -71,6 +71,26 @@ function boundary_face_name(mesh::IntervalMesh, elem::Integer, face)
     return nothing
 end
 
+function monotonic_check(faces)
+    if !(hasproperty(first(faces), :z) || eltype(faces) <: Real)
+        return nothing
+    end
+    z(face::AbstractFloat) = face
+    z(face::Geometry.AbstractPoint) = face.z
+    n = length(faces) - 1
+    monotonic_incr = all(map(i -> z(faces[i]) < z(faces[i + 1]), 1:n))
+    monotonic_decr = all(map(i -> z(faces[i]) > z(faces[i + 1]), 1:n))
+    if !(monotonic_incr || monotonic_decr)
+        error(
+            string(
+                "Faces in vertical mesh are not increasing monotonically. ",
+                "We need to have dz_bottom <= z_max / z_elem and dz_top >= z_max / z_elem.",
+            ),
+        )
+    end
+    return nothing
+end
+
 abstract type StretchingRule end
 
 """
@@ -89,6 +109,7 @@ function IntervalMesh(
         throw(ArgumentError("`nelems` must be ≥ 1"))
     end
     faces = range(domain.coord_min, domain.coord_max; length = nelems + 1)
+    monotonic_check(faces)
     IntervalMesh(domain, faces)
 end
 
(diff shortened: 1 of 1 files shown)


szy21:
Looks good, thanks! I think Tapio wants an error message saying something like `We need to have dz_bottom <= z_max/(nz+1) and dz_top >= z_max/(nz+1)` though.

kmdeck:
yep! for the land, we usually want finer resolution at the surface.

kmdeck:
> yep! for the land, we usually want finer resolution at the surface.

(this can be returned to when we need it, if that is easier!)

szy21:
> yep! for the land, we usually want finer resolution at the surface.

I think the question was if z would monotonically decrease for land, that is something like z=0 for the first grid and z<0 for the next grid?

kmdeck:
> > yep! for the land, we usually want finer resolution at the surface.
> 
> I think the question was if z would monotonically decrease for land, that is something like z=0 for the first grid and z<0 for the next grid?

z is always defined such that it is increasing upwards. But we want our mesh finer at the surface than at the bottom, which is opposite what the atmos has, I believe.

szy21:
> z is always defined such that it is increasing upwards. But we want our mesh finer at the surface than at the bottom, which is opposite what the atmos has, I believe.

Yep, but I think this PR is just to check if z is increasing upwards. It doesn't check what dz is doing?

kmdeck:
> > z is always defined such that it is increasing upwards. But we want our mesh finer at the surface than at the bottom, which is opposite what the atmos has, I believe.
> 
> Yep, but I think this PR is just to check if z is increasing upwards. It doesn't check what dz is doing?

makes sense! that is consistent with the land model, then

charleskawczynski:
Updated

src/Meshes/interval.jl
```diff
@@ -71,6 +71,20 @@ function boundary_face_name(mesh::IntervalMesh, elem::Integer, face)
...
 
+function monotonic_check(faces)
+    n = length(faces) - 1
+    monotonic_incr = all(map(i -> faces[i].z < faces[i + 1].z, 1:n))
+    monotonic_decr = all(map(i -> faces[i].z > faces[i + 1].z, 1:n))
+    if !(monotonic_incr || monotonic_decr)
+        error(
+            string(
+                "Faces in vertical mesh are not increasing monotonically. ",
+                "We need to have dz_bottom <= z_max/(nz+1) and dz_top >= z_max/(nz+1).",
```
szy21, line 82:
```suggestion
                "We need to have dz_bottom <= z_max/(z_elem+1) and dz_top >= z_max/(z_elem+1).",
```
Could you double check if this is correct (especially the +1 part)?

szy21 (approved):

src/Meshes/interval.jl
szy21: (z_elem is just to match the command line argument. I don't have a strong preference.)

charleskawczynski: I'm not familiar with this relationship, @trontrytel can you confirm?

szy21: Let me double check with Tapio. I actually think there should be no +1 here.

szy21: (but, yes, it would be useful if @trontrytel can test!)

szy21: Ok, I confirmed with Tapio, there should be no +1 here. I updated the suggestion.

PR #921, opened 2022-08-26
Add threading allocation test

charleskawczynski:
This PR adds a script for investigating allocations when threading. I confirmed that the allocation discrepancy is fixed by #922. We can still pursue this to fix the remaining allocations, but #922 hits the vast majority.


charleskawczynski:
I tried reviving this PR, and the summary is: the unthreaded version does not allocate, and the threading version does, but that's because launching threads allocates (which is a julia issue, not a ClimaCore issue). So, closing this PR.

PR #929, opened 2022-08-30
Add option for unthreaded bycolumn

charleskawczynski:
This will help us fix [CA#737](https://github.com/CliMA/ClimaAtmos.jl/issues/737), and it will help us unify the threaded / unthreaded versions of ClimaAtmos kernels without losing flame graphs. and I think it also just makes sense that threading is just on or off, regardless of whether we're threading for individual broadcast expressions or by columns.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -1,5 +1,6 @@
 module Fields
 
+import ..enable_threading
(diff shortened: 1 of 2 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try-

PR #933, opened 2022-09-04
Use ProfileCanvas, rm ChromeProfileFormat

charleskawczynski:
This PR changes our flame graphs from using ChromeProfileFormat to using ProfileCanvas, which allows us to directly generate html. This is nice because we can upload this as an artifact and don't need to upload the profile into a bucket on the google cloud in order to pass the URI to the firefox profile visualizer


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #935, opened 2022-09-08
revamp hypsography for HDF5 support

simonbyrne:
This simplifies some internals in the Hypsography interface, and adds support for HDF5 export.

Note that this is breaking, but since it isn't widely used (e.g. it isn't in ClimaAtmos yet), I think it should be okay. (@akshaysridhar is that okay?)


src/Hypsography/Hypsography.jl
```diff
@@ -2,10 +2,174 @@ module Hypsography
```
akshaysridhar, line 2: This combines `adaptspace.jl` and `terrainadapt.jl`. `Adaption` now takes the surface elevation as an argument. @simonbyrne It seems like this would change the ClimaAtmos API; this is fine with me. This is functionally (i.e. w.r.t mesh metrics) the same as the existing system.

simonbyrne: Yes, it is breaking: the main question is what code is using this API? I might be able to add some deprecations if need be.

test/InputOutput/hybrid3dcubedsphere_topography.jl
```diff
@@ -0,0 +1,94 @@
...
+    # write field vector to hdf5 file
+    writer = InputOutput.HDF5Writer(filename, comms_ctx)
+    InputOutput.write!(writer, Y, "Y")
+    close(writer)
+
+    reader = InputOutput.HDF5Reader(filename, comms_ctx)
+    restart_Y = InputOutput.read_field(reader, "Y") # read fieldvector from hdf5 file
+    close(reader)
+    @test restart_Y == Y # test if restart is exact
+end
```
akshaysridhar, line 94: Tested restarts and output hdf5 file locally to explore datastructures (Julia HDF5 reader) successfully. This is ready to be merged.

akshaysridhar (approved):

PR #940, opened 2022-09-13
Make slider interaction work for Makie Field recipe

SimonDanisch:
Use lift to visualize a changing field.
The updated code is designed in a way, that the mesh doesn't need recalculating if the sphere itself doesn't change, so only color values will get updated, which is much more efficient.
Also, started using the Makie native types to avoid conversions.
With this PR, a basic slider example works:
```julia
fig = Figure()
level_s = Slider(fig[2, 1], range = 1:n)
# plot level at slider
level_field = map(level_s.value) do n
    ClimaCore.level(T, n)
end
plot(fig[1, 1], level_field, axis=(show_axis=false,))
fig
```
clima-slider


SimonDanisch:
Added basic GeoMakie example:
geomakie
GeoMakie still needs https://github.com/MakieOrg/GeoMakie.jl/pull/123 to work with newest versions, and I will need to think about how to more easily support the GeoMakie mesh plot via a nice recipe API...
In principle, we just do what we do in the current recipe, but with a different coordinate transformation, so we could just add it to the existing recipe!

SimonDanisch:
Improved the dashboard:
clima-dashbaord

Uncovered this bug while switching between the fields:
https://github.com/MakieOrg/Makie.jl/pull/2297#issuecomment-1256065098

simonbyrne (approved):

PR #945, opened 2022-09-15
Inline `copyto!` to fix Ref allocations

### LINKED ISSUE #946: Allocations in broadcasts with Ref
## Describe the bug
We're incurring allocations in broadcast statements involving the use of `Ref`.

charleskawczynski:
This PR applies `@inline` to `copyto!`, which was found to fix allocations with `Ref`s inside broadcast expressions, and adds tests. I'm going to try this out with ClimaAtmos to see if one more minor change is needed, too.

This fixes our `bycolumn` allocations with `Ref`s from 24576 bytes to 0.

Closes #946.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -91,6 +91,16 @@ steps:
       slurm_mem: 64GB
       slurm_ntasks: 1
 
+  - label: ":computer: Field broadcast performance"
+    key: "cpu_field_broadcast_perf"
+    command:
+      - "julia --color=yes --project=test test/Fields/field_bc.jl"
+    agents:
+      config: cpu
+      queue: central
+      slurm_mem: 64GB
+      slurm_ntasks: 1
+
   - label: ":flower_playing_cards: unit tests"
     key: "gpu_unittests"
     command:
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Thanks for the investigatory work, @simonbyrne!

test/Fields/field.jl
```diff
@@ -468,3 +468,41 @@ end
...
+        Y = FieldFromNamedTuple(space, (; x = FT(2)))
+
+        # Plain broadcast
+        Yx = Y.x
+        foo!(Yx) # compile first
+        p = @allocated foo!(Yx)
+        @test_broken p == 0 # 16 bytes
+        # debug info:
+        # @info "space = $(nameof(typeof(space)))"
+        # @show p
```
charleskawczynski, line 501: cc @simonbyrne, not sure why this is still allocating, does anything stick out to you?

charleskawczynski: Fixed 🎉

charleskawczynski:
Oops, looks like that doesn't pass the allocation tests with `--check-bounds=yes`, so I'll move these tests into a separate file.

PR #947, opened 2022-09-16
Bump patch version for new release

charleskawczynski:


charleskawczynski:
looks like the Hadley circulation needs more memory. I’ll try to merge once more as is, and increase if it doesn’t go through.

Issue #949, opened 2022-09-19
`getproperty` on `FieldVector`s incurs allocations


charleskawczynski:
`getproperty` on `FieldVector`s incurs allocations, and they shouldn't.

charleskawczynski:
According to #950, this may not have been an issue afterall.

PR #950, opened 2022-09-19
Add broken allocation test

charleskawczynski:
This PR adds a broken allocation test for `getproperty` with FieldVectors.

I spent a few minutes trying to see what's going on. I tried applying `@inline` to `_values` and `unwrap`, but that didn't work. So, for now I figure we can just add it as a broken test.

Relevant issue: #949.

### DIFF
--- a/test/Fields/field_opt.jl
+++ b/test/Fields/field_opt.jl
@@ -1,3 +1,4 @@
+# These tests require running with `--check-bounds=[auto|no]`
 using Test
 using StaticArrays, IntervalSets
 import ClimaCore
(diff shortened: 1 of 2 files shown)


test/Fields/field_opt.jl
```diff
@@ -51,3 +52,26 @@ include(joinpath(@__DIR__, "util_spaces.jl"))
         @test p == 0
     end
 end
+
+# https://github.com/CliMA/ClimaCore.jl/issues/949
+@testset "Allocations with getproperty on FieldVectors" begin
+    FT = Float64
+    function allocs_test!(Y)
+        x = Y.x
```
simonbyrne, line 60: This could be getting optimized away a no-op. You may have to do something with `x`
```suggestion
        x = Y.x
        fill!(x, 2.0)
```

simonbyrne (approved):

PR #954, opened 2022-09-21
Add a perimeter iterator for 2D spectral elements.

sriharshakandala:

### DIFF
--- a/src/Spaces/spectralelement.jl
+++ b/src/Spaces/spectralelement.jl
@@ -23,6 +23,37 @@ end
 topology(space::AbstractSpectralElementSpace) = space.topology
 quadrature_style(space::AbstractSpectralElementSpace) = space.quadrature_style
 
+abstract type AbstractPerimeter end
+
+"""
+    Perimeter2D <: AbstractPerimeter
+
+Iterate over the perimeter degrees of freedom of a 2D spectral element.
+"""
+struct Perimeter2D{Nq} <: AbstractPerimeter end
+
+"""
+    Perimeter2D(Nq)
+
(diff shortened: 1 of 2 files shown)


simonbyrne (changes requested):
Can you add a test?

src/Spaces/spectralelement.jl
```diff
@@ -23,6 +23,41 @@ end
...
+    ip::V
+    jp::V
+end
+"""
+    Perimeter2D(Nq)
+
+Construct a perimeter iterator for a 2D spectral element of degree `(Nq-1)`.
+"""
+function Perimeter2D(Nq)
+    ip, jp = Int[], Int[]
```
simonbyrne, line 44: Do we need to allocate an array here?

simonbyrne (approved):

src/Spaces/spectralelement.jl
```diff
@@ -128,6 +160,7 @@ struct SpectralElementSpace2D{
     ghost_dss_weights::D
     internal_surface_geometry::IS
     boundary_surface_geometries::BS
+    perimeter::P
```
simonbyrne, line 163: you don't really need to store this, as it's a singleton object (i.e. it doesn't take any space). But no harm in it.

simonbyrne: if you do add it, you then need to add it to the default constructors (that's the cause of the failing CI)

but a better option might be to just define:
```julia
perimeter(space::SpectralElementSpace2D) = 
Perimeter2D(Quadratures.degrees_of_freedom(space.quadrature_style))
```

simonbyrne (approved):

PR #959, opened 2022-09-22
Add and fix broken tests for domain-point broadcasting

charleskawczynski:
This PR adds and fixes (most) broken tests for domain-surface broadcasting.

I don't think julia has something like `@test_broken_throws`, so my work-around is to use `@test_broken` with `try-catch`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.10.13"
+version = "0.10.14"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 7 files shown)


charleskawczynski:
Alright, I think I fixed a bunch of the tests, I'm not sure if this is optimal, but it looks like it may be the right way. The only case that I didn't fix is the "PlanarSpace" and "3D" space (i.e., we have full functionality inside `bycolumn` loops now, because `PointSpace`s broadcast nicely with everything)

charleskawczynski:
I apparently need one more fix:

```julia
@inline Base.Broadcast.broadcast_shape(
    pointspace::AbstractPointSpace,
    space::AbstractPointSpace,
) = space
@inline Base.Broadcast.broadcast_shape(
    space::AbstractPointSpace,
    pointspace::AbstractPointSpace,
) = space
```
is needed to avoid a method ambiguity. I'll see if I can make the test slightly more complicated to account for this case.

charleskawczynski:
I just tried this out locally in ClimaAtmos, and it's working

charleskawczynski:
@kmdeck, @braghiere , I imagine this may help fix missing support for the land model

kmdeck:
thanks @charleskawczynski! Ill have time to take a closer look early next week.

simonbyrne (approved):

simonbyrne:
https://buildkite.com/clima/climacore-ci/builds/1480#01836bbb-bd85-4fa3-a2b9-85850879899a/131-139

Issue #943, opened 2022-09-14
Unintuitive behavior of `sum`


kmdeck:
## Describe the bug
I am trying to use `sum` to compute integrals of different fields over different spaces. In particular, I am taking `sum(ones(space))` - which should be volume or area or column or "point" integrals. But,
- I get funny results when using `level` of the 3d space vs the horizontal space for the spherical shell. For example, I would expect that \int (ones(horizontal_space)) = 4pi R^2, and that \int ones(level) = 4pi(R+z)^2, where z is the height of the level. 
- I would expect sum(point_field) = 0, but I get a nonzero result.
- I also found that using `by column` to take a column integral over each column of the spherical shell - returning something that lives on the horizontal space - also seemed to give unintuitive results.
## To Reproduce
MWE below
```julia
using ClimaCore
using Test
@testset "Spherical Shell" begin
    FT = Float32
    radius = FT(10.0)
    height =FT(2.0)
    nelements = (10,10)
    npolynomial = 3
    vertdomain = ClimaCore.Domains.IntervalDomain(
        ClimaCore.Geometry.ZPoint(FT(0)),
        ClimaCore.Geometry.ZPoint(FT(height));
        boundary_tags = (:bottom, :top),
    )
    
    vertmesh = ClimaCore.Meshes.IntervalMesh(
        vertdomain,
        ClimaCore.Meshes.Uniform(),
        nelems = nelements[2],
    )
    vert_center_space = ClimaCore.Spaces.CenterFiniteDifferenceSpace(vertmesh)

    horzdomain = ClimaCore.Domains.SphereDomain(radius)
    horzmesh = ClimaCore.Meshes.EquiangularCubedSphere(horzdomain, nelements[1])
    horztopology = ClimaCore.Topologies.Topology2D(horzmesh)
    quad = ClimaCore.Spaces.Quadratures.GLL{npolynomial + 1}()
    horzspace = ClimaCore.Spaces.SpectralElementSpace2D(horztopology, quad)

    hv_center_space = ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace(
        horzspace,
        vert_center_space,
    )
    hv_face_space = ClimaCore.Spaces.FaceExtrudedFiniteDifferenceSpace(hv_center_space)
    # Sum behavior
    ones_volume = ones(hv_center_space)
    ones_horizontal_space = ones(hv_center_space.horizontal_space)
[...]

Issue #748, opened 2022-06-06
Add support for vertical integrals


charleskawczynski:
**Is your feature request related to a problem? Please describe.**

We need vertical integrals in various places, including TurbulenceConvection and ClimaAtmos. The current `sum` is just a volume integral which is equivalent for a single column of area 1, but definitely not what we want to use for things like liquid water path.

Please link to any related issues/PRs.

**Describe the solution you'd like**

Define both
 - `vertical∫` that computes a vertical definite integral
 - `vertical_indefinite∫` that computes a vertical indefinite integral

**Describe alternatives you've considered**

**Additional context**

simonbyrne:
As discussed on slack, I think definite integrals are well-defined: you sum multiplying by the dxdxi[3,3] term, and return a field on the horizontal space.

For indefinite integrals, I think we should return a face-aligned field? I think we also want an option for upward or downward.

simonbyrne:
Duplicate of #693.

Issue #693, opened 2022-04-29
Vertical integrals


simonbyrne:
From @jiahe23:

> basically to compute sum_i (rho_q[i]) / sum_i (rho[i]) where q is the cloud water in the unit of kg/kg and rho is density, i represents the ith level

This should be waited by Jacobians to deal with mesh stretching.

charleskawczynski:
This is what we added in TC:
```julia
function Base.cumsum(field::CC.Fields.FiniteDifferenceField)
    Base.cumsum(parent(CC.Fields.weighted_jacobian(field)) .* vec(field), dims = 1)
end
function Base.cumsum!(fieldout::CC.Fields.FiniteDifferenceField, fieldin::CC.Fields.FiniteDifferenceField)
    Base.cumsum!(fieldout, parent(CC.Fields.weighted_jacobian(fieldin)) .* vec(fieldin), dims = 1)
end
```
We didn't put it in ClimaCore since this is really only valid for `VF` data layouts. Also, I think `vertical_integral` would be a better name.

simonbyrne:
The problem with `weighted_jacobian` in the general case is that it will include the GLL quadrature weights. we may need to rethink how we use weights.

PR #962, opened 2022-09-26
Add support for definite column integrals

charleskawczynski:
This PR adds `column_integral_definite!`, a non-allocating definite integral for columns. The idea is that it will replace the existing (allocating) definite integral [here](https://github.com/CliMA/ClimaAtmos.jl/blob/44cee55def2a51433c5dcdcdae010b7777cd741b/examples/hybrid/sphere/baroclinic_wave_utilities.jl#L167-L186). However, this implementation has several advantages:
 - It's compatible with `bycolumn`, so the computation can occur in parallel
 - It's allocation-free
 - It's _tested_ on single column and sphere configurations

Looking at the flame graph for allocations (using the Julia 1.8 Allocs module with `sample_rate=1`), this is responsible for most of the remaining allocations in ClimaAtmos:
Screen Shot 2022-09-26 at 7 57 05 AM

The interface for the added function captures two cases:

```julia
function column_integral_definite!(col∫field::Field, field::Field)
    bycolumn(axes(field)) do colidx
        column_integral_definite!(col∫field[colidx], field[colidx])
        nothing
    end
    return nothing
end
```
and
```julia
function column_integral_definite!(
    col∫field::PointField,
    field::ColumnField,
)
    @inbounds col∫field[] = column_integral_definite(field)
    return nothing
end
```

A step towards closing #943, #748, [CA#686](https://github.com/CliMA/ClimaAtmos.jl/issues/686).

## A note on an alternative approach
Ideally, we would write this function as `column_integral_definite!(fn, args...)` where we might be able to write a broadcast statement like:

```julia
@. f2D = column_integral_definite(f3D) do z
     f3D.x * z^2
end
```

however, this would require us to define broadcasting between planar and 3D domains, which is not trivial (or maybe not possible) because of ambiguities. The ambiguities arise because (2D, 3D) broadcasting may want different things in different cases, for example:

 - (f2D, f3D) -> f3D: mul full 3D field by planar surface value
 - (f2D, f3D) -> f2D: perform reduction over z-coordinate to yield 2D field

The situation is similar when thinking about what happens when we make views. For example,

```julia
Fields.bycolumn(axes(f3D)) do colidx
     @. f2D[colidx] = column_integral_definite(f3D[colidx]) do z
         f3D[colidx].x * z^2
      end
end
```

Now, we have to define how `DataF` data layouts broadcast with `VF`. Again, we have two cases:

 - (f0D, f1D) -> f1D: mul full 1D field by 0D field
 - (f0D, f1D) -> f0D: perform reduction over z-coordinate to yield 0D field

My vote/preference is to support the first cases (which is partially supported already) and write custom functions (e.g., reductions) that operate on single fields for the second case.

A step towards closing [CA#686](https://github.com/CliMA/ClimaAtmos.jl/issues/686)

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -256,6 +256,14 @@ local_geometry_field(space::AbstractSpace) =
 local_geometry_field(field::Field) = local_geometry_field(axes(field))
 
 
+"""
+    dz_field(field::Field)
+    dz_field(space::AbstractSpace)
+
+A `Field` containing the `Δz` values on the same space as the given field.
+"""
+dz_field(field::Field) = dz_field(axes(field))
+dz_field(space::AbstractSpace) = Field(Spaces.dz_data(space), space)
 
 include("broadcast.jl")
 include("mapreduce.jl")
--- a/src/Operators/integrals.jl
+++ b/src/Operators/integrals.jl
@@ -1,0 +1,50 @@
+#####
+##### Column integrals (definite)
+#####
+
+"""
+    column_integral_definite(field::Field)
+
+The definite vertical column integral of field `field`.
+"""
+column_integral_definite(field::Fields.CenterExtrudedFiniteDifferenceField) =
+    column_integral_definite(field, 1)
+column_integral_definite(field::Fields.FaceExtrudedFiniteDifferenceField) =
+    column_integral_definite(field, PlusHalf(1))
+
+column_integral_definite(field::Fields.Field, one_index::Union{Int, PlusHalf}) =
+    column_integral_definite(similar(level(field, one_index)), field)
+
+"""
+    column_integral_definite!(col∫field::Field, field::Field)
+
+The definite vertical column integral, `col∫field`, of field `field`.
+"""
+function column_integral_definite!(col∫field::Fields.Field, field::Fields.Field)
+    Fields.bycolumn(axes(field)) do colidx
+        column_integral_definite!(col∫field[colidx], field[colidx])
+        nothing
+    end
+    return nothing
+end
+
+function column_integral_definite!(
+    col∫field::Fields.PointField,
+    field::Fields.ColumnField,
+)
+    @inbounds col∫field[] = column_integral_definite(field)
+    return nothing
+end
+
+function column_integral_definite(field::Fields.ColumnField)
+    field_data = Fields.field_values(field)
+    Δz_data = Spaces.dz_data(axes(field))
+    Nv = Spaces.nlevels(axes(field))
+    ∫field = zero(eltype(field))
+    @inbounds for j in 1:Nv
+        ∫field += field_data[j] * Δz_data[j]
+    end
+    return ∫field
+end
+
+# TODO: add support for indefinite integrals
--- a/src/Spaces/finitedifference.jl
+++ b/src/Spaces/finitedifference.jl
@@ -153,6 +153,27 @@ local_geometry_data(space::CenterFiniteDifferenceSpace) =
 local_geometry_data(space::FaceFiniteDifferenceSpace) =
     space.face_local_geometry
 
+
+"""
+    z_component(::Type{<:Goemetry.AbstractPoint})
+
+The index of the z-component of an abstract point
+in an `AxisTensor`.
+"""
+z_component(::Type{<:Geometry.LatLongZPoint}) = 9
+z_component(::Type{<:Geometry.ZPoint}) = 1
+
+"""
+    dz_data(space::AbstractSpace)
+
+A DataLayout containing the `Δz` on a given space `space`.
+"""
+function dz_data(space::AbstractSpace)
+    lg = local_geometry_data(space)
+    data_layout_type = eltype(lg.coordinates)
+    return getproperty(lg.∂x∂ξ.components.data, z_component(data_layout_type))
+end
+
 function left_boundary_name(space::FiniteDifferenceSpace)
     boundaries = Topologies.boundaries(Spaces.topology(space))
     propertynames(boundaries)[1]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.10.14"
+version = "0.10.15"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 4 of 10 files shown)


src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
+#####
```
simonbyrne, line 1: I'd prefer that this go in Operators, rather than Fields

src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
...
+##### Column integrals (definite)
+#####
+
+##### Allocating interface:
+column_∫_definite(field::CenterExtrudedFiniteDifferenceField) =
```
simonbyrne, line 6: I'd prefer an ASCII name
```suggestion
column_integrate_definite(field::CenterExtrudedFiniteDifferenceField) =
```

src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
+#####
+##### Column integrals (definite)
+#####
+
+##### Allocating interface:
```
simonbyrne, line 5: docstrings?

src/Spaces/finitedifference.jl
```diff
@@ -153,6 +153,16 @@ local_geometry_data(space::CenterFiniteDifferenceSpace) =
 local_geometry_data(space::FaceFiniteDifferenceSpace) =
     space.face_local_geometry
 
+
+z_component(::Type{<:Geometry.LatLongZPoint}) = 9
+z_component(::Type{<:Geometry.ZPoint}) = 1
+
+function get_dz(space::AbstractSpace)
```
simonbyrne, line 160: analogous to `coordinates_data`, `local_geometry_data`
```suggestion
function dz_data(space::AbstractSpace)
```

src/Fields/Fields.jl
```diff
@@ -256,9 +256,12 @@ local_geometry_field(space::AbstractSpace) =
 local_geometry_field(field::Field) = local_geometry_field(axes(field))
 
 
+get_dz(field::Field) = get_dz(axes(field))
+get_dz(space::AbstractSpace) = Field(Spaces.get_dz(space), space)
```
simonbyrne, line 260: Functios should be consistent in the object they return: `dz_data` returns a `DataLayout`, and `dz_field` returns a `Field`
```suggestion
dz_field(field::Field) = dz_field(axes(field))
dz_field(space::AbstractSpace) = Field(Spaces.dz_data(space), space)
```

src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
...
+
+function column_∫_definite!(
+    col∫field::Field{<:DataLayouts.DataF},
+    field::Field{<:DataLayouts.VF},
+)
+    @inbounds col∫field[] = column_∫_definite(field)
+    return nothing
+end
+
+function column_∫_definite(field::Field{<:DataLayouts.VF})
```
simonbyrne, line 31:
```suggestion
function column_integrate_definite(field::ColumnField)
```

src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
...
+    bycolumn(axes(field)) do colidx
+        column_∫_definite!(col∫field[colidx], field[colidx])
+        nothing
+    end
+    return nothing
+end
+
+function column_∫_definite!(
+    col∫field::Field{<:DataLayouts.DataF},
+    field::Field{<:DataLayouts.VF},
```
simonbyrne, line 25:
```suggestion
    col∫field::PointField,
    field::ColumnField,
```

src/Fields/integrals.jl
```diff
@@ -0,0 +1,40 @@
...
+    return nothing
+end
+
+function column_∫_definite(field::Field{<:DataLayouts.VF})
+    Δz = Spaces.get_dz(axes(field))
+    ∫field = zero(eltype(field))
+    @inbounds for j in eachindex(parent(field))
+        ∫field += parent(field)[j] * parent(Δz)[j]
+    end
+    return ∫field
```
simonbyrne, line 37: Try to avoid calling `parent`
```suggestion
    field_data = Fields.field_values(field)
    Δz_data = Spaces.dz_data(axes(field))
    Nv = size(parent(Δz_data), 5)
    ∫field = zero(eltype(field))
    @inbounds for j in 1:Nv
        ∫field += field_data[j] * Δz_data[j]
    end
    return ∫field
```

src/Fields/integrals.jl
charleskawczynski: How about `column_integral_definite`?

charleskawczynski: I went with `column_integral_definite`, alternatively I'd be fine with `integrate_column_definite`

simonbyrne: `vertical_integral_definite`?

simonbyrne:
Also cross referencing #693

src/Fields/integrals.jl
simonbyrne: lets keep `column`

src/Fields/integrals.jl
charleskawczynski: This suggestion fails the convergence test

```
  Expression: 2 < sum(abs.(cr)) / length(cr) < 2.01
   Evaluated: 2 < 0.09232257384130037 < 2.01
```

It looks like `Nv = size(parent(Δz_data), 5)` needs to be `Nv = size(parent(Δz_data), 1)`

charleskawczynski: (this is after `bycolumn` is called on the field)

src/Fields/integrals.jl
charleskawczynski: Discussed more, going with `column_integral_definite`

src/Operators/integrals.jl
```diff
@@ -0,0 +1,50 @@
...
+    field::Fields.ColumnField,
+)
+    @inbounds col∫field[] = column_integral_definite(field)
+    return nothing
+end
+
+function column_integral_definite(field::Fields.ColumnField)
+    field_data = Fields.field_values(field)
+    Δz_data = Spaces.dz_data(axes(field))
+    Nv = size(parent(Δz_data), 1)
```
simonbyrne, line 42: Sorry, this should have been
```suggestion
    Nv = size(Δz_data, 5)
```

charleskawczynski: See https://github.com/CliMA/ClimaCore.jl/pull/962#discussion_r981742559

src/Operators/integrals.jl
```diff
@@ -0,0 +1,50 @@
...
+    field::Fields.ColumnField,
+)
+    @inbounds col∫field[] = column_integral_definite(field)
+    return nothing
+end
+
+function column_integral_definite(field::Fields.ColumnField)
+    field_data = Fields.field_values(field)
+    Δz_data = Spaces.dz_data(axes(field))
+    Nv = size(Δz_data, 5)
```
simonbyrne, line 42:
```suggestion
    field_data = Fields.field_values(field)
    Δz_data = Spaces.dz_data(axes(field))
    Nv = Spaces.nlevels(axes(field))
```

Issue #963, opened 2022-09-26
Allocations in StencilCoefs broadcasting


charleskawczynski:
Allocations occur when broadcasting over `StencilCoefs`, this needs to be fixed. See PR #964, which adds a broken test.

charleskawczynski:
Current MWE / debugging script:

```julia
julia --project=test

include("test/Fields/field_opt.jl")

cspace = all_spaces(Float64)[8];
fspace = Spaces.FaceExtrudedFiniteDifferenceSpace(cspace);
bidiag_type = Operators.StencilCoefs{-half, half, NTuple{2, Float64}};
∂ᶠ𝕄ₜ∂ᶜρ = Fields.Field(bidiag_type, fspace);
∂ᶜρₜ∂ᶠ𝕄 = Fields.Field(bidiag_type, cspace);
tridiag_type = Operators.StencilCoefs{-1, 1, NTuple{3, Float64}};
S = Fields.Field(tridiag_type, fspace);

IR = Ref(Operators.StencilCoefs{-1, 1}((zero(Float64), one(Float64), zero(Float64))));

# I think this is the broadcasted object (without the assignment)
bc = Broadcast.broadcasted(-, Broadcast.broadcasted(Operators.ComposeStencils(), ∂ᶠ𝕄ₜ∂ᶜρ, ∂ᶜρₜ∂ᶠ𝕄), IR);

@code_warntype allocs_test_Ref_with_compose!(S, ∂ᶠ𝕄ₜ∂ᶜρ, ∂ᶜρₜ∂ᶠ𝕄)

using Cthulhu
@descend allocs_test_Ref_with_compose!(S, ∂ᶠ𝕄ₜ∂ᶜρ, ∂ᶜρₜ∂ᶠ𝕄)

using JET
@test_opt allocs_test_Ref_with_compose!(S, ∂ᶠ𝕄ₜ∂ᶜρ, ∂ᶜρₜ∂ᶠ𝕄)
```

Issue #923, opened 2022-08-26
Propagate inbounds for apply/compose operators


charleskawczynski:
I tried properly apply `Base.@propagate_inbounds` to `apply_stencil_at_idx` and `compose_stencils_at_idx` in #917 (to elide `@boundscheck` in `getidx`). However, CI times for [this job](https://github.com/CliMA/ClimaCore.jl/blob/208d8d8ade0acabf9a5999ee5d14f9947bd561e3/.buildkite/pipeline.yml#L123-L130) was going from 6 mins to more than 57 mins when applied to `apply_stencil_at_idx`, and basically hung (for all three `implicit_stencil` tests) when applied to `compose_stencils_at_idx`.

We should probably investigate why this is happening. Interestingly, ClimaAtmos ran fine on the #917 branch with both `apply_stencil_at_idx` and `compose_stencils_at_idx`, so it could be that the issue was related to using `--check-bounds=yes`. I'd be fine disabling bounds check in the opt test, but it's probably important that we keep it in the correctness test, so we should dig more into this if there's significant performance implications here. I'll leave this as a later effort for now, but wanted to track latest details if this needs to be revived.

charleskawczynski:
It may be that the ambiguities were exacerbating the compile times from when this was originally tried (cross ref: #961)

PR #975, opened 2022-10-01
Fix some tests, try to fix broken tests

charleskawczynski:
This PR
 - "fixes" the broken allocation tests for broadcasting with `Ref` by moving functions outside of `@testset`
 - Adds a "working" broken allocation test, `Allocations StencilCoefs Ref with ComposeStencils broadcasting`, which reproduces the allocations we see in ClimaAtmos' linear solve. It seems to be an interaction between broadcasting over `StencilCoefs`, `ComposeStencils`, and `Ref`s. Removing the `compose` eliminates the allocations, and eliminating the `Ref(StencilCoefs(...))` eliminates the allocations.

Related issue: #963.

### DIFF
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -402,11 +402,7 @@ end
             domain_surface_bc!(Y.x, ᶜz_surf, ᶜx_surf)
         end
         # Skip spaces incompatible with Fields.bycolumn:
-        (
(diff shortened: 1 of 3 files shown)


charleskawczynski:
I now suspect that these allocations are the result of check-bounds due to not inlining the `ComposeStencils` kernels (issue #923)

PR #976, opened 2022-10-01
Propagate inbounds to apply and compose operators

### LINKED ISSUE #923: Propagate inbounds for apply/compose operators

charleskawczynski:
Try to close #923.

### DIFF
--- a/src/Operators/operator2stencil.jl
+++ b/src/Operators/operator2stencil.jl
@@ -88,8 +88,7 @@ stencil_interior_width(op::Operator2Stencil, args...) =
 boundary_width(op::Operator2Stencil, bc::BoundaryCondition, args...) =
     boundary_width(op.op, bc, args...)
 
-# TODO: find out why using Base.@propagate_inbounds blows up compilation time
-function stencil_interior(
+Base.@propagate_inbounds function stencil_interior(
     ::Operator2Stencil{<:Union{InterpolateF2C, InterpolateC2F}},
     loc,
     idx,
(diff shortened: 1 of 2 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

simonbyrne (approved):

charleskawczynski:
I'll rebase after #978 merges and further update mem requests if needed

simonbyrne:
@charleskawczynski want to rebase now?

charleskawczynski:
Closing until we have a way to measure that these macros actually do anything.

PR #977, opened 2022-10-02
Add `number_of_columns` for computing number of columns

charleskawczynski:
This PR adds a new function, `number_of_columns`, for computing the number of columns in a given field / space.


simonbyrne:
I'm not sure about the name. How about `ncolumns`, as that would be consistent with other names (`nlevels`)?

simonbyrne (approved):

PR #978, opened 2022-10-02
Extend and fix Ref broadcasting with StencilCoefs

### LINKED ISSUE #963: Allocations in StencilCoefs broadcasting

charleskawczynski:
This PR extends and fixes `Ref` broadcasting with StencilCoefs (#963), solution was basically the same as the one applied in #945. Closes #963.

I tried applying only `copyto!`, and only `apply_stencil!`, but that didn't work so I applied to all three methods (which does work).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -118,6 +118,7 @@ steps:
       - "julia -O0 --color=yes --check-bounds=yes --project=test test/Operators/finitedifference/implicit_stencils.jl --float_type Float32"
     agents:
       config: cpu
+      slurm_mem: 10GB
       queue: central
       slurm_ntasks: 1
 
@@ -127,6 +128,7 @@ steps:
       - "julia -O0 --color=yes --check-bounds=yes --project=test test/Operators/finitedifference/implicit_stencils.jl --float_type Float64"
     agents:
       config: cpu
+      slurm_mem: 10GB
       queue: central
       slurm_ntasks: 1
 
@@ -775,6 +777,7 @@ steps:
       - "examples/hybrid/sphere/output/deformation_flow/*"
     agents:
       config: cpu
+      slurm_mem: 20GB
       queue: central
       slurm_ntasks: 1
 
@@ -786,6 +789,7 @@ steps:
       - "examples/hybrid/sphere/output/hadley_circulation/*"
     agents:
       config: cpu
+      slurm_mem: 20GB
       queue: central
       slurm_ntasks: 1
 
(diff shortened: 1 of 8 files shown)


simonbyrne (approved):

charleskawczynski:
I think I need to just fix some of the other tests first, since those seem to be now failing. I'll take a look

charleskawczynski:
Ok, so the tests that were breaking were using a closure pattern that I thought might be useful, but we're not actually using anywhere, so it seems like a reasonable trade. That is, the test that previously passed was something of the form:

```julia
    c∇closure() = @. cz = cx * cy * ∇c(wvec(fy)) * ∇c(wvec(fx)) * cϕ * cψ
    c∇closure()
    p = @allocated begin
        c∇closure()
    end
    @test p == 0 # now converted to @test_broken p == 0
```

To make sure we couldn't keep all the tests working, I
 - moved `if VERSION >= v"1.7.0"` to `runtests.jl` (since all the tests were inside that and it resulted in a very large indent)
 - Removed the closures over `cfield`, `ffield` etc. so that most of the functions could be defined outside `@testset` (see #975) to ensure allocations weren't due to function definitions inside `@testset`

This didn't work, so I'm thinking that we just merge and replace the previously passing test with a broken test (since we don't currently use that closure pattern anyway).

charleskawczynski:
I'll keep the commits separate in case we'd like to revert the second commit.

test/runtests.jl
```diff
@@ -64,7 +64,9 @@ if !Sys.iswindows()
 @safetestset "FD ops - opt" begin @time include("Operators/finitedifference/opt.jl") end
 @safetestset "FD ops - wfact" begin @time include("Operators/finitedifference/wfact.jl") end
 @safetestset "FD ops - linsolve" begin @time include("Operators/finitedifference/linsolve.jl") end
-@safetestset "FD ops - examples" begin @time include("Operators/finitedifference/opt_examples.jl") end
+if VERSION >= v"1.7.0"
+    @safetestset "FD ops - examples" begin @time include("Operators/finitedifference/opt_examples.jl") end
+end
```
simonbyrne, line 69: Our current Project.toml requires julia 1.7, so this is redundant.
```suggestion
@safetestset "FD ops - examples" begin @time include("Operators/finitedifference/opt_examples.jl") end
```

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
I had to increase the mem requests for several jobs, is that alright? cc @simonbyrne

simonbyrne (approved):

charleskawczynski:
Maybe the depot was entangled with the other job that was closing?

Issue #983, opened 2022-10-06
Allocations with broadcasting field vectors, including `fill!(::FieldVector, ::Number)` and `zero(eltype(::FieldVector))`


charleskawczynski:
ClimaCore incurs allocations with:

 - [ ] `Y .= 1` where `Y::FieldVector`
 - [ ] `fill!(::FieldVector, ::Number)` and
 - [ ] `Y .= zero(eltype(Y))` where `Y::FieldVector`

PR with broken tests are pending.

charleskawczynski:
Also, I just realized, #949 was closed because #950 accidentally tested `Field`s and not `FieldVector`s.

charleskawczynski:
Reproducer:

```julia
# These tests require running with `--check-bounds=[auto|no]`
using Test, StaticArrays, IntervalSets
import ClimaCore: ClimaCore, Fields, slab, Domains, Topologies, Meshes, Operators, Spaces, Geometry
function FieldFromNamedTuple(space, nt::NamedTuple)
    cmv(z) = nt
    return cmv.(Fields.coordinate_field(space))
end
include(joinpath(pkgdir(ClimaCore), "test", "Fields", "util_spaces.jl"))
space = first(all_spaces(Float64));
Y = Fields.FieldVector(; c = FieldFromNamedTuple(space, (; x = Float64(0))),
                         f = FieldFromNamedTuple(space, (; x = Float64(0))));
function bc_val!(Y)
    Y .= 1
    nothing
end
bc_val!(Y) # compile first

using Profile
Profile.Allocs.@profile sample_rate=1 bc_val!(Y)
import PProf
PProf.Allocs.pprof(from_c = false)

using Cthulhu
@descend bc_val!(Y)

using JET
bc_val!(Y)
@test_opt bc_val!(Y)
```

PR #984, opened 2022-10-06
Add broken allocation tests for broadcasting with FieldVectors

charleskawczynski:
This PR adds broken allocation tests for broadcasting with `FieldVector`s. Related issue: #983.


charleskawczynski:
Need a minor fix to make sure that this is working correctly.

charleskawczynski:
I was going to see if this was an easy fix, but I couldn't find anything obvious. It looks like Cthulhu complains about `Base.axes(::FieldVector)` (there's a PartialStruct with `Any[]` in the types), but I'm not sure if this is culprit.

PR #993, opened 2022-10-18
Expand dz_field tests, add more coords support

charleskawczynski:
This PR:
 - Expands `dz_field` tests
 - Adds `dz_field` support for more coords types

I've been running into situations in ClimaAtmos when unifying edmf and dycore radiation pieces where `z_component(::Type{<:Geometry.XZPoint}) = 4` is needed. I figured it would be better to just add support across the board.

It'd be good to confirm that I'm interpreting `coordinate_axis` correctly (which was needed for some of the tests).

### DIFF
--- a/src/Spaces/finitedifference.jl
+++ b/src/Spaces/finitedifference.jl
@@ -161,7 +161,12 @@ The index of the z-component of an abstract point
 in an `AxisTensor`.
 """
 z_component(::Type{<:Geometry.LatLongZPoint}) = 9
+z_component(::Type{<:Geometry.Cartesian3Point}) = 1
+z_component(::Type{<:Geometry.Cartesian13Point}) = 4
+z_component(::Type{<:Geometry.Cartesian123Point}) = 9
+z_component(::Type{<:Geometry.XYZPoint}) = 9
 z_component(::Type{<:Geometry.ZPoint}) = 1
+z_component(::Type{<:Geometry.XZPoint}) = 4
 
 """
     dz_data(space::AbstractSpace)
--- a/src/Spaces/finitedifference.jl
+++ b/src/Spaces/finitedifference.jl
@@ -153,30 +153,36 @@ local_geometry_data(space::CenterFiniteDifferenceSpace) =
 local_geometry_data(space::FaceFiniteDifferenceSpace) =
     space.face_local_geometry
 
+Base.@deprecate z_component(::Type{T}) where {T} Δz_metric_component(T) false
 
 """
-    z_component(::Type{<:Goemetry.AbstractPoint})
+    Δz_metric_component(::Type{<:Goemetry.AbstractPoint})
 
 The index of the z-component of an abstract point
 in an `AxisTensor`.
 """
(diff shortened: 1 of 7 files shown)


valeriabarra:
Thanks Charlie for adding/expanding these. I've seen this uses the local geometry info. Does it mean that this uses the generalized coordinates only and we don't need to test this on both uniform and stretched meshes?

charleskawczynski:
> Thanks Charlie for adding/expanding these. I've seen this uses the local geometry info. Does it mean that this uses the generalized coordinates only and we don't need to test this on both uniform and stretched meshes?

Yes, this should work for uniform or stretched meshes AFAIU.

src/Spaces/finitedifference.jl
```diff
@@ -161,7 +161,12 @@ The index of the z-component of an abstract point
 in an `AxisTensor`.
 """
 z_component(::Type{<:Geometry.LatLongZPoint}) = 9
```
simonbyrne, line 163: should this be deltaz component?

charleskawczynski: I'm fine with renaming it if you thing `dz_component` is a better name. It's used in `getproperty(lg.∂x∂ξ.components.data, z_component(data_layout_type))`, so I interpreted that as the `z`-component of `dx` (the multi-dimensional vector)

simonbyrne: `Δz_metric_component`

src/Spaces/finitedifference.jl
```diff
@@ -161,7 +161,12 @@ The index of the z-component of an abstract point
...
+z_component(::Type{<:Geometry.Cartesian3Point}) = 1
+z_component(::Type{<:Geometry.Cartesian13Point}) = 4
+z_component(::Type{<:Geometry.Cartesian123Point}) = 9
+z_component(::Type{<:Geometry.XYZPoint}) = 9
 z_component(::Type{<:Geometry.ZPoint}) = 1
+z_component(::Type{<:Geometry.XZPoint}) = 4
 
 """
     dz_data(space::AbstractSpace)
```
simonbyrne, line 172:
```suggestion
    Δz_data(space::AbstractSpace)
```

simonbyrne: and add a deprecation

charleskawczynski:
Ran into 
```
--------------------------------------------------------------------------
--
  | An ORTE daemon has unexpectedly failed after launch and before
  | communicating back to mpirun. This could be caused by a number
  | of factors, including an inability to create a connection back
  | to mpirun due to a lack of common network interfaces and/or no
  | route found between them. Please check network connectivity
  | (including firewalls and network routing requirements).
  | --------------------------------------------------------------------------
```

charleskawczynski:
Grr, now ran into:

> 🚨 Error: The command was interrupted by a signal

https://buildkite.com/clima/climacore-ci/builds/1542#0183f74f-5646-4e36-9b0e-b99f04ebcc34/333-475

PR #968, opened 2022-09-29
Add invalidations script

charleskawczynski:
This PR adds a script for monitoring invalidations, so that we can incrementally fix egregious method invalidations (which can lead to slow compile times).

PR #999, opened 2022-10-24
Improve show invalidations

charleskawczynski:
This PR is an attempt to fix some of the method invalidations shown in [this comment](https://github.com/CliMA/ClimaCore.jl/pull/968#issuecomment-1287559723).

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -465,7 +465,7 @@ ClimaCore.Fields.truncate_printing_field_types() = true
 =#
 truncate_printing_field_types() = false
 
-function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
+function Base.show(io::IOContext, ::Type{T}) where {T <: Fields.Field}
     if truncate_printing_field_types()
         print(io, truncated_field_type_string(T))
     else
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -287,7 +287,8 @@ ClimaCore.Fields.truncate_printing_field_types() = true
 @testset "Truncated printing" begin
     nt = (; x = Float64(0), y = Float64(0))
     Y = FieldFromNamedTuple(spectral_space_2D(), nt)
-    @test sprint(show, typeof(Y)) == "Field{(:x, :y)} (trunc disp)"
+    @test sprint(show, typeof(Y); context = IOContext(stdout)) ==
+          "Field{(:x, :y)} (trunc disp)"
 end
 ClimaCore.Fields.truncate_printing_field_types() = false
 


charleskawczynski:
```
echo "Invalidations on default branch: 3098 (3098 via deps)" >> $GITHUB_STEP_SUMMARY
echo "This branch: 2302 (2302 via deps)" >> $GITHUB_STEP_SUMMARY
```
🎉

src/Fields/Fields.jl
```diff
@@ -437,11 +437,11 @@ ClimaCore.Fields.truncate_printing_field_types() = true
 =#
 truncate_printing_field_types() = false
 
-function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
```
simonbyrne, line 440: This is probably because `Union{} <: Fields.Field`: I've seen this in other cases.

charleskawczynski:
This fix actually breaks the functionality, here's a MWE:

```julia
struct Baz end
struct Bar{B} end
struct Foo{B} end
bar(::Foo, a) = error("Done")
Base.show(io::IO, f::Foo) = print(io, "JustFoo") # stack trace NOT adjusted
# Base.show(io::IO, ::Type{<:Foo}) = print(io, "JustFoo") # stack trace adjusted
f = Foo{Bar{Baz}}()
bar(f)
```

src/Fields/Fields.jl
```diff
@@ -437,11 +437,11 @@ ClimaCore.Fields.truncate_printing_field_types() = true
...
+function Base.show(io::IO, f::Fields.Field)
```
simonbyrne, line 440: What about
```suggestion
function Base.show(io::IO, T::Type{Fields.Field{V, S}) where {V,S}
```

charleskawczynski: This still results in invalidations 🤔

charleskawczynski: Ah, it looks like we can use `Base.show(io::IOContext, ::Type{<:Fields.Field})`. Is that alright, @simonbyrne?

charleskawczynski: Hmm, there's still invalidations with that, but it's down from 1229 to 149 (at least in the MWE)

charleskawczynski: Ok, and based on the invalidations script, this invalidation site will drop to the bottom of the list.

src/Fields/Fields.jl
```diff
@@ -465,7 +465,7 @@ ClimaCore.Fields.truncate_printing_field_types() = true
 =#
 truncate_printing_field_types() = false
 
-function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
+function Base.show(io::IOContext, ::Type{T}) where {T <: Fields.Field}
```
simonbyrne, line 468: does this work? why not my suggestion above?

charleskawczynski: This works (truncates the stack traces), but invalidates ~150 methods instead of ~1k.

The suggestion above (`Base.show(io::IO, T::Type{Fields.Field{V, S}) where {V,S}`) still invalidates ~1k methods

charleskawczynski: This was all tested with the above MWE (which seems to well reproduce what we're seeing).

simonbyrne: I'd like to figure out a bit more what is being invalidated

charleskawczynski: I think we're invalidating https://github.com/JuliaLang/julia/blob/36034abf26062acad4af9dcec7c4fc53b260dbb4/base/show.jl#L881

simonbyrne:
Honestly, I don't see an easy way to do this without causing a bunch of invalidations.

charleskawczynski:
> Honestly, I don't see an easy way to do this without causing a bunch of invalidations.

I agree, I think it's an inherent "extensibility" problem. From what I've read on SnoopCompile, "fixing" invalidations is focused on fixing egregious ones rather than all of them. Is there a problem with switching to overloading / invalidating a more specific method (with `IOContext` instead of `IO`) so that we don't invalidate as many methods?

charleskawczynski:
Bump!

simonbyrne:
But why `IOContext`? Is that all that is needed?

Since we don't enable this by default, what if we were to use some sort of [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl/) option:
```julia
if @load_preference("truncate field printing", false)
    function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
        ...
    end
end
```

charleskawczynski:
> But why `IOContext`? Is that all that is needed?

My thought was that we should use it because it still (AFAICT) delivers the desired behavior: truncated printing / stack traces of fields, and invalidates fewer methods because `IOContext` is a more specific type than `IO`

> Since we don't enable this by default, what if we were to use some sort of [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl/) option:
> 
> ```julia
> if @load_preference("truncate field printing", false)
>     function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
>         ...
>     end
> end
> ```

I'd be fine with using Preferences.jl instead of the internal method to handle code loading, but I think that this will still invalidate 1k methods if `@load_preference("truncate field printing", false) == true`

simonbyrne:
Okay, if it still works we can just overload `IOContext`. But I think making it a preference is the way to go.

charleskawczynski:
Ah, I just had an idea of what might be happening: I was looking at the [invalidations table](https://buildkite.com/clima/climaatmos-ci/builds/5160#01843520-1d29-4054-9bd7-bc94ecb780e0/140-210) and saw that we also invalidate `truncate_printing_field_types`, which is only defined in ClimaCore. Perhaps invalidating _this_ method triggers the invalidation of `show`.

I'll try out a branch that uses preferences and see how the ClimaAtmos invalidations table looks with that instead.

PR #1001, opened 2022-10-25
Reverse stretched mesh for land model

### LINKED ISSUE #995: Reverse stretched grids for land model
The atmospheric model uses a stretched grid (via exponential or generalized exponential stretching functions) for which the smallest grid spacing happens near the surface (called `dz_bottom` in the code and in ClimaAtmos CLI options) and the largest near the top of the atmosphere (called `dz_top`). 

However, the land model requires the opposite behavior. That is, the smallest grid spacing near the top surface (just at subsurface level) and largest grid spacing at the bottom (deep down towards the Earth core). 

**Describe alternatives you've considered**

- For the `ExponentialStretching` function, which currently has only one parameter (the height scaling factor, H), I propose adding the optional boolean parameter `reverse_mode`, so that the typical user of the land model would specify

  `stretching = Meshes.ExponentialStretching(H; reverse_mode = true)` 

  Note that the boolean parameter (`false` by default) will align with the existing use-case for the atmosphere. So only users who want to reverse this function will have to set this to `true`. 

  Internally, we would just use the opposite `η(ζ)` function, i.e., instead of the current [`η(ζ)`](https://github.com/CliMA/ClimaCore.jl/blob/aa15859549770caf317d7ea786576758cdc26f96/src/Meshes/interval.jl#L147), defined as `η(ζ) = -h * log1p((expm1(-1 / h)) * ζ)`, we will use,  `η(ζ) = h * log1p((expm1(-1 / h)) * ζ)` and this change will be easily enabled by the boolean flag.

- The `GeneralizedExponentialStretching` requires a bit more thinking. 
  The nature of the signature of this function would not change, and I would still add an optional boolean parameter, i.e., 

  `stretching = Meshes.GeneralizedExponentialStretching(dz_bottom, dz_top; reverse_mode)`

  `dz_bottom` will still be specified as the first argument and `dz_top` as the second one. But internally, we have to account for the fact that what was the role of the smallest grid spacing for the atmosphere (`dz_bottom`) is now the role of `dz_top` and vice versa. 

**Additional context**
I added a section to the design-docs, section _3.3.3 Reversed exponential stretching (for land model)_ with the proposed model, but I would appreciate more guidance to know if I got off track (@tapios , @kmdeck , @charleskawczynski )

Reviewers: @kmdeck and @charleskawczynski

valeriabarra:
This will close #995

This PR adds a new command line argument for the creation of a stretched mesh, both with `ExponentialStretching` and `GeneralizedExponentialStretching` stretching types. 

To activate the reverse mode for land models, the user will have to pass a boolean flag (`false` by default, to reflect the existing mode for atmos configurations) to the stretching mode function. The new interface will look like: 

```stretching = Meshes.ExponentialStretching(H; reverse_mode)```

and 

```stretching = Meshes.GeneralizedExponentialStretching(dz_bottom, dz_top; reverse_mode)```

In addition to the new feature and new unit tests, this PR improves existing documentation and tests. In the `ExponentialStretching` unit tests, we were originally only checking the generated mesh faces against a reference profile. I left these checks and added checks without a reference profile, so that we effectively check that the transformation (and its inverse) give us the desired behavior. 

### DIFF
--- a/src/Meshes/interval.jl
+++ b/src/Meshes/interval.jl
@@ -83,7 +83,7 @@ function monotonic_check(faces)
     if !(monotonic_incr || monotonic_decr)
         error(
             string(
-                "Faces in vertical mesh are not increasing monotonically. ",
+                "Faces in vertical mesh are not increasing or decreasing monotonically. ",
                 "We need to have dz_bottom <= z_max / z_elem and dz_top >= z_max / z_elem.",
             ),
         )
@@ -115,7 +115,7 @@ end
 
 
 """
-    ExponentialStretching(H)
+    ExponentialStretching(H::FT)
 
 Apply exponential stretching to the domain when constructing elements. `H` is
 the scale height (a typical atmospheric scale height `H ≈ 7.5`km).
@@ -126,7 +126,13 @@ For an interval ``[z_0,z_1]``, this makes the elements uniformally spaced in
 \\zeta = \\frac{1 - e^{-\\eta/h}}{1-e^{-1/h}},
 ```
 where ``\\eta = \\frac{z - z_0}{z_1-z_0}``, and ``h = \\frac{H}{z_1-z_0}`` is
-the non-dimensional scale height.
+the non-dimensional scale height. If `reverse_mode` is `true`, the smallest
+element is at the top, and the largest at the bottom (this is typical for land
+model configurations).
+
+Then, the user can define a stretched mesh via
+
+    ClimaCore.Meshes.IntervalMesh(interval_domain, ExponentialStretching(H); nelems::Int, reverse_mode = false)
 """
 struct ExponentialStretching{FT} <: StretchingRule
     H::FT
@@ -136,6 +142,7 @@ function IntervalMesh(
     domain::IntervalDomain{CT},
     stretch::ExponentialStretching{FT};
     nelems::Int,
+    reverse_mode::Bool = false,
 ) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
     if nelems < 1
         throw(ArgumentError("`nelems` must be ≥ 1"))
@@ -144,19 +151,36 @@ function IntervalMesh(
     cmax = Geometry.component(domain.coord_max, 1)
     R = cmax - cmin
     h = stretch.H / R
-    η(ζ) = ζ == 1 ? ζ : -h * log1p((expm1(-1 / h)) * ζ)
-    faces =
-        [CT(cmin + R * η(ζ)) for ζ in range(FT(0), FT(1); length = nelems + 1)]
+
+    η(ζ) = ζ == 1 ? ζ : (-h) * log1p((expm1(-1 / h)) * ζ)
+    faces = [
+        CT(reverse_mode ? cmax + R * η(ζ) : cmin + R * η(ζ)) for
+        ζ in range(FT(0), FT(1); length = nelems + 1)
+    ]
+
+    if reverse_mode
+        faces = map(f -> eltype(faces)(-f.z), faces)
+        faces[1] = faces[1] == -cmax ? cmax : faces[1]
+        reverse!(faces)
+    end
     monotonic_check(faces)
     IntervalMesh(domain, faces)
 end
 
 """
-    GeneralizedExponentialStretching(dz_bottom, dz_top)
+    GeneralizedExponentialStretching(dz_bottom::FT, dz_top::FT)
 
 Apply a generalized form of exponential stretching to the domain when constructing elements.
-`dz_bottom` and `dz_top` are target element grid spacings at surface and at the top of the
-vertical column domain (m).
+`dz_bottom` and `dz_top` are target element grid spacings at the bottom and at the top of the
+vertical column domain (m). In typical atmosphere configurations, `dz_bottom` is the smallest
+grid spacing and `dz_top` the largest one. On the other hand, for typical land configurations,
+`dz_bottom` is the largest grid spacing and `dz_top` the smallest one.
+
+For land configurations, use `reverse_mode` = `true` (default value `false`).
+
+Then, the user can define a generalized stretched mesh via
+
+    ClimaCore.Meshes.IntervalMesh(interval_domain, GeneralizedExponentialStretching(dz_bottom, dz_top); nelems::Int, reverse_mode = false)
 """
 struct GeneralizedExponentialStretching{FT} <: StretchingRule
     dz_bottom::FT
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Meshes/interval.jl [hunk 1]
```diff
@@ -136,102 +138,211 @@ function IntervalMesh(
...
-    faces =
-        [CT(cmin + R * η(ζ)) for ζ in range(FT(0), FT(1); length = nelems + 1)]
+    reverse_coeff = reverse_mode ? -1 : 1
+    η(ζ) =
+        ζ == 1 ? reverse_coeff * ζ :
+        reverse_coeff * (-h) * log1p((expm1(-1 / h)) * ζ)
+    faces = [
+        CT(reverse_mode ? cmax + R * η(ζ) : cmin + R * η(ζ)) for
+        ζ in range(FT(0), FT(1); length = nelems + 1)
+    ]
```
charleskawczynski, line 161: Can we get away without these changes?

charleskawczynski (commented):
It might be nice to do a peel offf PR and change the `dz_surface` to `dz_bottom`, first. It's difficult to spot which changes are due to name changes only vs added feature

src/Meshes/interval.jl
```diff
@@ -167,34 +189,54 @@ function IntervalMesh(
...
+    reverse_mode::Bool = false,
 ) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
     if nelems ≤ 1
         throw(ArgumentError("`nelems` must be ≥ 2"))
     end
+
+    if nelems ≤ 1
+        throw(ArgumentError("`nelems` must be ≥ 2"))
+    end
+
```
charleskawczynski, line 201:
```suggestion
```

test/Meshes/interval.jl
```diff
@@ -126,6 +136,47 @@ end
...
+    )
+    mesh = unit_intervalmesh(
+        stretching = Meshes.ExponentialStretching(0.25),
+        nelems = 1,
+        reverse_mode = true,
+    )
+    @test Meshes.coordinates(mesh, 1, 1) == Geometry.ZPoint(-1.0)
+    @test Meshes.coordinates(mesh, 1, 2) ≈ Geometry.ZPoint(0.0)
+    # approx equal to gcm ref profile height fraction given the vert domain
+    # ~7.5 km of 45 km vertical domain extent
```
tapios, line 153: =```suggestion
    # height of 45 km and scale height of ~7.5 km 
```

File: test/Meshes/interval.jl [hunk 4]
```diff
@@ -126,6 +136,47 @@ end
...
+    # we normalize to unit length for easy comparison
+    mesh = unit_intervalmesh(
+        stretching = Meshes.ExponentialStretching(7.5 / 45.0),
```
tapios, line 156: Rather than hard coding a mesh for testing, it's better to test that the grid spacing is equidistant in exp(-z/H), where H is the scale height.

valeriabarra: Oh, I followed the same test pattern that was already there for the non-reversed stretched grid. I am not sure I entirely follow the suggestion though. Can you elaborate a bit more? We have the separate tests for the `ExponentialStretch` which scales with the height scale H

src/Meshes/interval.jl
valeriabarra: Why would you remove this? `GeneralizedExponentialStretching` only works with two arguments that specify the bottom and top spacing so I can see why the original implementer thought about not allowing a mesh with fewer than 2 elements because you would not have a way to specify both bottom and top spacing with fewer than 2 elements

test/Meshes/interval.jl
valeriabarra: Oh, I think I got it now! Basically check that the inverse transformation gives ewuispaced points.

src/Meshes/interval.jl
charleskawczynski: It's a duplicate of lines directly above, it looks like a rebase error

valeriabarra: Ah, thanks for the good catch. Yeah, it must have been from the rebase. I'll remove it and squash

valeriabarra:
Thanks @charleskawczynski . I have now cleaned it up and tests (both existing and improved ones) all pass. If approved, I'll squash commits

src/Meshes/interval.jl
```diff
@@ -167,34 +191,50 @@ function IntervalMesh(
...
     nelems::Int,
+    reverse_mode::Bool = false,
 ) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
     if nelems ≤ 1
         throw(ArgumentError("`nelems` must be ≥ 2"))
     end
+
     dz_bottom, dz_top = stretch.dz_bottom, stretch.dz_top
-    if !(dz_bottom ≤ dz_top)
+    if !(dz_bottom ≤ dz_top) && reverse_mode == false
```
charleskawczynski, line 201:
```suggestion
    if !(dz_bottom ≤ dz_top) && !reverse_mode
```
I think this reads a bit easier

src/Meshes/interval.jl
```diff
@@ -167,34 +191,50 @@ function IntervalMesh(
...
         throw(ArgumentError("dz_bottom must be ≤ dz_top"))
     end
-    # surface coord height value
+
+    if !(dz_bottom ≥ dz_top) && reverse_mode == true
```
charleskawczynski, line 205:
```suggestion
    if !(dz_bottom ≥ dz_top) && reverse_mode
```

src/Meshes/interval.jl
```diff
@@ -167,34 +191,50 @@ function IntervalMesh(
...
     z_top = Geometry.component(domain.coord_max, 1)
+    # but in case of reverse_mode, we temporarily swap them so that the following root solve algorithm does not need to change
+    if reverse_mode
+        z_bottom, z_top = Geometry.component(domain.coord_max, 1),
+        -Geometry.component(domain.coord_min, 1)
+    end
+    # for the same reason, in case of reverse_mode, we swap the roles of dz_bottom (largest for land, smallest for atmos) and dz_top (largest for atmos, smallest for land)
+    if reverse_mode
+        dz_top, dz_bottom = dz_bottom, dz_top
+    end
```
charleskawczynski, line 221:
```suggestion
    # but in case of reverse_mode, we temporarily swap them so that the following root solve algorithm does not need to change
    if reverse_mode
        z_bottom, z_top = Geometry.component(domain.coord_max, 1),
        -Geometry.component(domain.coord_min, 1)
        dz_top, dz_bottom = dz_bottom, dz_top
    end
```

src/Meshes/interval.jl
valeriabarra: It's subjective. For me that I have vision problem the version spelled out is easier to read than the one with the bang that could be easily missed at a first glance. But it's ok. I changed it as suggested in [0bf0a5a](https://github.com/CliMA/ClimaCore.jl/pull/1001/commits/0bf0a5ac9a4bc648a7796b7efd5465fe2d2beebf)

src/Meshes/interval.jl
```diff
@@ -167,34 +191,48 @@ function IntervalMesh(
...
+    # so that the following root solve algorithm does not need to change
+    if reverse_mode
+        z_bottom, z_top = Geometry.component(domain.coord_max, 1),
+        -Geometry.component(domain.coord_min, 1)
+        dz_top, dz_bottom = dz_bottom, dz_top
+    end
 
     # define the inverse σ⁻¹ exponential stretching function
-    exp_stretch(ζ, h) = -h * log(1 - (1 - exp(-1 / h)) * ζ)
+    exp_stretch(ζ, h) = ζ == 1 ? ζ : -h * log(1 - (1 - exp(-1 / h)) * ζ)
```
charleskawczynski, line 222: What is this change needed for? Is this to mirror what's done in the exponential stretching?

src/Meshes/interval.jl
```diff
@@ -167,34 +191,48 @@ function IntervalMesh(
...
     nelems::Int,
+    reverse_mode::Bool = false,
 ) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
     if nelems ≤ 1
         throw(ArgumentError("`nelems` must be ≥ 2"))
     end
+
     dz_bottom, dz_top = stretch.dz_bottom, stretch.dz_top
-    if !(dz_bottom ≤ dz_top)
+    if !(dz_bottom ≤ dz_top) && !reverse_mode
```
charleskawczynski, line 201: Alternatively, we could leave this as

```julia
    if !(dz_bottom ≤ dz_top)
        throw(ArgumentError("dz_bottom must be ≤ dz_top"))
    end
```
and move it to below where `dz_bottom` and `dz_top` are recomputed, that may actually be simpler, what do you think?

src/Meshes/interval.jl
valeriabarra: Yes, because otherwise the corner case of the extremum might have not be exactly computed

charleskawczynski (approved):

PR #1008, opened 2022-11-01
Use Preferences.jl for truncated field printing

charleskawczynski:
This PR attempts to use Preferences.jl instead of using an existing pattern used in a few places, where internally defined methods are expected to be (optionally) overloaded:

```julia
truncate_printing_field_types() = false
enable_threading() = false
allow_mismatched_diagonalized_spaces() = false
assert_exact_transform() = false
allow_mismatched_fd_spaces() = false
```

I'm not super familiar with Preferences.jl, but it may be a better way of handling these global / package options


charleskawczynski:
bors try

src/Fields/Fields.jl
```diff
@@ -459,17 +459,14 @@ long.
...
+import Preferences
+Preferences.@set_preferences!(Pair("TruncateClimaCoreFieldPrinting" => false))
 ```
 =#
-truncate_printing_field_types() = false
-
-function Base.show(io::IO, ::Type{T}) where {T <: Fields.Field}
-    if truncate_printing_field_types()
+import Preferences
+if Preferences.@load_preference("TruncateClimaCoreFieldPrinting", false)
```
simonbyrne, line 467: Preferences are local to the package, so don't need to be qualified
```suggestion
if Preferences.@load_preference("TruncateFieldPrinting", false)
```

test/Fields/field.jl
```diff
@@ -283,13 +283,14 @@ end
 end
 
 # Test truncated field type printing:
-ClimaCore.Fields.truncate_printing_field_types() = true
+import Preferences
+Preferences.@set_preferences!(Pair("TruncateClimaCoreFieldPrinting" => true))
```
simonbyrne, line 287: This probably won't work, as it needs to be set before the module is loaded.

src/Fields/Fields.jl
charleskawczynski: Ah, good to know. That's what I originally had but thought it'd be safer to qualify. I'll change it back, thanks!

test/Fields/field.jl
charleskawczynski: Yeah, I wasn't sure but I thought I'd try. What's a good way to test this, then? Can I add a separate `@safetestset "Recursive" begin @time include("test_truncated_field_printing.jl") end`?

charleskawczynski: Let me try that

charleskawczynski:
bors try

test/Fields/field.jl
simonbyrne: no, it has to be in a whole new session.

simonbyrne: (it will also trigger precompilation)

charleskawczynski:
Julia 1.10 will support depth limited stack printing, so let’s just close this and wait for improvements.

Issue #1020, opened 2022-11-22
Compute geometric area of an element


valeriabarra:
**Is your feature request related to a problem? Please describe.**
The geometric area of an element is needed to implement the bubble correction (Issue #1019)

**Describe the solution you'd like**
To compute the geometric area of an element similar to Tempest Remap's implementation.

- In Tempest Remap, this is done in the function [`CalculateFaceAreas`](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L129), which in turn calls the
    - [`CalculateFaceArea_Concave`](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L2586) function for concave faces, which in turn calls the 
        - [`ConvexifyFace`](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L2639) (which seems to use an external `Triangle` package (+16K LOC!)) and the
    - [`CalculateFaceArea`](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L2620) for convex ones that, in turn, uses
       - [`CalculateFaceAreaQuadratureMethod`](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L2208)
       
Where should all these geometry utilities live? Should we create a separate file in `src/Spaces/`, say `src/Spaces/bubble_utilities.jl`? 

cc: @simonbyrne , @charleskawczynski

valeriabarra:
How to test this? Against a quadrilateral spherical cap (i.e., one patch of a cubed-sphere with 1 horizontal element)?  I need to find resources

simonbyrne:
If you want it to use `SphereDomain`, it will have to live in a later module. The obvious place is `Meshes`, as that is where we break the sphere into quads.

Tests:
 - a panel of the cubed sphere (quadrilateral spherical cap) should be 1/6th of the spherical surface area (by symmetry)
 - the sum of all elements in a panel should equal the total area of the panel

simonbyrne:
[...]

Issue #1019, opened 2022-11-17
Bubble correction


valeriabarra:
## Purpose
The purpose of this SDI is to describe the development plan to add the bubble correction feature. To the best of my current knowledge this feature request is not well documented, so this SDI draft will probably be modified as I learn more about the specs and problem statement. 

## Cost/benefits/risks
Benefits: Bubble correction is supposed to improve the numerical surface area computation of any given (discrete) mesh. A more accurate surface area, in turn, will improve the accuracy of surface fluxes calculation. For low-order geometries (like the one we use) the numerical area of the domain (and of an element) is given by the sum of integration weights. But because of the low-order discretization, the numerical area is not equal to the geometric area. The "epsilon bubble correction" is supposed to fix this discrepancy by modifying the inner weights in each element so that geometric and numerical areas of each element match. 

Risks: there is not sufficient documentation that we could find on the topic, so it is possible that we might have a naive conceptual map and are not fully able to list the details or challenges of the implementation. 

## Producers
@valeriabarra with potential collaborators and reviewers: @simonbyrne and @charleskawczynski 

## Components
Implement the bubble correction algorithm and related tests. Initially, we may want to "augment" the Jacobian so that we keep both the corrected and non-corrected versions for testing/debugging/comparison purposes. 

- First we need to compute the geometric face area to know how far off the numerical one is. TempestRemap has a reference [implementation](https://github.com/ClimateGlobalChange/tempestremap/blob/b622790cc993998dff93e038af38d17b786c98b7/src/GridElements.cpp#L129). Not sure at the moment how many other utilities from the +3K LOC in [GridElements.cpp](https://github.com/ClimateGlobalChange/tempestremap/blob/master/src/GridElements.cpp) are going to be needed as well.

## Inputs
Implementation references in other projects: 

-  [HOMME](https://github.com/E3SM-Project/E3SM/blob/19b51f71af7dc39ba6bb344a8543f7dd5203644d/components/homme/src/share/geometry_mod.F90#L79) 
- [TempestRemap](https://github.com/ClimateGlobalChange/tempestremap/blob/master/src/FiniteElementTools.cpp#L376) 

## Results and deliverables
A description of the key results, deliverables, quality expectations, and performance metrics.

## Task breakdown
A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results.
- [ ] Task 1: Implement a calculation of an element's face geometric area. 
[...]

PR #1033, opened 2022-12-07
Valeria/bubble correction

valeriabarra:
This is a ~~(WIP for now)~~ PR that addresses #1020 and #1019. 

It implements the bubble correction a la TempestRemap and HOMME. 
 

### DIFF
--- a/src/Spaces/spectralelement.jl
+++ b/src/Spaces/spectralelement.jl
@@ -163,11 +163,42 @@ struct SpectralElementSpace2D{
 end
 
 """
-    SpectralElementSpace2D(topology, quadrature_style)
+    SpectralElementSpace2D(topology, quadrature_style; enable_bubble)
 
-Construct a `SpectralElementSpace2D` instance given a `topology` and `quadrature`.
+Construct a `SpectralElementSpace2D` instance given a `topology` and `quadrature`. The
+flag `enable_bubble` enables the `bubble correction` for more accurate element areas.
+
+# Input arguments:
+- topology: Topology2D
+- quadrature_style: QuadratureStyle
+- enable_bubble: Bool
+
+The idea behind the so-called `bubble_correction` is that the numerical area
+of the domain (e.g., the sphere) is given by the sum of nodal integration weights
+times their corresponding Jacobians. However, this discrete sum is not exactly
+equal to the exact geometric area  (4pi*radius^2 for the sphere). To make these equal,
+the "epsilon bubble" approach modifies the inner weights in each element so that
+geometric and numerical areas of each element match.
+
+Let ``\\Delta A^e := A^e_{exact} - A^e_{approx}``, then, in
+the case of linear elements, we correct ``W_{i,j} J^e_{i,j}`` by:
+```math
+\\widehat{W_{i,j} J^e}_{i,j} = W_{i,j} J^e_{i,j} + \\Delta A^e * W_{i,j}  .
+```
+and the case of non linear elements, by
+```math
+\\widehat{W_{i,j} J^e}_{i,j} = W_{i,j} J^e_{i,j} \\left( 1 + \\tilde{A}^e \\right) ,
+```
+where ``\\tilde{A}^e`` is the approximated area given by the sum of the interior nodal integration weights.
+
+Note: This is accurate only for cubed-spheres of the [`Meshes.EquiangularCubedSphere`](@ref) and
+[`Meshes.EquidistantCubedSphere`](@ref) type, not for [`Meshes.ConformalCubedSphere`](@ref).
 """
-function SpectralElementSpace2D(topology, quadrature_style)
+function SpectralElementSpace2D(
+    topology,
+    quadrature_style;
+    enable_bubble = false,
+)
 
     # 1. compute localgeom for local elememts
     # 2. ghost exchange of localgeom
@@ -185,8 +216,6 @@ function SpectralElementSpace2D(topology, quadrature_style)
     ### How to DSS multiple fields?
     # 1. allocate buffers externally
 
-
-
     domain = Topologies.domain(topology)
     if domain isa Domains.SphereDomain
         CoordType3D = Topologies.coordinate_type(topology)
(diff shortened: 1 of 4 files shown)


valeriabarra:
bors try

valeriabarra:
@charleskawczynski I decided to change the assert with a warning and do not perform the bubble correction algorithm at all in that case (similar to TempestRemap's check). This is the only clean way to deal with this in the case of low precision tests. I verified that this works as well in the ClimaAtmos case that was failing w/ Float32 earlier

charleskawczynski:
> @charleskawczynski I decided to change the assert with a warning and do not perform the bubble correction algorithm at all in that case (similar to TempestRemap's check). This is the only clean way to deal with this in the case of low precision tests. I verified that this works as well in the ClimaAtmos case that was failing w/ Float32 earlier

ClimaAtmos uses Float32 by default, are you saying that the bubble correction won't work with Float32? I'm thinking that we may need feedback about this. My impression was that we could use this bubble correction with either precision (given the default precision in ClimaAtmos).

valeriabarra:
> > @charleskawczynski I decided to change the assert with a warning and do not perform the bubble correction algorithm at all in that case (similar to TempestRemap's check). This is the only clean way to deal with this in the case of low precision tests. I verified that this works as well in the ClimaAtmos case that was failing w/ Float32 earlier
> 
> ClimaAtmos uses Float32 by default, are you saying that the bubble correction won't work with Float32? I'm thinking that we may need feedback about this. My impression was that we could use this bubble correction with either precision (given the default precision in ClimaAtmos).

No, I am not saying that. In lower precision test cases, only a few elements (I believe those that are flattest, near the center of the cube panels, not near the corners) won't "qualify" for bubble correction, meaning, it would be pointless and wasteful to apply it to them (this is similar to TempestRemap's [behavior](https://github.com/ClimateGlobalChange/tempestremap/blob/ddee889eacf4635c19abb3468bdc5d884e2c190c/src/FiniteElementTools.cpp#L377), that in that case does not apply the correction at all). This is because in lower precision test cases, the difference between the high-order element area and regular element area is not representable in that precision. So even correcting that, won't have any effect because literally in those cases we can't represent any better than that.

src/Spaces/spectralelement.jl
```diff
@@ -211,20 +242,138 @@ function SpectralElementSpace2D(topology, quadrature_style)
...
+                    u, ∂u∂ξ = compute_local_geometry(
+                        global_geometry,
+                        topology,
+                        elem,
+                        ξ,
+                        AIdx,
+                    )
+                    J = det(Geometry.components(∂u∂ξ))
+                    WJ = J * quad_weights[i] * quad_weights[j]
+                    elem_area += WJ
```
charleskawczynski, line 297:
```suggestion
```

charleskawczynski:
> ...In lower precision test cases, only a few elements (I believe those that are flattest, near the center of the cube panels, not near the corners) won't "qualify" for bubble correction, meaning, it would be pointless and wasteful to apply it to them (this is similar to TempestRemap's [behavior](https://github.com/ClimateGlobalChange/tempestremap/blob/ddee889eacf4635c19abb3468bdc5d884e2c190c/src/FiniteElementTools.cpp#L377), that in that case does not apply the correction at all). This is because in lower precision test cases, the difference between the high-order element area and regular element area is not representable in that precision. So even correcting that, won't have any effect because literally in those cases we can't represent any better than that.

Ah, ok, that makes sense given that these assertions are quite small. Sounds good to me. I've added one more comment about deleting an unused variable, aside from that I think we can rebase / fix the merge conflict and merge. Thank you!

charleskawczynski (approved):

test/Spaces/sphere.jl
```diff
@@ -5,36 +5,51 @@ import ClimaCore: Domains, Topologies, Meshes, Spaces, Geometry, column
...
+        topology = Topologies.Topology2D(context, mesh)
+        quad = Spaces.Quadratures.GLL{Nq}()
+        space = Spaces.SpectralElementSpace2D(topology, quad)
+
+        # surface area
+        @test sum(ones(space)) ≈ FT(4pi * radius^2) rtol = 1e11 * eps(FT)
+
+        enable_bubble = false
+        no_bubble_space =
+            Spaces.SpectralElementSpace2D(topology, quad; enable_bubble)
```
simonbyrne, line 24:
```suggestion
        no_bubble_space =
            Spaces.SpectralElementSpace2D(topology, quad; enable_bubble = false)
```

test/Spaces/sphere.jl
```diff
@@ -5,36 +5,51 @@ import ClimaCore: Domains, Topologies, Meshes, Spaces, Geometry, column
...
+        # Now check that constructor with enable_buble = false falls back on existing behavior
+        @test sum(ones(no_bubble_space)) ≈ FT(4pi * radius^2) rtol =
+            1e11 * eps(FT)
 
-    # surface area
-    @test sum(ones(space)) ≈ 4pi * radius^2 rtol = 1e-3
+        # Now check constructor with bubble enabled
+        enable_bubble = true
+        bubble_space =
+            Spaces.SpectralElementSpace2D(topology, quad; enable_bubble)
```
simonbyrne, line 33:
```suggestion
        bubble_space =
            Spaces.SpectralElementSpace2D(topology, quad; enable_bubble = true)
```

test/Spaces/sphere.jl
```diff
@@ -5,36 +5,51 @@ import ClimaCore: Domains, Topologies, Meshes, Spaces, Geometry, column
...
-    domain = Domains.SphereDomain(radius)
-    mesh = Meshes.EquiangularCubedSphere(domain, ne)
-    topology = Topologies.Topology2D(context, mesh)
-    quad = Spaces.Quadratures.GLL{Nq}()
-    space = Spaces.SpectralElementSpace2D(topology, quad)
+    for FT in (Float64, Float32)
+        context = ClimaComms.SingletonCommsContext()
+        radius = FT(3)
+        ne = 4
+        Nq = 4
```
simonbyrne, line 12: Add a test for Nq = 2 as well?

test/Spaces/sphere.jl
valeriabarra: Sorry, I am just seeing your comments now, after this has been merged.

src/Spaces/spectralelement.jl
```diff
@@ -211,20 +242,137 @@ function SpectralElementSpace2D(topology, quadrature_style)
...
+                            elem,
+                            ξ,
+                            AIdx,
+                        )
+                        J = det(Geometry.components(∂u∂ξ))
+                        WJ = J * quad_weights[i] * quad_weights[j]
+                        # Modify WJ only for interior nodes
+                        if i != 1 && j != 1 && i != Nq && j != Nq
+                            WJ *= 1 + interior_elem_area
+                        end
```
simonbyrne, line 366:
```suggestion
                        # Modify J only for interior nodes
                        if i != 1 && j != 1 && i != Nq && j != Nq
                            J *= 1 + interior_elem_area
                        end
                        WJ = J * quad_weights[i] * quad_weights[j]
```

Issue #1034, opened 2022-12-07
Some Docs improvements


valeriabarra:
Our ClimaCore.jl documentation needs some improvements, including (but possibly not limited to):

- [x] Better landing page with brief description (right now this is empty)
- [x] Inclusion of "Installation instructions" or "How to" guides
- [x] Resolution (to a good extent, but maybe not entirely) of build Warnings (e.g., "missing docstring"
- [x] Add contributing guides, code of conduct?
- [x] Reshuffle some sections
- [x] Add missing `lib/` entries

valeriabarra:
@sriharshakandala this [image](https://github.com/CliMA/ClimaCore.jl/blob/5a99983e5494032f02f0ec0f43a9bfcbb03437ef/docs/src/api.md?plain=1#L118) is not linked properly in the docs. Do you know the correct path to locate/link it properly?

sriharshakandala:
Yes. This is located at `https://github.com/CliMA/ClimaCore.jl/tree/main/docs/src/assets`

valeriabarra:
> Yes. This is located at `https://github.com/CliMA/ClimaCore.jl/tree/main/docs/src/assets`

Okay, thanks. It needs to be in the upper-level directory to be linked correctly. I am going to fix it.

sriharshakandala:
> > Yes. This is located at `https://github.com/CliMA/ClimaCore.jl/tree/main/docs/src/assets`
> 
> Okay, thanks. It needs to be in the upper-level directory to be linked correctly. I am going to fix it.

Thanks.

valeriabarra:
Does anyone know how to fix these types of warnings: 

![image](https://user-images.githubusercontent.com/39932030/206274372-8cb217d7-1d8e-4c27-89b3-24909509fc80.png)

The build log is not very informative and I checked that the docstring is there. `domain` is defined similarly in the `Meshes` module but there are not warnings there

valeriabarra:
> Does anyone know how to fix these types of warnings:
> 
> ![image](https://user-images.githubusercontent.com/39932030/206274372-8cb217d7-1d8e-4c27-89b3-24909509fc80.png)
> 
> The build log is not very informative and I checked that the docstring is there. `domain` is defined similarly in the `Meshes` module but there are not warnings there

Oh, I've just figured it out. It needed to be properly namespaced in the docstring!

valeriabarra:
[...]

PR #1035, opened 2022-12-07
Docs: some improvements

valeriabarra:
This PR provides some doc improvements. Addresses #1034 

It fixes a figure linking problem. The space-filling curve was in the wrong path and was not found in the linking. Now it displays correctly (see screenshot of preview) 

image

And aims to provide solution of other warnings/issues.


valeriabarra:
I think this is in a good shape now (pending my comment above). 

Merging this will close #1034

src/Topologies/Topologies.jl
```diff
@@ -43,9 +44,9 @@ abstract type AbstractDistributedTopology <: AbstractTopology end
 coordinate_type(topology::AbstractTopology) = coordinate_type(domain(topology))
 
 """
-    domain(topology)
+    Topologies.domain(topology::AbstractTopology)
 
-Returns the domain of the `topology` from the underlying `mesh`
+Returns the [`Domains.AbstractDomain`](@ref) of the `topology` from the underlying `mesh`
```
charleskawczynski, line 49:
```suggestion
Returns the domain (a subtype of [`Domains.AbstractDomain`](@ref)) of the `topology` from the underlying `mesh`
```

I might leave this as it was for preciseness, but we can include the ref and type information like this?

docs/src/index.md
```diff
@@ -1 +1,11 @@
 # ClimaCore.jl
+
```
charleskawczynski, line 2: Looks nice, thanks!

docs/src/installation_instructions.md
```diff
@@ -0,0 +1,57 @@
+# Installation
+
```
charleskawczynski, line 2: This page is fine with me

docs/src/lib/ClimaCoreMakie.md
```diff
@@ -0,0 +1,7 @@
+# ClimaCoreMakie.jl
```
charleskawczynski, line 1: LGTM

docs/src/lib/ClimaCorePlots.md
```diff
@@ -0,0 +1,7 @@
+# ClimaCorePlots.jl
```
charleskawczynski, line 1: LGTM

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -2,7 +2,7 @@ import ClimaCore.DataLayouts: IJFH
 
 
 """
-    LinearTempestRemap{T, S, M, C, V}
+    ClimaCoreTempestRemap.LinearMap{T, S, M, C, V}
```
charleskawczynski, line 5: I don't think we need to qualify these methods in the doc string, Documenter already provides a qualified name above the doc string (see the API docs for an example).

The upside is that the name exactly matches, the downside is that all the function doc strings get a bit too wordy. Also, it's not a pattern that we're currently using across this and other repos (it doesn't seem to me to be a common pattern in other Julia packages either)

src/Topologies/Topologies.jl
```diff
@@ -43,9 +44,9 @@ abstract type AbstractDistributedTopology <: AbstractTopology end
 coordinate_type(topology::AbstractTopology) = coordinate_type(domain(topology))
 
 """
-    domain(topology)
+    Topologies.domain(topology::AbstractTopology)
```
charleskawczynski, line 47:
```suggestion
    domain(topology::AbstractTopology)
```

lib/ClimaCoreTempestRemap/src/onlineremap.jl
valeriabarra: This was one of the "missing docrstring" errors that I tried to chase down and saw that this was a frequent issue among other Documenter users and found that qualifying would solve it. It did solve it in this instance

src/Topologies/Topologies.jl
valeriabarra: This was again my attempt at resolving the error (see above [comment](https://github.com/CliMA/ClimaCore.jl/pull/1035#issuecomment-1341645780)). I found some threads that discuss this (e.g., see [this](https://discourse.julialang.org/t/calling-docs-twice-for-the-same-function-in-documenter-jl/37791))

lib/ClimaCoreTempestRemap/src/onlineremap.jl
charleskawczynski: We can do qualifying in the docs API

valeriabarra: Yeah, but then again, it was my attempt, before discovering that actually the docstring had an outdated name! Before, apparently it used to be called `LinearTempestRemap` and then it was later changed to `LinearMap` and people forgot to update the docstring

valeriabarra: So it's probably ok now to remove it altogether. I am going to try

charleskawczynski: We should probably update https://github.com/CliMA/ClimaCore.jl/blob/main/docs/make.jl#L59 to [:example_block, :export] (if that's the syntax), to ensure this doesn't keep popping up.

simonbyrne:
For the topology image, can we plot the whole thing as one line?

sriharshakandala:
I can modify it per our requirements. I just pushed temporary branch with the plotting code at https://github.com/CliMA/GilbertCurves.jl/tree/sk/cubed_sphere_sfc

lib/ClimaCoreTempestRemap/src/onlineremap.jl
valeriabarra: @charleskawczynski , According to [Documenter's docs](https://documenter.juliadocs.org/stable/lib/public/index.html#Documenter.makedocs), the keywords are: `:autodocs_block`, `:cross_references`, `:docs_block`, `:doctest`, `:eval_block`, `:example_block`, `:footnote`, `:linkcheck`, `:meta_block`, `:missing_docs`, `:parse_error`, and `:setup_block`. So there is no `:export` or similar. What did you have in mind?

src/Topologies/Topologies.jl
valeriabarra: @charleskawczynski , I have applied the suggestion, but the problem is still present (see my comment in the PR convo)

simonbyrne:
actually, let me add some numbering

valeriabarra:
Thanks @simonbyrne . I updated the image.

valeriabarra:
> actually, let me add some numbering

Ok, I'll update again

simonbyrne:
```julia
using GilbertCurves, Plots

n = 8

panel_right = gilbertindices((n,n), majdim=1)
panel_up = gilbertindices((n,n), majdim=2)

list = vcat(panel_right,
            [idx + CartesianIndex(n,0) for idx in panel_up],
            [idx + CartesianIndex(n,n) for idx in panel_right],
            [idx + CartesianIndex(2n,n) for idx in panel_up],
            [idx + CartesianIndex(2n,2n) for idx in panel_right],
            [idx + CartesianIndex(3n,2n) for idx in panel_up])


p = plot([(c[1]-0.5)/n for c in list], [(c[2]-0.5)/n for c in list], line_z=1:length(list), legend=false, xaxis=false, yaxis=false, grid=false, alpha=0.5)

plot!(p,
      [1,2,2,3,3,4,4,3,3,2,2,1,1,0,0,1,1,2,2,3],
      [0,0,1,1,2,2,3,3,2,2,1,1,0,0,1,1,2,2,3,3];
      color=:black)

annotate!(p, 0.5,0.5, text("1", 30))
annotate!(p, 1.5,0.5, text("2", 30))
annotate!(p, 1.5,1.5, text("3", 30))
annotate!(p, 2.5,1.5, text("4", 30))
annotate!(p, 2.5,2.5, text("5", 30))
annotate!(p, 3.5,2.5, text("6", 30))

png(p, "panel.png")
```
panel

simonbyrne (approved):

src/Topologies/Topologies.jl
charleskawczynski: Ah, I see, we're importing the function and then re-defining the function stub and documenting that one. That actually seems a bit odd to me, @simonbyrne, should we just remove the duplicate function stub or leave it and document both qualified functions (e.g., `Topologies.domain(topology::AbstractTopology)`).

I suppose it's benign, so I'm fine with either one

charleskawczynski (approved):

simonbyrne:
@valeriabarra feel free to merge whenever you're happy with it

src/Topologies/Topologies.jl
valeriabarra: I ended up removing the duplicated stub docstring. There was no way that by qualifying it, Documenter would be happy with it.

PR #1096, opened 2023-01-11
Add broken gpu broadcasting test

charleskawczynski:
This PR adds a broken gpu broadcasting test, which works on the cpu.


test/Fields/bc_gpu.jl
```diff
@@ -0,0 +1,42 @@
...
+            Geometry.YPoint(FT(1)),
+            periodic = true,
+        ),
+    )
+    mesh = Meshes.RectilinearMesh(domain, 2, 2)
+    topology = Topologies.Topology2D(context, mesh)
+    quad = Spaces.Quadratures.GLL{3}()
+    space = Spaces.SpectralElementSpace2D(topology, quad)
+    (; x) = Fields.coordinate_field(space)
+    ϕ = @. sin(2 * FT(π) * x)
```
simonbyrne, line 31: The issue appears to be the `FT` here: we will need to add a fix like:
https://github.com/JuliaGPU/CUDA.jl/blob/master/src/broadcast.jl#L17-L19

charleskawczynski:
I believe we've fixed this with

```julia
Base.Broadcast.broadcasted(
    fs::AbstractFieldStyle,
    ::Type{T},
    args...,
) where {T} = Base.Broadcast.broadcasted(fs, (x...) -> T(x...), args...)
```

Issue #188, opened 2021-09-15
Optionally interpolate extruded spectral spaces to a regular grid in the horizontal


jakebolewski:
Currently we are not doing that, leads to confusion in interpretation:
https://github.com/CliMA/ClimaCore.jl/blob/main/src/Plots/plots.jl#L126

PR #1107, opened 2023-01-19
Add parallel matrix multiplication for remapping (WIP)

juliasloan25:
We wish to be able to use MPI to parallelize matrix multiplication for remapping. This involves using the existing halo exchange code from ClimaComms for information exchange, then performing local matrix multiplications.

Coupler SDI [#188](https://github.com/CliMA/ClimaCoupler.jl/issues/188)


sriharshakandala:
This general communication approach is reasonable for the preliminary implementation!

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
```diff
@@ -0,0 +1,139 @@
...
+        Spaces.SpectralElementSpace2D(topology_o_singleton, quad_o)
+
+    # generate test data in the Field format (on non-distributed topo. for non-halo exchange version)
+    field_i_singleton = sind.(Fields.coordinate_field(space_i_singleton).long)
+
+    # global exchange (no buffer fill/send here) - convert to superhalo in next implementation
+    root_pid = 0
+    ClimaComms.gather(comms_ctx, parent(field_i_singleton))
+    field_i_singleton =
+        MPI.bcast(field_i_singleton, root_pid, comms_ctx.mpicomm)
```
simonbyrne, line 57: We should define this in ClimaComms: https://github.com/CliMA/ClimaComms.jl/issues/21

docs/Manifest.toml
```diff
@@ -230,7 +236,7 @@ uuid = "c2caaa1d-32ae-4754-ba0d-80e7561362e9"
 version = "0.1.0"
 
 [[deps.ClimaCoreTempestRemap]]
-deps = ["ClimaComms", "ClimaCore", "Dates", "LinearAlgebra", "NCDatasets", "PkgVersion", "TempestRemap_jll", "Test"]
+deps = ["ClimaComms", "ClimaCommsMPI", "ClimaCore", "ClimaCorePlots", "Dates", "LinearAlgebra", "MPI", "NCDatasets", "PkgVersion", "Plots", "TempestRemap_jll", "Test"]
```
LenkaNovak, line 239: It would be good for `ClimaCoreTempestRemap` to be independent of Plots.

lib/ClimaCoreTempestRemap/Project.toml
```diff
@@ -7,10 +7,13 @@ version = "0.3.7"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCommsMPI = "5f86816e-8b66-43b2-912e-75384f99de49"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaCorePlots = "cf7c7e5a-b407-4c48-9047-11a94a308626"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 PkgVersion = "eebad327-c553-4316-9ea0-9fa01ccd7688"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
LenkaNovak, line 16: Can this be removed?

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
...
 end
 
+# This version of this function is used for distributed remapping
+function remap!(
+    target::Fields.Field,
+    R::LinearMap,
+    source::Fields.Field,
+    context::ClimaComms.AbstractCommsContext,
+    target_mesh::Meshes.AbstractMesh{2},
+    space_o,
```
LenkaNovak, line 80: could add the ::Spaces.Space type

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
...
+)
+    pid = ClimaComms.mypid(context)
+    nprocs = ClimaComms.nprocs(context)
+
+    target_array = parent(target)
+    source_array = parent(source)
+
+    # Create mapping from global to local xelement indices (for distributed runs)
```
LenkaNovak, line 88: xelement > element ?

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
...
+    target_elemorder = Meshes.elements(target_mesh)
+    target_elempid = nothing
```
LenkaNovak, line 90: is this needed?

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
```diff
@@ -0,0 +1,139 @@
...
+
+    # use TempestRemap to generate weightson root process, then broadcast to all processes
+    if ClimaComms.iamroot(comms_ctx)
+        R = CCTR.generate_map(space_o_singleton, space_i_singleton)
+
+        # remap without MPI (for testing comparison) and plot solution
+        field_o_singleton = CCTR.remap(R, field_i_singleton)
```
LenkaNovak, line 64: It would be nice to output the time taken for both `remap!` cases to demonstrate the change in performance of this first version.

LenkaNovak (approved):
Looks good, Julia. I only have a few minor comments.

docs/Manifest.toml
juliasloan25: I think it needs Plots to generate the plots [here].(https://github.com/CliMA/ClimaCore.jl/pull/1107/files#diff-1e4f6dd4582cafce906c9edc5c99f075f342be492fd2ab82b278fb07a5efb0b1R125)

lib/ClimaCoreTempestRemap/src/onlineremap.jl
juliasloan25: fixed

lib/ClimaCoreTempestRemap/src/onlineremap.jl
juliasloan25: good catch! I'll remove it

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
juliasloan25: In the distributed case, would we want to output the time info of the process that takes the longest? Since that will determine how long the overall remapping takes?

LenkaNovak: That's one way. Or you could apply a barrier and then report it with the root node.

docs/Manifest.toml
LenkaNovak: Ahh yes, we should comment out these `plot` and `savefig` functions (since they are not actually being inspected by the CI, but they're still useful to have there for debugging).

lib/ClimaCoreTempestRemap/Project.toml
juliasloan25: Yes, will do

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
juliasloan25: I assume the barrier would go right after the `remap!` call. I'm not sure what I would call `@benchmark` on though. In the serial case I was planning on doing something like `time_serial = @benchmark remap!(...)`, but I'm not sure how the barrier would fit in with this in the distributed case.

LenkaNovak: How about just calling `@btime` of a function that includes the `remap!` and the barrier?

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
...
+    # To make IO easier, we enforce the partitioning to be sequential in the element ordering.
+    target_elempid, target_ranges =
+        Topologies.simple_partition(length(target_elemorder), nprocs)
+    @assert issorted(target_elempid)
+
+    target_local_elem_gidx = target_ranges[pid] # gidx = local_elem_gidx[lidx]
+    target_global_elem_lidx = Dict{Int, Int}() # inverse of local_elem_gidx: lidx = global_elem_lidx[gidx]
+    for (lidx, gidx) in enumerate(target_local_elem_gidx)
+        target_global_elem_lidx[gidx] = lidx
+    end
```
simonbyrne, line 100: we should minimize the amount of work we do in `remap!`: either store `target_global_elem_lidx` in the `LinearMap` case, or expose it as part of `Topology2D`

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
     return target
 end
 
+# This version of this function is used for distributed remapping
+function remap!(
+    target::Fields.Field,
+    R::LinearMap,
+    source::Fields.Field,
+    context::ClimaComms.AbstractCommsContext,
```
simonbyrne, line 78: we should be able to access this from the underlying spaces.

lib/ClimaCoreTempestRemap/src/onlineremap.jl
simonbyrne, line 80: these should all be accessible from the `LinearMap` object.

lib/ClimaCoreTempestRemap/src/onlineremap.jl
simonbyrne, line 100: 1. Don't recompute this: use the information: use the `local_elem_gidx` from the topology
2. Store the dictionary in the `remap` object (or store it in the topology)

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,6 +70,83 @@ function remap!(
...
+    target_mesh::Meshes.AbstractMesh{2},
+    space_o,
+)
+    pid = ClimaComms.mypid(context)
+    nprocs = ClimaComms.nprocs(context)
+
+    target_array = parent(target)
+    source_array = parent(source)
```
simonbyrne, line 86: For now you are assuming that `source` is field on a singleton comms context: add an assertion and a comment to make clear this is what is happening.

juliasloan25:
bors try

lib/ClimaCoreTempestRemap/src/onlineremap.jl
juliasloan25: Addressed in updated version of code - `target_global_elem_lidx` is created in `generate_map` using `topology.local_elem_gidx`, and is stored in the returned `LinearMap`

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
```diff
@@ -0,0 +1,150 @@
...
+            1e-20
+
+        # display remapping times for serial and distributed implementations
+        @show "serial remapping min, max times (ns): $(extrema(time_serial.times))"
+        @show "serial remapping median time (ns): $(median(time_serial.times))"
+        @show "serial remapping mean time (ns): $(mean(time_serial.times))"
+
+        @show "distr. remapping min, max times (ns): $(extrema(time_distr.times))"
+        @show "distr. remapping median time (ns): $(median(time_distr.times))"
+        @show "distr. remapping mean time (ns): $(mean(time_distr.times))"
```
simonbyrne, line 148: Use the `@info` macro, e.g.
```
@info "serial remapping times (ns)" min=min(time_serial.times) max=max(time_serial.times) median=...
```

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
```diff
@@ -0,0 +1,150 @@
...
+using ClimaCommsMPI
+using MPI
+using ClimaCore: Geometry, Meshes, Domains, Topologies, Spaces, Fields
+using Test
+using BenchmarkTools
+# use these packages for manual inspection of solutions
+# using Plots
+# using ClimaCorePlots
+
+OUTPUT_DIR = mkpath("CI_OUTPUT_DIR")
```
simonbyrne, line 13:
```suggestion
OUTPUT_DIR = mkpath(get(ENV, "CI_OUTPUT_DIR", tempname()))
```

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -100,72 +162,111 @@ Generate the remapping weights from TempestRemap, returning a `LinearMap` object
...
     out_type = "cgll",
 )
+    if (target_space_distr != nothing)
+        comms_ctx = target_space_distr.topology.context
+    else
+        comms_ctx = target_space.topology.context
+    end
+
+    if ClimaComms.iamroot(comms_ctx)
+        # TODO for next distr. regrid impl. do this on root process then bcast weight data
```
LenkaNovak, line 180: Is this TODO still valid? It looks like this is already done below.

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,18 +73,75 @@ function remap!(
...
+
+        # use unweighted dss to broadcast the so-far unpopulated (redundant) nodes from their unique node counterparts
+        # (we could get rid of this if we added the redundant nodes to the matrix)
+        # using out_type == "cgll"
+        if R.out_type == "cgll"
+            topology = Spaces.topology(axes(target))
+            hspace = Spaces.horizontal_space(axes(target))
+            quadrature_style = hspace.quadrature_style
+
+            # copy remapped target data onto non-distributed space
```
LenkaNovak, line 136: IIRC the target is on the distributed topology? 🤔

LenkaNovak (approved):
Looks good! I only have a couple of doc clarification comments. Thank you, Julia!

lib/ClimaCoreTempestRemap/src/onlineremap.jl
juliasloan25: Ahh good catch, I wasn't expecting to do this in this implementation but ending up changing it. I'll remove the TODO

juliasloan25:
bors try

simonbyrne:
bors try

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
```diff
@@ -0,0 +1,149 @@
...
+    end
+
+    # setup and apply the remap using distributed approach
+    field_o_distr = Fields.zeros(space_o_distr)
+
+    # apply the remapping to field_i_singleton and store the result in field_o_distr
+    # time the distributed remapping (can't save result when interpolating)
+    time_distr =
+        @benchmark (f -> CCTR.remap!($field_o_distr, $R, $field_i_singleton);
+        ClimaComms.barrier($comms_ctx))
```
simonbyrne, line 83: this doesn't look right:

1. `f -> CCTR.remap!($field_o_distr, $R, $field_i_singleton)` is just defining an anonymous function, it isn't actually calling it
2. I don't think it's a good idea to `@benchmark` anything that is calling MPI code, as different processes might run different numbers of times.

simonbyrne:
bors try

simonbyrne:
bors try

LenkaNovak:
Is this good to go now?

lib/ClimaCoreTempestRemap/test/mpi_tests/online_remap.jl
juliasloan25: Is there a better way to quantitatively compare the performance of the MPI vs. serial implementations?

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -68,18 +73,73 @@ function remap!(
...
+                x -> x[1],
+                (
+                    view(R.source_idxs[1], n),
+                    view(R.source_idxs[2], n),
+                    view(R.source_idxs[3], n),
+                ),
+            )
+
+            # choose only the target inds local to this process (skip inds from other processes)
+            t_gidx3 = view(R.target_idxs[3], n)[1]
```
LenkaNovak, line 112: I wonder if we could be a bit more verbose here. Maybe `target_elem_gidx` or something like that?

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -100,72 +160,110 @@ Generate the remapping weights from TempestRemap, returning a `LinearMap` object
...
+    else
+        weights = nothing
+        source_unique_idxs = nothing
+        target_unique_idxs = nothing
+        col_indices = nothing
+        row_indices = nothing
+    end
+
+    if !(comms_ctx isa ClimaComms.SingletonCommsContext)
+        root_pid = 0
```
LenkaNovak, line 247: Would a barrier make sense here? I guess this is only called during init, so there shouldn't be a huge performance hit.

LenkaNovak (approved):
Hi Julia, thanks for doing this! I only had a couple of minor comments. Then, if the visual checks are still reasonable, then feel free to merge.

lib/ClimaCoreTempestRemap/src/onlineremap.jl
juliasloan25: Yeah, I think this is a good idea to make sure all processes have the correct information before creating the `LinearMap`.

juliasloan25:
> Hi Julia, thanks for doing this! I only had a couple of minor comments. Then, if the visual checks are still reasonable, then feel free to merge.

The plots still look good after adding your review comments, so I'll go ahead and merge.

juliasloan25:
bors try

juliasloan25:
bors try

PR #1113, opened 2023-01-31
Fix memory issues

### LINKED ISSUE #1064: Simple integration tests/examples cannot be run locally for excessive memory usage
## Describe the bug
I tried to run some integration tests/examples locally on my machine, that I used to run frequently just fine until some time ago (some in September). But now these examples get killed on my machine for running out-of-memory. 

## To Reproduce
From `main`, try to run locally `examples/hybrid/sphere/deformation_flow.jl` or `examples/hybrid/sphere/hadley_circulation.jl`

```julia
julia --project=examples examples/hybrid/sphere/deformation_flow.jl
```

<details><summary>Project</summary>
<p>
Using the `examples` project.

</p>
</details>

## System details

Any relevant system information:
- Julia version: 1.8.3
- operating system: Ubuntu 20.04
- system memory: 16 GB

## Related issues / PRs

I tried to go back in commit history to see when this started to appear, and up until  `5273149` (PR #972  - merged on October 4th), I can get them to run. But the following merged commit, i.e., `1151fca`  (PR #978 - merged same date, later on October 4th) is the one that kills it. So the latter PR seems to be the culprit in the incremented memory usage.

dennisYatunin:
This PR reduces ClimaCore memory usage due to excessive code inlining, and it closes #1064.

Specifically, it

- Removes the `@inline` annotations to `Base.copyto!` that were added in #945 and #978.
- Removes all `slurm_mem` increases from the CI pipeline, so that the default value of 8GB applies to every task.
- Extends broadcasting over `Field`s and `DataLayout`s so that it can handle single-element `Tuple`s. In theory, broadcasting with a single-element `Tuple` should be equivalent to broadcasting with a `Ref`. However, since a `Ref` is mutable, it looks like the compiler needs to perform runtime allocations when using a `Ref`, unless we force it to inline the majority of the broadcasting machinery.
- Replaces every `Ref` in `test/Fields/field_opt.jl` with a single-element `Tuple`, so that the runtime allocation unit tests can pass.
- Replaces two of the `@test_broken`s in `test/Operators/finitedifference/opt_examples.jl` with `@test`s, since those tests are now passing, even though they don't appear to have anything to do with `Ref`s.

Note: It is still necessary to go through the rest of our code (both in ClimaCore and elsewhere) and replace every `Ref` in a `Field`/`DataLayout` broadcast with a single-element `Tuple`. Now that the `@inline` annotations have been removed, broadcasting with a `Ref` will trigger runtime allocations, which will pose a problem for our long runs.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,8 +46,6 @@ steps:
         key: "cpu_unittests"
         command:
           - "julia --color=yes --check-bounds=yes --project=test test/runtests.jl"
-        agents:
-          slurm_mem: 64GB
 
       - label: ":computer: unit tests expensive"
         key: "cpu_unittests_expensive"
@@ -55,15 +53,11 @@ steps:
           - "julia --color=yes --check-bounds=yes --project=test test/runtests_expensive.jl"
         env:
           OMPI_MCA_rmaps_base_oversubscribe: 1
-        agents:
-          slurm_mem: 64GB
 
       - label: ":computer: Field broadcast performance"
         key: "cpu_field_perf"
         command:
           - "julia --color=yes --project=test test/Fields/field_opt.jl"
-        agents:
-          slurm_mem: 20GB
 
       - label: ":flower_playing_cards: unit tests"
         key: "gpu_unittests"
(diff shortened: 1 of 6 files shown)


simonbyrne:
bors try

dennisYatunin:
bors try

simonbyrne (approved):

charleskawczynski:
I think it would have been better if we had added new tests in this PR, and converted any passing tests to failing tests, rather than change existing tests, which makes understanding the changes more complicated.

PR #1116, opened 2023-02-01
Update compat for ClimaCoreTempestRemap

sriharshakandala:

### DIFF
--- a/lib/ClimaCoreTempestRemap/Project.toml
+++ b/lib/ClimaCoreTempestRemap/Project.toml
@@ -15,7 +15,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 ClimaComms = "0.3.1"
-ClimaCore = "0.10"
+ClimaCore = "0.10.25"
 NCDatasets = "0.11, 0.12"
 PkgVersion = "0.1, 0.2"
 TempestRemap_jll = "2.1.2"


lib/ClimaCoreTempestRemap/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ClimaCoreTempestRemap"
 uuid = "d934ef94-cdd4-4710-83d6-720549644b70"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.3.7"
+version = "0.3.8"
```
simonbyrne, line 5: We never actually tagged 0.3.7

simonbyrne: So you don't need to update this

sriharshakandala: I seems to be tagged here: https://github.com/CliMA/ClimaCore.jl/commit/e5a2ec9c89f835198c25e44fe8c578c2f439b427#diff-1fa6758b8a92bc430e50639e6151ba4f28598df076aee726a46f8c821e86acce
Not sure why this wasn't in the registry!

simonbyrne: looks like registrator didn't pick it up

Just up the min version, then we can tag it

PR #1121, opened 2023-02-06
Rename dss2! to dss!

sriharshakandala:
Rename dss2 functions to dss

### DIFF
--- a/src/Spaces/dss_transform.jl
+++ b/src/Spaces/dss_transform.jl
@@ -196,6 +196,13 @@ function weighted_dss_start! end
 function weighted_dss_internal! end
 function weighted_dss_ghost! end
 
+# for backward compatibility
+function weighted_dss2! end
+function weighted_dss_start2! end
+function weighted_dss_internal2! end
+function weighted_dss_ghost2! end
+function dss2! end
+
 # helper functions for DSS2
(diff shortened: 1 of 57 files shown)


sriharshakandala:
bors try

src/Spaces/dss_transform.jl
```diff
@@ -196,6 +196,13 @@ function weighted_dss_start! end
 function weighted_dss_internal! end
 function weighted_dss_ghost! end
 
+# for backward compatibility
+function weighted_dss2! end
+function weighted_dss_start2! end
+function weighted_dss_internal2! end
+function weighted_dss_ghost2! end
+function dss2! end
```
simonbyrne, line 204: where are these methods defined?

sriharshakandala: These methods are defined in `src/Fields/Fields.jl`. These are just used for backward compatibility and will be eventually removed.

simonbyrne: But it looks like not all the same methods are defined as before?

sriharshakandala: Currently, only dss functions defined in the fields are currently called from external packages. So, backward compatible definitions are only added for these functions. The remaining functions are only used internally!

PR #1129, opened 2023-02-24
cuda mapreduce

### LINKED ISSUE #1118: GPU reductions over fields
Cannot call `sum`,`norm`, etc on GPU-based fields.

Need `mapreduce` implementation for GPU-based `Field`s.

- Nvidia slides: https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf
- CUDA.jl `mapreduce`: https://github.com/JuliaGPU/CUDA.jl/blob/8cd13d63ee02a4290193ba97a96927ae3672a121/src/mapreduce.jl

**Describe alternatives you've considered**
We can't just fall back on CuArrays implementation unless we're on a scalar field (even then, `sum` etc won't take into account quadrature weights).

**Additional context**
Used in a lot of tests, and some OrdinaryDiffEq timesteppers call `norm`

sriharshakandala:
Fixes #1118

### DIFF
--- a/src/Fields/mapreduce_cuda.jl
+++ b/src/Fields/mapreduce_cuda.jl
@@ -1,0 +1,231 @@
+
+Base.sum(
+    field::Union{Field, Base.Broadcast.Broadcasted{<:FieldStyle}},
+    ::ClimaComms.CUDA,
+) = mapreduce_cuda(identity, +, field, weighting = true) #TODO: distributed support to be added
+
+Base.sum(fn, field::Field, ::ClimaComms.CUDA) =
+    mapreduce_cuda(fn, +, field, weighting = true) #TODO: distributed support to be added
+
+Base.maximum(fn, field::Field, ::ClimaComms.CUDA) =
+    mapreduce_cuda(fn, max, field) #TODO: distributed support to be added
+
+Base.maximum(field::Field, ::ClimaComms.CUDA) =
+    mapreduce_cuda(identity, max, field) #TODO: distributed support to be added
+
+Base.minimum(fn, field::Field, ::ClimaComms.CUDA) =
+    mapreduce_cuda(fn, min, field) #TODO: distributed support to be added
+
+Base.minimum(field::Field, ::ClimaComms.CUDA) =
+    mapreduce_cuda(identity, min, field) #TODO: distributed support to be added
+
+Statistics.mean(
+    field::Union{Field, Base.Broadcast.Broadcasted{<:FieldStyle}},
+    ::ClimaComms.CUDA,
+) = Base.sum(field) ./ Spaces.local_area(axes(field)) #TODO: distributed support to be added
+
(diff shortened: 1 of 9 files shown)


src/Fields/mapreduce_cuda.jl
```diff
@@ -0,0 +1,236 @@
...
+    n_ops_on_load = cld(nitems, nthreads) == 1 ? 0 : 7
+    effective_blksize = nthreads * (n_ops_on_load + 1)
+    nblocks = cld(nitems, effective_blksize)
+
+    reduce_cuda = CuArray{T}(undef, nblocks, Nf)
+    reduce_cuda_fld = CuArray{T}(undef, Nf)
+    shmemsize = nthreads
+    # place each field on a different block
+    @cuda threads = (nthreads) blocks = (nblocks, Nf) shmem =
+        shmemsize * sizeof(T) mapreduce_cuda_kernel!(
```
simonbyrne, line 63: I think `shmem` is only for dynamic shared memory: we should be able to use static here.
```suggestion
    @cuda threads = (nthreads) blocks = (nblocks, Nf) mapreduce_cuda_kernel!(
```

simonbyrne:
For broadcasting, I think we need to define a method like
https://github.com/CliMA/ClimaCore.jl/blob/e361847ef0a9752378d8986b2d036ea399509738/src/DataLayouts/broadcast.jl#L311-L326
`where A <: CuArray` to call `mapreduce_cuda`

src/Fields/mapreduce_cuda.jl
```diff
@@ -0,0 +1,236 @@
...
+    if nblocks > 1
+        nthreads = min(32, nblocks)
+        shmemsize = nthreads
+        @cuda threads = (nthreads) blocks = (Nf) shmem = shmemsize * sizeof(T) reduce_cuda_blocks_kernel!(
+            reduce_cuda_fld,
+            reduce_cuda,
+            op,
+            Val(shmemsize),
+        )
+    end
```
simonbyrne, line 83: if `nblocks == 1` then `reduce_cuda_fld` will be uninitialized.

src/Fields/mapreduce_cuda.jl
simonbyrne: Or we use dynamic shared memory and don't `Val`-specialize it?

src/Fields/mapreduce_cuda.jl
```diff
@@ -0,0 +1,236 @@
...
+    @cuda threads = (nthreads) blocks = (nblocks, Nf) shmem =
+        shmemsize * sizeof(T) mapreduce_cuda_kernel!(
+        reduce_cuda,
+        f,
+        op,
+        pdata,
+        pwt,
+        weighting,
+        Val(shmemsize),
+        Val(n_ops_on_load),
```
simonbyrne, line 71: Does this need to be `Val`-specialized?

simonbyrne (approved):

PR #1178, opened 2023-04-09
Mv bubble warning outside loop

charleskawczynski:
This PR moves the bubble warning to outside the element loop, so that we don't bloat the log with the same repeated warning.

### DIFF
--- a/src/Spaces/spectralelement.jl
+++ b/src/Spaces/spectralelement.jl
@@ -275,6 +275,7 @@ function SpectralElementSpace2D(
 
     local_geometry = DataLayouts.IJFH{LG, Nq}(Array{FT}, nlelems)
     ghost_geometry = DataLayouts.IJFH{LG, Nq}(Array{FT}, ngelems)
+    bubble_warn = false
 
     quad_points, quad_weights =
         Quadratures.quadrature_points(FT, quadrature_style)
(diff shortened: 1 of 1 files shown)


valeriabarra:
I've just come across this (now merged) PR. This the error message is not entirely clear, since it refers to the element (because it used to be inside an element loop). I agree that it is a good idea to move the warning outside, but we should update/clarify the error message specifying that there was one (or more) element for which the bubble correction was not performed. The way it currently reads, the second part of the preposition "The bubble correction will not be performed." sounds like the bubble correction was not performed altogether, which is not true.

charleskawczynski:
Hi @valeriabarra. Yeah, I didn't change the error message because I figured that this change was easy enough to make, non-controversial and separate from improving the language of the error message itself. I only removed the duplicate error messages (see [here](https://buildkite.com/clima/climaatmos-ci/builds/8929#018776d3-9688-44f5-9792-0a556457fc79/140-177) for example) because it was bloating the log.

I agree that the message itself can be improved.

simonbyrne:
Since it is the default, can we just disable it altogether? Why would a user care?

valeriabarra:
> Since it is the default, can we just disable the warning altogether? Why would a user care?

We can. I think it was suggested when I was developing this. We certainly can remove it if not deemed useful.

PR #1189, opened 2023-04-14
Add more inbounds and inlining

charleskawczynski:
This PR applies more `@inbounds` and inlining to
 - some of the spectral element operators (written for the cpu)
 - some of the dss kernels

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -477,7 +477,7 @@ Divergence{()}(space) = Divergence{operator_axes(space)}()
 operator_return_eltype(op::Divergence{I}, ::Type{S}) where {I, S} =
     RecursiveApply.rmaptype(Geometry.divergence_result_type, S)
 
-function apply_operator(op::Divergence{(1,)}, space, slabidx, arg)
+Base.@propagate_inbounds function apply_operator(op::Divergence{(1,)}, space, slabidx, arg)
     FT = Spaces.undertype(space)
     QS = Spaces.quadrature_style(space)
     Nq = Quadratures.degrees_of_freedom(QS)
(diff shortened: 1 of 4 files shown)


charleskawczynski:
bors try

src/Operators/spectralelement.jl
```diff
@@ -1355,7 +1405,7 @@ function apply_operator(op::WeakCurl{(1, 2)}, space, slabidx, arg)
     else
         error("invalid return type")
     end
-    for j in 1:Nq, i in 1:Nq
+    @inbounds for j in 1:Nq, i in 1:Nq
```
simonbyrne, line 1408: you don't need `@inbounds` here if you use `@propagate_inbounds`

charleskawczynski: Ah, interesting, I thought `@inbounds` was needed any time a new index was introduced. Good catch!

charleskawczynski:
Interestingly, this doesn’t seem to have much of an impact. Any differences could easily be variation.

charleskawczynski:
The CPU vector dss is the slowest of the spectral kernels, and using IntrospectionTools along with `julia --check-bounds=no`:

```julia
using Revise; using ClimaCore
include(joinpath(pkgdir(ClimaCore), "test", "Operators", "spectralelement", "benchmark_utils.jl"))
include(joinpath(pkgdir(ClimaCore), "test", "Operators", "spectralelement", "benchmark_kernels.jl"))
args = setup_kernel_args(["--device", "CPU"]);
kernel_vector_dss!(args)
using IntrospectionTools
@code_summary kernel_vector_dss!(args)
```
shows that we're currently  generating the same native code as if we used `julia --check-bounds=no`, so these changes are not needed.

charleskawczynski:
Actually, the allocation job shows that this fixes some allocation sites. So perhaps we should reconsider.

charleskawczynski:
Superseded by #1338

PR #1192, opened 2023-04-15
Distr. regridding - local indices in LinearMap

### LINKED ISSUE #1191: Distributed regridding - store local indices in LinearMap
We want to simplify the implementation of `remap!` by converting unique indices (TR convention) to local indices in `generate_map` and storing then in the `LinearMap` struct. This conversion is currently being done in the remap! function, but the code will be cleaner and a bit faster if we do this just once during setup of the `LinearMap`.

Part of ClimaCoupler SDI [#188](https://github.com/CliMA/ClimaCoupler.jl/issues/188).
PR #1192.

We currently convert the unique indices (`target_idxs`) using `target_global_elem_lidx` in `remap!`. We will move this conversion to `generate_map`, remove the unique indices ('target_idxs') from `LinearMap`, store the local indices (`target_local_idxs`) in the `LinearMap` struct, and use these indices directly when performing the multiplication in `remap!`.

`source_global_elem_lidx` is not yet implemented for the source indices, but we will add that as part of this issue as well. We will use it to convert source indices analogously to the target indices, described above.

**Potential pitfall**
If we just store the local element indices in `source_local_idxs` and in `target_local_idxs`, their lengths won't agree with that of `weights`.
* Can we use either `target_global_elem_lidx` or `source_global_elem_lidx` to choose only the local elements of the weights returned by TempestRemap and store those in `LinearMap`?
* We may need to wait until the next PR (when we use `MPI.scatter` instead of `MPI.bcast`) to store only the local values
* For now, I'm storing an value of 0 at non-local indices in `source_local_idxs` and `target_local_idxs`

**Additional context**
To keep backwards compatibility with serial spaces, we should return the unique indices unchanged in the case of a serial space. This is important particularly for the source indices, since the source data is still generated on a serial space at this point. This shouldn't be difficult since `target_global_elem_lidx` and `source_global_elem_lidx` are identity mappings for serial spaces, but we should test this feature.

juliasloan25:
This PR stores local target indices in `LinearMap` instead of the unique indices used by TempestRemap. This will make the `remap!` function cleaner and isolate these internals to `generate_map`.

Note: This function includes one workaround which is not ideal but will be removed in the next implementation. On each process, the source indices, target indices, and weight vectors stored in `LinearMap` have to have the same length so that the values align in the correct positions to perform the multiplication in `remap!`. However, if each process only stored its local source/target indices, these vectors would be smaller than the weights. To get around this, I set the entire target index vector to 0 on each process and only filled in the correct values for local indices. In the next implementation, we will instead store only the local values of the indices and weights on each process.

Note 2: This implementation only works for remappings where the partitions of the source and target spaces are collocated. For non-aligned meshes, we'll need to implement the super-halo exchange for source data to get correct results.

This closes #1191.

## Quality assurance
Distribued remapping is compared against serial remapping (i.e. the previous implementation), as well as the analytical solution. The distributed and serial solutions are quantitatively found to have error of less than 10^-20, and are also compared visually in the following plots:

regrid_source_data
regrid_target_data_mpi
regrid_target_data_serial
regrid_target_data_analytical

### DIFF
--- a/lib/ClimaCoreTempestRemap/src/onlineremap.jl
+++ b/lib/ClimaCoreTempestRemap/src/onlineremap.jl
@@ -5,33 +5,30 @@ using MPI
 
 
 """
-    LinearMap{T, S, M, C, V, M}
+    LinearMap{S, T, W, I, V}
 
 stores information on the TempestRemap map and the source and target data:
 
 where:
  - `source_space` and `target_space` are ClimaCore's 2D spaces.
- - `weights` is a vector of remapping weights. (length = number of overlap-mesh nodes)
- - `source_idxs` a 3-element Tuple with 3 index vectors, representing (i,j,elem) indices on the source mesh. (length of each index vector = number of overlap-mesh nodes)
- - `target_idxs` is the same as `source_idxs` but for the target mesh.
- - `col_indices` are the source column indices from TempestRemap. (length = number of overlap-mesh nodes)
+ - `weights` is a vector of remapping weights. (length = number of overlap-mesh nodes).
+ - `source_local_idxs` a 3-element Tuple with 3 index vectors, representing local (i,j,elem) indices on the source mesh. (length of each index vector = number of overlap-mesh nodes)
+ - `target_local_idxs` is the same as `source_local_idxs` but for the target mesh.
  - `row_indices` are the target row indices from TempestRemap. (length = number of overlap-mesh nodes)
- - `out_type` string that defines the output type
- - `target_global_elem_lidx` is a mapping from global to local indices on the target space
+ - `out_type` string that defines the output type.
 
 """
-struct LinearMap{S, T, W, I, V, M} # make consistent with / move to regridding.jl
+struct LinearMap{S, T, W, I, V} # make consistent with / move to regridding.jl
     source_space::S
     target_space::T
     weights::W
-    source_idxs::I
-    target_idxs::I
-    col_indices::V
+    source_local_idxs::I
+    target_local_idxs::I
     row_indices::V
     out_type::String
-    target_global_elem_lidx::M
 end
 
+# This version of this function is used for serial remapping
 function remap!(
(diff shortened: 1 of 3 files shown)


simonbyrne (approved):

lib/ClimaCoreTempestRemap/src/onlineremap.jl
```diff
@@ -97,34 +93,36 @@ function remap!(target::Fields.Field, R::LinearMap, source::Fields.Field)
...
-            if !(target_elem_gidx in keys(R.target_global_elem_lidx))
-                continue
-            else
-                # get global i, j inds but local elem index e
-                it = view(R.target_idxs[1], n)[1]
-                jt = view(R.target_idxs[2], n)[1]
-                et = R.target_global_elem_lidx[target_elem_gidx]
+            # extract local target indices
+            it, jt, et = map(
+                x -> x[1],
```
simonbyrne, line 108: What is this for?

juliasloan25: This is just extracting the local indices on the target space so that we can index into the target data array at the location corresponding to the current weight.

juliasloan25: `view(source_local_idxs[i], n)` returns a `view(::Vector{Int64}, 1)`, so we have to index into it to actually extract the value, which is what the `map` does. We could instead just do `source_local_idxs[1][n]` and remove the `map` and `view`s. I inherited this way of doing it from the previous serial implementation, and I wasn't sure if it improves the performance or memory usage so I kept it even though it's harder to read.

juliasloan25:
bors try

juliasloan25:
bors try

juliasloan25:
bors try

juliasloan25:
approved by @LenkaNovak offline 4/24/23

PR #1194, opened 2023-04-18
update dependencies, remove CUDA_PATH var

simonbyrne:
This fixes bug in CUDA_Runtime_Discovery.jl


simonbyrne:
@charleskawczynski the benchmarks are failing: I think we need to rethink this approach.

charleskawczynski:
@simonbyrne, is it okay to just increase the buffer?

charleskawczynski:
@simonbyrne, I think it'll stagnate after we update it a few times. I know it's not perfect, but I'm not sure there is a perfect way to do this to ensure we catch if performance does degrade

PR #1198, opened 2023-04-19
Add and fix JET tests for DSS

### LINKED ISSUE #1285: Widen spectral element tests to extruded spaces
## Describe the bug
Our current spectral element tests only target 2D spectral element spaces (in `test/Operators/spectralelement/benchmark_ops.jl`), and we should probably test them for full 3D extruded spaces. This would help widen the coverage, and ensure that we're not missing performance issues for our actual use case.

charleskawczynski:
This PR adds and fixes JET tests for DSS.

This is actually also going to fix a bug: `map(zero, slab(perimeter_data, 1, 1)[1])` is not the correct eltype for ntuple fields + when there are ghost vertices, but our test suite does not cover this situation (until now).

Closes #1285, #1283.

I've added a TODO for widening the perf tests for the GPU. Right now, this fixes the CPU inference issues, and hooks in allocation tests.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -856,6 +856,13 @@ steps:
         soft_fail:
           - exit_status: 1
 
+      - label: ":rocket: Spectral element operator benchmarks (extruded CPU Float64)"
+        key: "gpu_spectral_ops_bm_extruded_cpu_float64"
+        command:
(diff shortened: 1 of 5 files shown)


simonbyrne (approved):

charleskawczynski:
bors try

simonbyrne (approved):

PR #1200, opened 2023-04-24
GPU finite difference

### LINKED ISSUE #1171: GPU finite-difference operators
Two options

1. Oceananigans approach: launch a thread for every node. Conceptually this is the simplest, as each point can be evaluated independently.
   - Unlike Oceananigans we don't use halos for boundary conditions. To avoid branches, we might want a separate kernel launch for each boundary, similar to how we split up the loop in `apply_stencil!` (https://github.com/CliMA/ClimaCore.jl/blob/dd55a7bb2b8aa0bd5f768573c1f9c9ea8c76fd32/src/Operators/finitedifference.jl#L3337-L3345)

2. Use a single thread for each column, and have a loop over levels within the kernel. This should be the simplest (no synchronization is required), and allows dependencies.
  - I think this is what ClimateMachine.jl did
  - potentially the easiest, as we could probably just call `apply_stencil!` directly from within the kernel

Perhaps start with option 2 to get something working.

simonbyrne:
Fixes #1171 

### DIFF
--- a/src/DataLayouts/cuda.jl
+++ b/src/DataLayouts/cuda.jl
@@ -92,3 +92,14 @@ function Base.copyto!(
     end
     return dest
 end
+
+function Base.copyto!(
+    dest::VF{S},
+    bc::Union{VF{S, A}, Base.Broadcast.Broadcasted{VFStyle{A}}},
+) where {S, A <: CUDA.CuArray}
+    _, _, _, Nv, Nh = size(bc)
+    if Nv > 0 && Nh > 0
+        CUDA.@cuda threads = (1, 1) blocks = (Nh, Nv) knl_copyto!(dest, bc)
+    end
+    return dest
+end
--- a/examples/hybrid/box/bubble_3d_invariant_rhoe.jl
+++ b/examples/hybrid/box/bubble_3d_invariant_rhoe.jl
@@ -52,7 +52,11 @@ function hvspace_3D(
         boundary_tags = (:bottom, :top),
     )
     vertmesh = Meshes.IntervalMesh(vertdomain, nelems = zelem)
-    vert_center_space = Spaces.CenterFiniteDifferenceSpace(vertmesh)
+    verttopo = Topologies.IntervalTopology(
+        ClimaComms.SingletonCommsContext(comms_ctx.device),
+        vertmesh,
+    )
+    vert_center_space = Spaces.CenterFiniteDifferenceSpace(verttopo)
 
     horzdomain = Domains.RectangleDomain(
         Geometry.XPoint{FT}(xlim[1]) .. Geometry.XPoint{FT}(xlim[2]),
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -1195,6 +1195,14 @@ function Base.fill!(data::VIFH, val)
     return data
 end
 
+@propagate_inbounds function Base.getindex(
+    data::VIFH{S},
+    I::CartesianIndex{5},
+) where {S}
+    i, _, _, v, h = I.I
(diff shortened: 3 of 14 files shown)


simonbyrne:
bors try

src/DataLayouts/cuda.jl
```diff
@@ -92,3 +92,14 @@ function Base.copyto!(
     end
     return dest
 end
+
+function Base.copyto!(
+    dest::VF{S},
+    bc::Union{VF{S, A}, Base.Broadcast.Broadcasted{VFStyle{A}}},
+) where {S, A <: CUDA.CuArray}
+    _, _, _, Nv, Nh = size(bc)
```
sriharshakandala, line 100: How are we querying `Nf` here? Should `Nh` be one?

sriharshakandala: In general, `1D` threadblocks, with appropriate `slab` calls when direct indexing is not perimitted, may be more efficient!
```
_max_threads_cuda() = 256

function _configure_threadblock(nitems)
    nthreads = min(_max_threads_cuda(), nitems)
    nblocks = cld(nitems, nthreads)
    return (nthreads, nblocks)
end

function Base.copyto!(
    dest::VF{S},
    bc::Union{VF{S, A}, Base.Broadcast.Broadcasted{VFStyle{A}}},
) where {S, A <: CUDA.CuArray}
    nitems = length(parent(dest))
    nthreads, nblocks = _configure_thread_blocks(nitems)
    if nitems > 0
        CUDA.@cuda threads = (nthreads,) blocks = (nblocks) knl_copyto_vf!(dest, bc)
    end
    return dest
end

function knl_copyto_vf!(dest, src)
    gidx = CUDA.threadIdx().x + (CUDA.blockIdx().x - 1) * CUDA.blockDim().x
    _, _, _, Nv, _ = size(dest)
    f = cld(gidx, Nv)
    v = gidx - Nv * (f - 1)
    dest_slab = slab(dest, f)
    src_slab = slab(src, f)
    @inbounds dest_slab[v] = src_slab[v]
    return nothing
end
```

sriharshakandala: If operators are applies to fields (`f`s) recursively, then we may only need to distribute threads based on `Nv`!

simonbyrne: > How are we querying `Nf` here? Should `Nh` be one?

We don't need `Nf` (each thread can access all fields), and yes, `Nh == 1`

I agree this is inefficient (we're only doing a fraction of the work we should be doing)

simonbyrne: Actually, i have an idea on how to fix this.

simonbyrne: See #1216

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

src/DataLayouts/cuda.jl
simonbyrne: I realise what we have is far from optimal, but we can clean this up once we have other pieces in.

sriharshakandala: Sure. We can revisit this once we have other pieces in.

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
Okay, this appears to be passing. @sriharshakandala any more comments?

sriharshakandala:
> Okay, this appears to be passing. @sriharshakandala any more comments?

Looks good to me. The bug fix, effecting extruded 3D case, for reduction code has been added in PR https://github.com/CliMA/ClimaCore.jl/pull/1221

PR #1203, opened 2023-04-27
Make threading more flexible

charleskawczynski:
This PR splits off threading from column threading.


charleskawczynski:
I tried this out and did not see any improvement, but it's not clear to me that this was an exhaustive way to exercise this idea.

Issue #1196, opened 2023-04-18
One test failure 


waywardpidgeon:
I ran the two tests, runtests.jl and expensive_runtests.jl, for ClimaCore in a Windows 10 environment successfully, with only one error "Test composed stencils" at line 192 in column.jl:
Expression \del . (w .* I.(\theta))
  Expected: BoundsError
  No exception thrown

A brief examination of the code indicates this just means the program "fell out the bottom" without finding where to compute. It relates to issue #1183, but I can't find any occurrence of "div" in the related code. A secti
[runtests.docx](https://github.com/CliMA/ClimaCore.jl/files/11266988/runtests.docx)
on of the output from the Windows Power Shell is attached.
Thanks - waywardpidgeon
[runtests.docx](https://github.com/CliMA/ClimaCore.jl/files/11266993/runtests.docx)

waywardpidgeon:
I used Pkg.test() as suggested and runtests.jl completed successfully
except for FD opts - linsolve which published 56 passes and 4 broken tests.
Part of the output is attached - its runtests2.docx
Thanks - waywardpidgeon

simonbyrne:
Can you paste the output in a comment (quoted by ```)?

waywardpidgeon:
Sorry about the docx output file. I'm on the Julia learning curve still, and now know how to get the history in a more accessible form, but that will have to wait for next time! Oh, I forgot to say I removed the "!" from !Sys.iswindows() on line 13 of runtests.jl (and the corresponding line of expensive_runtests.jl) of course.

charleskawczynski:
Closed by #1206

PR #1206, opened 2023-04-28
Test check bounds only when forced

charleskawczynski:
I've found many users find this unintuitive (e.g., [here](https://github.com/CliMA/ClimaCore.jl/issues/1196)), perhaps we can skip this test if bounds checks are not forced?

This will make

```julia
include("test/runtests.jl")
```
Pass with or without `--check-bounds=true`.

### DIFF
--- a/test/Operators/finitedifference/column.jl
+++ b/test/Operators/finitedifference/column.jl
@@ -189,7 +189,12 @@ end
         ∂ = Operators.GradientF2C()
 
         # TODO: should we throw something else?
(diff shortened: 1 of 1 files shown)


simonbyrne:
Ideally it would still throw an error. Until we fix it, we should throw an error, but we can't combine `@test_throws` and `broken` (https://github.com/JuliaLang/julia/issues/21098).

This seems like a reasonable option.

simonbyrne (approved):

PR #1211, opened 2023-05-01
Remove bubble warning

### LINKED ISSUE #1210: Remove bubble warning
## Purpose
It was recently suggested that we remove the bubble correction [warning](https://github.com/CliMA/ClimaCore.jl/blob/main/src/Spaces/spectralelement.jl#L410-L412) since now it is used as default in the downstream repo (ClimaAtmos), so the user might not care about this warning.

Remove the warning and its flag

valeriabarra:
It was recently suggested that we remove the bubble correction [warning](https://github.com/CliMA/ClimaCore.jl/blob/main/src/Spaces/spectralelement.jl#L410-L412) since now it is used as default in the downstream repo (ClimaAtmos), so the user might not care about this warning.

This will close #1210 


valeriabarra:
bors try

simonbyrne (approved):

PR #1214, opened 2023-05-02
Bump version, update dependencies

charleskawczynski:
This PR bumps the ClimaCore patch version and updates the dependencies.

PR #1215, opened 2023-05-02
Remove Device mod, use ClimaComms

charleskawczynski:
This PR removes the `Device` module, and uses ClimaComms. xref: https://github.com/CliMA/ClimaCore.jl/pull/1214#issuecomment-1531941357


valeriabarra (approved):

simonbyrne:
we should add a deprecation

charleskawczynski:
> we should add a deprecation

Does that mean we need to keep the module around?

simonbyrne:
maybe create a file `src/deprecated.jl`, and put the module and function definitions in there (once you have this working)

charleskawczynski:
@simonbyrne, how does that look?

charleskawczynski:
bors try

src/deprecated.jl
```diff
@@ -0,0 +1,24 @@
+# Device module
+module Device
+using ClimaComms
+using CUDA
+
+Base.@deprecate device(; disablegpu = false) ClimaComms.device(;
+    disablegpu = false,
+) false
```
simonbyrne, line 8: ClimaComms.device doesn't take a `disablegpu` arg.
```suggestion
Base.@deprecate device(; disablegpu = false) if disablegpu
    ClimaComms.CPUDevice()
else
    ClimaComms.device()
end false
```

charleskawczynski: Ah, whoops, I thought I had looked at the definition in ClimaComms, must have been in ClimaCore. Thanks!

charleskawczynski:
bors try

simonbyrne (approved):

charleskawczynski:
( I canceled the try once this was approved and seemed to be working )

PR #1216, opened 2023-05-03
maximize number of active threads by doing multiple slabs per block

simonbyrne:
This should be more efficient.

### DIFF
--- a/src/DataLayouts/cuda.jl
+++ b/src/DataLayouts/cuda.jl
@@ -40,33 +40,41 @@ Base.similar(
     dims::Dims{N},
 ) where {T, N, B} = similar(CUDA.CuArray{T, N, B}, dims)
 
-function knl_copyto!(dest, src)
-
-    #=
-    nij, nh = size(dest)
-
-    thread_idx = CUDA.threadIdx().x
-    block_idx = CUDA.blockIdx().x
-    block_dim = CUDA.blockDim().x
-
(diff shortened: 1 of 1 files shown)


simonbyrne:
bors try

src/DataLayouts/cuda.jl
```diff
@@ -88,7 +100,11 @@ function Base.copyto!(
 ) where {S, Nij, A <: CUDA.CuArray}
     _, _, _, Nv, Nh = size(bc)
     if Nv > 0 && Nh > 0
-        CUDA.@cuda threads = (Nij, Nij) blocks = (Nh, Nv) knl_copyto!(dest, bc)
+        Nl = fld(CUDA.warpsize(), Nij * Nij)
```
sriharshakandala, line 103: I think we can use up to `1024` threads. 
If we are hard-wired to use a `(Nij, Nij)` threadblock, we can just add an assertion `Nij * Nij <= 1024 (or max_threads)`. 
For more general code, `Nl` can be calculated inside the kernel! (`Nl = cld(Nij * Nij, blockDim.x()*blockDim.y()`)

simonbyrne: Good point. We probably need to play around with it and do some benchmarking to see what works.

simonbyrne:
The other question is how best to use this to take advantage of coalesced loads/stores

PR #1219, opened 2023-05-03
Fix logic in Zalesak

### LINKED ISSUE #1218: Logic bug in Zalesak operator
## Describe the bug
There seems to be a bug in the logic of the Zalesak operator, on this [line](https://github.com/CliMA/ClimaCore.jl/blob/cd3a6fec9f5e8044140bfc14056dfb85ad759d1e/src/Operators/finitedifference.jl#L1722). Although the Durran book only talks about a cosmetic correction, we can use the `&&` to match the reference and see if it improves.

Screen Shot 2023-05-03 at 12 42 40 PM

valeriabarra:
There seems to be a bug in the logic of the Zalesak operator, on this [line](https://github.com/CliMA/ClimaCore.jl/blob/cd3a6fec9f5e8044140bfc14056dfb85ad759d1e/src/Operators/finitedifference.jl#L1722). Although the Durran book only talks about a cosmetic correction, we can use the && to match the reference and see if it improves.

This will close #1218 

### DIFF
--- a/examples/hybrid/sphere/deformation_flow.jl
+++ b/examples/hybrid/sphere/deformation_flow.jl
@@ -309,7 +309,7 @@ lim_first_upwind_ρ_err, lim_first_upwind_ρq_errs =
 lim_centered_ρ_err, lim_centered_ρq_errs = conservation_errors(lim_centered_sol)
 
 # Check that the conservation errors are not too big.
-max_err = 40 * eps(FT)
+max_err = 42 * eps(FT)
 @test abs(third_upwind_ρ_err) < max_err
 @test all(abs.(third_upwind_ρq_errs) .< max_err)
 @test all(abs.(fct_ρq_errs) .< max_err)
@@ -353,7 +353,7 @@ max_q5_roundoff_err = 2 * eps(FT)
     tracer_ranges(lim_third_upwind_sol) .<
     0.6 .* tracer_ranges(third_upwind_sol),
 )
-@test all(tracer_ranges(lim_fct_sol) .< 0.5 .* tracer_ranges(third_upwind_sol))
+@test all(tracer_ranges(lim_fct_sol) .< 0.55 .* tracer_ranges(third_upwind_sol))
 @test all(
     tracer_ranges(lim_first_upwind_sol) .<
     0.5 .* tracer_ranges(third_upwind_sol),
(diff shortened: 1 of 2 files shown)


valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

charleskawczynski (approved):
Code changes look fine to me.

simonbyrne:
It seems like Zalasek doesn't have any unit tests. Could we add some?

valeriabarra:
> It seems like Zalasek doesn't have any unit tests. Could we add some?

I had added the [1D column advection example](https://github.com/CliMA/ClimaCore.jl/blob/main/examples/column/zalesak_fct_advection.jl) as a unit/integration test. But if you want it in `test/`, sure we can add that too.

valeriabarra:
> > It seems like Zalasek doesn't have any unit tests. Could we add some?
> 
> I had added the [1D column advection example](https://github.com/CliMA/ClimaCore.jl/blob/main/examples/column/zalesak_fct_advection.jl) as a unit/integration test. But if you want it in `test/`, sure we can add that too.

Ah, now I remember. I think we hadn't added it in there, because it's not a spatial operator only. It relies on the integrator `dt`, so we decided the best place was in the `examples/` directory

simonbyrne:
I think you could test the tendency given an assumed dt? e.g. we could check that the desired properties hold after a single forward Euler step for a variety of dts

Would be good to have both test cases from the Durran book.

simonbyrne:
@valeriabarra Lets merge this for now, as it's clear that it more closely matches the Durran book. But can you please add some tests (in particular Durran notes that this code path doesn't affect the square wave, so it would be good to have a case that does). Also we should add unit tests that it satisfies the necessary properties.

valeriabarra:
> @valeriabarra Lets merge this for now, as it's clear that it more closely matches the Durran book. But can you please add some tests (in particular Durran notes that this code path doesn't affect the square wave, so it would be good to have a case that does). Also we should add unit tests that it satisfies the necessary properties.

I am trying to check the tolerance of the failing deformation flow test. I can't get the example to run locally on my machine though, for some CUDA error (even though my environment is in order)

valeriabarra:
The tolerance for the deformation flow test needed to be adjusted very slightly. 

This is how the `q3` tracer looks like at day 12 in this PR: 

q3_day12_fct

Vs `main`:
main_q3_day12_fct

Looks reasonable?

PR #1209, opened 2023-05-01
use ClimaComms.bcast instead of MPI function

simonbyrne:
This was added to ClimaComms (https://github.com/CliMA/ClimaComms.jl/pull/48), so we don't need to dip into the MPI module.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1220, opened 2023-05-04
Add MPI compat for CCTR

charleskawczynski:
This PR adds a MPI compat bound for ClimaCoreTempestRemap (auto-merging of the latest version failed: https://github.com/JuliaRegistries/General/pull/82879)


simonbyrne:
Actually, I think with #1209, we can get rid of the MPI direct dependency (will need to remove `using MPI` in a few places)

simonbyrne (approved):

Issue #1170, opened 2023-03-31
GPU support for extruded spaces on distributed context


simonbyrne:
Need to make sure that the vertical context is always a `SingletonCommsContext`, and the horizontal should be a `MPICommsContext`.

charleskawczynski:
We're first updating to the latest version of ClimaComms so that we don't write gpu code that is incompatible with the latest update of ClimaComms.

simonbyrne:
Can we call this done?

charleskawczynski:
I'm still working on a PR for this

charleskawczynski:
It's ready for review now

PR #1222, opened 2023-05-04
Deprecate comm_context, use ClimaComms.context

charleskawczynski:
This PR deprecates
 - `comm_context` in favor of using `ClimaComms.context`
and methods that assume the device is internally determined using `CUDA.functional()`, which is problematic if CUDA is available but we want to run on the CPU:
 - `HDF5Reader`
 - `HDF5Writer`
 - `Topology2D`
 - (fixes `create_dss_buffer` to use the topology context)

A step towards #1170.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -1,6 +1,5 @@
 module Fields
 
-import ..comm_context
 import ClimaComms
 import ..enable_threading
 import ..slab, ..slab_args, ..column, ..column_args, ..level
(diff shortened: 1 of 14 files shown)


src/Fields/Fields.jl
```diff
@@ -45,11 +45,11 @@ comm_context(space::Spaces.ExtrudedFiniteDifferenceSpace) =
 comm_context(space::Spaces.SpectralElementSpace2D) =
     comm_context(space.topology)
 comm_context(space::S) where {S <: Spaces.AbstractSpace} =
-    ClimaComms.SingletonCommsContext()
+    comm_context(space.topology)
```
simonbyrne, line 48: could deprecate `comm_context` altogether, and overload `ClimaComms.context` (same we do for `ClimaComms.device`)?

charleskawczynski: Isn't `comm_context` still a useful getter for users if all spaces contain a context?

charleskawczynski: Oh, I misread, so change this function name to `context` @simonbyrne?

charleskawczynski: Yeah, that would make sense, we can then just overload it.

simonbyrne: I mean overload `ClimaComms.context`

simonbyrne (approved):

charleskawczynski:
@simonbyrne, do we want to add topology to `PointSpace`s? On the one hand, `ClimaComms.context` will work for all spaces, on the other hand it kind of seems a bit awkward (a single point cannot be distributed).

simonbyrne:
it's a bit of a funny one, not sure.

simonbyrne:
Can we just define it to be `SingletonContext` for now? That would be backward compatible.

charleskawczynski:
> Can we just define it to be `SingletonContext` for now? That would be backward compatible.

The problem with that is that SingletonContext needs to inherit the device. Should we add a `device` property to `PointSpace`s?

charleskawczynski:
After looking around, I think adding at least `context` to the `PointSpace` is the most maintainable path forward. Otherwise we'll need to specialize on PointField for `sum`, `mean`, `norm` etc. (and that list is relatively long).

charleskawczynski:
bors try

PR #1225, opened 2023-05-05
Make implicit test variables readable

charleskawczynski:
Hi @dennisYatunin, can you please take this PR over and replace the characters that are being replaced with ones that are readable? These characters don't seem to render in my browser, vscode, or in the REPL-- so I have no idea if they are distinguishable or not.

If someone else knows what these characters are supposed to be, I would really appreciate the help here.


szy21 (approved):
@dennisYatunin  says those characters are used on purpose - ⍰ represents either F or C. He said he won't do it again though!

charleskawczynski:
There’s already F/C in some of those variables, the question mark symbol is a wild card for F/C too?

szy21:
> There’s already F/C in some of those variables, the question mark symbol is a wild card for F/C too?

Yes, that’s how I understand it.

PR #1229, opened 2023-05-05
Break up implicit stencil tests for improving inference

charleskawczynski:
CI times have increased and are a bit of a drag, mostly due to the implicit stencil tests, which take ~1hr.

This PR breaks up these tests into separate functions to help more easily utilize inference tools to understand why it's taking so long.


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #1239, opened 2023-05-12
Improve test utilities

charleskawczynski:
This PR:
 - Moves `util_spaces.jl` into a module, called `TestUtilities`
 - Breaks `all_spaces` into separate functions and adds a `all_spaces` function that returns an array of each space.

I noticed that there are edge cases missing in our DSS tests (see #1198, which breaks some examples but passes all unit tests). I think we need to get a handle on the duplication in our test suite, and `TestUtilities` is where I'd like to collect pieces that are being duplicated.


valeriabarra (approved):

test/TestUtilities/TestUtilities.jl
```diff
@@ -0,0 +1,197 @@
...
+        SpectralElementSpace1D(FT; context),
+        SpectralElementSpace2D(FT; context),
+        # TODO: add these
+        # SpectralElementRectilinearSpace2D(FT; context),
+        # SpectralElementFiniteDifferenceRectilinearSpace2D(FT; context),
+        ColumnCenterFiniteDifferenceSpace(FT; zelem, context),
+        ColumnFaceFiniteDifferenceSpace(FT; zelem, context),
+        SphereSpectralElementSpace(FT; context),
+        ExtrudedCenterFiniteDifferenceSpace(FT; zelem, context),
+        # ExtrudedFaceFiniteDifferenceSpace(FT; zelem, context), # errors on sum
```
charleskawczynski, line 153: @sriharshakandala, @simonbyrne, sorry, I realize now that I squashed this into the previous commit. Please uncomment to reproduce the error

Issue #1242, opened 2023-05-17
OOM errors in CI initialization


charleskawczynski:
## Describe the bug
It seems that we're having rather frequent OOM errors in CI initialization: https://buildkite.com/clima/climacore-ci/builds/2053#01882bce-ccd8-4590-b3e7-23a77ce73e2f/1815-2036

## To Reproduce
Open a PR.

charleskawczynski:
😞

https://buildkite.com/clima/climacore-ci/builds/2063#018830dd-21e6-4075-862b-155b6c1d28fb/154-698.

This was with adding 32GB for initialization, cc @simonbyrne.

Should I just add a `ClimaCore.jl/Manifest.toml` and revert the versions? Or adjust the compat bounds to restrict CUDA / Plots? I suppose I can try to figure out which one is OOMing on us.

charleskawczynski:
Ah, it _has_ to be CUDA-- Plots isn't in the `ClimaCore.jl/Project.toml`

charleskawczynski:
Looking back, it seems like we've been on CUDA 4.2.0 for a while. I may have looked at a local unchecked Manifest and thought it was related to CUDA. So this issue is perhaps unrelated to the CUDA version.

charleskawczynski:
Closing since I've not seen this in a while. I'll reopen if it pops up again.

PR #1244, opened 2023-05-17
Add Manifest for benchmarks environment

charleskawczynski:
Our benchmarks environment is erroring during initialization, and the compat bounds are unbounded, so it's perhaps breaking due to versions it's updating to. This PR adds compats and a Manifest in hopes that we have a more reproducible environment.

A workaround for #1242.


charleskawczynski:
bors try

simonbyrne (approved):

PR #1246, opened 2023-05-18
Add proposed new buffer to spectral element benchmarks

charleskawczynski:
This PR adds a proposed new buffer for when the spectral element benchmarks fail. This should help make updating the buffer a bit easier.


charleskawczynski:
Let's just manually set this..

PR #1248, opened 2023-05-18
strip out unnecessary space metadata in spectral element operators

### LINKED ISSUE #1226: Composed operators use too much parameter space
## Describe the bug
When composing multiple operators on the GPU, I often get the error
```
ERROR: Failed to compile PTX code (ptxas exited with code 255)
Invocation arguments: --generate-line-info --verbose --gpu-name sm_60 --output-file /tmp/jl_WvsZ8X6v8x.cubin /tmp/jl_WZ85G5u5Yp.ptx
...
uses too much parameter space (0x1208 bytes, 0x1100 max).
ptxas fatal   : Ptx assembly aborted due to errors
If you think this is a bug, please file an issue and attach /tmp/jl_WZ85G5u5Yp.ptx
```

What this is saying is that the GPU kernel parameters (the information passed from the CPU to the GPU kernel on each launch) being too large. The limit appears to be about 4KB, but our vector Laplacian operator is over 8KB.

This is caused by our representation of the broadcasted expression trees: we duplicate a lot of information. Each `Field` carries its space, as does each node in the expression tree.

Some ways we can fix this:
- remove the space when we Adapt the `Field` (we could have a special `DeviceField`?).
  - the main challenge here is how to make `getidx` work (in particular the `half` offset/periodic pieces)
- remove the space when we Adapt the bc (except for the top-level?)
  - this gets challenging as some operations change spaces: FD can alternate center <-> face, SE can change order (but this is not yet supported on GPUs).

simonbyrne:
Fixes #1226.

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -493,8 +493,6 @@ Base.@propagate_inbounds function set_node!(
     fv = Fields.field_values(field)
     fv[i, j, nothing, v, h] = val
 end
-set_node!(field, ij, slabidx, val) =
-    set_node!(axes(field), field, ij, slabidx, val)
 
 Base.Broadcast.BroadcastStyle(
     ::Type{<:SpectralBroadcasted{Style}},
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -186,7 +186,7 @@ Base.@propagate_inbounds function copyto_slab!(out, bc, slabidx)
     Nq = Quadratures.degrees_of_freedom(QS)
     rbc = resolve_operator(bc, slabidx)
     @inbounds for ij in node_indices(axes(out))
-        set_node!(out, ij, slabidx, get_node(rbc, ij, slabidx))
+        set_node!(space, out, ij, slabidx, get_node(space, rbc, ij, slabidx))
     end
     return nothing
 end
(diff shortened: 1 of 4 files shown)


src/Operators/spectralelement.jl
```diff
@@ -426,7 +493,8 @@ Base.@propagate_inbounds function set_node!(
     fv = Fields.field_values(field)
     fv[i, j, nothing, v, h] = val
 end
-
+set_node!(field, ij, slabidx, val) =
```
charleskawczynski, line 496: Do we need this?

test/Operators/hybrid/cuda.jl
```diff
@@ -91,14 +91,28 @@ end
...
+    coords = Fields.coordinate_field(hv_face_space_gpu)
+    z = coords.z
 
     gradc = Operators.GradientF2C()
 
     @test parent(Geometry.WVector.(gradc.(z))) ≈
           parent(Geometry.WVector.(ones(hv_center_space_gpu)))
-end
 
+
```
charleskawczynski, line 102: Can we add a comment what this issue this fixes? (https://github.com/CliMA/ClimaCore.jl/issues/1226)

charleskawczynski (approved):
Overall looks good!

My preference would be to eagerly unpack fieldvalues so that we don't accidentally grab the field space in some broadcast expressions but not others. But it should be fine assuming it's thoroughly tested.

My only other concern is the `set_node!(field, ij, slabidx, val) = set_node!(axes(field), field, ij, slabidx, val)` fallback, won't we want to always avoid this?

simonbyrne:
bors try

simonbyrne:
bors try

src/Operators/spectralelement.jl
```diff
@@ -233,6 +233,53 @@ Base.@propagate_inbounds _resolve_operator_args(slabidx, arg, xargs...) = (
...
+) where {Style}
+    current_space = axes(bc)
+    if current_space == parent_space
+        new_space = PlaceholderSpace()
+    else
+        new_space = current_space
+    end
+    return Base.Broadcast.Broadcasted{Style}(
+        bc.f,
+        map(arg -> strip_space(arg, current_space), bc.args),
```
charleskawczynski, line 263: Does this need to be recursive?

simonbyrne: Probably, but it's not in a tight loop, so we're probably fine.

charleskawczynski: Just so I understand correctly, does this mean that if a BC expression has 8 levels of recursion, then this would strip it down to 7?

charleskawczynski: Ah, ok, got it.

simonbyrne:
bors try

charleskawczynski (approved):

PR #1250, opened 2023-05-19
Make TestUtilities a package

charleskawczynski:
It occurred to me that perhaps #1242 could also be influenced by the fact that including the TestUtilities module many times could trigger lots of recompilation-- I'm not sure why the GC wouldn't take care of this, though. One way to circumvent the potential code loading issue, I think, would be to make TestUtilities a package, and simply dev it in the test directory. Trying that here.


charleskawczynski:
Perhaps we don't need this.

Issue #1240, opened 2023-05-17
CUDA sum incorrect on V100


charleskawczynski:
## Describe the bug
The CUDA `sum` is producing different results between CPU and GPU for certain spaces. See [this build](https://buildkite.com/clima/climacore-ci/builds/2052#01882b5c-45b4-4528-a55e-88200c868acb), which is the result of widening the test spaces in [this PR](https://github.com/CliMA/ClimaCore.jl/pull/1239). For now, I'll check the array type and note these as broken tests.

cc @sriharshakandala, @simonbyrne

simonbyrne:
@sriharshakandala can you please look into this ASAP?

sriharshakandala:
These tests seem to be running correctly on the GPU cluster (single GPU). Please see below:
```
[skandala@hpc-26-18 test]$ git status
On branch ck/test_utils
Your branch is up to date with 'origin/ck/test_utils'.

nothing to commit, working tree clean
[skandala@hpc-26-18 test]$ srun julia --project Operators/finitedifference/column.jl 
Scalar Field FiniteDifferenceSpaces: Error During Test at /central/home/skandala/ClimaCore.jl/test/Operators/finitedifference/column.jl:36
 Unexpected Pass
 Expression: sum(fones) ≈ pi
 Got correct result, please change to @test if no longer broken.

Scalar Field FiniteDifferenceSpaces: Error During Test at /central/home/skandala/ClimaCore.jl/test/Operators/finitedifference/column.jl:47
 Unexpected Pass
 Expression: ≈(sum(ffield), FT(2.0), atol = 0.01)
 Got correct result, please change to @test if no longer broken.

Scalar Field FiniteDifferenceSpaces: Error During Test at /central/home/skandala/ClimaCore.jl/test/Operators/finitedifference/column.jl:36
 Unexpected Pass
 Expression: sum(fones) ≈ pi
 Got correct result, please change to @test if no longer broken.

Scalar Field FiniteDifferenceSpaces: Error During Test at /central/home/skandala/ClimaCore.jl/test/Operators/finitedifference/column.jl:47
 Unexpected Pass
 Expression: ≈(sum(ffield), FT(2.0), atol = 0.01)
 Got correct result, please change to @test if no longer broken.

Test Summary:                       | Pass  Error  Total   Time
Scalar Field FiniteDifferenceSpaces |   18      4     22  59.3s
[...]

PR #1251, opened 2023-05-19
Test GPU sum bug

charleskawczynski:
@sriharshakandala, this should yield the error in #1240. Sorry, I think I squashed this commit into "fixes".


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
Ah, this is actually failing on the CPU, too cc @sriharshakandala, @simonbyrne. I'm a bit confused about the one build that did fail, and I'm now not convinced that this is the same failure.

PR #1259, opened 2023-05-22
Distributed regridding - concrete example (WIP)

juliasloan25:
Part of ClimaCoupler SDI [#188](https://github.com/CliMA/ClimaCoupler.jl/issues/188).

The goal of this PR is to implement a concrete example of distributed regridding. This will allow us to implement the logic for a specific case while interactively viewing and verifying weights, indices, output values, etc. It will also help us better understand the relevant MPI functions and how/when they need to be used.

This implementation is still a work in progress and does not work as desired. The current blocking issue is that after generating the weights on the root process, they are not correctly being broadcast to other processes, so the `weights` object is `nothing` on non-root processes. I have tried using both `MPI.bcast` and `ClimaComms.bcast`, but neither solves this problem.


src/Operators/remapping.jl
```diff
@@ -268,10 +268,29 @@ end
...
     size(Meshes.elements(topology.mesh), 1)
-nyelems(topology::Topologies.Topology2D{<:ClimaComms.SingletonCommsContext}) =
+end
+function nyelems(
+    topology::T,
+) where {
+    T <: Union{
+        Topologies.Topology2D{<:ClimaComms.SingletonCommsContext},
+        Topologies.Topology2D{<:ClimaComms.MPICommsContext},
+    },
```
charleskawczynski, line 287:
```suggestion
    T <: Topologies.Topology2D{<:ClimaComms.AbstractCommsContext}
```
Maybe just use the abstract comms context here if it applies?

PR #1261, opened 2023-05-23
Conditionally overload Krylov.ktypeof

charleskawczynski:
This PR conditionally overloads `Krylov.ktypeof` using Requires. This is added here to avoid type piracy in ClimaAtmos. Alternatively to this PR, we could wait until we upgrade to Julia 1.9 and use package extensions.


simonbyrne (approved):
This seems reasonable. Once we switch to 1.9 we can use extensions.

charleskawczynski:
Ah, I need to update the manifests since this PR adds the `Requires` dependency.

Issue #837, opened 2022-07-27
Decrease CI times


charleskawczynski:
Now that the column performance has been improved, we could actually significantly improve CI times by figuring out how to reduce only two jobs:
 - implicit stencil tests (which look like we could easily split into 2)
 - The planar IGW (which, for some reason, takes 40 min)

charleskawczynski:
I think we can close this. CI is about 40 minutes, 20 of which is instantiation, and 20 minutes runtime seems reasonable to me. If we want to use a depot to speed up initialization, CI will be stateful and we'll need to periodically clear the depot path, which seems more like a trade-off than a clear win.

PR #1281, opened 2023-05-31
Break down large tuples in implicit tests

charleskawczynski:
A step towards #837. I believe that this may help improve inference, which I've seen has an impact on the smaller tuple size (in the apply).

### DIFF
--- a/test/Operators/finitedifference/implicit_stencils.jl
+++ b/test/Operators/finitedifference/implicit_stencils.jl
@@ -13,10 +13,10 @@ include(joinpath("test", "Operators", "finitedifference", "implicit_stencils.jl"
 # For interactive use:
 ```julia
 julia -O0 --check-bounds=yes --project=test
-push!(ARGS, "--float_type", "Float32")
+push!(ARGS, "--float_type", "Float32");
 include(joinpath("test", "Operators", "finitedifference", "implicit_stencils_utils.jl"))
-
-const FT = parsed_args["float_type"] == "Float64" ? Float64 : Float32
+(s, parsed_args) = parse_commandline();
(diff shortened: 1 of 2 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
Notes for future self:

 - A local version is working that runs in ~10 min
 - The main problems with runtime have been: 1) poor inference and 2) working with huge non-homogeneous tuples
 - Constructing `all_ops` in the test suite is relatively cheap, but calling methods with it is _very_ expensive. So, I think the simple fix is to manually unroll the loops (using a handful of `Base.Cartesian.@nexprs` calls) in global scope (or in the `@testset`s)
 - Right now I have a version that writes to a file, but we should probably use expressions + eval to avoid writing and including a file at runtime.

charleskawczynski:
bors try

charleskawczynski:
Oof, this is still taking 40 min. Interesting, running this interactively takes, IIRC, 13 min? (Including the flags used in the pipeline)

maybe something else is going on? It seems like this is worth understanding.

charleskawczynski:
Interesting update: I celebrated early. It turns out that some of these individual expressions take a very long time to compile compared to others. The first few compose functions took ~0.2 seconds, so I estimated that this PR would really speed things up.

However, there was one that took **25 seconds**, several that took ~18, several more 15 etc.  So, it seems to be an issue with the implicit solver infrastructure. Interestingly, when broken down in this way, these methods appear to be fully inferred, however, they can still take a very long time to compile.

I think we may need to look at what's going on at compile time to understand where all the time is spent.

charleskawczynski:
Superseded by #1305

Issue #1278, opened 2023-05-30
Make performance tests more systematic


charleskawczynski:
**Is your feature request related to a problem? Please describe.**
Our existing performance tests are currently a mixture of allocation tests, inference tests, and benchmarks written at different times and with different goals in mind. It might be nice if we more systematically organize our performance tests.

Looking at our test suite, it seems like there are a few performance tests that we could apply to all of our operators:

##### Deterministic
 - Inference (JET)
 - Allocations `@allocated`
 - GFlops (can replace the comparison against ordinary arrays)
 - Test counted number of `" inbounds "` occurrences in llvm output. (IntrospectionTools.jl)

##### Non-deterministic, 
 - BenchmarkTools.@btime (with all the deterministic tests in place, benchmarks can be made soft fail)

PR #1288, opened 2023-06-05
Bump patch version

charleskawczynski:


charleskawczynski:
I've also skipped on the benchmark tests, and removed the soft-fail (this way they are at least required to _run_). It's a bit too much of a drag to depend on non-deterministic tests. Hopefully between the other tests, we can still have a strong signal that benchmarks will be good. Relevant issue: #1278.

PR #1291, opened 2023-06-05
interpolate to regular array

simonbyrne:
Should make it easier to visualize slices.

cc @szy21 


szy21:
Thank you!

simonbyrne:
Not sure `Operators` is the right place for this. Maybe it should go in a different module?

docs/src/operators.md
```diff
@@ -105,6 +105,11 @@ SetDivergence
 Extrapolate
 ```
 
+## Interpolation
+
+```@docs
+array_interpolate
```
charleskawczynski, line 111:
```suggestion
interpolate_array
```

src/Operators/arrayinterpolate.jl
```diff
@@ -0,0 +1,165 @@
...
+latpts = range(Geometry.LatPoint(-80.0), Geometry.LatPoint(80.0), length = 21)
+zpts = range(Geometry.ZPoint(0.0), Geometry.ZPoint(1000.0), length = 21)
+
+interpolate_array(field, longpts, latpts, zpts)
+```
+
+!!! note
+    Hypsography is not currently handled correctly.
+"""
+interpolate_array
```
charleskawczynski, line 96:
```suggestion
function interpolate_array end
```

I'd prefer we declare this a function for clarity

src/Geometry/coordinates.jl
```diff
@@ -129,6 +133,10 @@ product_coordinates(yp::YPoint, zp::ZPoint) = YZPoint(promote(yp.y, zp.z)...)
 product_coordinates(xyp::XYPoint, zp::ZPoint) =
     XYZPoint(promote(xyp.x, xyp.y, zp.z)...)
 
+product_coordinates(latp::LatPoint, longp::LongPoint) =
+    LatLongPoint(promote(latp.lat, longp.long)...)
+product_coordinates(longp::LongPoint, latp::LatPoint) =
```
charleskawczynski, line 138: Does `promote` preserve the order? it looks like the order is swapped here, does that matter?

src/Operators/Operators.jl
```diff
@@ -26,5 +26,6 @@ include("operator2stencil.jl")
 include("pointwisestencil.jl")
 include("remapping.jl")
 include("integrals.jl")
+include("arrayinterpolate.jl")
```
charleskawczynski, line 29: Should we rename this file to `interpolate_array.jl` just so that the names match better?

src/Operators/arrayinterpolate.jl
```diff
@@ -0,0 +1,165 @@
...
+function interpolate_array(
+    field::Fields.ExtrudedFiniteDifferenceField,
+    xpts,
+    ypts,
+    zpts,
+)
+    space = axes(field)
+    @assert ClimaComms.context(space) isa ClimaComms.SingletonCommsContext
+
+    horz_topology = Spaces.topology(space)
```
charleskawczynski, line 139:
```suggestion
    horz_topology = Spaces.topology(Spaces.horizontal_space(space))
```
Don't we need to extract the horizontal space for this?

charleskawczynski: (same for above)

charleskawczynski (approved):
The src code changes look good. I didn't look too closely for off-by-1 index errors, but it looks like the tests should catch such errors.

src/Operators/arrayinterpolate.jl
```diff
@@ -0,0 +1,165 @@
...
+    field::Fields.Field,
+    slabidx::Fields.SlabIndex,
+    (I1, I2)::Tuple{<:AbstractArray, <:AbstractArray},
+)
+    space = axes(field)
+    x = zero(eltype(field))
+    QS = Spaces.quadrature_style(space)
+    Nq = Spaces.Quadratures.degrees_of_freedom(QS)
+
+    for j in 1:Nq, i in 1:Nq
```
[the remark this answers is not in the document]
simonbyrne: It isn't. It's mainly for post-processing

test/Operators/arrayinterpolate.jl
```diff
@@ -0,0 +1,149 @@
...
+
+    vertmesh = Meshes.IntervalMesh(vertdomain, nelems = 30)
+    verttopo = Topologies.IntervalTopology(vertmesh)
+    vert_center_space = Spaces.CenterFiniteDifferenceSpace(verttopo)
+
+    horzdomain = Domains.SphereDomain(1e6)
+
+    quad = Spaces.Quadratures.GLL{4}()
+    horzmesh = Meshes.EquiangularCubedSphere(horzdomain, 6)
+    horztopology = Topologies.Topology2D(horzmesh)
```
charleskawczynski, line 114:
```suggestion
    horztopology = Topologies.Topology2D(context, horzmesh)
```
Don't we need a context here?

simonbyrne:
After more thought, I think we should move this to a new module: it's not really an operator in the usual sense, and if we want to support Hypsography, it has to come after. Any thoughts on a name?

- `Remapping`/`Regridding` (we could fold in the remapping pieces as well)?
- `ArrayInterpolation`?

simonbyrne:
bors try

charleskawczynski:
> After more thought, I think we should move this to a new module: it's not really an operator in the usual sense, and if we want to support Hypsography, it has to come after. Any thoughts on a name?
> 
> * `Remapping`/`Regridding` (we could fold in the remapping pieces as well)?
> * `ArrayInterpolation`?

By usual sense do you mean that it just doesn't use the broadcasting machinery? The distinction seems a bit more back-end related. I think it's fine, but I'm also fine with moving it to a separate module

simonbyrne:
> By usual sense do you mean that it just doesn't use the broadcasting machinery?

Basically, yes. Also if we want to support hypsography, it has to come after that module.

simonbyrne:
bors try

szy21:
The way we construct 3D box in ClimaAtmos seems different from that in ClimaCore. In ClimaAtmos for the horizontal topology we use the spacefillingcurve (both for box and sphere). Would slice also work in this case?
```
topology = Topologies.Topology2D(
            comms_ctx,
            mesh,
            Topologies.spacefillingcurve(mesh),
        )
```

Issue #1283, opened 2023-06-02
DSS fails inference


charleskawczynski:
## Describe the bug
dss fails inference:
```julia
julia> JET.@test_opt CA.dss!(Yₜ, p, t)
JET-test failed at REPL[66]:1
  Expression: #= REPL[66]:1 =# JET.@test_opt CA.dss!(Yₜ, p, t)
  ═════ 4 possible errors found ═════
  ┌ @ /Users/charliekawczynski/Dropbox/Caltech/work/dev/CliMA/ClimaAtmos.jl/src/prognostic_equations/dss.jl:17 ClimaCore.Spaces.weighted_dss_start2!(Y.c, p.ghost_buffer.c)
  │┌ @ /Users/charliekawczynski/.julia/packages/ClimaCore/A0lxf/src/Fields/Fields.jl:400 ClimaCore.Spaces.weighted_dss_start!(field, ghost_buffer)
  ││┌ @ /Users/charliekawczynski/.julia/packages/ClimaCore/A0lxf/src/Fields/Fields.jl:366 ClimaCore.Spaces.weighted_dss_start!(ClimaCore.Fields.field_values(field), ClimaCore.Fields.axes(field), dss_buffer)
  │││┌ @ /Users/charliekawczynski/.julia/packages/ClimaCore/A0lxf/src/Spaces/dss.jl:261 ClimaCore.Spaces.weighted_dss_start!(data, space, ClimaCore.Spaces.horizontal_space(space), dss_buffer)
  ││││┌ @ /Users/charliekawczynski/.julia/packages/ClimaCore/A0lxf/src/Spaces/dss.jl:282 ClimaCore.Spaces.dss_local_ghost!(device, dss_buffer.perimeter_data, ClimaCore.Spaces.perimeter(hspace), hspace.topology)
  │││││┌ @ /Users/charliekawczynski/.julia/packages/ClimaCore/A0lxf/src/Spaces/dss.jl:874 zero_data = ClimaCore.Spaces.map(ClimaCore.Spaces.zero, ClimaCore.Spaces.slab(perimeter_data, 1, 1)[1])
  ││││││┌ @ namedtuple.jl:219 map(tuple(f), map(Base.Tuple, tuple(tuple(nt), nts...))...)
  │││││││┌ @ tuple.jl:224 map(f, Base.tail(t))
  ││││││││┌ @ tuple.jl:224 map(f, Base.tail(t))
[...]

PR #1295, opened 2023-06-06
Add more perf dss tests

charleskawczynski:
Unfortunately, even though #1198 fixed some JET failures, others (added here) ended up with worse performance. The fields tested here were taken from ClimaAtmos, where this inference failure was observed. Hopefully fixing this JET test will close out #1283 for good.

### DIFF
--- a/test/Operators/spectralelement/benchmark_kernels.jl
+++ b/test/Operators/spectralelement/benchmark_kernels.jl
@@ -105,43 +105,41 @@ end
 ##### scalar dss!
 kernel_scalar_dss_array!(args) = kernel_copyto!(args)
 function kernel_scalar_dss!(args)
-    (; ϕ_buffer) = args.buffers
-    (; ϕ) = args
-    Spaces.weighted_dss_start!(ϕ, ϕ_buffer)
-    Spaces.weighted_dss_internal!(ϕ, ϕ_buffer)
-    Spaces.weighted_dss_ghost!(ϕ, ϕ_buffer)
-    return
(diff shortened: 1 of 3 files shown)


charleskawczynski:
This is actually seg-faulting for me locally

charleskawczynski:
Turns out this is a DSS bug: we’re getting an OOB error (when running with check bounds, for certain types of fields). I’ll open an issue and add this test as a broken example

charleskawczynski:
It was my mistake, I passed in the wrong buffer. That seems a bit dangerous. I'll open an issue to add some checks for this.

charleskawczynski:
bors try

test/Operators/spectralelement/benchmark_kernels.jl
```diff
@@ -145,3 +145,25 @@ function kernel_ntuple_field_dss!(args)
...
+end
+
+##### complicated_field dss!
+kernel_complicated_field_dss_array!(args) = kernel_copyto!(args)
+function kernel_complicated_field_dss!(args)
+    (; f_comp_buffer) = args.buffers
+    (; f_comp) = args
+    Spaces.weighted_dss_start!(f_comp, f_comp_buffer)
+    Spaces.weighted_dss_internal!(f_comp, f_comp_buffer)
+    Spaces.weighted_dss_ghost!(f_comp, f_comp_buffer)
```
sriharshakandala, line 167: We can just do `Spaces.weighted_dss(...)` unless there is a good reason to separate them into three function calls!

PR #1305, opened 2023-06-09
remove final column calls in FD operators

simonbyrne:
Avoids creating unnecessary views. Makes consistent with changes in #1200.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -38,6 +38,7 @@ right_face_boundary_idx(arg) = right_face_boundary_idx(axes(arg))
 left_center_boundary_idx(arg) = left_center_boundary_idx(axes(arg))
 right_center_boundary_idx(arg) = right_center_boundary_idx(axes(arg))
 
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
Good catch!!!

src/Operators/finitedifference.jl
```diff
@@ -59,11 +62,12 @@ Base.@propagate_inbounds function Geometry.LocalGeometry(
     idx::PlusHalf,
     hidx,
 )
-    i = idx.i + 1
+    v = idx + half
     if Topologies.isperiodic(Spaces.vertical_topology(space))
-        i = mod1(i, length(space))
+        v = mod1(v, length(space))
```
charleskawczynski, line 67: Wait, is `mod1` defined for `PlusHalf`?

charleskawczynski: Oh, right, `v = (idx + half)::Int` so this is fine

charleskawczynski:
@simonbyrne! Did you see what happened to the implicit stencil tests? We went from 36 min (main) to 14 min (this PR) 🎉

PR #1306, opened 2023-06-09
Fix dss inference

charleskawczynski:
This PR fixes dss inference for the complicated field (🎉) by removing the extra function call to `tuplemap`.

I remember we kept `tuplemap` around for some reason-- I think it was so that we could try forcing specialization when we wanted to. Being that it simply calls `map`, I'm wondering if we should just remove the extra function call altogether and delete the `tuplemap` function. Thoughts @simonbyrne?

Closes #1283.

PR #1308, opened 2023-06-11
Remove tuplemap

charleskawczynski:
This PR removes `tuplemap`, which was shown to improve inference in #1306. Perhaps we should just get rid of it?


charleskawczynski:
bors try

simonbyrne (approved):

PR #1321, opened 2023-06-13
Run all unit tests in buildkite

charleskawczynski:
This PR:
 - moves all of the unit tests into buildkite.
 - Deletes the `test/gpu/device.jl` file, since it's from a deprecated module
 - Moves `test/gpu/data.jl → test/DataLayouts/cuda.jl`
 - Attempts to improve the grouping and labeling of the tests in the buildkite pipeline

I've also incorporated #1319.

This is a pretty big step towards #837. Making bors not depend on GHA would effectively close the issue.


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
I think these failures are due to flakey CI. bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

simonbyrne:
can we split out the tests which are triggered from `runtests` and those which are not? maybe move them to a `test/extra` directory?

sriharshakandala:
Is it practical to separate out tests that need to be run with `--check-bounds=yes`?

charleskawczynski:
> can we split out the tests which are triggered from `runtests` and those which are not? maybe move them to a `test/extra` directory?

I was hoping to handle that more systematically in a follow-up PR. Right now in main, we don't have a clear partition what is included in `runtests.jl` / `runtests_expensive.jl` and what is not (basically, the distinction is what is included in buildkite vs GHA).

My thought was that we could parse the buildkite yaml file and include all unit tests (based on `Unit: ` groups) in `runtests.jl`. We could also make sure to use the same environment variables, so that the tests are actually doing the same thing (they're not, at the moment) except for running on a different OS.

charleskawczynski:
> Is it practical to separate out tests that need to be run with `--check-bounds=yes`?

Ah, good catch, I wasn't consistent with that. I think all unit tests should have `--check-bounds=yes` and all non-unit tests can omit `--check-bounds`. Does that sound okay?

simonbyrne:
It is possible that #1322 may fix the `--check-bounds` issue.

charleskawczynski:
> It is possible that #1322 may fix the `--check-bounds` issue.

What's the issue with `--check-bounds`?

juliasloan25 (approved):
This looks good to me

simonbyrne (approved):

sriharshakandala (approved):

Issue #1230, opened 2023-05-06
Implicit Solver Interface Updates


dennisYatunin:
## Purpose

The interface we use to specify matrices and solve linear systems of equations for the implicit solver needs to be refactored and extended. This interface, which is currently spread across several files in ClimaCore and ClimaAtmos (`ClimaCore.jl/src/Operators/stencilcoefs.jl`, `ClimaCore.jl/src/Operators/pointwisestencil.jl`, `ClimaCore.jl/src/Operators/operator2stencil.jl`, `ClimaAtmos.jl/src/tendencies/implicit/wfact.jl`, and `ClimaAtmos.jl/src/tendencies/implicit/schur_complement_W.jl`) is unnecessarily convoluted and involves a large number of hardcoded assumptions, which makes it extremely challenging for new users to experiment with the implicit solver and to add new implicit tendencies. In particular, the next stage of EDMF development will involve adding many implicit tendencies to the atmosphere model, and our approximation of the total implicit tendency's Jacobian matrix will end up with a highly non-trivial sparsity pattern. In order for more than a single developer to be able to understand and implement the implicit solver for the dycore+EDMF, we will first need to make the changes outlined in this SDI. In addition, the land modeling team has been experimenting with their own implicit solver, and these changes will speed up their development and add the new functionality they require.

There is currently a draft PR that contains a detailed sketch of all the proposed changes: #1190

## Cost/benefits/risks

The cost/risk is development time. The benefit will be a significantly reduced complexity of implicit solver implementations, both in ClimaAtmos and in ClimaLSM. There will be a simple, well-documented interface to all of the numerical algorithms required by implicit solvers, which will allow user-facing code to be relatively short and easily extensible.

## Producers

@dennisYatunin

## Components

- Simplify the interface for band matrix fields and matrix multiplication using fields.
- Add support for specifying derivatives of two-argument operators and for specifying derivatives of operator boundary conditions. 
- Add a user-friendly representation of block matrices.
- Add an efficient linear solver for block matrices.
- Add thorough documentation and unit tests.

### `BandMatrixRow` and `MultiplyColumnwiseBandMatrixField`

The type we currently use to represent an element of a band matrix field is called `StencilCoefs`, and it is extremely confusing and poorly designed. Upon refactoring, this will become `BandMatrixRow`, which will have the following improvements:
[...]

PR #1322, opened 2023-06-13
simplify operator boundary condition handling

simonbyrne:
I needed to rework the operator boundary condition logic to get the Jacobian stencil computation to work on GPUs.

Fixes #1173 

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1326, opened 2023-06-14
Add new MatrixFields module, along with unit tests and performance tests

dennisYatunin:
# Purpose

This PR encompasses the first two tasks of the implicit interface updates [SDI](https://github.com/CliMA/ClimaCore.jl/issues/1230). That is, it refactors `StencilCoefs` into `BandMatrixRow` and `ApplyStencil`/`ComposeStencil` into `MultiplyColumnwiseBandMatrixField`, and it also adds a comprehensive set of tests. In the process of refactoring, the code for low-level matrix operations has been simplified and useful new functionality has been added. In order to avoid conflicts with pre-existing code, the new data structures and functions have been placed in a separate module called `MatrixFields`; this module can be removed once we are ready to replace all of the old code.

### API improvements
- Matrix field rows now have simple constructors (e.g., `DiagonalMatrixRow` and `TridiagonalMatrixRow`) that can be used to construct matrix fields on the fly in broadcast expressions.
- Matrix fields are now automatically promoted to common types for basic arithmetic operations. This involves promoting the entries in the rows, as well as padding rows with zeros when rows have unequal numbers of entries. In addition, `LinearAlgebra.I` (or, rather, `LinearAlgebra.UniformScaling`) is automatically promoted to a `DiagonalMatrixRow` when used with `BandMatrixRow`s. So, for example, it is now possible to evaluate `DiagonalMartrixRow(1) - 2 * TridiagonalMatrixRow(3, 4, 5) - I / 6`.
- Matrix fields are now fully integrated with `RecursiveApply`, which allows a single matrix field to store multiple matrices by using `Tuple`s or `NamedTuple`s as entries instead of scalars. This is achieved by using the new `rzero` function instead of `zero`, and also by making `+`, `-`, `*`, and `/` call `radd`, `rsub`, `rmul`, and `rdiv` when used on `BandMatrixRow`s. This change will make it possible to, e.g., store the derivatives of all tracer tendencies with respect to vertical velocity as a single matrix field in `ClimaAtmos`, which will lead to simpler code and better performance.
- All matrix-matrix and matrix-vector multiplications can now be handled using the `⋅` operator, which is an alias for `MultiplyColumnwiseBandMatrixField()`. The implementation of this operator is much simpler than the implementation of `ApplyStencil`/`ComposeStencil`, and it is effectively localized to the single function `multiply_matrix_at_index`. In order to avoid inference failures for complex matrix field broadcast expressions, the recursion limit for this function needed to be disabled. Unfortunately, just as with the pre-existing code, inlining this function with `@propagate_inbounds` drastically increased compilation time, and it also caused inference failures for particularly complicated matrix broadcast expressions. Although not using `@propagate_inbounds` causes a 3--4x slowdown in the evaluation of matrix broadcast expressions, it avoids a 10--100x slowdown in their compilation.
- The matrix of the divergence operator can now be properly represented as a matrix of covectors, rather than a matrix of scalars. Since the divergence operator turns vectors into scalars, multiplying its matrix by a matrix/vector of vectors should return a matrix/vector of scalars. However, this is not as simple as just multiplying a covector by a vector and returning a scalar, since the vector may need to be projected onto the dual axis of the covector before the multiplication. So, matrix-matrix and matrix-vector multiplication is now implemented using the `rmul_with_projection` function instead of `rmul`. In general, this function works just like `rmul`, but, when the first argument is a covector (an `AdjointAxisVector`) and the second is a vector or higher-order tensor (an `AxisTensor`), it uses the local geometry to project the second argument before the multiplication. In the future, we may want to generalize this to higher-order cotensors.
- The `MatrixFields` module also includes functions for converting matrix fields to more conventional arrays that are compatible with `LinearAlgebra.jl`. The function `column_field2array` converts a matrix field defined on a column space to a `BandedMatrix` from `BandedMatrices.jl`, and it also converts a regular field defined on a column space to a `Vector`. The function `field2arrays` works for fields defined on multiple columns by calling `column_field2array` on each column. 
- Matrix fields now get displayed in the REPL as actual matrices. Specifically, when a matrix field contains `Number` entries, its first column is passed to `column_field2array_view`, which is identical to `column_field2array`, except that it allocates less memory by returning a view of the field's underlying data. Although `column_field2array` and `column_field2array_view` work for matrix fields with `Tuple`/`NamedTuple` entries, the `show` method for a `BandedMatrix` crashes when displaying such entries. In addition, this `show` method results in somewhat unreadable output for other struct entries, like vectors and covectors. So, all non-`Number` matrix fields still get displayed in the REPL like regular fields.

### Related ClimaCore Changes
- To ensure that promotion and conversion of `BandMatrixRow` entries works correctly for nested `Tuple`/`NamedTuple` entries, the `rpromote_type` and `rconvert` functions have been added to RecursiveApply. The first function uses `rmaptype` to recursively call `promote_type`, and the second function uses `rmap` and `rzero` to make a more type-stable version of Julia's built-in `convert` function.
- To support manipulating matrices of covectors (and, in the future, matrices of higher-order cotensors), all basic arithmetic operations have been defined for `AdjointAxisTensor`. This includes `+`, `-`, `*`, `/`, `\`, and `==`, as well as the `zero` function. The new methods undo the adjoint and fall back to the methods for `AxisTensor`.
- To simplify the setup of the non-scalar matrix tests, `map` has been generalized from only working for a single `Field` to working for multiple `Field`s.

### Tests
- All of the following test files check for correctness, allocations, and type instabilities.
- The file `test/MatrixFields/band_matrix_row.jl` tests low-level operations with `BandMatrixRow`s, ensuring that constructors, conversions, arithmetic operations, and combinations with `LinearAlgebra.I` work for numbers and nested `Tuple`/`NamedTuple` values.
- The file `test/MatrixFields/rmul_with_projection.jl` tests `rmul_with_projection` for all currently supported combinations of scalars, vectors, tensors, and covectors, as well as several combinations of nested `Tuple`/`NamedTuple` values.
- The file `test/MatrixFields/field2array.jl` checks that `column_field2array`, `column_field2array_view`, and `field2arrays` work as expected for very simple matrix fields.
- The file `test/MatrixFields/matrix_field_broadcasting.jl` tests a wide range of scalar and non-scalar matrix field broadcast expressions. It compares each scalar matrix field broadcast expression against an equivalent implementation using `field2arrays` and LinearAlgebra's `mul!`, making sure that the matrix field implementation is roughly as performant as the array implementation. It also compares each non-scalar matrix field broadcast expression (e.g., an expression with matrices of covectors or `NamedTuple`s) against an equivalent scalar matrix field implementation.
- I have also checked that the new implementation is at least as performant as the pre-existing implementation for those cases that the pre-existing API can support. For example, to compare the two implementations for the "diagonal matrix times bi-diagonal matrix times tri-diagonal matrix times quad-diagonal matrix times vector" test case, run the following code after `test/MatrixFields/matrix_field_broadcasting.jl`:
    ```
    using ClimaCore: MatrixFields, Operators
    function compare_to_old_implementation()
        ᶜᶜmat, ᶜᶠmat, ᶠᶠmat, ᶠᶜmat, ᶜvec, ᶠvec = random_test_fields(Float64)
        result = @. ᶜᶜmat ⋅ ᶜᶠmat ⋅ ᶠᶠmat ⋅ ᶠᶜmat ⋅ ᶜvec
        inputs = (ᶜᶜmat, ᶜᶠmat, ᶠᶠmat, ᶠᶜmat, ᶜvec)
        func!(result, ᶜᶜmat, ᶜᶠmat, ᶠᶠmat, ᶠᶜmat, ᶜvec) =
            @. result = ᶜᶜmat ⋅ ᶜᶠmat ⋅ ᶠᶠmat ⋅ ᶠᶜmat ⋅ ᶜvec
        time, allocs = @benchmark call_func(func!, result, inputs)
        
        to_old_version(row::BMR) where {BMR <: MatrixFields.BandMatrixRow} =
            Operators.StencilCoefs{MatrixFields.outer_diagonals(BMR)...}(row.entries)
        apply = Operators.ApplyStencil()
        compose = Operators.ComposeStencils()
        ᶜᶜmat_old = to_old_version.(ᶜᶜmat)
        ᶜᶠmat_old = to_old_version.(ᶜᶠmat)
        ᶠᶠmat_old = to_old_version.(ᶠᶠmat)
        ᶠᶜmat_old = to_old_version.(ᶠᶜmat)
        result_old = @. apply(compose(compose(compose(ᶜᶜmat_old, ᶜᶠmat_old), ᶠᶠmat_old), ᶠᶜmat_old), ᶜvec)
        inputs_old = (ᶜᶜmat_old, ᶜᶠmat_old, ᶠᶠmat_old, ᶠᶜmat_old, ᶜvec)
        func!_old(result_old, ᶜᶜmat_old, ᶜᶠmat_old, ᶠᶠmat_old, ᶠᶜmat_old, ᶜvec) =
            @. result_old = apply(compose(compose(compose(ᶜᶜmat_old, ᶜᶠmat_old), ᶠᶠmat_old), ᶠᶜmat_old), ᶜvec)
        time_old, allocs_old = @benchmark call_func(func!_old, result_old, inputs_old)

        @assert allocs == allocs_old == 0

        @info "New implementation time: $time s"
        @info "Old implementation time: $time_old s"
    end
    compare_to_old_implementation()
    ```
    This code prints out the following:
    ```
    [ Info: New implementation time: 1.5848e-5 s
    [ Info: Old implementation time: 1.5283e-5 s
    ```


src/RecursiveApply/RecursiveApply.jl
```diff
@@ -9,17 +9,60 @@ module RecursiveApply
 
 export ⊞, ⊠, ⊟
 
+# These functions need to be generated for type stability (since T.parameters is
```
charleskawczynski, line 12: Let's add these changes (in RecursiveApply) as a separate PR, with tests.

src/Geometry/axistensors.jl
```diff
@@ -269,10 +269,30 @@ Base.propertynames(x::AxisVector) = symbols(axes(x, 1))
...
+@inline +(a::AdjointAxisTensor) = (+a')'
+@inline -(a::AdjointAxisTensor) = (-a')'
+@inline +(a::AdjointAxisTensor, b::AdjointAxisTensor) = (a' + b')'
+@inline -(a::AdjointAxisTensor, b::AdjointAxisTensor) = (a' - b')'
+@inline *(a::Number, b::AdjointAxisTensor) = (a * b')'
+@inline *(a::AdjointAxisTensor, b::Number) = (a' * b)'
+@inline /(a::AdjointAxisTensor, b::Number) = (a' / b)'
+@inline \(a::Number, b::AdjointAxisTensor) = (a' \ b)'
+
+@inline (==)(a::AdjointAxisTensor, b::AdjointAxisTensor) = a' == b'
```
charleskawczynski, line 292: I'd like for us to add tests in the test suite for confirming the correct return type.

src/MatrixFields/band_matrix_row.jl
```diff
@@ -0,0 +1,153 @@
...
+        "TridiagonalMatrixRow{$T}"
+    elseif BMR <: QuaddiagonalMatrixRow
+        "QuaddiagonalMatrixRow{$T}"
+    elseif BMR <: PentadiagonalMatrixRow
+        "PentadiagonalMatrixRow{$T}"
+    else
+        "BandMatrixRow{$ld, $bw, $T}"
+    end
+    print(io, string)
+end
```
charleskawczynski, line 50: Unfortunately, overloading `Base.show` for `::Type{T}` invalidates `show` in Base. Can we change this to `Base.show(io::IO, ::BandMatrixRow)` instead?

src/RecursiveApply/RecursiveApply.jl
dennisYatunin: Should I include `rmul_with_projection` in that PR? It would probably make more sense to have it be part of RecursiveApply, though that would require loading the RecursiveApply module after the Geometry module.

src/MatrixFields/band_matrix_row.jl
dennisYatunin: The goal here was actually to change the way matrix fields are printed, rather than individual rows. For example:
```
julia> ᶠᶠmat
TridiagonalMatrixRow{Float64}-valued Field:
  entries: 
    1: [0.667658, 0.766307, 0.866266, 0.0461284, 0.439407, 0.390036, 0.0762841, 0.878829, 0.0061787, 0.0188659  …  0.398943, 0.871208, 0.0441692, 0.878441, 0.982707, 0.0220367, 0.712676, 0.0229924, 0.560176, 0.692064]
    2: [0.95014, 0.690321, 0.463986, 0.189937, 0.00750967, 0.00224279, 0.5567, 0.830047, 0.527776, 0.519096  …  0.65025, 0.284222, 0.0891259, 0.00904144, 0.115214, 0.612847, 0.625339, 0.38462, 0.0864744, 0.369243]
    3: [0.503339, 0.620339, 0.343298, 0.918134, 0.0660757, 0.776861, 0.3916, 0.0791701, 0.496441, 0.0822908  …  0.735152, 0.521021, 0.970042, 0.0290453, 0.650956, 0.53432, 0.610975, 0.368205, 0.930298, 0.133497]
```
Should I just directly overwrite `Base.show(io::IO, ::Field)` instead?

src/RecursiveApply/RecursiveApply.jl
charleskawczynski: No, just the RecursiveApply part. Projections aren't loaded until Geometry, so better keep that where it is for now.

src/MatrixFields/band_matrix_row.jl
charleskawczynski: > Should I just overwrite ...

Yeah, that's probably a lot better

src/MatrixFields/matrix_field_utils.jl
```diff
@@ -0,0 +1,85 @@
...
+    space = axes(field)
+    column_indices = if space isa Spaces.FiniteDifferenceSpace
+        (((1, 1), 1),)
+    elseif space isa Spaces.ExtrudedFiniteDifferenceSpace
+        (Spaces.all_nodes(Spaces.horizontal_space(space))...,)
+    else
+        error("Invalid space type: $(typeof(space).name.wrapper)")
+    end
+    return map(column_indices) do ((i, j), h)
+        return column_field2array(Spaces.column(field, i, j, h))
```
charleskawczynski, line 83:
```suggestion
        column_field2array(Spaces.column(field, i, j, h))
```

src/MatrixFields/matrix_multiplication.jl
```diff
@@ -0,0 +1,338 @@
...
+
+function boundary_width(
+    ::MultiplyColumnwiseBandMatrixField,
+    ::TopLeftMatrixCorner,
+    matrix1,
+    arg,
+)
+    ld1, _ = outer_diagonals(eltype(matrix1))
+    return max((left_idx(axes(arg)) - ld1) - left_idx(axes(matrix1)), 0)
+end
```
simonbyrne, line 154: I've changed how this works in #1322 which simplifies the logic, can you use that approach?

dennisYatunin:
Couple of updates:
- All of the reviewer comments have been addressed.
- I added a file that tests low-level operations on `BandMatrixRow`s. This showed that type promotion was not working as expected for `Tuple`/`NamedTuple` entries, and it showed that Julia's default `convert` is not type stable for nested `Tuple`/`NamedTuple` entries. So, I also added `rpromote_type` and `rconvert` to RecursiveApply.
- I added a file that tests field-to-array conversions.
- I removed `rmul_with_projection_return_type`, which I was using to get the return type of `rmul_with_projection`, and replaced it with `Base._return_type`, which is how `Base.Broadcast` gets the return types of broadcast expressions.
- I made matrix fields of `Number`s display in the REPL as matrices by overwriting the `show` method for `Fields.Field{<:Fields.AbstractData{<:BandMatrixRow}}` and calling `column_field2array_view`. @simonbyrne, Charlie said you might want to take a look at this change.

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

test/runtests.jl
```diff
@@ -75,6 +75,12 @@ if !Sys.iswindows()
     @safetestset "Hybrid - dss opt" begin @time include("Operators/hybrid/dss_opt.jl") end
     @safetestset "Hybrid - opt" begin @time include("Operators/hybrid/opt.jl") end
 
+    @safetestset "MatrixFields - BandMatrixRow" begin @time include("MatrixFields/band_matrix_row.jl") end
+    @safetestset "MatrixFields - rmul_with_projection" begin @time include("MatrixFields/rmul_with_projection.jl") end
+    @safetestset "MatrixFields - field2arrays" begin @time include("MatrixFields/field2arrays.jl") end
```
charleskawczynski, line 80: Let's add these tests separately to buildkite. Please see https://github.com/CliMA/ClimaCore.jl/pull/1340 for the direction I'm hoping runtests.jl moves to.

dennisYatunin:
bors try

dennisYatunin:
bors try

test/MatrixFields/matrix_field_broadcasting.jl
```diff
@@ -0,0 +1,816 @@
...
+                maximum(zip(array, ref_array)) do (value, ref_value)
+                    Int(abs(value - ref_value) / eps(ref_value))
+                end
+            end
+
+        @info "$test_name:\n\tBest Time = $best_time s\n\tBest Reference Time \
+               = $best_ref_time s\n\tMaximum Error = $max_error eps"
+
+        # Test that set_result! is performant compared to ref_set_result!.
+        @test best_time < time_ratio_limit * best_ref_time
```
charleskawczynski, line 85: @dennisYatunin, we don't need to use a test here-- these tests are unfortunately pretty flakey. I'm fine with changing this to be an `@info` print to the log of the results. Keep the correctness, allocation, and JET tests though!

dennisYatunin: I think I've actually been getting fairly consistent timing results, since I'm taking the minimum time from a large number of trials. There's a big difference between running locally and on Buildkite, but that's probably just because the cluster can do vectorized operations more quickly than my computer. If you want, though, I can just comment out this test; the info statement above it prints all the relevant results.

charleskawczynski: Okay, either way is fine with me-- if you leave the test in, just include a buffer that we can bump to help alleviate flakiness from timing variance.

charleskawczynski:
@dennisYatunin, there's still invalidations, can you look into/fix this?

charleskawczynski:
You can use

```julia
using SnoopCompileCore
invalidations = @snoopr begin
    using ClimaCore
    nothing
end;

import SnoopCompile
import PrettyTables # load report_invalidations
SnoopCompile.report_invalidations(;
    invalidations,
    n_rows = 0,
)
```

But I think there's a lot of invalidations in Julia 1.8.5, so you may want to run this on Julia 1.9

charleskawczynski:
@dennisYatunin, if @simonbyrne is not able to review this, can you test this branch out on ClimaAtmos before merging? Then paste a link to the build in the comments on this PR. Thank you!

simonbyrne:
Looking at the high-level, I think this is okay. As @charleskawczynski mentioned, it would be good to figure out what the invalidations are.

Can you add a GPU test as well? At a high-level, it looks like it should work, but may require some tweaks.

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

src/MatrixFields/band_matrix_row.jl
```diff
@@ -0,0 +1,138 @@
...
+denoting the type of the row's entries):
+- `DiagonalMatrixRow{T}`
+- `BidiagonalMatrixRow{T}`
+- `TridiagonalMatrixRow{T}`
+- `QuaddiagonalMatrixRow{T}`
+- `PentadiagonalMatrixRow{T}`
+"""
+struct BandMatrixRow{ld, bw, T} # bw is the bandwidth (the number of diagonals)
+    entries::NTuple{bw, T}
+    BandMatrixRow{ld, bw, T}(entries::NTuple{bw, Any}) where {ld, bw, T} =
```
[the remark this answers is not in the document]
dennisYatunin: I was hoping to match Julia's default constructor as closely as possible, so that there would be minimal behavioral changes when we get rid of this inner constructor. Specifically, the default constructor is `BandMatrixRow{ld, bw, T}(entries) where {ld, bw, T} = new{ld, bw, T}(convert(NTuple{bw, T}, entries))`. In theory, I could have just replaced `convert` with `rconvert` and left it as is, but I decided to also check that `entries` is already an `NTuple{bw}` just to be safe. If I were to replace the `Any` with `T`, then this would be much more restrictive than the default constructor, and we wouldn't be able to do something like `@. TridiagonalMatrixRow(0, diagonal_field, 0)`, since we'd need to wrap the `0`'s with `FT`s.

src/MatrixFields/matrix_field_utils.jl
```diff
@@ -0,0 +1,141 @@
...
+    n_rows = Spaces.nlevels(space)
+    n_cols = n_rows + bottom_corner_matrix_d
+    matrix_ld = field_ld + matrix_d_minus_field_d
+    matrix_ud = field_ud + matrix_d_minus_field_d
+    matrix_ld <= 0 && matrix_ud >= 0 ||
+        error("BandedMatrices.jl does not yet support matrices that have \
+               diagonals with indices in the range $matrix_ld:$matrix_ud")
+
+    return n_rows, n_cols, matrix_ld, matrix_ud
+end
```
charleskawczynski, line 27:
```suggestion
# Find the diagonal index of the value that ends up in the bottom-right
# corner of the matrix, as well as the amount by which the field's diagonal
# indices get shifted when it is converted into a matrix.
bottom_corner_matrix_d(::Spaces.Staggering, ::Int) = 0          # field is either a center-to-center or face-to-face matrix
bottom_corner_matrix_d(::Spaces.CellCenter, i::PlusHalf) = 1    # field is a face-to-center matrix
bottom_corner_matrix_d(::Spaces.CellFace, i::PlusHalf) = -1     # field is a center-to-face matrix
matrix_d_minus_field_d(::Spaces.Staggering, ::Int) = 0          # field is either a center-to-center or face-to-face matrix
matrix_d_minus_field_d(::Spaces.CellCenter, i::PlusHalf) = half # field is a face-to-center matrix
matrix_d_minus_field_d(::Spaces.CellFace, i::PlusHalf) = -half  # field is a center-to-face matrix

function banded_matrix_info(field)
    field_ld, field_ud = outer_diagonals(eltype(field))
    space = axes(field)
    staggering = space.staggering
    n_rows = Spaces.nlevels(space)
    n_cols = n_rows + bottom_corner_matrix_d(staggering, field_ld)
    matrix_ld = field_ld + matrix_d_minus_field_d(staggering, field_ld)
    matrix_ud = field_ud + matrix_d_minus_field_d(staggering, field_ld)
    matrix_ld <= 0 && matrix_ud >= 0 ||
        error("BandedMatrices.jl does not yet support matrices that have \
               diagonals with indices in the range $matrix_ld:$matrix_ud")
    return n_rows, n_cols, matrix_ld, matrix_ud
end
```
I might break that double union splitting into dispatch, I think it maybe reads a bit easier. (please double check my work there)

src/MatrixFields/matrix_field_utils.jl
```diff
@@ -0,0 +1,141 @@
...
+            _BandedMatrix(field_data_transpose, n_cols, matrix_ud, -matrix_ld)
+        return permutedims(matrix_transpose)
+        # TODO: Despite not copying any data, this function still allocates a
+        # small amount of memory because of _BandedMatrix and permutedims.
+    else # field represents a vector
+        return vec(reinterpret(eltype(field), parent(field)'))
+    end
+end
+
+all_columns(::Fields.FiniteDifferenceField) = (((1, 1), 1),)
```
charleskawczynski, line 89: Is this supposed to be `::Fields.ColumnField`?

src/MatrixFields/matrix_field_utils.jl
```diff
@@ -0,0 +1,141 @@
...
+all_columns(field::Fields.ExtrudedFiniteDifferenceField) =
+    (Spaces.all_nodes(Spaces.horizontal_space(axes(field)))...,)
```
charleskawczynski, line 91: Does this need to be splatted? Splatting large collections can be very costly. `all_nodes` returns an iterator, can we not map over it?

src/MatrixFields/matrix_field_utils.jl
dennisYatunin: It turns out that specializing on `Fields.ColumnField` and `Fields.ExtrudedFiniteDifferenceField` results in a method ambiguity, since the first is specializing on the data container type and the second is specializing on the space type. In order to avoid failing the Aqua tests, the specialization needs to be on `Fields.FiniteDifferenceField` and `Fields.ExtrudedFiniteDifferenceField`.

src/MatrixFields/matrix_field_utils.jl
dennisYatunin: Ok, yeah, I agree the current version is hard to read. Will do!

src/MatrixFields/matrix_field_utils.jl
dennisYatunin: I had this return a tuple to avoid allocations in the `matrix_field_broadcasting.jl` reference benchmarks. However, it looks like I can also avoid those allocations by using a `for`-loop over the arrays instead of calling `foreach`, so I'll do that instead and remove the splatting.

dennisYatunin:
bors try

src/MatrixFields/matrix_field_utils.jl
charleskawczynski: So, `all_columns(::Fields.FiniteDifferenceField) = (((1, 1), 1),)` was written generically to avoid a method ambiguity?

dennisYatunin: Yes. And I also wanted to make sure that the only two valid options are `Fields.ColumnField` and `Fields.ExtrudedFiniteDifferenceField`, since a `BandMatrixRow` field on any other space cannot be interpreted as a matrix.

dennisYatunin:
bors try

src/MatrixFields/matrix_field_utils.jl
charleskawczynski: Method ambiguities can be fixed by defining the intersecting case, so, I think what we want is: 
```julia
all_columns(::DataLayouts.DataColumn, ::Spaces.AbstractSpace) = (((1, 1), 1),)
all_columns(::DataLayouts.AbstractData, space::Spaces.ExtrudedFiniteDifferenceSpace) =
    Spaces.all_nodes(Spaces.horizontal_space(space))
all_columns(::DataLayouts.DataColumn, ::Spaces.ExtrudedFiniteDifferenceSpace) = (((1, 1), 1),)
```
This last one should fix the ambiguity.

charleskawczynski: This way the code paths are a bit more explicit / clear.

charleskawczynski: (at the cost of verbosity)

dennisYatunin: Isn't a field on a FiniteDifferenceSpace always going to be stored on a DataColumn, though? That's the assumption I made when writing [this code](https://github.com/CliMA/ClimaCore.jl/blob/eef71dc05a0672d1736e128aeb500c4a18e57727/src/Operators/thomas_algorithm.jl#L23-L49) for the Thomas algorithm. I don't see why we need to handle a column of data stored on a generic AbstractSpace.

charleskawczynski: I think it depends how it's sliced. We can use `bycolumn` to make column views, or `byslab` to make slab views-- those will have different data layouts

charleskawczynski: I'm not making the argument that this will ever be done, I think it's just more defensive/explicit

dennisYatunin: Don't `bycolumn` and `byslab` also change the space, though? If someone were to call `column_field2array` from a `byslab`, the correct behavior would be to throw an error, since it is not possible to interpret a horizontal slab of `BandMatrixRows` as a matrix.

dennisYatunin: With the version you suggested, the `all_columns` function wouldn't throw an error in that case, but the `column_field2array` function still would.

dennisYatunin: If I make the changes you mentioned, how should I change the following condition used lower down in this file:
```
    is_matrix_field =
        field isa Fields.FiniteDifferenceField ||
        field isa Fields.ExtrudedFiniteDifferenceField
```
If a field of `BandMatrixRow`s isn't located on one of these two spaces, then it can't be interpreted as a matrix, and it needs to be printed differently. Should I use `ColumnField` instead of `FiniteDifferenceField` here?

dennisYatunin:
bors try

src/MatrixFields/matrix_field_utils.jl
charleskawczynski: > Isn't a field on a FiniteDifferenceSpace always going to be stored on a DataColumn, though?

Not based on its definition: 
```julia
const FiniteDifferenceField{V, S} =
    Field{V, S} where {V <: AbstractData, S <: Spaces.FiniteDifferenceSpace}
```

I'll take a look at `column_field2array` and the `is_matrix_field` part today.

src/MatrixFields/matrix_field_utils.jl
```diff
@@ -0,0 +1,139 @@
...
+
+            # Find the rows for which diagonal_data[row] is in the matrix.
+            # Note: The matrix index (1, 1) corresponds to the diagonal index 0,
+            # and the matrix index (n_rows, n_cols) corresponds to the diagonal
+            # index n_cols - n_rows.
+            first_row = matrix_d < 0 ? 1 - matrix_d : 1
+            last_row = matrix_d < n_cols - n_rows ? n_rows : n_cols - matrix_d
+
+            diagonal_data_view = view(diagonal_data, first_row:last_row)
+            @allowscalar copyto!(matrix_diagonal, diagonal_data_view)
```
charleskawczynski, line 58: Is this never called in a hot loop? Why do we need `@allowscalar`?

charleskawczynski: Ah, this is just used for printing, nevermind.

src/MatrixFields/matrix_multiplication.jl
```diff
@@ -0,0 +1,381 @@
...
+Operators.right_interior_idx(
+    space::Spaces.AbstractSpace,
+    ::MultiplyColumnwiseBandMatrixField,
+    ::BottomRightMatrixCorner,
+    matrix1,
+    arg,
+) = matrix_right_interior_index(space, outer_diagonals(eltype(matrix1))[2])
+
+function rmul_type(::Type{T1}, ::Type{T2}, ::Type{LG}) where {T1, T2, LG}
+    type = Base._return_type(rmul_with_projection, Tuple{T1, T2, LG})
```
simonbyrne, line 203: I'm not so keen on reaching into internals like this. Can we work around it somehow?

dennisYatunin:
bors try

dennisYatunin:
@simonbyrne @charleskawczynski I've removed the use of `Base._return_type` and replaced it with `mul_return_type` and `rmul_return_type`. I also added a few new tests for this. Could you please take a look at the relevant changes?

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
Also, it looks like the new invalidations are due to adding BandedMatrices.jl as a dependency to ClimaCore, because they remain unchanged when I uncomment the new code in the `MatrixFields` module. Since this dependency is needed for printing and array-based utility functions, there isn't really a good way to get rid of the invalidations. The total invalidation count is only increasing by a fraction of a percent, though, so this isn't a big issue.

simonbyrne:
I'm broadly okay with this, but it is pretty complex, so it could do with a few more signposts so that others can find their way around. Could you expand the docs a bit to explain how it works?

src/MatrixFields/rmul_with_projection.jl
```diff
@@ -0,0 +1,160 @@
...
+    mul_with_projection(x, y, lg)
+
+axis_tensor_type(::Type{T}, ::Type{Tuple{A1}}) where {T, A1} =
+    Geometry.AxisVector{T, A1, SVector{length(A1.instance), T}}
+function axis_tensor_type(::Type{T}, ::Type{Tuple{A1, A2}}) where {T, A1, A2}
+    N1 = length(A1.instance)
+    N2 = length(A2.instance)
+    S = SMatrix{N1, N2, T, N1 * N2}
+    return Geometry.Axis2Tensor{T, Tuple{A1, A2}, S}
+end
```
charleskawczynski, line 42: Should we use `StaticArraysCore.Size` instead of `A1/A2.instance`? I would even prefer writing a small helper, `sa_len(::Type{SA.SVector{N}}) where {N} = N` to get the length.
Also, it might be nice if we added type restrictions on what `A1`/`A2` are (I assume StaticArrays?)

dennisYatunin: `A1` and `A2` are both of type `Geometry.AbstractAxis`, and the `Geometry` module only has a method for `Base.length(ax::AbstractAxis{I}) where {I} = length(I)`. We could just add a method to `Geometry` for getting the length from the type, like `_length(::Type{<:AbstractAxis{I}}) where {I} = length(I)`?

charleskawczynski: Yeah, I think I'd prefer that

dennisYatunin:
@simonbyrne @charleskawczynski I've added a new `MatrixFields` section to the docs, where I put all the docstrings from this module. I also moved the derivation of the matrix multiplication operator's algorithm there, and I added a summary docstring for the entire module. Is there anything else I should do before merging?

dennisYatunin:
bors try

dennisYatunin:
bors try-

dennisYatunin:
bors try

dennisYatunin:
bors try

simonbyrne (approved):

PR #1335, opened 2023-06-16
Cache intermediate variables in topo example

charleskawczynski:
This PR caches intermediate variables in topo example (this example currently allocates 444.616 GiB).

Also, this example ([examples/hybrid/plane/topo_dc_2dinvariant_rhoe.jl](https://github.com/CliMA/ClimaCore.jl/pull/1335/files#diff-72d2194427b23722f2ca13c4686e0bfe553a5c42b4825390ac018de45b9dff80)) seems to be exiting early due to numerical instability.

PR #1337, opened 2023-06-19
Inline some small dss functions

charleskawczynski:
Peeled off from #1335, this PR inlines some dss functions and applies `Base.@propagate_inbounds` to ensure we elide bounds checks. These are pretty small functions, so inlining is likely worth it.


charleskawczynski:
bors try

PR #1338, opened 2023-06-19
Apply propagate_inbounds to remaining SEM apply_operators

charleskawczynski:
This was peeled off from #1335, this PR applies `Base.propagate_inbounds` to ensure that we elide bound checks. These are kind of big functions, so maybe it's not worth the code gen? Let's at least see how the performance benchmarks look.


charleskawczynski:
bors try

simonbyrne:
I'm not sure we need it, since we have the `@inbounds inside

charleskawczynski:
Okay, closing until we have a way to measure that these macros actually do anything.

PR #1348, opened 2023-06-26
Add GPU Thomas solver

sriharshakandala:

### DIFF
--- a/examples/hybrid/schur_complement_W.jl
+++ b/examples/hybrid/schur_complement_W.jl
@@ -154,16 +154,7 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
 
         @. xᶠ𝕄 = bᶠ𝕄 + dtγ * (apply(∂ᶠ𝕄ₜ∂ᶜρ, bᶜρ) + apply(∂ᶠ𝕄ₜ∂ᶜ𝔼, bᶜ𝔼))
 
-        # TODO: Do this with stencil_solve!.
-        Ni, Nj, _, _, Nh = size(Spaces.local_geometry_data(axes(xᶜρ)))
-        for h in 1:Nh, j in 1:Nj, i in 1:Ni
-            xᶠ𝕄_column_view = parent(Spaces.column(xᶠ𝕄, i, j, h))
-            S_column = Spaces.column(S, i, j, h)
-            @views S_column_array.dl .= parent(S_column.coefs.:1)[2:end]
-            S_column_array.d .= parent(S_column.coefs.:2)
-            @views S_column_array.du .= parent(S_column.coefs.:3)[1:(end - 1)]
-            ldiv!(lu!(S_column_array), xᶠ𝕄_column_view)
-        end
+        Operators.column_thomas_solve!(S, xᶠ𝕄)
 
         @. xᶜρ = -bᶜρ + dtγ * apply(∂ᶜρₜ∂ᶠ𝕄, xᶠ𝕄)
         @. xᶜ𝔼 = -bᶜ𝔼 + dtγ * apply(∂ᶜ𝔼ₜ∂ᶠ𝕄, xᶠ𝕄)
--- a/src/Operators/thomas_algorithm.jl
+++ b/src/Operators/thomas_algorithm.jl
@@ -1,0 +1,105 @@
+"""
+    column_thomas_solve!(A, b)
+
+Solves the linear system `A * x = b`, where `A` is a tri-diagonal matrix
+(represented by a `Field` of tri-diagonal matrix rows), and where `b` is a
+vector (represented by a `Field` of numbers). The data in `b` is overwritten
+with the solution `x`, and the upper diagonal of `A` is also overwritten with
+intermediate values used to compute `x`. The algorithm is described here:
+https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm.
+"""
+column_thomas_solve!(A, b) =
+    column_thomas_solve!(ClimaComms.device(axes(A)), A, b)
+
+column_thomas_solve!(::ClimaComms.AbstractCPUDevice, A, b) =
+    thomas_algorithm!(A, b)
+
+function column_thomas_solve!(::ClimaComms.CUDADevice, A, b)
+    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+    nthreads, nblocks = Spaces._configure_threadblock(Ni * Nj * Nh)
+    @cuda threads = nthreads blocks = nblocks thomas_algorithm_kernel!(A, b)
+end
+
+function thomas_algorithm_kernel!(
+    A::Fields.ExtrudedFiniteDifferenceField,
+    b::Fields.ExtrudedFiniteDifferenceField,
+)
+    idx = threadIdx().x + (blockIdx().x - 1) * blockDim().x
+    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+    if idx <= Ni * Nj * Nh
+        i, j, h = Spaces._get_idx((Ni, Nj, Nh), idx)
+        thomas_algorithm!(Spaces.column(A, i, j, h), Spaces.column(b, i, j, h))
+    end
+    return nothing
+end
+
+thomas_algorithm_kernel!(
+    A::Fields.FiniteDifferenceField,
+    b::Fields.FiniteDifferenceField,
+) = thomas_algorithm!(A, b)
+
+thomas_algorithm!(
+    A::Fields.ExtrudedFiniteDifferenceField,
+    b::Fields.ExtrudedFiniteDifferenceField,
+) = Fields.bycolumn(colidx -> thomas_algorithm!(A[colidx], b[colidx]), axes(A))
+
+function thomas_algorithm!(
+    A::Fields.FiniteDifferenceField,
+    b::Fields.FiniteDifferenceField,
+)
+    nrows = Spaces.nlevels(axes(A))
+    lower_diag = A.coefs.:1
+    main_diag = A.coefs.:2
+    upper_diag = A.coefs.:3
+
+    # first row
+    denominator = _getindex(main_diag, 1)
+    _setindex!(upper_diag, 1, _getindex(upper_diag, 1) / denominator)
+    _setindex!(b, 1, _getindex(b, 1) / denominator)
+
+    # interior rows
+    for row in 2:(nrows - 1)
+        numerator =
+            _getindex(b, row) -
+            _getindex(lower_diag, row) * _getindex(b, row - 1)
+        denominator =
+            _getindex(main_diag, row) -
+            _getindex(lower_diag, row) * _getindex(upper_diag, row - 1)
+        _setindex!(upper_diag, row, _getindex(upper_diag, row) / denominator)
+        _setindex!(b, row, numerator / denominator)
+    end
+
+    # last row
+    numerator =
+        _getindex(b, nrows) -
+        _getindex(lower_diag, nrows) * _getindex(b, nrows - 1)
+    denominator =
+        _getindex(main_diag, nrows) -
+        _getindex(lower_diag, nrows) * _getindex(upper_diag, nrows - 1)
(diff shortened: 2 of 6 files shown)


examples/hybrid/schur_complement_W.jl
```diff
@@ -206,3 +201,90 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
...
+    return nothing
+end
+
+function column_thomas_solve!(::ClimaComms.CPUSingleThreaded, A, xᶠ𝕄, dims)
+    (Ni, Nj, Nh) = dims
+    (; S, S_column_array) = A
+    for h in 1:Nh, j in 1:Nj, i in 1:Ni
+        xᶠ𝕄_column_view = parent(Spaces.column(xᶠ𝕄, i, j, h))
+        S_column = Spaces.column(S, i, j, h)
+        @views S_column_array.dl .= parent(S_column.coefs.:1)[2:end]
```
simonbyrne, line 240: i'd like to be able to do this directly in place on `S_columns.coefs`

dennisYatunin:
bors try

dennisYatunin:
bors try

examples/hybrid/schur_complement_W.jl
```diff
@@ -206,3 +198,82 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
...
+    # back substitution
+    for row in (nrows - 1):-1:1
+        b[row] -= b[row + 1] * A[row, 3]
+    end
+    return nothing
+end
+
+function column_thomas_solve!(::ClimaComms.CPUSingleThreaded, A, b)
+    Ni, Nj, _, _, Nh = size(Spaces.local_geometry_data(axes(A)))
+    for h in 1:Nh, j in 1:Nj, i in 1:Ni
```
charleskawczynski, line 234: Is there a reason we're not using `bycolumn` here? It would allow us to combine this and the multithreaded implementations

examples/hybrid/schur_complement_W.jl
```diff
@@ -206,3 +198,82 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
...
+
+"""
+    thomas_algorithm!(A, b)
+
+Thomas algorithm for solving a linear system A x = b,
+where A is a tri-diagonal matrix.
+A and b are overwritten.
+Solution is written to b
+"""
+function thomas_algorithm!(A, b)
```
charleskawczynski, line 210: Let's move this, and the remaining functions added in this file into `src/`, since we want it useable in ClimaAtmos.

perf/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
charleskawczynski, line 4: What's this needed for? Maybe we can do this in a separate PR?

examples/hybrid/schur_complement_W.jl
simonbyrne: Yes, this should be part of ClimaCore itself.

examples/hybrid/schur_complement_W.jl
```diff
@@ -206,3 +198,82 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
...
+    # A[_, 1] is lower diag, A[_, 2] is main diag, A[_, 3] is upper diag
+    nrows = size(A, 1)
+    # first row
+    A[1, 3] /= A[1, 2]
```
simonbyrne, line 214: Accessing by `A[row, diag]`, is a bit confusing. Instead of accessing by `parent`, can we instead use `getidx` on the `Field`?

@dennisYatunin can you help with this?

dennisYatunin: Since the matrix is represented by a field of matrix rows (`StencilCoefs`), we do have that `parent(A)[row, diag] == getidx(A, row, ...)[diag]`. However, there isn't a good way to use `setidx` to set an individual entry on a row; we can only use it to set all the entries in a row at once. This algorithm, as it is written, uses the third diagonal of `A` as a cache for computing `A \ b`, so it isn't really compatible with `setidx`. However, we could also allocate a cache instead of overwriting the third diagonal of `A`, and then use `setidx` to set the entries in that cache (each entry would just be a floating-point value). Would that be a good alternative?

dennisYatunin: Also, we probably want to avoid modifying `A` in this function, because we might want to reuse the same Jacobian matrix for multiple Newton iterations. I think allocating a cache and leaving the third diagonal of `A` unchanged would be better than what we have here. Note that the current code doesn't modify the original matrix field; instead, it copies the field data into an array, and then modifies the third diagonal in that array.

simonbyrne: No, I would prefer that we don't allocate a cache for now: the extra data movement will start to get expensive. If we need it, we can add it as an option.

charleskawczynski: Can we clarify this: can we write this as `thomas_algorithm!(A, b, cache)`? so that `thomas_algorithm!` is non-allocating, but we have additional storage for handling things more intuitively?

examples/hybrid/test_thomas.jl
```diff
@@ -0,0 +1,80 @@
+using Test
+using IntervalSets
+import Random: seed!
+import LinearAlgebra: Tridiagonal, norm
+import ClimaComms
+import ClimaCore:
+    Geometry, Domains, Meshes, Topologies, Spaces, Fields, Operators
+
+include("schur_complement_W.jl")
```
charleskawczynski, line 9:
```suggestion
```

examples/hybrid/schur_complement_W.jl
charleskawczynski: Also, I'm not a fan of overwriting `b` with `x`. It's very awkward from a user POV. We have to compute tendencies, pass them in, and suddenly it becomes the state vector. If, for some reason, we want to keep the tendencies around (say for computing the residual, or for limiters or something), then this function corrupts that data. One extra field is not a performance issue here.

simonbyrne: This isn't trying to be a generic `ldiv!` (which is why I gave it a different name).

Look, there are lots of ways we can do this. For now, just make it completely in-place, with no cache. That solves the immediate problem in the simplest way possible. If we decide we need something else, we can do that then.

simonbyrne: As for the `getidx`/`setidx!`, you can define
```
ld = S.coefs.:1
d = S.coefs.:2
ud = S.coefs.:3
```
and operate on those

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

src/Operators/thomas_algorithm.jl
```diff
@@ -0,0 +1,89 @@
...
+    end
+end
+
+# This is the same as @inbounds Fields.field_values(column_field)[index]
+_getindex(column_field, index) = @inbounds Operators.getidx(
+    axes(column_field),
+    column_field,
+    Operators.Interior(),
+    index - 1 + Operators.left_idx(axes(column_field)),
+)
```
simonbyrne, line 80: why not just used `getidx` directly?

dennisYatunin: I just though it would be a bit hard to read with all the `space`s and `loc`s and `- 1 + left_idx`s floating around everywhere.

src/Operators/thomas_algorithm.jl
```diff
@@ -0,0 +1,89 @@
...
+    idx = threadIdx().x + (blockIdx().x - 1) * blockDim().x
+    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+    if idx <= Ni * Nj * Nh
+        i, j, h = Spaces._get_idx((Ni, Nj, Nh), idx)
+        thomas_algorithm!(Spaces.column(A, i, j, h), Spaces.column(b, i, j, h))
+    end
+    return nothing
+end
+
+function thomas_algorithm!(A, b)
```
charleskawczynski, line 33:
```suggestion
function thomas_algorithm!(A::Fields.ColumnField, b::Fields.ColumnField)
```

src/Operators/thomas_algorithm.jl
```diff
@@ -0,0 +1,89 @@
...
+column_thomas_solve!(::ClimaComms.AbstractCPUDevice, A, b) =
+    Fields.bycolumn(colidx -> thomas_algorithm!(A[colidx], b[colidx]), axes(A))
+
+function column_thomas_solve!(::ClimaComms.CUDADevice, A, b)
+    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+    nthreads, nblocks = Spaces._configure_threadblock(Ni * Nj * Nh)
+    @cuda threads = nthreads blocks = nblocks thomas_algorithm_kernel!(A, b)
+end
+
+function thomas_algorithm_kernel!(A, b)
```
charleskawczynski, line 23:
```suggestion
thomas_algorithm_kernel!(A::Fields.ColumnField, b::Fields.ColumnField) = thomas_algorithm!(A, b)
function thomas_algorithm_kernel!(A::Fields.ExtrudedFiniteDifferenceField, b::Fields.ExtrudedFiniteDifferenceField)
```

charleskawczynski (changes requested):
We need to make `column_thomas_solve!` work for `Fields.ExtrudedFiniteDifferenceField` and `Fields.ColumnField`. Let's handle this with dispatch (let's add types for these methods so it's clear what types of fields they're valid for, and so that we get better error messages), we'll need an extra method or two. Let me know if you need help.

dennisYatunin:
bors try

dennisYatunin:
bors try

charleskawczynski (commented):
Unfortunately, nothing seems to be converging, so there seems to be a translational bug somewhere. Maybe we can add a test against the original code to identify the issue?

test/Operators/thomas_algorithm.jl
```diff
@@ -0,0 +1,61 @@
...
+    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+    for i in 1:Ni, j in 1:Nj, h in 1:Nh
+        A_column_data = Array(parent(Spaces.column(A_copy, i, j, h)))
+        A_column_array = Tridiagonal(
+            A_column_data[2:end, 1],
+            A_column_data[:, 2],
+            A_column_data[1:(end - 1), 3],
+        )
+        b_column_array = Array(parent(Spaces.column(b_copy, i, j, h)))[:]
+        x_column_array = Array(parent(Spaces.column(b, i, j, h)))[:]
```
charleskawczynski, line 43:
```suggestion
    Fields.bycolumn(axes(A)) do colidx
        A_column_data = Array(parent(A_copy[colidx]))
        A_column_array = Tridiagonal(
            A_column_data[2:end, 1],
            A_column_data[:, 2],
            A_column_data[1:(end - 1), 3],
        )
        b_column_array = Array(parent(b_copy[colidx]))[:]
        x_column_array = Array(parent(b[colidx]))[:]
```

dennisYatunin: I tried this earlier, and it doesn't work on the GPU

charleskawczynski: Why is `[:]` necessary here?

dennisYatunin: It's because the vector field is stored as a 5D array, and I need to reshape it into a 1D array. I think putting `[:]` at the end is equivalent to using the `vec` function; it creates a 1D view of the 5D parent array.

charleskawczynski:
New build at https://buildkite.com/clima/climaatmos-ci/builds/10949

charleskawczynski (approved):

Issue #1097, opened 2023-01-11
GPU broadcasting sometimes breaks with type casting


charleskawczynski:
Sometimes GPU broadcasting breaks with type casting. See https://github.com/CliMA/ClimaCore.jl/pull/1096

charleskawczynski:
I think that this is fixed with CUDA’s adapt on types.

PR #1365, opened 2023-07-11
Fix type-casting allocations in broadcasting

### LINKED ISSUE #1355: Broadcasting or-or incurs allocations
MWE:

```julia
using Revise, Test
import ClimaCore
if !(@isdefined TU)
    include(joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"))
    import .TestUtilities as TU
end
foo(a, c) = a;
FT_broadcast_allocating!(a) = FT_broadcast_allocating!(a, eltype(a));
function FT_broadcast_allocating!(a, ::Type{FT}) where FT
    @. a = foo(a, false || a < FT(1))
    # @. a = foo(a, a < FT(1)) # unexpected pass
    return nothing
end;
space = TU.CenterExtrudedFiniteDifferenceSpace(Float32);
a = ClimaCore.Fields.level(fill(Float32(0), space), 1);

FT_broadcast_allocating!(a) # compile first

p_allocated = @allocated FT_broadcast_allocating!(a)
@test_broken p_allocated == 0
```

charleskawczynski:
Closes #1355 and #1097.

I'm not thrilled that this is basically type piracy, but it does fix the allocations. It's actually a bit surprising to me that a simple definition doesn't exist in Base.

This should fix the GPU moisture issue in ClimaAtmos (among other isbits errors and allocations).

### DIFF
--- a/src/Fields/broadcast.jl
+++ b/src/Fields/broadcast.jl
@@ -301,6 +301,8 @@ Base.Broadcast.broadcasted(
     ::Val{n},
 ) where {n} = Base.Broadcast.broadcasted(x -> Base.literal_pow(^, x, Val(n)), f)
 
+Base.Broadcast.broadcasted(::Type{T}, x::Number) where {T <: Number} = x
+
 # Specialize handling of +, *, muladd, so that we can support broadcasting over NamedTuple element types
 # Required for ODE solvers
 
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -140,7 +140,7 @@ end
 
         ifelse_broadcast_allocating(a, b, c)
         p_allocated = @allocated ifelse_broadcast_allocating(a, b, c)
-        @test_broken p_allocated == 0
+        @test p_allocated == 0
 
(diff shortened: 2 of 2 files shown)


src/Fields/broadcast.jl
```diff
@@ -301,6 +301,8 @@ Base.Broadcast.broadcasted(
     ::Val{n},
 ) where {n} = Base.Broadcast.broadcasted(x -> Base.literal_pow(^, x, Val(n)), f)
 
+Base.Broadcast.broadcasted(::Type{T}, x::Number) where {T <: Number} = x
```
charleskawczynski, line 304:
```suggestion
Base.Broadcast.broadcasted(::Type{T}, x::Number) where {T <: Number} = T(x)
```
I guess we should actually do this instead

charleskawczynski:
We can work around this on the implementation side, without type piracy. Maybe that's a better solution for now, until https://github.com/JuliaLang/julia/issues/50554 is fixed.

PR #1358, opened 2023-07-06
Widen OrdinaryDiffEq and SciMLBase compat range

charleskawczynski:
This is kind of a peel off from #1235. This PR widens the compat range for OrdinaryDiffEq and SciMLBase

PR #1375, opened 2023-07-14
Use ClimaTimeSteppers in implicit hydro balance example

charleskawczynski:
This PR changes the implicit hydro balance example to use ClimaTimeSteppers.

This is a peel off from https://github.com/CliMA/ClimaCore.jl/pull/1358.

This will technically have behavior changes because we're switching from `Rosenbrock23` to `ARS343`

### DIFF
--- a/examples/column/hydrostatic_implicit.jl
+++ b/examples/column/hydrostatic_implicit.jl
@@ -1,4 +1,5 @@
 using LinearAlgebra
+import ClimaTimeSteppers as CTS
 import ClimaCore:
     Fields,
     Domains,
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

charleskawczynski:
This is actually breaking (we're using `soft_fail` on this job)

charleskawczynski:
bors try

dennisYatunin:
Oh, sorry, I was looking at the ClimaAtmos code. In this file, it would be `Wfact = jacobian!`. The `jac = jacobian!` is not used by ClimaTimeSteppers (because the thing we're actually specifying here is `I - dt * gamma * J`, rather than `J` itself).

dennisYatunin:
Actually, looking more closely at this, I don't think that will work. The `jacobian!(J, Y, p, t)` function needs to be rewritten as a `Wfact!(W, Y, p, dtγ, t)` function (so `J` needs to be replaced with `I - dtγ * J`), and then `linsolve!` needs to be modified to solve `A * x = b`, where `A = I - dtγ * J`, instead of whatever it's solving now.

charleskawczynski:
Superseded by #1400

Issue #867, opened 2022-08-05
Re-design AxisTensors


charleskawczynski:
## Purpose
The purpose of this SDI is to improve the design of AxisTensors. There are several issues with the existing design:
 1. A single "mega-type" (a type-heavy struct) is used to allow for _many_ different types of axis tensors-- (some previous tests create 3486 different kinds, to be exact).
 2. Due to the "mega-type" approach, we end up stuck having to generate _many_ methods for performant runtime code, rather than using _far_ fewer methods and relying on dispatch. One price we pay with that is significantly higher latency. #853 makes some ClimaAtmos examples go from 2 min to 7 min. But other, larger jobs, reduce in runtime by ~10%.
 3. Documentation. We can't easily document this type because there are so many potential type parameters. We sort of mask this by creating many user-facing aliases, but those are not easily documentable, and users often only see this in source code, and not in stack traces, because stack traces will show `typeof(::AxisTensor)` which will not show the alias, but the concrete type of `AxisTensor`, creating a visual mismatch for people debugging.
 4. There are many wrappers in Geometry which perform nested operations on specific types and simply promote types in order to function, at the cost of doing all operations in a dense manner. This results in a huge number of flops relative to how many may actually be needed. One key aspect of the re-design will be to offer functions that do operations with a minimal number of flops.

Link to any relevant PRs/issues.

## Cost/benefits/risks
 - The cost of this is likely in the ~1-2 months of work for 1-2 people, as the AxisTensors is a fundamental component, is pretty , and will likely require breaking changes.
 - The benefit of fixing this is mostly mentioned in the purpose: better performance, latency, and documentation
 - The risk of not fixing this is that it's a fundamental building block for ClimaCore users, and the compilation latency issue cannot be solved upstream, so it's imperative that this is redesigned

## Producers
 - I know that @simonbyrne has ideas on how to improve the design, I'd be happy to help, too.

## Components
 - `AxisTensor`s

## Inputs
There are many relations (e.g., conversions) that will need to be re-written, but most of this is related to the type design

## Results and deliverables
A better design to AxisTensors

## Task breakdown
A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results.
- [ ] Decide how to decompose the 3486 AxisTensors into composeable parts that can result in a better design for conversions
- [ ] Implement and test
...

## Reviewers
 - @simonbyrne, @charleskawczynski, @akshaysridhar, @valeriabarra

tapios:
[...]

PR #1392, opened 2023-07-24
Update vector conversions

sriharshakandala:
Add custom `CovariantVector` to `Contravariant123Vector` transformation functions 

### DIFF
--- a/src/Geometry/conversions.jl
+++ b/src/Geometry/conversions.jl
@@ -7,7 +7,7 @@
 ContravariantVector(u::ContravariantVector, ::LocalGeometry) = u
 CovariantVector(u::CovariantVector, ::LocalGeometry) = u
 LocalVector(u::LocalVector, ::LocalGeometry) = u
-
+# conversions between Covariant/Contravariant vectors and local vectors
 ContravariantVector(
     u::LocalVector{T, I},
     local_geometry::LocalGeometry{I},
@@ -24,6 +24,29 @@ LocalVector(
     u::CovariantVector{T, I},
     local_geometry::LocalGeometry{I},
 ) where {T, I} = local_geometry.∂ξ∂x' * u
+
+# conversions between Covariant and Contravariant vectors
+Contravariant123Vector(
+    u::CovariantVector{T, (1, 2)},
+    local_geometry::LocalGeometry{(1, 2, 3)},
+) where {T} = local_geometry.gⁱʲ * Covariant123Vector(u[1], u[2], zero(u[1]))
+
+Contravariant123Vector(
+    u::CovariantVector{T, (3,)},
+    local_geometry::LocalGeometry{(1, 2, 3)},
+) where {T} =
+    local_geometry.gⁱʲ * Covariant123Vector(zero(u[1]), zero(u[1]), u[1])
+
+
+ContravariantVector(
+    u::CovariantVector{T, I},
+    local_geometry::LocalGeometry{I},
+) where {T, I} = local_geometry.gⁱʲ * u
+
+CovariantVector(
+    u::ContravariantVector{T, I},
+    local_geometry::LocalGeometry{I},
+) where {T, I} = local_geometry.gᵢⱼ * u
 
 # Converting to specific dimension types
 (::Type{<:ContravariantVector{<:Any, I}})(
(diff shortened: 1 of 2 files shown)


sriharshakandala:
bors try

charleskawczynski (commented):
This PR looks good to me. It's a bit unfortunate that these are not captured in the axis tensor benchmarks. Can you point to the methods that these new functions are overriding / specializing on?

I'm curious if we can write a more general overloading, like we did before for `project / transform`. It seems that one crux of this is to cache `∂x∂ξ' * ∂x∂ξ`, which seems like a good idea (I forgot that we left that on the table before).

Related issue: #867

sriharshakandala:
> This PR looks good to me. It's a bit unfortunate that these are not captured in the axis tensor benchmarks. Can you point to the methods that these new functions are overriding / specializing on?
> 
> I'm curious if we can write a more general overloading, like we did before for `project / transform`. It seems that one crux of this is to cache `∂x∂ξ' * ∂x∂ξ`, which seems like a good idea (I forgot that we left that on the table before).

In the context of the GPU version, we noticed that this function uses much less register space per thread compared to the pre-existing version. Looking at the Nsight Systems profile for the pre-existing version, the fundamental bottleneck seems to be register space. This helps us use more threads per block or enable the GPU to schedule multiple blocks on the same SM resulting in significant speedup.

sriharshakandala:
> This PR looks good to me. It's a bit unfortunate that these are not captured in the axis tensor benchmarks. Can you point to the methods that these new functions are overriding / specializing on?
> 
> I'm curious if we can write a more general overloading, like we did before for `project / transform`. It seems that one crux of this is to cache `∂x∂ξ' * ∂x∂ξ`, which seems like a good idea (I forgot that we left that on the table before).

I believe we provide a specialization that overrides https://github.com/CliMA/ClimaCore.jl/blob/main/src/Geometry/conversions.jl#L102 for this specification. Please correct me if I am missing something!

sriharshakandala:
bors try

charleskawczynski:
https://github.com/CliMA/ClimaCore.jl/blob/main/src/Geometry/conversions.jl#L102 for this specific case. Please correct me if I am missing something!

Ah, yeah, that's a pretty bad method (it calls `project` twice). In fact, all of the generic ones look pretty bad (cc @bloops)

https://github.com/CliMA/ClimaCore.jl/blob/main/src/Geometry/conversions.jl#L73-L109

src/Geometry/conversions.jl
```diff
@@ -25,6 +25,29 @@ LocalVector(
...
+
+ContravariantVector(
+    u::CovariantVector{T, I},
+    local_geometry::LocalGeometry{I},
+) where {T, I} = local_geometry.gⁱʲ * u
+
+CovariantVector(
+    u::ContravariantVector{T, I},
+    local_geometry::LocalGeometry{I},
+) where {T, I} = local_geometry.gᵢⱼ * u
```
charleskawczynski, line 49: Is diagonalization along `I` required here?

charleskawczynski: (just out of curiosity)

sriharshakandala: This works the full vector, with all `ndim` components. The first two deals with partial vectors!

charleskawczynski: Ok, I think you're saying yes. The generic method already defined does not diagonalize:

```julia
CovariantVector(
    u::ContravariantVector,
    local_geometry::LocalGeometry{I},
) where {I} = project(
    CovariantAxis{I}(),
    project(ContravariantAxis{I}(), u),
    local_geometry,
)
```

charleskawczynski (approved):
This is fine for now, but we should really improving the performance of these operators more systematically (like we did with `project` / `transform`).

src/Geometry/conversions.jl
```diff
@@ -25,6 +25,29 @@ LocalVector(
...
+Contravariant123Vector(
+    u::CovariantVector{T, (1, 2)},
+    local_geometry::LocalGeometry{(1, 2, 3)},
+) where {T} = local_geometry.gⁱʲ * Covariant123Vector(u[1], u[2], zero(u[1]))
+
+Contravariant123Vector(
+    u::CovariantVector{T, (3,)},
+    local_geometry::LocalGeometry{(1, 2, 3)},
+) where {T} =
+    local_geometry.gⁱʲ * Covariant123Vector(zero(u[1]), zero(u[1]), u[1])
```
charleskawczynski, line 38: @sriharshakandala, can we hand-write this and avoid the dense mul?

PR #1400, opened 2023-07-28
Remove incorrect hydrostatic_implicit.jl test

charleskawczynski:
I went to convert `hydrostatic_implicit.jl` to use ClimaTimesteppers, and @dennisYatunin and I discovered a bug that renders the test useless:

```julia
function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
    function _linsolve!(x, A, b, update_matrix = false; kwargs...)
        x = copy(b) #
        ...
    end
end
```
This does not mutate the incoming `x`, so `linsolve!` simply doesn't do what it's supposed to. Furthermore, since this test is meant to ensure the solution doesn't change, it "passes" because `linsolve!` doesn't actually do what it's suppose to.

Since this depends on an older version of OrdinaryDiffEq, we think it's best to simply remove this test.


charleskawczynski:
Supersedes #1375.

charleskawczynski:
I just realized that @simonbyrne is out this week, going to move forward with this

PR #1402, opened 2023-08-01
Error on NaNs in driver solution

charleskawczynski:
This PR adds a check for NaNs in the hybrid examples driver.


charleskawczynski:
bors try

charleskawczynski:
bors try

PR #1416, opened 2023-08-07
Widen scope of column benchmarks

### LINKED ISSUE #1409: Add column benchmarks on sphere
It looks like the column benchmarks are only on pure columns and rectangular domains. We should add them for a sphere, too.

charleskawczynski:
This PR widens the scope of column benchmarks, and tests on a sphere space (closes #1409).


charleskawczynski:
bors try

PR #1235, opened 2023-05-10
Upgrade to julia 1.9

charleskawczynski:
This PR upgrades to julia 1.9

Blockers
- [x] Get latest TaylorSeries.jl: https://github.com/CliMA/CubedSphere.jl/pull/29
- [x] https://github.com/JuliaSIMD/HostCPUFeatures.jl/issues/12 (required to get `JULIA_CPU_TARGET` to work properly)

PR #1417, opened 2023-08-08
Make GHA run on julia 1.9

charleskawczynski:
It might be nice to incrementally merge #1235. The main thing holding it back is issues with the shared depot path. One thing we can do is start running ClimaCore on 1.9 in GHA, and fix the inference issues that occur from upgrading, without having to deal with the shared depot path issue.

I don't love the idea of being straddled between versions, but the only breaking differences seem to be inference, which I suspect we can fix for both versions without too much headache.

### DIFF
--- a/.github/workflows/ClimaCoreMakie.yml
+++ b/.github/workflows/ClimaCoreMakie.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1.8'
+          version: '1.9'
       - uses: actions/cache@v1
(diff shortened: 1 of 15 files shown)


simonbyrne (approved):

charleskawczynski:
Ah, bors is not matching on the status, since we've added (not just replaced 1.8.5 with 1.9). I guess the simples solution would be to replace it since that should make the status look the same, do we want to keep GHA tests on 1.8? cc @simonbyrne

charleskawczynski:
Being that this PR doesn't fully convert us to 1.9, I'll try to keep 1.8, too

Issue #1418, opened 2023-08-08
Fix many documentation warnings


charleskawczynski:
The doc build is riddled with warnings, this should be fixed.

PR #1419, opened 2023-08-08
Fix a bunch of docs issues

charleskawczynski:
There are more, but I think this is a batch that was kind of quick / easier to fix. A step towards #1418.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -122,6 +122,17 @@ Topologies.AbstractTopology
 Topologies.IntervalTopology
 Topologies.Topology2D
 Topologies.spacefillingcurve
+Topologies.nelems
+Topologies.nneighbors
+Topologies.nsendelems
+Topologies.nghostelems
+Topologies.localelemindex
+Topologies.face_node_index
+Topologies.ghost_faces
+Topologies.vertex_node_index
+Topologies.local_vertices
+Topologies.ghost_vertices
+Topologies.neighbors
 ```
 
 ### Interfaces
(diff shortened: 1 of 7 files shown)


valeriabarra:
I can't visualize/review the preview of the docs via the `documenter/deploy` GHA. It throws a 404:

image

charleskawczynski:
> I can't visualize/review the preview of the docs via the `documenter/deploy` GHA. It throws a 404:

I think you maybe clicked it too early? I'm seeing the page show up

valeriabarra:
> I think you maybe clicked it too early? I'm seeing the page show up

Oh, it was already showing the green check mark and it was marked as completed and I refreshed it a few times. But yeah, I am able to visualize now, too

docs/src/api.md
```diff
@@ -199,19 +210,18 @@ Spaces.Quadratures.orthonormal_poly
...
 Spaces.dss_local_ghost!
 Spaces.dss_ghost!
 Spaces.create_dss_buffer
 Spaces.DSSBuffer
+Spaces.create_ghost_buffer
 Spaces.weighted_dss_start!
 Spaces.weighted_dss_internal!
 Spaces.weighted_dss_ghost!
 Spaces.weighted_dss!
+Spaces.weighted_dss
```
valeriabarra, line 224: There is a 
image

docs/src/api.md
```diff
@@ -303,3 +313,10 @@ InputOutput.read_field
 ```@docs
 Remapping.interpolate_array
 ```
+
+## Other
+
+
+```@docs
+Limiters.quasimonotone_limiter
```
valeriabarra, line 321: We already have the `Limiters` section with this API docstring 
https://clima.github.io/ClimaCore.jl/previews/PR1419/api/#ClimaCore.Limiters.QuasiMonotoneLimiter
Why add an "Other" section?

valeriabarra: And the latter one added errors:
image

docs/src/api.md
charleskawczynski: Yeah, this PR isn't fixing all of them. I'll try to close the rest out in a folloup PR, where we can make `strict = true`. I'll remove this since `weighted_dss` is not an existing function.

docs/src/api.md
charleskawczynski: Yes, but `quasimonotone_limiter` is not in the API section. Actually, I misinterpreted the error--
```julia
"""
    AbstractLimiter

Supertype for all limiters.

# Interfaces

- [`quasimonotone_limiter!`](@ref)
"""
abstract type AbstractLimiter end
```

`quasimonotone_limiter!` is referencing a non-existing function. I think it's supposed to be `apply_limiter!`.

valeriabarra (approved):

PR #1433, opened 2023-08-21
Store 1D SE differentiation matrix

sriharshakandala:
Store 1D SE differentiation matrix to avoid recomputation in spectral element kernels

### DIFF
--- a/src/Spaces/extruded.jl
+++ b/src/Spaces/extruded.jl
@@ -241,6 +241,7 @@ Base.@propagate_inbounds function level(
             horizontal_space.global_geometry,
             level(space.center_local_geometry, v),
             horizontal_space.dss_weights,
+            horizontal_space.differentiation_matrix,
         )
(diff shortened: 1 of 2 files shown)


sriharshakandala:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

simonbyrne:
Does this make any difference? Since it's an `SArray`, it should just get baked into the kernel.

sriharshakandala:
> Does this make any difference? Since it's an `SArray`, it should just get baked into the kernel.

This will 1) reduce op count for each kernel, 2) eliminate storage of full differentiation matrix on each thread (insteading of just load one entry per thread onto shmem)!

simonbyrne:
I don't quite see how this would help with that: basically the only difference is that this would pass it through as a kernel parameter, rather than as a compile-time constant. Do you have a benchmark?

PR #1436, opened 2023-08-25
Add FieldMatrix and linear solvers

dennisYatunin:
## Purpose

Third PR of #1230. Adds an interface for specifying block matrices of matrix fields and solving linear systems with these matrices. This will replace what is currently in `ClimaAtmos.jl/src/prognostic_equations/implicit/schur_complement_W.jl`, generalizing it for implicit diffusion and implicit EDMF.

### Main Changes

- Add the `FieldName` struct, which is a singleton type that represents a chain of `getproperty` calls
  - Add the `@name` macro for constructing `FieldName`s, which checks whether its input expression is a syntactically valid chain of `getproperty` calls before calling the default constructor.
  - A `name` can be used to access a property or sub-property of an object `x` by calling `get_field(x, name)`.
  - An `internal_name` can be appended onto another `name` in order to access a property or sub-property of `get_field(x, name)`.
- Add the `FieldNameDict` struct, which maps each key in a set of `FieldVectorKeys` or `FieldMatrixKeys` (see below) to a `Field` or some other object.
  - There are currently four subtypes of `FieldNameDict`:
    - `FieldMatrix` (the only user-facing subtype), which maps `NTuple{2, FieldName}`s to `ColumnwiseBandMatrixField`s or multiples of `LinearAlgebra.I`
    - `FieldVectorView`, which maps `FieldName`s to `Field`s; this is used to wrap a `FieldVector` so that it can be used in conjunction with a `FieldMatrix`
    - `FieldVectorViewBroadcasted` and `FieldMatrixBroadcasted`, each of which can store unevaluated `Base.AbstractBroadcasted` objects, in addition to what `FieldVectorView` and `FieldMatrix` can already store
  - Supports standard `AbstractDict` functions like `keys` and `pairs`.
  - An individual block of a `FieldNameDict` can be accessed by calling `dict[key]`, and a range of blocks can be accessed by calling `dict[set]`, where `set` is a `FieldNameSet`.
  - Given a `FieldMatrix` `A`, a similar matrix that only contains identity matrix blocks can be constructed with `one(A)`.
  - `FieldNameDict`s can be used in broadcast expressions, which support the following operations:
    - `+`, `-`, or `*`, where each input is either a `FieldNameDict` or a `FieldVector`
    - `inv`, where the input is a diagonal `FieldMatrix`
  - The new methods for `Base.Broadcast.broadcasted` construct chains of `Field` broadcast expressions from `FieldNameDict` broadcast expressions on the fly, somewhat similarly to how broadcasting works for ClimaCore operators.
- Add the `FieldMatrixSolver` struct, which solves an equation of the form `A * x = b`, where `A` is a `FieldMatrix` and where `x` and `b` are `FieldVector`s.
  - Add the `field_matrix_solve!` function, which works just like `ldiv!(x, A, b)`, except that it also takes a `FieldMatrixSolver` as its first argument.
  - Add four `FieldMatrixSolverAlgorithm`s, which can be nested inside of each other to build up more specialized algorithms:
    - `BlockDiagonalSolve`, which runs a "single field solver" for each block of the block diagonal matrix `A`; the single field solver can handle the four types of blocks:
      - Multiples of `LinearAlgebra.I`
      - Diagonal `ColumnwiseBandMatrixField`s
      - Tri-diagonal `ColumnwiseBandMatrixField`s (implementation of the Thomas algorithm)
      - Penta-diagonal `ColumnwiseBandMatrixField`s (implementation of the PTRANS-I algorithm)
    - `BlockLowerTriangularSolve`, which uses forward substitution to solve the equation for a block lower triangular matrix `A`
    - `SchurComplementSolve`, which generalizes what is currently in ClimaAtmos's `schur_complement_W.jl` file to any block matrix `A` with a diagonal block in the top-left corner
    - `ApproximateFactorizationSolve`, which lets us use "operator splitting" to approximately solve the equation for a diagonally dominant block matrix `A`
- Add documentation for how to specify a `FieldMatrix` and use it in a linear solver, along with internal documentation for the new `FieldName`-based infrastructure.
- Add unit tests for correctness, type stability, and allocations, and run them on both CPUs and GPUs through CI.
  - Test each single field solver on both a cell-center and a cell-face field.
  - Test each `FieldMatrixSolverAlgorithm` on block diagonal, block lower triangular, and block dense matrices.
  - Test solvers with identical structures to what we will use in ClimaAtmos for the following examples:
    - Dry dycore with implicit acoustic waves
    - Dry dycore with implicit acoustic waves and diffusion
    - Dry dycore + prognostic EDMF with implicit acoustic waves and SGS fluxes
    - Moist dycore + prognostic EDMF + tracers with implicit acoustic waves and SGS fluxes

### Internal Chagnes

- Add a collection of "unrolled functions", whose return values can be inferred during compilation if their input values are all singleton types.
  - These are all implemented as combinations of `unrolled_zip`, `unrolled_map`, and `unrolled_foldl`.
  - Several of these need to have their recursion limits disabled for the unit tests to be type stable.
- Add the `FieldNameTree` struct, which stores every `FieldName` that can be used to access `x` with `get_field(x, name)`.
  - A `name` can be checked for validity by calling `has_subtree_at_name(tree, name)`.
  - The children of `name` (the `FieldName`s that can be used to access the properties of `get_field(x, name)`) can be obtained by calling `child_names(name, tree)`.
- Add the `FieldNameSet` struct, which stores a set of `FieldVectorKeys` (each of which is a `FieldName`) or a set of `FieldMatrixKeys` (each of which is an `NTuple{2, FieldName}`).
  - Roughly equivalent to the built-in `KeySet` for `AbstractDict`s, but specialized for `FieldNameDict`s.
  - Supports standard `AbstractSet` functions like `union` and `setdiff`, as well as custom functions like `set_complement` and `matrix_product_keys`.
  - Handles overlaps between `FieldName`s (that is, situations where one property of `x` lies inside another property of `x`) by storing a `FieldNameTree` that contains all available `FieldName`s.
- Disable the recursion limits for several functions used to manipulate `FieldName`s, `FieldNameTree`s, and `FieldNameSet`s, as this is necessary in order for the unit tests to be type stable.
- Remove the methods for `RecursiveApply.rmul` that specialize on `Number`, which is also necessary in order for the unit tests to be type stable.
  - These methods are no longer required, now that #1454 has been merged in.
- Add support for calling `inv` on `BandMatrixRow`s.
- Qualify the use of `CUDA.@allowscalar`, per Charlie's suggestion.
- Fix some type instabilities in `matrix_field_test_utils.jl`.
- Remove an unused variable name in `operator_matrices.jl`.

### DIFF
--- a/src/MatrixFields/MatrixFields.jl
+++ b/src/MatrixFields/MatrixFields.jl
@@ -2,7 +2,7 @@
     MatrixFields
 
 This module adds support for defining and manipulating `Field`s that represent
-matrices. Specifically, it specifies the `BandMatrixRow` type, which can be used
+matrices. Specifically, it adds the `BandMatrixRow` type, which can be used
 to store the entries of a band matrix. A `Field` of `BandMatrixRow`s on a
 `FiniteDifferenceSpace` can be interpreted as a band matrix by vertically
 concatenating the `BandMatrixRow`s. Similarly, a `Field` of `BandMatrixRow`s on
@@ -19,46 +19,60 @@ for them:
 - Integration with `RecursiveApply`, e.g., the entries of `matrix_field` can be
     `Tuple`s or `NamedTuple`s instead of single values, which allows
     `matrix_field` to represent multiple band matrices at the same time
+- Integration with `Operators`, e.g., the `matrix_field` that gets applied to
+    the argument of any `FiniteDifferenceOperator` `op` can be obtained using
+    the `FiniteDifferenceOperator` `operator_matrix(op)`
 - Conversions to native array types, e.g., `field2arrays(matrix_field)` can
     convert each column of `matrix_field` into a `BandedMatrix` from
     `BandedMatrices.jl`
 - Custom printing, e.g., `matrix_field` gets displayed as a `BandedMatrix`,
     specifically, as the `BandedMatrix` that corresponds to its first column
+
+This module also adds support for defining and manipulating sparse block
+matrices of `Field`s. Specifically, it adds the `FieldMatrix` type, which is a
+dictionary that maps pairs of `FieldName`s to `ColumnwiseBandMatrixField`s or
+multiples of `LinearAlgebra.I`. This comes with the following functionality:
+- Addition and subtraction, e.g., `@. field_matrix1 + field_matrix2`
+- Matrix-vector multiplication, e.g., `@. field_matrix * field_vector`
+- Matrix-matrix multiplication, e.g., `@. field_matrix1 * field_matrix2`
+- Integration with `RecursiveApply`, e.g., the entries of `field_matrix` can be
+    specified either as matrix `Field`s of `Tuple`s or `NamedTuple`s, or as
+    separate matrix `Field`s of single values
+- The ability to solve linear equations using `FieldMatrixSolver`, which is a
+    generalization of `ldiv!` that is designed to optimize solver performance
 """
 module MatrixFields
 
-import CUDA: @allowscalar
-import LinearAlgebra: UniformScaling, Adjoint, AdjointAbsVec
+import CUDA
+import LinearAlgebra: I, UniformScaling, Adjoint, AdjointAbsVec, mul!, inv
 import StaticArrays: SMatrix, SVector
 import BandedMatrices: BandedMatrix, band, _BandedMatrix
+import ClimaComms
 
 import ..Utilities: PlusHalf, half
 import ..RecursiveApply:
     rmap, rmaptype, rpromote_type, rzero, rconvert, radd, rsub, rmul, rdiv
+import ..RecursiveApply: ⊠, ⊞, ⊟
 import ..DataLayouts: AbstractData
 import ..Geometry
 import ..Spaces
 import ..Fields
 import ..Operators
 
-export ⋅
 export DiagonalMatrixRow,
     BidiagonalMatrixRow,
     TridiagonalMatrixRow,
     QuaddiagonalMatrixRow,
     PentadiagonalMatrixRow
+export FieldVectorKeys, FieldVectorView, FieldVectorViewBroadcasted
+export FieldMatrixKeys, FieldMatrix, FieldMatrixBroadcasted
+export ⋅, FieldMatrixSolver, field_matrix_solve!
 
 # Types that are teated as single values when using matrix fields.
 const SingleValue =
     Union{Number, Geometry.AxisTensor, Geometry.AdjointAxisTensor}
 
 include("band_matrix_row.jl")
-include("rmul_with_projection.jl")
-include("matrix_shape.jl")
-include("matrix_multiplication.jl")
-include("lazy_operators.jl")
-include("operator_matrices.jl")
-include("field2arrays.jl")
 
 const ColumnwiseBandMatrixField{V, S} = Fields.Field{
     V,
@@ -71,6 +85,19 @@ const ColumnwiseBandMatrixField{V, S} = Fields.Field{
     },
 }
 
+include("rmul_with_projection.jl")
+include("matrix_shape.jl")
+include("matrix_multiplication.jl")
+include("lazy_operators.jl")
+include("operator_matrices.jl")
+include("field2arrays.jl")
+include("unrolled_functions.jl")
+include("field_name.jl")
+include("field_name_set.jl")
+include("field_name_dict.jl")
+include("field_matrix_solver.jl")
+include("single_field_solver.jl")
+
 function Base.show(io::IO, field::ColumnwiseBandMatrixField)
     print(io, eltype(field), "-valued Field")
     if eltype(eltype(field)) <: Number
@@ -82,7 +109,10 @@ function Base.show(io::IO, field::ColumnwiseBandMatrixField)
         end
         column_field = Fields.column(field, 1, 1, 1)
         io = IOContext(io, :compact => true, :limit => true)
-        @allowscalar Base.print_array(io, column_field2array_view(column_field))
+        CUDA.@allowscalar Base.print_array(
+            io,
+            column_field2array_view(column_field),
+        )
     else
         # When a BandedMatrix with non-number entries is printed, it currently
         # either prints in an illegible format (e.g., if it has AxisTensor or
--- a/src/MatrixFields/band_matrix_row.jl
+++ b/src/MatrixFields/band_matrix_row.jl
@@ -135,3 +135,9 @@ Base.:*(value::SingleValue, row::BandMatrixRow) =
 
 Base.:/(row::BandMatrixRow, value::Number) =
     map(entry -> rdiv(entry, value), row)
+
+inv(row::DiagonalMatrixRow) = DiagonalMatrixRow(inv(row[0]))
+inv(::BandMatrixRow{ld, bw}) where {ld, bw} = error(
+    "The inverse of a matrix with $bw diagonals is (usually) a dense matrix, \
+     so it cannot be represented using BandMatrixRows",
+)
--- a/src/RecursiveApply/RecursiveApply.jl
+++ b/src/RecursiveApply/RecursiveApply.jl
@@ -157,9 +157,6 @@ rconvert(::Type{T}, X) where {T} =
 Recursively scale each element of `X` by `Y`.
 """
 rmul(X, Y) = rmap(*, X, Y)
-rmul(w::Number, X) = rmap(x -> w * x, X)
-rmul(X, w::Number) = rmap(x -> x * w, X)
-rmul(w1::Number, w2::Number) = w1 * w2
 const ⊠ = rmul
 
 """
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -570,6 +570,23 @@ steps:
           slurm_gpus: 1
           slurm_mem: 40GB
 
+      - label: "Unit: field names"
+        key: unit_field_names
+        command: "julia --color=yes --check-bounds=yes --project=test test/MatrixFields/field_names.jl"
+
+      - label: "Unit: field matrix solvers (CPU)"
+        key: unit_field_matrix_solvers_cpu
+        command: "julia --color=yes --check-bounds=yes --project=test test/MatrixFields/field_matrix_solvers.jl"
+        agents:
+          slurm_mem: 40GB
+
+      - label: "Unit: field matrix solvers (GPU)"
+        key: unit_field_matrix_solvers_gpu
+        command: "julia --color=yes --project=test test/MatrixFields/field_matrix_solvers.jl"
+        agents:
+          slurm_gpus: 1
+          slurm_mem: 40GB
+
   - group: "Unit: Hypsography"
     steps:
 
--- a/docs/src/matrix_fields.md
+++ b/docs/src/matrix_fields.md
@@ -26,6 +26,18 @@ MultiplyColumnwiseBandMatrixField
 operator_matrix
 ```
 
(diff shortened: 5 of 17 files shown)


dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

src/MatrixFields/MatrixFields.jl
```diff
@@ -27,10 +27,11 @@ for them:
 """
 module MatrixFields
 
-import CUDA: @allowscalar
-import LinearAlgebra: UniformScaling, Adjoint, AdjointAbsVec
+import CUDA: @allowscalar, @cuda, threadIdx, blockIdx, blockDim
```
charleskawczynski, line 30: Can we just qualify these where used instead?

src/MatrixFields/band_matrix_row.jl
```diff
@@ -135,3 +135,9 @@ Base.:*(value::SingleValue, row::BandMatrixRow) =
 
 Base.:/(row::BandMatrixRow, value::Number) =
     map(entry -> rdiv(entry, value), row)
+
+inv(row::DiagonalMatrixRow) = DiagonalMatrixRow(inv(row[0]))
```
charleskawczynski, line 139:
```suggestion
inv(row::DiagonalMatrixRow) = DiagonalMatrixRow(inv(row[1]))
```
is there a custom `getindex` for `DiagonalMatrixRow`?

dennisYatunin: Yes, it indexes using the diagonal index. So -1 corresponds to the first lower diagonal, 1 corresponds to the first upper diagonal, etc.

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

src/MatrixFields/band_matrix_row.jl
```diff
@@ -135,3 +135,9 @@ Base.:*(value::SingleValue, row::BandMatrixRow) =
...
+inv(::BMR) where {BMR <: BandMatrixRow} = error(
```
charleskawczynski, line 140:
```suggestion
inv(::BandMatrixRow) = error(
```

dennisYatunin:
bors try

dennisYatunin:
bors try

charleskawczynski:
Before we move forward with this, @dennisYatunin, please move https://github.com/CliMA/ClimaAtmos.jl/blob/16c2a66b6146a3229fc4a594e084c05b8a92151b/src/utils/abbreviations.jl#L79-L85 into ClimaCore (and then remove them from ClimaAtmos), so that we are sure that this new code is compatible with these additional definitions.

dennisYatunin:
@charleskawczynski Sure thing. I think this PR is in a decent state now (I just need to add some more tests and docs), so I'll leave it open for review while I do that and #1445. @simonbyrne, could also you take a look at this in the meantime?

dennisYatunin:
bors try

src/MatrixFields/var_and_var_tree.jl
```diff
@@ -0,0 +1,157 @@
...
+    VariableName(name_chain...)
+
+A singleton type that represents a chain of `getproperty` calls, which can be
+used to access a property or sub-property of an object `x` using the function
+`get_var(x, var)`. The entire object `x` can also be accessed with the empty
+`VariableName()`.
+"""
+struct VariableName{name_chain} end
+VariableName() = VariableName{()}() # This is required for type stability.
+VariableName(name_chain...) = VariableName{name_chain}()
```
simonbyrne, line 11: It looks like this largely supplants the stuff in https://github.com/CliMA/ClimaCore.jl/blob/5215d8dd3496bc01db7f0aa53fd2c06b61a19a99/src/Fields/field_iterator.jl

@charleskawczynski can we combine it?

src/MatrixFields/var_set.jl
```diff
@@ -0,0 +1,336 @@
...
+    - `FieldVectorKeys`
+    - `FieldMatrixKeys`
+
+TODO
+"""
+struct VarSet{
+    T <: Union{VariableName, VariableNameTup},
+    V <: NTuple{<:Any, T},
+    VT <: Union{VarTree, Nothing},
+} <: AbstractSet{T}
```
simonbyrne, line 16: how (and where) is a `VarSet` used?

dennisYatunin: It's the equivalent of the built-in `KeySet` for `VarDict`, but with additional functionality for handling `VariableName`s and row-column pairs of `VariableNames`.

dennisYatunin: I added a blurb about each of the new structs in the PR description.

src/MatrixFields/var_and_var_tree.jl
dennisYatunin: Yes I think `field_iterator` can be replaced by (or turned into a wrapper of) `FieldVectorView`, and `property_chains` can be replaced by `top_level_vars`.

simonbyrne: would it make sense to move it there?

simonbyrne: I also don't see any tests or examples on how to use it?

charleskawczynski: Agreed, and yes, we should definitely add tests

dennisYatunin: Sure, I can move it and add more docs/tests. Two questions, though:
- Should we try to maintain the old API of `property_chains` and `single_field` (and just make those functions into wrappers for the new functions), or should we make this a breaking change and have users switch to writing `for field in FieldVectorView(Y) ... end`?
- The `VariableName` infrastructure (`VarTree` + `VarSet` + `VarDict`) depends on both `Field`s and `ColumnwiseBandMatrixField`s, the latter of which is only defined in the MatrixFields module. Should we move the `ColumnwiseBandMatrixField` definition to the Fields module (and maybe move all the matrix multiplication code to the Operators module)? If so, maybe that can be a future PR?

simonbyrne:
To clarify, which pieces are not working at the moment on the GPU?

dennisYatunin:
> To clarify, which pieces are not working at the moment on the GPU?

On the [last run](https://buildkite.com/clima/climacore-ci/builds/2578#018ab929-35b8-4663-a879-9533f86324bd), everything seemed to work fine on GPUs. I'm a bit worried about the allocations with the tests designed to replicate ClimaAtmos, but those are on the CPU (there are always allocations on the GPU).

simonbyrne:
Can we update this with https://github.com/CliMA/ClimaAtmos.jl/pull/2044?

src/RecursiveApply/RecursiveApply.jl
```diff
@@ -157,9 +157,9 @@ rconvert(::Type{T}, X) where {T} =
 Recursively scale each element of `X` by `Y`.
 """
 rmul(X, Y) = rmap(*, X, Y)
-rmul(w::Number, X) = rmap(x -> w * x, X)
-rmul(X, w::Number) = rmap(x -> x * w, X)
-rmul(w1::Number, w2::Number) = w1 * w2
+# rmul(w::Number, X) = rmap(x -> w * x, X)
+# rmul(X, w::Number) = rmap(x -> x * w, X)
+# rmul(w1::Number, w2::Number) = w1 * w2
```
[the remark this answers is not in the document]
dennisYatunin: Yeah the JET tests for some of the unit tests were failing before I did this.

src/MatrixFields/var_dict.jl
```diff
@@ -0,0 +1,368 @@
...
+function Base.Broadcast.broadcasted(
+    ::FieldMatrixStyle,
+    f::Union{typeof(+), typeof(-)},
+    vector_or_matrix1::FieldMatrixStyleType,
+    vector_or_matrix2::FieldMatrixStyleType,
+)
+    var_dict_type(vector_or_matrix1) == var_dict_type(vector_or_matrix2) ||
+        field_matrix_style_error(f, (vector_or_matrix1, vector_or_matrix2))
+    VarDictType = var_dict_type(vector_or_matrix1)
+    all_keys = union(keys(vector_or_matrix1), keys(vector_or_matrix2))
```
charleskawczynski, line 219: `union` necessarily allocates, so this is likely one source of allocations. If you actually need the unique set of these keys, then this likely will either allocate or be type unstable. This doesn't seem ideal.

dennisYatunin: I wrote a custom `union` (and `intersect`, `setdiff`, etc.) for `VarSet` that is type-stable and handles variable name overlaps.

dennisYatunin:
bors try

dennisYatunin:
bors cancel

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

src/MatrixFields/MatrixFields.jl
```diff
@@ -19,46 +19,60 @@ for them:
...
 
-export ⋅
 export DiagonalMatrixRow,
     BidiagonalMatrixRow,
     TridiagonalMatrixRow,
     QuaddiagonalMatrixRow,
     PentadiagonalMatrixRow
+export FieldVectorKeys, FieldVectorView, FieldVectorViewBroadcasted
+export FieldMatrixKeys, FieldMatrix, FieldMatrixBroadcasted
+export @name, ⋅, FieldMatrixSolver, field_matrix_solve!
```
charleskawczynski, line 69: `@name` is _very_ generic. I don't think this should be exported.

dennisYatunin: Ok, I'll change it!

dennisYatunin: (Or `@fieldname`?)

charleskawczynski: Idk, I'm inclined to say no because it looks an awful lot like an individual macro for something like `fieldnames`. The entire field names part of this code looks like it could/should be a part of a separate package. The API is not clear, and there are a lot of methods (this is perhaps a Julia issue, though). Maybe @simonbyrne can chime in here?

dennisYatunin: It's essentially a way to store nested `propertynames`; i.e., a type-stable version of `propertychains`.

dennisYatunin: It will probably be useful to move the `FieldName`s infrastructure to ClimaUtilities in the future. For now, though, we might want to leave it in here to sort out the interface issues more easily. @charleskawczynski Can I just leave things as they are for now, but with `@name` unexported? That would let people start testing new Jacobians in ClimaAtmos while I make further interface changes in other PRs.

dennisYatunin: Is there anything I can improve in the docs to make the API more clear?

charleskawczynski: > Can I just leave things as they are for now, but with @name unexported?

I think that's fine

PR #1441, opened 2023-09-01
Enable indexing FieldVectors with column indices

### LINKED ISSUE #1442: Enable indexing FieldVectors with column indices
We need to be able to index `FieldVectors` so we can easily pass large state vectors to columnwise functions. This will avoid having to splat all fields in function arguments. 

Extend the `Base.getindex` function

LenkaNovak:
Closes #1442 

### DIFF
--- a/src/Fields/indices.jl
+++ b/src/Fields/indices.jl
@@ -17,7 +17,13 @@ end
 
 Base.@propagate_inbounds Base.getindex(field::Field, colidx::ColumnIndex) =
     column(field, colidx)
-
+Base.@propagate_inbounds function Base.getindex(
+    fv::FieldVector{T},
+    colidx::ColumnIndex,
+) where {T}
+    values = map(x -> x[colidx], _values(fv))
+    return FieldVector{T, typeof(values)}(values)
+end
 Base.@propagate_inbounds function column(
     field::SpectralElementField1D,
     colidx::ColumnIndex{1},
--- a/test/Fields/field.jl
+++ b/test/Fields/field.jl
@@ -1,4 +1,6 @@
 using Test
+using JET
+
 using ClimaComms
 using OrderedCollections
 using StaticArrays, IntervalSets
@@ -12,6 +14,7 @@ using LinearAlgebra: norm
 using Statistics: mean
 using ForwardDiff
 using CUDA
+using CUDA: @allowscalar
 
 include(
     joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"),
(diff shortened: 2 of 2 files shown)


akshaysridhar (approved):
LGTM. Thanks for adding this method.

akshaysridhar:
bors try

charleskawczynski (commented):
Can we add an inference test to this? I don't think that the FieldVector outer constructor is type stable

LenkaNovak:
> don't think that the FieldVector outer constructor is type stable

Hi @charleskawczynski , thanks for reviewing. And sure we can add an interface test! What exactly do you have in mind? There are other functions in the same file that test methods dispatched on `FieldVector`, so I assumed it was safe.

charleskawczynski:
> Hi @charleskawczynski , thanks for reviewing. And sure we can add an interface test! What exactly do you have in mind? There are other functions in the same file that test methods dispatched on `FieldVector`, so I assumed it was safe.

It's safe, but calling that particular outer constructor is not type-stable. Please see https://github.com/CliMA/ClimaAtmos.jl/pull/2079 for an attempted fix where we're hitting the same problem with the old TC code. It has a huge performance penalty-- note that it's completely fine to use once or twice (outside of `step!`), but (at least) the constructor that calls `promote_type(map(RecursiveArrayTools.recursive_bottom_eltype, values)...)` should really not be put inside `step!`.

charleskawczynski (approved):
Alright, this new version is fully inferred and only allocates 32 bytes (per column). So I think this can work.

Thanks for fixing this @LenkaNovak!

charleskawczynski:
Actually, I'm not sure that this is working, see [this build](https://buildkite.com/clima/climaatmos-ci/builds/12802#018a76b6-9aa9-4b84-b3e7-197884051208), which fails 🤔

LenkaNovak:
Awesome, thank you for your help, @charleskawczynski !

> Actually, I'm not sure that this is working, see [this build](https://buildkite.com/clima/climaatmos-ci/builds/12802#018a76b6-9aa9-4b84-b3e7-197884051208), which fails 🤔

Ahh, is that indexing a FieldVector of FieldVectors?

charleskawczynski:
Actually, I think the breaking part is due to how we handled names in the old TC code. I think this should be fine, so long we add a test that `propertynames(field_vector) == propertynames(field_vector[colidx])`.

LenkaNovak:
bors try

LenkaNovak:
bors try

LenkaNovak:
bors try

LenkaNovak:
bors try

LenkaNovak:
bors try

LenkaNovak:
bors try

LenkaNovak:
bors try

PR #1445, opened 2023-09-02
Migrate CA column ops to ClimaCore

charleskawczynski:
This PR moves the column indefinite integrals and column mapreduce from ClimaAtmos to ClimaCore. These are needed for a few different parameterizations (e.g., radiation, gravity wave, turbconv etc.).

@dennisYatunin, I didn't see tests for this in ClimaAtmos, can you please take this PR over and add some tests?

### DIFF
--- a/src/Operators/integrals.jl
+++ b/src/Operators/integrals.jl
@@ -65,17 +65,16 @@ column_integral_definite!(
     ᶜfield::Fields.CenterFiniteDifferenceField,
 ) = _column_integral_definite!(∫field, ᶜfield)
 
-function _column_integral_definite!(
-    ∫field,
-    ᶜfield::Fields.ColumnField,
-)
+function _column_integral_definite!(∫field, ᶜfield::Fields.ColumnField)
     Δz = Fields.Δz_field(ᶜfield)
     first_level = Operators.left_idx(axes(ᶜfield))
     last_level = Operators.right_idx(axes(ᶜfield))
     ∫field_data = Fields.field_values(∫field)
     Base.setindex!(∫field_data, rzero(eltype(∫field)))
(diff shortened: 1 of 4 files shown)


src/Operators/integrals.jl
```diff
@@ -83,4 +85,190 @@ reduction_getindex(column_field, index) = @inbounds getidx(
     index - 1 + left_idx(axes(column_field)),
 )
 
-# TODO: add support for indefinite integrals
+"""
+    column_indefinite_integral!(ᶠintegral::Field, ᶜintegrand::Field)
+
+Set `ᶠintegral(z) = ∫_0^z ᶜintegrand(z) dz`, where `ᶠintegral` is on cell
+faces and `ᶜintegrand` is on cell centers.
+"""
```
simonbyrne, line 93: add to docs?

dennisYatunin:
bors try

dennisYatunin:
bors try

charleskawczynski:
bors try

charleskawczynski:
I've updated the names to match the existing ones. The tests look good to me, I think this is ready to merge.

charleskawczynski:
bors try

charleskawczynski:
It turned out that the gpu tests were not exercised in the original commit, and some changes were needed. I got the `column_mapreduce!` working in the column case, but it's failing compilation on the gpu for the extruded space case.

That function is only currently needed / used in the old EDMF dycoms case, however, it's a very useful function and we should get it working at some point (e.g., it will come in handy in the gravity wave parameterization). So, I'm included to leave it as a broken test for now. I think this PR is ready to go, I'm happy to squash if that's preferred.

dennisYatunin (approved):
New changes look good to me!

PR #1448, opened 2023-09-06
Allow HDF5Writer to append output

Sbozzolo:
This change adds an option to `HDF5Writer` so that data can be appended to an existing file instead of overwriting it.

From https://juliaio.github.io/HDF5.jl/stable/#Opening-and-closing-files

cw = read-write, create file if not existing, preserve existing contents

### DIFF
--- a/src/InputOutput/writers.jl
+++ b/src/InputOutput/writers.jl
@@ -1,7 +1,8 @@
 abstract type AbstractWriter end
 
 """
-    HDF5Writer(filename::AbstractString[, context::ClimaComms.AbstractCommsContext])
+    HDF5Writer(filename::AbstractString[, context::ClimaComms.AbstractCommsContext];
+               overwrite::Bool = true)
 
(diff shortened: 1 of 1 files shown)


src/InputOutput/writers.jl
```diff
@@ -45,11 +45,14 @@ end
 function HDF5Writer(
     filename::AbstractString,
     context::ClimaComms.AbstractCommsContext,
+    overwrite::Bool = true,
```
simonbyrne, line 48:
```suggestion
    context::ClimaComms.AbstractCommsContext;
    overwrite::Bool = true,
```
this might make more sense as a keyword argument.

simonbyrne (commented):
Seems reasonable, but we should add docs.

Sbozzolo:
I made the option a keyword and added three lines in the docstring to explain it.

simonbyrne (approved):

PR #1452, opened 2023-09-08
Support dss for Covariant123Vector

sriharshakandala:
Provide support for DSS of Covariant123Vector

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -434,6 +434,7 @@ steps:
           CLIMACOMMS_CONTEXT: "MPI"
         agents:
           slurm_ntasks: 2
+          slurm_mem: 20GB
 
       - label: "Unit: run sphere geometry distributed (3)"
         key: unit_run_sphere_geometry_distributed3
@@ -442,6 +443,7 @@ steps:
           CLIMACOMMS_CONTEXT: "MPI"
         agents:
           slurm_ntasks: 3
+          slurm_mem: 20GB
 
       - label: "Unit: run sphere geometry distributed (4)"
         key: unit_run_sphere_geometry_distributed4
@@ -450,6 +452,7 @@ steps:
           CLIMACOMMS_CONTEXT: "MPI"
         agents:
           slurm_ntasks: 4
+          slurm_mem: 20GB
 
       - label: "Unit: rectilinear cuda"
         key: unit_rectilinear_cuda
--- a/src/Spaces/dss.jl
+++ b/src/Spaces/dss.jl
@@ -69,6 +69,12 @@ function create_dss_buffer(
     nfacedof = Nij - 2
     T = eltype(parent(data))
     TS = _transformed_type(data, local_geometry, local_weights, DA) # extract transformed type
+    # Add TS for Covariant123Vector
+    # For DSS of Covariant123Vector, the third component is treated like a scalar
+    # and is not transformed
+    if eltype(data) <: Geometry.Covariant123Vector
+        TS = Geometry.UVWVector{T}
+    end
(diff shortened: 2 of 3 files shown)


sriharshakandala:
bors try

sriharshakandala:
bors try-

sriharshakandala:
bors try

simonbyrne:
Can you add a test?

simonbyrne:
A quick note as to why this is valid: our metric terms are computed from the partial derivative matrix
```math
\frac{\partial x}{\partial \xi} =
\begin{bmatrix}
H & 0 \\
(\nabla_h z)^\top & \Delta_v z
\end{bmatrix}
```
where 
- $H = \frac{\partial x_h }{\partial \xi_{1,2}}$ is the horizontal derivative matrix, which may be different at colocated points.
- $(\nabla_h z)^\top = \frac{\partial z}{\partial \xi_{1,2}}$ is the horizontal gradient of $z$ across levels. This has been DSS-ed (with respect to the horizontal metric terms), which means that ${H^{-1}}^\top \nabla_h z$ are the same at colocated points.
- $\Delta_v z$ is the vertical gradient of $z$: since all points in the same column are colocated, this should also be the same at colocated points.

This means that when we do the conversion to $(u,v,w)$:
```math
\begin{bmatrix}
u \\ v \\ w
\end{bmatrix}
=
\left(\frac{\partial \xi}{\partial x}\right)^\top
\begin{bmatrix}
u_1 \\ u_2 \\ u_3
\end{bmatrix}
=
\begin{bmatrix}
{H^{-1}}^\top & 
\frac{{H^{-1}}^\top \nabla_h z}{\Delta_v z} \\
0 & \frac{1}{\Delta_v z}
\end{bmatrix}
\begin{bmatrix}
u_h \\ u_3
\end{bmatrix}
```
the coefficients of the $u_3$ term will be the same at all colocated points.

Similarly, we have the inverse map
```math
\begin{bmatrix}
u_h \\ u_3
\end{bmatrix}
=
\left(\frac{\partial x}{\partial \xi}\right)^\top
\begin{bmatrix}
u \\ v \\ w
\end{bmatrix}
=
\begin{bmatrix}
H^\top & \nabla_h z \\
0 & \Delta_v z
\end{bmatrix}
\begin{bmatrix}
u \\ v \\ w
\end{bmatrix}
```

First consider just $u_3$: if $u_3^{(i)}$ are the colocated points we're DSS-ing, and $W^{(i))}$ are the normalized weights the DSS-ed values $\hat{u}^{(i)}_3$ are
```math
\hat{u}^{(i)}_3 
= (\Delta_v z) \sum_j W^{(j)} \frac{1}{(\Delta_v z)} u^{(j)}_3
=  \sum_j w^{(j)} u^{(j)}_3
```
since $(\Delta_v z)$ is the same at all colocated points.

Similarly for $u_h = (u_1, u_2)$, let $Q = \frac{{H^{-1}}^\top \nabla_h z}{\Delta_v z}$ (which as noted above, is the same at all colocated points), then
```math
\hat u_h^{(i)}
=  H^\top \sum_j W^{(j)} \left([(H^{-1})^\top u_h]^{(j)} +  Q u_3^{(j)} \right) + (\Delta_v z) \sum_j W^{(j)} \frac{1}{(\Delta_v z)} u^{(j)}_3
```
which I'm pretty sure cancels out to just the horizontal terms (to be continued...)

sriharshakandala:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

sriharshakandala:
bors try

simonbyrne (approved):

Issue #1453, opened 2023-09-08
NamedTuple-Field r-operations are broken


charleskawczynski:
MWE:

```julia
import ClimaCore.Geometry as Geometry
import ClimaCore.Domains as Domains
import ClimaCore.Meshes as Meshes
import ClimaCore.Topologies as Topologies
import ClimaCore.Spaces as Spaces
import ClimaCore.Fields as Fields
import ClimaCore.Operators as Operators

function foo(::Type{FT}) where {FT}
    nelems = 20
    npoly = 3
    domain = Domains.SphereDomain(FT(6371220))
    mesh = Meshes.EquiangularCubedSphere(domain, nelems)
    topology = Topologies.Topology2D(mesh)
    quad = Spaces.Quadratures.GLL{npoly + 1}()
    space = Spaces.SpectralElementSpace2D(topology, quad)
    coords = Fields.coordinate_field(space)

    ρ = ones(space)
    q = map(coords) do coord
        (; gaussian_hills=FT(0), cosine_bells = FT(0), slotted_cylinders = FT(0))
    end
    init_state = Fields.FieldVector(; ρ, ρq = ρ .* q)

    # current wind vector (Section 2.3)
    current_wind_vector = Fields.Field(Geometry.UVVector{FT}, space)
    wind_vector(coord, t) = Geometry.UVVector(FT(0), FT(0))

    div = Operators.Divergence()
    function T_lim!(tendency, state, _, t)
        @. current_wind_vector = wind_vector(coords, t)
        @. tendency.ρ = -div(state.ρ * current_wind_vector)
        @. tendency.ρq = -div(state.ρq * current_wind_vector)
        return nothing
    end
    tendency = similar(init_state)
    state = init_state
    T_lim!(tendency, state, (;), FT(0))
end
foo(Float64)
nothing
```
This used to work in version 0.10.39 (found in ClimaTimesteppers).

charleskawczynski:
@simonbyrne, it may be that this limiter test in ClimaTimeSteppers was silently not failing (until recently). Can you confirm?

In particular:

```julia
@. tendency.ρq = -div(state.ρq * current_wind_vector)
```
looks to me like an ill-defined operation (`state.ρq` is a NamedTuple-valued Field). However, I suppose that this could just operate point-wise over the values in the NamedTuple.

charleskawczynski:
Stack trace is:

```julia
julia> include("../cc_inference_problem.jl")
[...]

PR #1454, opened 2023-09-09
Fix and add test for 1453

charleskawczynski:
This PR fixes and adds a test for #1453.

### DIFF
--- a/test/aqua.jl
+++ b/test/aqua.jl
@@ -22,7 +22,7 @@ using Aqua
     for method_ambiguity in ambs
         @show method_ambiguity
     end
-    @test length(ambs) ≤ 16
+    @test length(ambs) ≤ 15
 end
 
 @testset "Aqua tests (additional)" begin
--- a/src/RecursiveApply/RecursiveApply.jl
+++ b/src/RecursiveApply/RecursiveApply.jl
@@ -46,8 +46,8 @@ rmap(fn::F, X) where {F} = fn(X)
 rmap(fn::F, X::Tuple{}) where {F} = ()
 rmap(fn::F, X::Tuple) where {F} =
     (rmap(fn, first(X)), rmap(fn, Base.tail(X))...)
-rmap(fn::F, X::NamedTuple{names}) where {F, names} =
-    NamedTuple{names}(rmap(fn, Tuple(X)))
+rmap(fn::F, X::NamedTuple) where {F} =
+    NamedTuple{nt_names(X)}(rmap(fn, Tuple(X)))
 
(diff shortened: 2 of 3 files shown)


dennisYatunin (approved):
This looks good! I think you can also delete all the methods that specialize on `::Number` in this file, e.g.,
```
rmul(w::Number, X) = rmap(x -> w * x, X)
rmul(X, w::Number) = rmap(x -> x * w, X)
rmul(w1::Number, w2::Number) = w1 * w2
```

charleskawczynski:
> This looks good! I think you can also delete all the methods that specialize on `::Number` in this file, e.g.,

I'm going to leave them there, just to be overly cautious and make sure that this is not a breaking change

simonbyrne (approved):

charleskawczynski:
I had to fix some method ambiguities, and I'm actually not 100% what behavior we want in these edge cases. cc @simonbyrne. For example:

For 2-argument empty tuple cases, do we want:
```julia
rmap(fn::F, X::Tuple{}, Y::Tuple) where {F} =
    (rmap(fn, first(Y)), rmap(fn, Base.tail(Y))...)
```
or
```julia
rmap(fn::F, X::Tuple{}, Y::Tuple) where {F} = ()
```
or
```julia
rmap(fn::F, X::Tuple{}, Y::Tuple) where {F} = error("oops")
```
?

For now I put the first option, but arguably this should be empty or error.

charleskawczynski:
I also introduced a new function, `nt_names` so that we can specialize on that, and remove the forced specialization on all the `rmap` methods. This way, forced specialization on `nt_names` can be reused, potentially resulting in fewer specializations at the cost of one more method.

test/aqua.jl
```diff
@@ -22,7 +22,7 @@ using Aqua
     for method_ambiguity in ambs
         @show method_ambiguity
     end
-    @test length(ambs) ≤ 16
+    @test length(ambs) ≤ 13
```
simonbyrne, line 25: this is failing https://github.com/CliMA/ClimaCore.jl/actions/runs/6214646588/job/16866787715?pr=1454#step:7:686
```
Aqua tests (performance): Test Failed at /home/runner/work/ClimaCore.jl/ClimaCore.jl/test/aqua.jl:25
  Expression: length(ambs) ≤ 13
   Evaluated: 15 ≤ 13
```

charleskawczynski: Ah, interesting, it was passing for me locally. I can bump it back up. @simonbyrne, can you re-review the functionality aspect I introduced here?

PR #1459, opened 2023-09-12
accessing components of AxisVectors which are zero should return zero

simonbyrne:
cc @charleskawczynski 

### DIFF
--- a/test/Geometry/axistensors.jl
+++ b/test/Geometry/axistensors.jl
@@ -8,6 +8,10 @@ ClimaCore.Geometry.assert_exact_transform() = true
     x = Geometry.Covariant12Vector(1.0, 2.0)
     y = Geometry.Contravariant12Vector(1.0, 4.0)
 
+    @test x.u₁ === 1.0
+    @test x.u₂ === 2.0
+    @test x.u₃ === 0.0
+
     M = Geometry.Axis2Tensor(
         (Geometry.Cartesian12Axis(), Geometry.Covariant12Axis()),
         [1.0 0.0; 0.5 2.0],
--- a/test/Geometry/axistensors.jl
+++ b/test/Geometry/axistensors.jl
@@ -1,4 +1,4 @@
-using Test
+using Test, JET
 using ClimaCore.Geometry, ClimaCore.DataLayouts
 using LinearAlgebra, StaticArrays
 import ClimaCore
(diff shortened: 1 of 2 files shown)


charleskawczynski (commented):
I think returning `nothing` makes this function type-unstable. Can we return `-1` and check for that in `getproperty`?

test/Geometry/axistensors.jl
```diff
@@ -8,6 +8,10 @@ ClimaCore.Geometry.assert_exact_transform() = true
     x = Geometry.Covariant12Vector(1.0, 2.0)
     y = Geometry.Contravariant12Vector(1.0, 4.0)
 
+    @test x.u₁ === 1.0
+    @test x.u₂ === 2.0
+    @test x.u₃ === 0.0
```
charleskawczynski, line 13:
```suggestion
    @test x.u₃ === 0.0
    @inferred Geometry.idxin((1,2), 1)
```

simonbyrne:
> I think returning `nothing` makes this function type-unstable. Can we return `-1` and check for that in `getproperty`?

It's inlined, so should be handled by small union splitting.

charleskawczynski:
Ok, if it happens to be an issue, I think we can just inline some `symidx` logic (it's only used in one function) and write `getproperty` as:

```julia
@inline function Base.getproperty(x::AxisVector, name::Symbol)
    ax = axes(x, 1)
    n = symidx(ax, name)
    S = symbols(ax)
    if name == S[1]
        return if 1 in I ? components(idxin(I, 1))[n] : zero(eltype(x))
    elseif name == S[2]
        return if 2 in I ? components(idxin(I, 2))[n] : zero(eltype(x))
    elseif name == S[3]
        return if 3 in I ? components(idxin(I, 3))[n] : zero(eltype(x))
    else
        throw(PropertyError(ax, name))
    end
end
```

simonbyrne:
symidx is already inlined?
https://github.com/CliMA/ClimaCore.jl/blob/a12d0f6c3db57942533ed17029aa5564ca0e11cd/src/Geometry/axistensors.jl#L96

charleskawczynski:
> symidx is already inlined?
> 
> https://github.com/CliMA/ClimaCore.jl/blob/a12d0f6c3db57942533ed17029aa5564ca0e11cd/src/Geometry/axistensors.jl#L96

Yeah, I meant to manually inline

charleskawczynski:
> It's inlined, so should be handled by small union splitting.

I remember this particular function still being an inference issue a while back, maybe it's since improved in Base.

charleskawczynski (approved):

PR #1463, opened 2023-09-18
Add read/write HDF5 attributes

Sbozzolo:
- [X] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [X] Unit tests are included OR N/A.
- [] Code is exercised in an integration test OR N/A.
- [X] Documentation has been added/updated OR N/A.

### DIFF
--- a/src/InputOutput/readers.jl
+++ b/src/InputOutput/readers.jl
@@ -396,3 +396,11 @@ function read_field(reader::HDF5Reader, name::AbstractString)
         error("Unsupported type $type")
     end
 end
+
+"""
+    read_attributes(reader::AbstractReader, name::AbstractString, data::Dict)
+
+Return the attributes associated to the object at `name` in the given HDF5 file.
+"""
+read_attributes(reader::HDF5Reader, name::AbstractString) =
+    h5readattr(reader.file.filename, name)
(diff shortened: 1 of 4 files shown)


src/InputOutput/readers.jl
```diff
@@ -396,3 +396,11 @@ function read_field(reader::HDF5Reader, name::AbstractString)
         error("Unsupported type $type")
     end
 end
+
+"""
+    read_attributes!(reader::AbstractReader, name::AbstractString, data::Dict)
```
simonbyrne, line 401:
```suggestion
    read_attributes(reader::AbstractReader, name::AbstractString, data::Dict)
```

simonbyrne:
This requires knowing the name in the HDF5 path. For convenience, why don't we add a `attributes` keyword arg to `write!` which will attach the relevant attributes to the object?

Sbozzolo:
`write!` can write multiple groups, which one do we attach the attribute to?

simonbyrne:
i meant for writing fields.

Sbozzolo:
> i meant for writing fields.

When we write a field, we also write its space:
https://github.com/CliMA/ClimaCore.jl/blob/0577957e7fc0c8ca32ca29aa8de75637075d9c69/src/InputOutput/writers.jl#L401

I think it makes sense that the attribute gets attached to `fields/$name`, but it's not as straightforward.

PR #1464, opened 2023-09-18
Check if "ClimaCore version" is already attribute

Sbozzolo:
HDF5.jl complains if you try writing an attribute that already exists, so that appending to `ClimaCore`-generated files can be a problem

### DIFF
--- a/src/InputOutput/writers.jl
+++ b/src/InputOutput/writers.jl
@@ -59,7 +59,16 @@ function HDF5Writer(
     else
         file = h5open(filename, mode, context.mpicomm)
     end
-    write_attribute(file, "ClimaCore version", string(VERSION))
+    # Add an attribute to the file if it doesn't already exist
+    if haskey(attributes(file), "ClimaCore version")
+        file_version = VersionNumber(attrs(file)["ClimaCore version"])
+        current_version = VERSION
+        if file_version != current_version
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Maybe in the future we should consider checking if the version of ClimaCore is the same and allow appending only if it is.

src/InputOutput/writers.jl
```diff
@@ -59,7 +59,9 @@ function HDF5Writer(
     else
         file = h5open(filename, mode, context.mpicomm)
     end
-    write_attribute(file, "ClimaCore version", string(VERSION))
+    # Add an attribute to the file if it doesn't already exist
+    haskey(attributes(file), "ClimaCore version") ||
+        write_attribute(file, "ClimaCore version", string(VERSION))
```
simonbyrne, line 64:
```suggestion
    if haskey(attributes(file), "ClimaCore version")
        file_version = VersionNumber(attrs(file)["ClimaCore version"])
        current_version = VERSION
        if file_version != current_version
            @warn "$filename was written using a different version of ClimaCore than is currently loaded" file_version current_version
        end
    else
        write_attribute(file, "ClimaCore version", string(VERSION))
    end
```

Sbozzolo:
I cannot see how this change led to the failing test.

PR #1471, opened 2023-09-21
Allow SphereDomain to take Ints

### LINKED ISSUE #1468: `SphereDomain` does not support `Int`s
```julia
import ClimaComms: context
import ClimaCore: Spaces, Domains, Meshes, Topologies, Geometry, Fields

radius = 1000

h_quad = Spaces.Quadratures.GLL{4}()
h_domain = Domains.SphereDomain(radius)
h_mesh = Meshes.EquiangularCubedSphere(h_domain, 6)
h_topology = Topologies.Topology2D(
            context(),
            h_mesh,
            Topologies.spacefillingcurve(h_mesh),
        )
h_space = Spaces.SpectralElementSpace2D(
            h_topology,
            h_quad
        )
```

Returns

```
ERROR: MethodError: no method matching legendre(::Type{Int64}, ::Int64, ::GaussQuadrature.EndPt)

Closest candidates are:
  legendre(::Type{T}, ::Integer, ::GaussQuadrature.EndPt) where T<:AbstractFloat
   @ GaussQuadrature ~/.julia/packages/GaussQuadrature/XPJQd/src/GaussQuadrature.jl:114
  legendre(::Any, ::Any)
   @ GaussQuadrature ~/.julia/packages/GaussQuadrature/XPJQd/src/GaussQuadrature.jl:126
  legendre(::Type{T}, ::Integer) where T<:AbstractFloat
   @ GaussQuadrature ~/.julia/packages/GaussQuadrature/XPJQd/src/GaussQuadrature.jl:114
  ...
```

Sbozzolo:
Closes #1468

### DIFF
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -149,9 +149,11 @@ coordinate_type(domain::RectangleDomain) = typeof(
 
 A domain representing the surface of a sphere with radius `radius`.
 """
-struct SphereDomain{FT} <: AbstractDomain where {FT <: AbstractFloat}
+struct SphereDomain{FT <: AbstractFloat} <: AbstractDomain
     radius::FT
 end
+SphereDomain(radius) = SphereDomain(float(radius))
+
 Base.show(io::IO, domain::SphereDomain) =
     print(io, nameof(typeof(domain)), ": radius = ", domain.radius)
 
--- a/test/Domains/domains.jl
+++ b/test/Domains/domains.jl
@@ -1,0 +1,15 @@
+import ClimaCore: Domains
+
+using Test
+
+@testset "Domain" begin
+    # Check that we can build a SphereDomain with all the reasonable types
+    for T in (Float64, Float32, Int)
+        radius = convert(T, 1)
+
+        domain = Domains.SphereDomain(radius)
+
+        @test typeof(domain) <: Domains.AbstractDomain
+        @test domain.radius == radius
+    end
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,6 +20,8 @@ if !Sys.iswindows()
     @safetestset "DataLayouts 1dx" begin @time include("DataLayouts/data1dx.jl") end
     @safetestset "DataLayouts 2dx" begin @time include("DataLayouts/data2dx.jl") end
 
+    @safetestset "Domains" begin @time include("Domains/domains.jl") end
+
     @safetestset "Geometry" begin @time include("Geometry/geometry.jl") end
     @safetestset "AxisTensors" begin @time include("Geometry/axistensors.jl") end
 


simonbyrne (approved):

charleskawczynski:
This won't work as is for `Float32`:

```julia
import ClimaCore.Fields as Fields
import ClimaComms
import ClimaCore.Utilities as Utilities
import ClimaCore.Geometry as Geometry
import ClimaCore.Meshes as Meshes
import ClimaCore.Spaces as Spaces
import ClimaCore.Topologies as Topologies
import ClimaCore.Domains as Domains

FT = Float32
zelem = 10;
context = ClimaComms.SingletonCommsContext();
helem = 4;
Nq = 4;
# radius = FT(128);
radius = 128;
zlim = (0, 1);
vertdomain = Domains.IntervalDomain(
    Geometry.ZPoint{FT}(zlim[1]),
    Geometry.ZPoint{FT}(zlim[2]);
    boundary_tags = (:bottom, :top),
);
vertmesh = Meshes.IntervalMesh(vertdomain, nelems = zelem);
vtopology = Topologies.IntervalTopology(context, vertmesh);
vspace = Spaces.CenterFiniteDifferenceSpace(vtopology);
hdomain = Domains.SphereDomain(radius);
hmesh = Meshes.EquiangularCubedSphere(hdomain, helem);
htopology = Topologies.Topology2D(context, hmesh);
quad = Spaces.Quadratures.GLL{Nq}();
hspace = Spaces.SpectralElementSpace2D(htopology, quad);
cspace = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace);
```
results in:
```julia
julia> cspace = Spaces.ExtrudedFiniteDifferenceSpace(hspace, vspace);
ERROR: MethodError: no method matching blockmat(::ClimaCore.Geometry.AxisTensor{Float64, 2, Tuple{ClimaCore.Geometry.LocalAxis{(1, 2)}, ClimaCore.Geometry.CovariantAxis{(1, 2)}}, StaticArraysCore.SMatrix{2, 2, Float64, 4}}, ::ClimaCore.Geometry.AxisTensor{Float32, 2, Tuple{ClimaCore.Geometry.LocalAxis{(3,)}, ClimaCore.Geometry.CovariantAxis{(3,)}}, StaticArraysCore.SMatrix{1, 1, Float32, 1}})
...
```

I generally find these sort of type promotions problematic unless they're applicable in the rest of the `space`. Otherwise we can't inherit them, and then we need to account for mixed precision in a bunch of places (assuming we want that)

charleskawczynski:
Alternatively, we could allow the domain to take ints, and promote later. My only concern with that is that eltypes won’t match in different parts of the space, but that might be fine if we never use the domain for the eltype. But, at some point along constructing spaces, I think we’ll (generally) need floats (but maybe not in the edge case of uniform meshes?)

Sbozzolo:
You are right. And we already have this problem: If `radius` in your example were a double, you would get the same error. 

There's many places where we attach the floating point type, so it is very easy to come up with conflicts. I don't know what the best way to handle this.

simonbyrne:
One thing we found with climatemachine is that it could be advantageous to compute all the metric terms in higher precision, then reduce them later. In that case, maybe it would make sense to move the float type option to the space constructor?

PR #1473, opened 2023-09-25
Fix lists in docs

Sbozzolo:
In the flavor of markdown used by Documenter

```
- ABC 
DEF
```
renders differently than

```
- ABC 
  DEF
```
For long lists, we want the second to avoid items that belong to the same item to be split new paragraphs.


Sbozzolo:
See problem here: https://clima.github.io/ClimaCore.jl/dev/math_framework/

PR #1475, opened 2023-09-26
Add distributed_remapping.jl

Sbozzolo:
This pull request introduces a new way to interpolate a field onto a given set of points that works for distributed simulations.

This change introduces a new object, a `Remapper`. `Remapper`s are process-local and are defined starting from the target points where we want to interpolate and the `Space`. `Remapper`s contain all the information necessary to perform interpolations of all the fields that are defined on the same `Space`. 

Once a `Remapper` is defined, it can be used as many times as needed. To use a `Remapper`, just call the new `interpolate` function with the field you want to remap:
```julia
remapper = Remapper(hcoords, zcoords, space)
interpolate(remapper, field) 
# => Array of values interpolated on hcoords × zcoords
```
Internally, `interpolate` computes the process-local points and then calls a `MPI.reduce` to return the final value in the root process.

### DIFF
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -151,23 +151,9 @@ function Remapper(target_hcoords, target_zcoords, space)
     horz_mesh = horizontal_topology.mesh
 
     interpolation_coeffs = map(local_target_hcoords) do hcoord
-        helem = Meshes.containing_element(horz_mesh, hcoord)
         quad = Spaces.quadrature_style(space)
         quad_points, _ = Spaces.Quadratures.quadrature_points(FT, quad)
-        ξ = Meshes.reference_coordinates(horz_mesh, helem, hcoord)
-        WI1 = Spaces.Quadratures.interpolation_matrix(
-            SVector(ξ[1]),
-            quad_points,
-        )
-        if is_1d
-            return (WI1,)
-        else
-            WI2 = Spaces.Quadratures.interpolation_matrix(
-                SVector(ξ[2]),
-                quad_points,
-            )
-            return (WI1, WI2)
-        end
+        return interpolation_weights(horz_mesh, helem, hcoord, quad_points)
     end
 
     # For 2D meshes, we have a notion of local and global indices. This is not the case for
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -153,7 +153,7 @@ function Remapper(target_hcoords, target_zcoords, space)
     interpolation_coeffs = map(local_target_hcoords) do hcoord
         quad = Spaces.quadrature_style(space)
         quad_points, _ = Spaces.Quadratures.quadrature_points(FT, quad)
-        return interpolation_weights(horz_mesh, helem, hcoord, quad_points)
+        return interpolation_weights(horz_mesh, hcoord, quad_points)
     end
 
     # For 2D meshes, we have a notion of local and global indices. This is not the case for
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -76,7 +76,15 @@ function target_hcoords_pid_bitmask(target_hcoords, topology, pid)
     return pid_hcoord.(target_hcoords) .== pid
 end
 
-struct Remapper{T <: ClimaComms.AbstractCommsContext, T1, T2, T3, T4, T5}
+struct Remapper{
+    T <: ClimaComms.AbstractCommsContext,
+    T1,
+    T2,
+    T3,
+    T4,
+    T5,
+    T6 <: Spaces.AbstractSpace,
+}
     comms_ctx::T
 
     # Target points that are on the process where this object is defined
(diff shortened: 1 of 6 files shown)


Sbozzolo:
bors try

Sbozzolo:
bors try

Sbozzolo:
bors try

Sbozzolo:
bors try

Sbozzolo:
I added support for planes as well. So, I think this PR is ready for review.

https://github.com/CliMA/ClimaAtmos.jl/pull/2179 is the ClimaAtmos side.

Here's an example of output:
rhoa_inst

src/Remapping/distributed_remapping.jl
```diff
@@ -0,0 +1,297 @@
...
+remapper = Remapper(hcoords, zcoords, space)
+
+int1 = interpolate(remapper, field1)
+int2 = interpolate(remapper, field2)
+
+```
+"""
+function interpolate(remapper::Remapper, field::T) where {T <: Fields.Field}
+
+    axes(field) == remapper.space ||
```
charleskawczynski, line 246:
```suggestion
    axes(field) === remapper.space ||
```
I think we want to compare the pointer here

src/Remapping/distributed_remapping.jl
```diff
@@ -0,0 +1,297 @@
...
+function Remapper(target_hcoords, target_zcoords, space)
+
+    comms_ctx = ClimaComms.context(space)
+    pid = ClimaComms.mypid(comms_ctx)
+    FT = Spaces.undertype(space)
+    if hasproperty(space, :horizontal_space)
+        horizontal_topology = space.horizontal_space.topology
+    else
+        horizontal_topology = space.topology
+    end
```
charleskawczynski, line 138:
```suggestion
    horizontal_topology = Spaces.topology(space)
```

src/Remapping/distributed_remapping.jl
```diff
@@ -0,0 +1,297 @@
...
+            WI1 = Spaces.Quadratures.interpolation_matrix(
+                SVector(ξ1),
+                quad_points,
+            )
+            WI2 = Spaces.Quadratures.interpolation_matrix(
+                SVector(ξ2),
+                quad_points,
+            )
+            return (WI1, WI2)
+        end
```
charleskawczynski, line 179:
```suggestion
        ξ = Meshes.reference_coordinates(horz_mesh, helem, hcoord)
        WI1 = Spaces.Quadratures.interpolation_matrix(
            SVector(ξ[1]),
            quad_points,
        )
        if is_1d
            return (WI1,)
        else
            WI2 = Spaces.Quadratures.interpolation_matrix(
                SVector(ξ[2]),
                quad_points,
            )
            return (WI1, WI2)
        end
```

src/Remapping/distributed_remapping.jl
Sbozzolo: I think that's what's already happening since `Space`s don't define `==`. 

From a philosophical point of view, I would like if two spaces are interchangeable, not if they are identical. The code works if you have to distinct but equivalent spaces (even if this concept is currently blurred because `Space`s are immutable.

src/Remapping/distributed_remapping.jl
Sbozzolo: Nice! Didn't know about that.

src/Remapping/distributed_remapping.jl
simonbyrne: eh, above seems clearer.

Sbozzolo: I am going to define a function `interpolation_weights` and dispatch over the function

src/Remapping/distributed_remapping.jl
charleskawczynski: Ah, right. And I agree with your philosophical point.

Sbozzolo:
bors try

test/Remapping/distributed_remapping.jl
```diff
@@ -0,0 +1,308 @@
...
+        horzmesh,
+    )
+    horzspace = Spaces.SpectralElementSpace1D(horztopology, quad)
+
+    hv_center_space =
+        Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_center_space)
+
+    coords = Fields.coordinate_field(hv_center_space)
+
+    xpts = range(-500.0, 500.0, length = 21)
```
simonbyrne, line 51: I think you should be able to just do
```suggestion
    xpts = range(-Geometry.XPoint(500.0), Geometry.XPoint(500.0), length = 21)
```

test/Remapping/distributed_remapping.jl
```diff
@@ -0,0 +1,308 @@
+using Logging
+using Test
+using IntervalSets
+
+import ClimaCore:
+    Domains, Fields, Geometry, Meshes, Operators, Spaces, Topologies, Remapping
+using ClimaComms
+const context = ClimaComms.MPICommsContext()
```
simonbyrne, line 8: Could we split the MPI tests from the singleton tests?

test/Remapping/distributed_remapping.jl
Sbozzolo: I can do this only in the case of a plane. 

I need to build `hcoords` with the correct type, and I cannot easily assemble, e.g.,  `XYPoint`s from `XPoint`s and `YPoint`s (see https://github.com/CliMA/ClimaCore.jl/issues/1476).

Sbozzolo:
bors try

simonbyrne (approved):

Sbozzolo:
bors try

Sbozzolo:
@charleskawczynski if you want to have another look, I think that this is ready to be merged

src/Remapping/interpolate_array.jl
```diff
@@ -144,22 +142,67 @@ function interpolate_array(
...
+)
+    helem = Meshes.containing_element(horz_mesh, hcoord)
+    ξ1, ξ2 = Meshes.reference_coordinates(horz_mesh, helem, hcoord)
+    WI1 = Spaces.Quadratures.interpolation_matrix(SVector(ξ1), quad_points)
+    WI2 = Spaces.Quadratures.interpolation_matrix(SVector(ξ2), quad_points)
+    return (WI1, WI2)
+end
+
+function interpolation_weights(
+    horz_mesh::Meshes.AbstractMesh1D,
```
charleskawczynski, line 182: At some point I’d like to rename this variable, maybe just use `hmesh`?

PR #1479, opened 2023-09-28
Update CUDA and dependencies

charleskawczynski:
This PR updates the CUDA compat entry and the dependencies, upgrading us to CUDA 5.


charleskawczynski:
bors try

Sbozzolo (approved):
Just leaving this here for our reference:

> 
> CUDA.jl 5.0 raises the required Julia and CUDA versions. The minimum Julia version is now 1.8, which should be enforced by the Julia package manager. The minimum CUDA toolkit version is now 11.4, but this cannot be enforced by the package manager. As a result, if you need to use an older version of the CUDA toolkit, you will need to pin CUDA.jl to v4.4 or below. [The README](https://github.com/JuliaGPU/CUDA.jl/blob/master/README.md) now contains a table of supported CUDA toolkit versions and the corresponding CUDA.jl release.
> 
> Most users will not be affected by this change: If you use the artifact-provided CUDA toolkit, you will automatically get the latest version supported by your CUDA driver.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

simonbyrne:
It doesn't look like we're using CUDA 5: do you know what's blocking the upgrade?

charleskawczynski:
> It doesn't look like we're using CUDA 5: do you know what's blocking the upgrade?

Oh, weird, I thought it was, let me check.

charleskawczynski:
@simonbyrne, alright, manifests were updated. I'll run CI to see how this goes.

charleskawczynski:
bors try

simonbyrne:
bors try-

simonbyrne:
bors try

PR #1495, opened 2023-10-11
Downgrade to CUDA 4

charleskawczynski:
CUDA 5 may have been causing some CI issues, so this PR downgrades back to CUDA 4.


simonbyrne:
bors try

PR #1497, opened 2023-10-11
Add support for topography in `distributed_remapping`

### LINKED ISSUE #1488: Add support for hypsography in `interpolate_array`
As the docstring for `interpolate_array` mentions:
```julia
!!! note
    Hypsography is not currently handled correctly.
```
The reason for this is in the call `velem = Meshes.containing_element(vert_mesh, zcoord)` in the `interpolate_slab_level_function`. With topography, this function looks up the `zcoord` above the local surface, not the `zcoord` above the sea level.

There's two valid ways to approach this problem (both required): 
- interpret `zcoord` as level. In this case, the `ClimaAtmos` NetCDF writer will have to handle this accordingly and output an elevation map
- find the correct vertical element that contains the coordinate (and return `NaN` or a fill value when not existing).

Sbozzolo:
This PR adds a new mode to the `interpolate` function for distributed remapping, `interpolate_topography`. When `interpolate_topography` is `true`, the target `zcoords` are assumed to be with respect to the mean sea level as opposed to with respect to the surface. `zcoords` that are below the surface are filled with `NaN`s.

Closes #1488

### DIFF
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -124,7 +124,6 @@ The `Remapper` is designed to not be tied to any particular `Field`. You can use
 `Remapper` for any `Field` as long as they are all defined on the same `topology`.
 
 `Remapper` is the main argument to the `interpolate` function.
-
 """
 function Remapper(target_hcoords, target_zcoords, space)
 
@@ -192,9 +191,18 @@ end
 
 
 """
-   interpolate(remapper, field)
+   interpolate(remapper, field; physical_z = false)
 
 Interpolate the given `field` as prescribed by `remapper`.
+
+
+Keyword arguments
+==================
+
+- `physical_z`: When `true`, interpolate to the physical z coordinates, taking into account
+                hypsography. If `false` (the default), interpolation will be based on the
+                reference z coordinate, i.e. will correspond to constant model levels.
+                `NaN`s are returned for values that are below the surface.
 
 Example
 ========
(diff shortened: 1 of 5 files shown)


Sbozzolo:
bors try

Sbozzolo:
bors try

Sbozzolo:
bors try

simonbyrne:
is this ready for review?

Sbozzolo:
> is this ready for review?

Yes, I forgot to mark it as such.

src/Remapping/distributed_remapping.jl
```diff
@@ -192,10 +191,19 @@ end
...
 
+
+Keyword arguments
+==================
+
+- `interpolate_topography`: When `true`, the `target_zcoords` to `remapper` are interpreted
+                            as "from mean sea level" (instead of "from surface") and
+                            topography is interpolated. `NaN`s are returned for values that
+                            are below the surface.
+
```
simonbyrne, line 206: A couple of points:
- elsewhere in ClimaCore we use "hypsography" not "topography"
- the distinction isn't actually MSL vs surface, but whether the Z coordinate corresponds to the "reference Z" (i.e. constant level) or the "physical Z" (constant altitude). 

So how about
```suggestion
- `physical_z`: When `true`, interpolate to the physical z coordinates, taking into account hypsography.
      If `false` (the default), interpolation will be based on the reference z coordinate, i.e. will correspond to constant model levels.
      `NaN`s are returned for values that are below the surface.

```

simonbyrne (approved):
Other than the above issue, it looks good.

Sbozzolo:
> Other than the above issue, it looks good.

Addressed.

Sbozzolo:
bors try

Sbozzolo:
bors try

simonbyrne:
Haven't seen that one fail before!

PR #1504, opened 2023-10-19
Add and test function-based indefinite integrals

### LINKED ISSUE #1492: Define indefinite column integrals for arbitrary functions
ClimaAtmos currently uses

```julia
    prob = SciMLBase.ODEProblem(dp_dz, p_0, (FT(0), z_max))
    return SciMLBase.solve(
        prob,
        ODE.Tsit5(),
        reltol = 10eps(FT),
        abstol = 10eps(FT),
    )
```
to compute a vertical indefinite integral. We currently only need this for initialization, and https://github.com/CliMA/ClimaAtmos.jl/pull/2183 implements this in a way that seems to work (MSE table changes are small). We could start by porting that over to ClimaCore and adding unit tests. Next (if we need) we could make this work on the GPU, which I think shouldn't be too difficult (it may work already as I've followed a similar pattern used for the other indefinite integral on fields).

charleskawczynski:
This PR adds an indefinite integral implementation that uses a function-based input argument. Concretely, given `ϕ(z)`, this interface allows us to solve for `∂ϕ/∂z = f(ϕ,z)`. The existing field-based implementation cannot be used in this situation, because `f` is a function of `ϕ`, which is what we're computing. So, this implementation uses RootSolvers to find `ϕ` at the next level in order to provide a somewhat generic interface.

Unfortunately, one limitation of this is that `ϕ` must be a scalar field, and it cannot be, e.g., a `NamedTuple`. We could try to lift this assumption, but that would require adjustments to RootSolvers.

Closes #1492.

### DIFF
--- a/src/Operators/integrals.jl
+++ b/src/Operators/integrals.jl
@@ -1,6 +1,8 @@
-import ..RecursiveApply: rzero, ⊠, ⊞
+import ..RecursiveApply: rzero, ⊠, ⊞, radd, rmul
 import ..Utilities
 import ..DataLayouts
+import RootSolvers
+import ClimaComms
 
 """
     column_integral_definite!(∫field::Field, ᶜfield::Field)
@@ -86,6 +88,21 @@ end
 Sets `ᶠ∫field```(z) = \\int_0^z\\,```ᶜfield```(z')\\,dz'``. The input `ᶜfield`
 must lie on a cell-center space, and the output `ᶠ∫field` must lie on the
 corresponding cell-face space.
+
+    column_integral_indefinite!(
+        f::Function,
+        ᶠ∫field::Fields.ColumnField,
+        ϕ₀ = 0,
+        average = (ϕ⁻, ϕ⁺) -> (ϕ⁻ + ϕ⁺) / 2,
+    )
+
+The indefinite integral `ᶠ∫field = ϕ(z) = ∫ f(ϕ,z) dz` given:
+
+- `f` the integral integrand function (which may be a function)
+- `ᶠ∫field` the resulting (scalar) field `ϕ(z)`
+- `ϕ₀` (optional) the boundary condition
+- `average` (optional) a function to compute the cell center
+   average between two cell faces (`ϕ⁻, ϕ⁺`).
 """
 column_integral_indefinite!(ᶠ∫field::Fields.Field, ᶜfield::Fields.Field) =
     column_integral_indefinite!(ClimaComms.device(ᶠ∫field), ᶠ∫field, ᶜfield)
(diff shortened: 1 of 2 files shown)


src/Operators/integrals.jl
```diff
@@ -155,6 +170,94 @@ function _column_integral_indefinite!(
...
+    Fields.bycolumn(axes(ᶠ∫field)) do colidx
+        column_integral_indefinite_cpu!(f, ᶠ∫field[colidx], args...)
+    end
+
+#=
+Function-based signature, solve for ϕ:
+```
+∂ϕ/∂z = f(ϕ,z)
+(ᶠϕ^{k+1}-ᶠϕ^{k})/ᶜΔz = ᶜf(ᶜϕ̄,ᶜz)
+ᶜϕ̄ = (ϕ^{k+1}+ϕ^{k})/2
```
Sbozzolo, line 223: Why is this called `ᶜϕ̄`? (with the minus?)

src/Operators/integrals.jl
```diff
@@ -86,6 +88,19 @@ end
 Sets `ᶠ∫field```(z) = \\int_0^z\\,```ᶜfield```(z')\\,dz'``. The input `ᶜfield`
 must lie on a cell-center space, and the output `ᶠ∫field` must lie on the
 corresponding cell-face space.
+
+    column_integral_indefinite!(
+        f::Function,
+        ᶠ∫field::Fields.ColumnField,
+        ϕ₀ = 0,
+        average = (ϕ⁻, ϕ⁺) -> (ϕ⁻ + ϕ⁺) / 2,
```
Sbozzolo, line 96: I think it would be good for the docstring to explain what `average` is and how it is used.

Also, explicitly mentioning the order of accuracy of the method would be useful, I think.

src/Operators/integrals.jl
```diff
@@ -155,6 +170,94 @@ function _column_integral_indefinite!(
...
+    f::Function,
+    ::ClimaComms.AbstractCPUDevice,
+    ᶠ∫field::Fields.FaceExtrudedFiniteDifferenceField,
+    args...,
+) =
+    Fields.bycolumn(axes(ᶠ∫field)) do colidx
+        column_integral_indefinite_cpu!(f, ᶠ∫field[colidx], args...)
+    end
+
+#=
```
Sbozzolo, line 218: (I didn't know that julia had multiline comments like this!)

Sbozzolo (commented):
Looks good to me.

Does the error in your test converge to zero if you increase the resolution of your space?

src/Operators/integrals.jl
charleskawczynski: Yeah, good point. I'll update that

src/Operators/integrals.jl
charleskawczynski: Yep! They're very handy

src/Operators/integrals.jl
charleskawczynski: It should just be an overbar? (I made it with `\^c<tab>\phi<tab>\bar<tab>`)

Sbozzolo: Ah! shouldn't be \bar\phi then? 
image
(When I copy and paste the symbol on GitHub, the bar gets lost)

src/Operators/integrals.jl
charleskawczynski: We should probably do that for all of our other operators, too..

src/Operators/integrals.jl
charleskawczynski: I don't think so? Here's what the two look like for me (left is `\bar\phi`, right is `\phi\bar`):

Screen Shot 2023-10-20 at 9 19 39 AM

I'm actually surprised that `\bar\phi` puts the bar on phi here. For example, if you do `a\bar\phi`, you'll see that the `bar` goes to `a`, and not `\phi`.

Sbozzolo: In LaTeX you put the \bar before the symbol, so I was expecting it to be the same, and if you do `\bar a`, you get a with a bar on top.

On my system, `\^c<tab>\phi<tab>\bar<tab>`, looks like:
image

charleskawczynski: Yeah, I think this works a bit differently compared to LaTeX in this regard

charleskawczynski:
> Does the error in your test converge to zero if you increase the resolution of your space?

I didn't add a convergence, that'd probably be a good idea though. I can confirm that the new operator is 2nd order accurate, but adding that test does make the test suite a bit more complicated. I'm fine with adding it if you think it's worth it.

Sbozzolo:
> > Does the error in your test converge to zero if you increase the resolution of your space?
> 
> I didn't add a convergence, that'd probably be a good idea though. I can confirm that the new operator is 2nd order accurate, but adding that test does make the test suite a bit more complicated. I'm fine with adding it if you think it's worth it.

For simple operations like this, I think that it would be a stronger test of correctness. We probably just need to run the same test with twice as many vertical elements and check the correct scaling in the error, and this might be a second for loop along with `for FT in (Float32, Float64)`. However, I agree that it makes the testsuite more complicated, so it is up to you if you want to implement it or leave it as a possible future improvement.

charleskawczynski:
> For simple operations like this, I think that it would be a stronger test of correctness. We probably just need to run the same test with twice as many vertical elements and check the correct scaling in the error, and this might be a second for loop along with `for FT in (Float32, Float64)`. However, I agree that it makes the testsuite more complicated, so it is up to you if you want to implement it or leave it as a possible future improvement.

I agree. In the interest in reducing package load times sooner than later, I'll open an issue to add convergence tests. Maybe they can be added more simply in some way.

charleskawczynski:
@Sbozzolo, alright to merge?

charleskawczynski:
Opened #1508.

Sbozzolo:
> @Sbozzolo, alright to merge?

Can you please add line to explain in the docstring what `average` is and why might want to change the default value?

charleskawczynski:
> Can you please add line to explain in the docstring what `average` is and why might want to change the default value?

Ah, thanks. I did update it but I must have accidentally stashed it. Updated, can you take another look @Sbozzolo?

Sbozzolo (approved):
Thank you!

PR #1505, opened 2023-10-19
use blocking synchronize to reduce poll waiting

simonbyrne:
Uses a blocking synchronize to synchronize GPU stream with CUDA. This appears to reduce time spent in `poll_wait` between threads (see https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/114)

### DIFF
--- a/src/Spaces/dss_cuda.jl
+++ b/src/Spaces/dss_cuda.jl
@@ -427,13 +427,13 @@ function fill_send_buffer!(::ClimaComms.CUDADevice, dss_buffer::DSSBuffer)
     if nsend > 0
         nitems = nsend * nlevels * nfid
         nthreads, nblocks = _configure_threadblock(nitems)
-        CUDA.synchronize() # CUDA MPI uses a separate stream. This will synchronize across streams
+        CUDA.synchronize(;blocking=true) # CUDA MPI uses a separate stream. This will synchronize across streams
         @cuda threads = (nthreads) blocks = (nblocks) fill_send_buffer_kernel!(
             send_data,
             send_buf_idx,
             pperimeter_data,
         )
-        CUDA.synchronize() # CUDA MPI uses a separate stream. This will synchronize across streams
+        CUDA.synchronize(;blocking=true) # CUDA MPI uses a separate stream. This will synchronize across streams
     end
     return nothing
 end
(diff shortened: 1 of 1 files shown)


src/Spaces/dss_cuda.jl
```diff
@@ -427,13 +427,13 @@ function fill_send_buffer!(::ClimaComms.CUDADevice, dss_buffer::DSSBuffer)
     if nsend > 0
         nitems = nsend * nlevels * nfid
         nthreads, nblocks = _configure_threadblock(nitems)
-        CUDA.synchronize() # CUDA MPI uses a separate stream. This will synchronize across streams
+        CUDA.synchronize(;blocking=true) # CUDA MPI uses a separate stream. This will synchronize across streams
```
simonbyrne, line 430: @sriharshakandala is this line even needed? the previous kernel is on the same stream

src/Spaces/dss_cuda.jl
```diff
@@ -468,13 +468,13 @@ function load_from_recv_buffer!(::ClimaComms.CUDADevice, dss_buffer::DSSBuffer)
     if nrecv > 0
         nitems = nrecv * nlevels * nfid
         nthreads, nblocks = _configure_threadblock(nitems)
-        CUDA.synchronize()
+        CUDA.synchronize(;blocking=true)
```
simonbyrne, line 471: I'm not sure we need this either, as `MPI.Waitall` should synchronize?

src/Spaces/dss_cuda.jl
```diff
@@ -468,13 +468,13 @@ function load_from_recv_buffer!(::ClimaComms.CUDADevice, dss_buffer::DSSBuffer)
...
         @cuda threads = (nthreads) blocks = (nblocks) load_from_recv_buffer_kernel!(
             pperimeter_data,
             recv_data,
             recv_buf_idx,
         )
-        CUDA.synchronize()
+        CUDA.synchronize(;blocking=true)
```
simonbyrne, line 477: And this one shouldn't be needed either?

src/Spaces/dss_cuda.jl
sriharshakandala: Originally, we just added the `synchronize` in 436. If I recall correctly, the addition `synchronize` was added in 430 has to do with `CUDA.jl` using a separate thread for MPI communications and making sure any earlier communication is completed!

simonbyrne:
I removed what I think are unnecessary synchronizations.

simonbyrne:
bors try

src/Spaces/dss_cuda.jl
sriharshakandala: Same here. Generally speaking this should not be required (as with CPU MPI).

simonbyrne:
Trying it in https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/116

sriharshakandala (approved):

charleskawczynski:
I think the error is unrelated to this PR?

PR #1507, opened 2023-10-20
Improve error messages for unknown types

### LINKED ISSUE #1502: Add `_first` callback for non-concrete eltype in broadcast expression
During the software meeting today, @dennisYatunin ran into a `no method matching` `_first(::Something)`. We should add a callback that throws the generic error (`BroadcastInferenceError: cannot infer eltype.`)

charleskawczynski:
Closes #1502.


Sbozzolo (approved):
Thank you!

Issue #1511, opened 2023-10-20
Distributed limiters aren't communicating correctly


simonbyrne:
## Describe the bug
The `QuasimonotoneLimiter` constructor calls `create_ghost_buffer`:
https://github.com/CliMA/ClimaCore.jl/blob/9c220472af5dab549275c183975a5b1fe735d0ba/src/Limiters/quasimonotone.jl#L53-L54

However in the only test, this currently returns `nothing`
https://github.com/CliMA/ClimaCore.jl/pull/1510

Probably not worth fixing at the moment: in #1494, we will probably change the limiters so that they will be applied only within the element (i.e. no communication should be required).

charleskawczynski:
Yeah, it looks like `create_ghost_buffer` needs to support `VIFH` data layouts.

simonbyrne:
it will also throw an error later, since we don't actually define any methods for `GhostBuffer` (I think I removed them all a while ago)

PR #1517, opened 2023-10-24
Add broken GPU limiters tests, and GPU dispatch mechanism

charleskawczynski:
This PR:
 - Adds GPU limiter tests
 - Fixes the limiter test to `ClimaComms.device()`
 - Adds a GPU dispatch mechanism, so that GPU paths are separated from CPU paths, but they currently do nothing (will follow up with implementation in a separate PR).

### DIFF
--- a/src/Limiters/quasimonotone.jl
+++ b/src/Limiters/quasimonotone.jl
@@ -1,3 +1,6 @@
+import ClimaComms
+import CUDA
+
 """
     QuasiMonotoneLimiter
 
@@ -85,12 +88,30 @@ storing it in `limiter.q_bounds`.
 Part of [`compute_bounds!`](@ref).
 """
 function compute_element_bounds!(limiter::QuasiMonotoneLimiter, ρq, ρ)
+    compute_element_bounds!(limiter, ρq, ρ, ClimaComms.device(ρ))
+end
+
+function compute_element_bounds!(
+    limiter::QuasiMonotoneLimiter,
+    ρq,
+    ρ,
+    ::ClimaComms.CUDADevice,
+) end
+
+function compute_element_bounds!(
+    limiter::QuasiMonotoneLimiter,
+    ρq,
+    ρ,
+    ::ClimaComms.AbstractCPUDevice,
+)
+    ρ_data = Fields.field_values(ρ)
+    ρq_data = Fields.field_values(ρq)
     q_bounds = limiter.q_bounds
-    (Ni, Nj, _, Nv, Nh) = size(ρq)
+    (Ni, Nj, _, Nv, Nh) = size(ρq_data)
     for h in 1:Nh
         for v in 1:Nv
-            slab_ρq = slab(ρq, v, h)
-            slab_ρ = slab(ρ, v, h)
+            slab_ρq = slab(ρq_data, v, h)
+            slab_ρ = slab(ρ_data, v, h)
             local q_min, q_max
             for j in 1:Nj
                 for i in 1:Ni
@@ -120,10 +141,21 @@ neighbors.
 
 Part of [`compute_bounds!`](@ref).
 """
+compute_neighbor_bounds_local!(limiter::QuasiMonotoneLimiter, ρ) =
+    compute_neighbor_bounds_local!(limiter, ρ, ClimaComms.device(ρ))
+
 function compute_neighbor_bounds_local!(
     limiter::QuasiMonotoneLimiter,
-    topology::Topologies.AbstractTopology,
+    ρ,
+    ::ClimaComms.CUDADevice,
+) end
+
+function compute_neighbor_bounds_local!(
+    limiter::QuasiMonotoneLimiter,
+    ρ,
+    ::ClimaComms.AbstractCPUDevice,
 )
+    topology = Spaces.topology(axes(ρ))
     q_bounds = limiter.q_bounds
     q_bounds_nbr = limiter.q_bounds_nbr
     (_, _, _, Nv, Nh) = size(q_bounds_nbr)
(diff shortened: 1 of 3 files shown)


src/Limiters/quasimonotone.jl
```diff
@@ -198,34 +228,20 @@ function compute_bounds!(
...
         ClimaComms.start(limiter.ghost_buffer.graph_context)
     end
-    compute_neighbor_bounds_local!(limiter, topology)
+    compute_neighbor_bounds_local!(limiter, ρ)
     if limiter.ghost_buffer isa Spaces.GhostBuffer
         ClimaComms.finish(ghost_buffer.graph_context)
-        compute_neighbor_bounds_ghost!(limiter, topology)
+        compute_neighbor_bounds_ghost!(limiter, Spaces.topology(axes(ρq)))
     end
 end
```
Sbozzolo, line 246: Maybe we can split this function into two methods and dispatch over `limiter.ghost_buffer`

charleskawczynski: I had thought about that, but I didn't change that because of https://github.com/CliMA/ClimaCore.jl/issues/1511, and I spoke with @simonbyrne about the possibility of just removing all things communication in the limiters code, because it's called after `dss!` has been called, so no communication think any communication is needed. Unless we need the entire element from neighboring processes, maybe that is what @simonbyrne was getting at.

I'm happy to change it, though.

Sbozzolo: Thanks. Sounds good.

(I'll try to look at the rest of the PR later)

test/Limiters/limiter.jl
```diff
@@ -113,23 +128,25 @@ end
...
         for h2 in 1:n2
             for h1 in 1:n1
                 s = slab(limiter.q_bounds, h1 + n1 * (h2 - 1))
-                q_min = s[1]
-                q_max = s[2]
-                @test q_min.x ≈ 2 * (h1 - 1)
-                @test q_min.y ≈ 3 * (h2 - 1)
-                @test q_max.x ≈ 2 * h1
-                @test q_max.y ≈ 3 * h2
+                # @show s
```
Sbozzolo, line 135: Is this a stray debugging comment?

test/Limiters/limiter.jl
```diff
@@ -194,8 +211,9 @@ end
         Limiters.compute_bounds!(limiter, ρq_ref, ρ)
         Limiters.apply_limiter!(ρq, ρ, limiter)
 
-        @test parent(ρq)[:, :, 1, 1] ≈
-              [FT(0.0) FT(0.0); FT(0.950005) FT(0.950005)] rtol = 10eps(FT)
+        @test Array(parent(ρq))[:, :, 1, 1] ≈
```
Sbozzolo, line 214: Why is an explicit `Array` needed here?

test/Limiters/limiter.jl
```diff
@@ -113,23 +128,25 @@ end
...
-                @test q_min.y ≈ 3 * (h2 - 1)
-                @test q_max.x ≈ 2 * h1
-                @test q_max.y ≈ 3 * h2
+                # @show s
+                CUDA.@allowscalar q_min = s[1]
+                CUDA.@allowscalar q_max = s[2]
+                is_cpu && @test q_min.x ≈ 2 * (h1 - 1)
+                is_cpu && @test q_min.y ≈ 3 * (h2 - 1)
+                is_cpu && @test q_max.x ≈ 2 * h1
+                is_cpu && @test q_max.y ≈ 3 * h2
```
Sbozzolo, line 141: It seems to me that all these tests are relevant only for the cpu case. It's that's the case, we could move the if statement before the two for loops. (But, since you haven't done it, I am guessing I am missing something...).

Sbozzolo (approved):

test/Limiters/limiter.jl
charleskawczynski: Ah, I thought I rebased this away, I'll fix this

test/Limiters/limiter.jl
charleskawczynski: It's needed (if data resides on the gpu) to move the data from the CPU to the GPU

Sbozzolo:
Also, please squash the commits! :)

test/Limiters/limiter.jl
charleskawczynski: Yeah, my plan is to remove these conditionals once I add the implementation in

test/Limiters/limiter.jl
Sbozzolo: Got it, thanks!

test/Limiters/limiter.jl
Sbozzolo: Sounds good!

src/Limiters/quasimonotone.jl
charleskawczynski: The split method looks like this:
```julia
compute_bounds!(
    limiter::QuasiMonotoneLimiter,
    ρq::Fields.Field,
    ρ::Fields.Field,
) = compute_bounds!(limiter, ρq, ρ, limiter.ghost_buffer)

function compute_bounds!(
    limiter::QuasiMonotoneLimiter,
    ρq::Fields.Field,
    ρ::Fields.Field,
    ghost_buffer::Spaces.GhostBuffer
)
    compute_element_bounds!(limiter, ρq, ρ)
    Spaces.fill_send_buffer!(
        Spaces.topology(axes(ρq)),
        limiter.q_bounds,
        ghost_buffer,
    )
    ClimaComms.start(ghost_buffer.graph_context)
    compute_neighbor_bounds_local!(limiter, ρ)
    ClimaComms.finish(ghost_buffer.graph_context)
    compute_neighbor_bounds_ghost!(limiter, Spaces.topology(axes(ρq)))
end

function compute_bounds!(
    limiter::QuasiMonotoneLimiter,
    ρq::Fields.Field,
    ρ::Fields.Field,
    ::Nothing
)
    compute_element_bounds!(limiter, ρq, ρ)
    compute_neighbor_bounds_local!(limiter, ρ)
end
```
I actually think I prefer the current form with union splitting. idk, what do you think?

Sbozzolo: I don't have strong opinions, especially if the code is likely to change. Pick the one your prefer.

PR #1520, opened 2023-10-27
Add gpu implementation of `QuasiMonotoneLimiter`

### LINKED ISSUE #1493: Add GPU support for horizontal limiters

charleskawczynski:
This PR adds a GPU implementation of the `QuasiMonotoneLimiter`.

Closes #1493.

### DIFF
--- a/src/Limiters/quasimonotone.jl
+++ b/src/Limiters/quasimonotone.jl
@@ -106,7 +106,71 @@ function compute_element_bounds!(
     ρq,
     ρ,
     ::ClimaComms.CUDADevice,
-) end
+)
+    S = size(Fields.field_values(ρ))
+    (Ni, Nj, _, Nv, Nh) = S
+    nthreads, nblocks = config_threadblock(Nv, Nh)
+
+    CUDA.@cuda threads = nthreads blocks = nblocks compute_element_bounds_kernel!(
+        limiter,
+        Fields.field_values(Operators.strip_space(ρq, axes(ρq))),
+        Fields.field_values(Operators.strip_space(ρ, axes(ρ))),
+        Nv,
+        Nh,
+        Val(Ni),
+        Val(Nj),
+    )
+    return nothing
+end
+
+function config_threadblock(Nv, Nh)
+    nitems = Nv * Nh
+    nthreads = min(256, nitems)
+    nblocks = cld(nitems, nthreads)
+    return (nthreads, nblocks)
+end
+
+function get_hv(Nv, Nh, blockIdx, threadIdx, blockDim, gridDim)
+    tidx = (blockIdx.x - 1) * blockDim.x + threadIdx.x
+    (h, v) = CartesianIndices((1:Nh, 1:Nv))[tidx].I
+    # @cuprintln("Nv,Nh,v,h=($Nv, $Nh,$v,$h)")
+    return (h, v)
+end
+
+function compute_element_bounds_kernel!(
+    limiter,
+    ρq,
+    ρ,
+    Nv,
+    Nh,
+    ::Val{Ni},
+    ::Val{Nj},
+) where {Ni, Nj}
+    (h, v) = get_hv(Nv, Nh, blockIdx(), threadIdx(), blockDim(), gridDim())
+    if h ≤ Nh && v ≤ Nv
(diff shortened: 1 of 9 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
CI passes for me locally. These errors are just CI failure issues.

src/Limiters/quasimonotone.jl
```diff
@@ -158,7 +222,54 @@ function compute_neighbor_bounds_local!(
...
     ::ClimaComms.CUDADevice,
-) end
+)
+    topology = Spaces.topology(axes(ρ))
+    Ni, Nj, _, Nv, Nh = size(Fields.field_values(ρ))
+    nthreads, nblocks = config_threadblock(Nv, Nh)
+    local_neighbor_elem_offset =
+        Adapt.adapt(CuVector{Int}, topology.local_neighbor_elem_offset)
+    local_neighbor_elem =
+        Adapt.adapt(CuVector{Int}, topology.local_neighbor_elem)
```
simonbyrne, line 232: This will do the conversion from a CPU array to a GPU array at every time step. Would be better to change this in the topology itself (AFAIK this isn't used by anything else)

simonbyrne: Also, I thought the plan was to change the limiters to be element-local in any case?

charleskawczynski: > This will do the conversion from a CPU array to a GPU array at every time step. Would be better to change this in the topology itself (AFAIK this isn't used by anything else)

We spoke about this last week, and I think you suggested to not adapt topology because it is such a huge struct. Regarding the element-local, I think that was specific to the water borrowing implementation. Being that this occurs after DSS, it's not clear to me why we would need any distributed support anyway. Want me to just delete `compute_neighbor_bounds_local!` and all of the ghost stuff?

simonbyrne: > I think we spoke about this last week, and I think you suggested to not adapt topology because it is such a huge struct.

You don't want to adapt topology itself, but you can make the `local_neighbor_elem`/`local_neighbor_elem_offset` fields Arrays/CuArrays depending on the device

simonbyrne: > Regarding the element-local, I think that was specific to the water borrowing implementation. Being that this occurs after DSS, it's not clear to me why we would need any distributed support anyway. Want me to just delete `compute_neighbor_bounds_local!` and all of the ghost stuff?

Maybe that's easiest? we would have to rework all the examples.

charleskawczynski: Ah, ok, and add them to the `QuasiMonotoneLimiter` struct? Sure, will do

simonbyrne: the only difference is how you compute `min`/`max`, so we could make that an option.

simonbyrne: > Ah, ok, and add them to the `QuasiMonotoneLimiter` struct? Sure, will do

No, i meant make them parametric inside the topology

charleskawczynski: I'm not sure I follow, make what parametric inside the topology struct?

charleskawczynski: > Maybe that's easiest? we would have to rework all the examples.

Why would we need to rework the examples?

charleskawczynski: @simonbyrne, where are you suggesting `local_neighbor_elem_offset` and `local_neighbor_elem` live?

charleskawczynski: Ah, I see, so just adapt those pieces, sounds good

simonbyrne (approved):

Sbozzolo:
Can we tag a ClimaCore release once this is merged?

simonbyrne:
Sure!

simonbyrne:
Scalar indexing errors:
https://buildkite.com/clima/climacore-ci/builds/2836#018bcf2d-94c3-4322-9ac1-0732487fa10e/165-199

PR #1521, opened 2023-10-27
Fix CUDA mapreduce bug, affecting fields with multiple variables.

sriharshakandala:


sriharshakandala:
bors try

src/Fields/mapreduce_cuda.jl
```diff
@@ -150,42 +150,33 @@ function mapreduce_cuda_kernel!(
     tidx = threadIdx().x
     bidx = blockIdx().x
     fidx = blockIdx().y
+    dataview = _dataview(pdata, fidx)
```
simonbyrne, line 153: Why not just pass the `data` object directly to the kernel? that way, we can just use regular datalayout indexing.

sriharshakandala: That would need a few additional changes, including moving away from linear indexing. I would prefer to do that in a different PR, if that is ok!

sriharshakandala:
bors try-

src/Fields/mapreduce_cuda.jl
simonbyrne: but i don't see how this fixes the issue?

simonbyrne (approved):
We can merge this as is, but we should do a proper fix.

PR #1529, opened 2023-11-06
Run CI on any given PR

Sbozzolo:


.github/workflows/ClimaCoreMakie.yml
```diff
@@ -1,13 +1,7 @@
...
 on:
   pull_request:
-    paths:
-      - ".github/workflows/ClimaCoreMakie.yml"
-      - "lib/ClimaCoreMakie/**"
-  push:
-    branches:
-      - staging
-      - trying
+
```
charleskawczynski, line 4: I think that deleting `push` here means that it's only built on the first push, is that what we want?

Sbozzolo: The block meant "build only when pushed on these two branches" (which bors used). Now, it means "build on PR". I have to check what happens if I push to an existing PR.

Sbozzolo: It restarts the checks. So, I think that this is what we want. Do you agree?

simonbyrne: > I think that deleting `push` here means that it's only built on the first push

No, it will build on whatever the head of the PR is (I think after it is merged into `main`)

charleskawczynski: Yep, I think so

simonbyrne: Specifically from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request:
> By default, a workflow only runs when a pull_request event's activity type is `opened`, `synchronize`, or `reopened`.

And from https://github.com/orgs/community/discussions/24567, `synchronize` is
> A `pull_request` event it’s only triggered when the pull request’s tracking branch is synchronized with the source branch for the pull request, and that happens when the source branch is updated.

PR #1538, opened 2023-11-08
simplify GitHub Actions

simonbyrne:
Will need to update branch protections to merge this


.github/workflows/UnitTests.yml
```diff
@@ -36,5 +43,6 @@ jobs:
       uses: julia-actions/julia-buildpkg@v1
 
     - name: Run Unit Tests
-      run: julia --project -e 'using Pkg; Pkg.test()'
-      continue-on-error: ${{ matrix.version == 'nightly' }}
+      uses: julia-actions/julia-runtest@v1
+      with:
+        annotate: true
```
Sbozzolo, line 48: Can you please add a carriage return here at the end of this line?

Sbozzolo:
Is the test on Julia 1.10 required to pass?

simonbyrne:
Currently this won't pass: for this to merge we will have to change the required checks.

.github/workflows/UnitTests.yml
simonbyrne:
```suggestion
        annotate: true

```

simonbyrne:
But to actually answer your question: we can just specify which ones we want to pass (so I can make it just the 1.9 ones for now)

Sbozzolo (approved):

simonbyrne:
I've changed the required checks, so this one needs to get merged before any others

PR #1542, opened 2023-11-13
Clean up hybrid plane examples directory. 

akshaysridhar:
Removes 4 experiments and their corresponding buildkite entries. 
 - density_current_2dinvariant_rhotheta.jl
- bubble_2d_invariant_rhotheta.jl
- topo_dc_2dinvariant_rhoe.jl
- bubble_2d_rhotheta.jl


charleskawczynski (approved):
I'm generally in favor of migrating examples to unit tests. My only concern is whether this exercises code that is otherwise not exercised. I ran into several such cases when making changes to metric terms (where unit tests missed broken parts that examples caught).

simonbyrne:
> I'm generally in favor of migrating examples to unit tests. My only concern is whether this exercises code that is otherwise not exercised. I ran into several such cases when making changes to metric terms (where unit tests missed broken parts that examples caught).

Can you link to them? My experience has been that the `examples` directory usually passes since the example don't have any `@test`s in them.

akshaysridhar:
cc @charleskawczynski https://buildkite.com/clima/climacore-ci/builds/2819#018bca05-66f7-45e0-8ba4-0b57763fc014 Here is perhaps one example of a failure in the `perf` env

charleskawczynski:
> Can you link to them?

I don't have any on hand, I'd have to go digging, and not all buildkite builds in the past are still available.

> My experience has been that the examples directory usually passes since the example don't have any @tests in them.

Yes, but any example that simply exercises code will still, e.g., throw method errors.

PR #1548, opened 2023-11-15
Make IntervalTopology(mesh) pick up ClimaComms.device()

Sbozzolo:
Otherwise everything is hardcoded to run on CPU


simonbyrne:
I think the rationale here was that we still don't support the GPU for 1D spectral element spaces. But since we use this for vertical stuff as well, it seems like an easy fix.

PR #1506, opened 2023-10-19
Fix implicit solver compilation issues

dennisYatunin:
This PR fixes a number of compilation issues in the implicit solver infrastructure, which have been causing inference failures and allocations in the ClimaAtmos-based unit tests (and in ClimaAtmos.jl itself).

# Content
- Replace all of the recursively unrolled functions used in the `MatrixFields` module with generatively unrolled functions, which gets rid of the `stack overflow in type inference` compiler errors. Some of these functions are taken from Unrolled.jl, while others build upon the functions defined there. Move all unrolled functions into `Utilities.UnrolledFunctions` and add some documentation. Update the `Manifest.toml` files to add Unrolled.jl.
- Rewrite `FieldNameDict` so that there are no constant `Union` types being passed around, which gets rid of the allocations in the prognostic EDMF unit test. This involves merging `FieldMatrix` and `FieldMatrixBroadcasted` into a single type, and also merging `FieldVectorView` and `FieldVectorViewBroadcasted`. Add the `is_lazy` function to distinguish between materialized and un-materialized `FieldNameDict`s. Update all relevant documentation.
- Modify the definitions of `norm_sqr(::FieldVector)` and `array_type(::FieldVector)` to make them type-stable for nested `FieldVector`s, which fixes the remaining type instabilities in a few unit tests. Nested `FieldVector`s can be generated while debugging a `StationaryIterativeSolve` that is part of a larger `FieldMatrixSolverAlgorithm`.
- Change some of the `FieldNameSet` method definitions to avoid specializations of the form `where {T}`, as was suggested by @charleskawczynski. This does not appear to affect compilation, but it does make the code a bit easier to read.

# Invalidations

Adding Unrolled.jl changes 4 rows in the invalidations table and brings the total number of invalidations up from 5529 to 5554. On the main branch, these 4 rows are

```
┌─────────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                       │ Function Name │ Invalidations │ Invalidations % │
│                                                 │               │               │     (xᵢ/∑x)     │
├─────────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ ColorTypes/1dGw6/src/operations.jl:24           │    isequal    │      50       │        1        │
│ InlineStrings/rlLZO/src/InlineStrings.jl:606    │     string    │      41       │        0        │
│ StaticArrays/yXGNL/src/abstractarray.jl:145     │    similar    │      35       │        0        │
[...]

PR #1551, opened 2023-11-16
Add matrix-free iterative solver and fix bugs in FieldNameSet

dennisYatunin:
This PR is peeled off from #1506, and it contains all of the changes we need to set up the implicit solver for diffusion (and diagnostic EDMF) in ClimaAtmos.

This does not include any changes related to improving compilation time, but it looks like some of those changes may not be needed anymore because upgrading Julia from 1.9.3 to 1.9.4 has reduced the compilation time of `test/MatrixFields/field_matrix_solvers.jl` by around an order of magnitude.

## Content
- Rename the `SchurComplementSolve` algorithm to `BlockArrowheadSolve`, since other algorithms can also make use of the Schur complement.
- Add the `SchurComplementReductionSolve` algorithm, which generalizes `BlockArrowheadSolve` to arbitrary block matrices by removing the restriction that the `A₁₁` block of `A` must be a diagonal matrix. In order to avoid computing the dense matrix `inv(A₁₁)`, this algorithm replaces the Schur complement matrix used in `BlockArrowheadSolve` with a `LazySchurComplement` object. The only operation this object supports is computing matrix-vector products by calling `lazy_mul(schur_complement, vector)`, so its corresponding linear system must be solved using a matrix-free algorithm.
- Add the `StationaryIterativeSolve` algorithm, along with a collection of `PreconditionerAlgorithm`s. These are implemented using "lazy" (i.e., matrix-free) operations, which means that they minimize the number of materialized broadcast expressions by using `FieldMatrixBroadcasted` and `FieldVectorViewBroadcasted` instead of `FieldMatrix` and `FieldVectorView` wherever possible. Currently, it looks like the best (in terms of both convergence rate and overall speed) preconditioner for ClimaAtmos will be a `BlockArrowheadSchurComplementPreconditioner` that wraps either a `MainDiagonalPreconditioner` (a Jacobi preconditioner) or a `MainDiagonalPreconditioner` wrapped in a `WeightedPreconditioner` (a damped Jacobi preconditioner). The `BlockDiagonalPreconditioner` and `BlockArrowheadPreconditioner` have also been added to facilitate further testing, as well as the catch-all `CustomPreconditioner`.
- Add the `ApproximateBlockArrowheadIterativeSolve` function, which is a shorthand for constructing a `SchurComplementReductionSolve` that wraps a `StationaryIterativeSolve` with a `BlockArrowheadSchurComplementPreconditioner`. This is sometimes called a "SIMPLE" solver in the context of computational fluid dynamics.
- Remove the `ApproximateFactorizationSolve` algorithm, since testing in ClimaAtmos has shown that it is not helpful.
- Add unit tests for `StationaryIterativeSolve` and `ApproximateBlockArrowheadIterativeSolve` with different preconditioners. Modify the ClimaAtmos-based unit tests to use `ApproximateBlockArrowheadIterativeSolve` instead of `ApproximateFactorizationSolve`.
- Add KrylovKit.jl to the Project.toml file so that we have access to its matrix-free `eigsolve` function. This is used to approximate the convergence rate of `StationaryIterativeSolve`, which will be helpful for performance tuning in ClimaAtmos. Update all the Manifest.toml files by calling `dev .`.
- Add support for `zero.(matrix_or_vector)` to `FieldNameDict`.
- Fix how `FieldNameTree`s get propagated through the solver to ensure that `eigsolve` can be evaluated for non-trivial preconditioners.
- Fix the `intersect`, `union`, `setdiff`, and `set_complement` methods for `FieldNameSet` to ensure that overlapping matrix keys are handled correctly. This allows the previously broken unit test for implicit prognostic EDMF to run, though it still fails the allocations test. Getting rid of the allocations will require a few other changes from #1506, which I'll add in a separate PR.
- Fix and expand the unit tests for `FieldNameSet`, and add some unit tests for `FieldNameDict`.
- Fix some minor bugs in the `==` and `summand_names_for_matrix_product` that were caught by the new tests.
- Improve the `show` methods for `FieldNameSet` and `FieldNameDict`.

---

The increase in invalidations is caused by adding KrylovKit.jl to the Project.toml file. This package has ChainRulesCore.jl as a dependency, and it adds the following line to the invalidations table:
```
┌────────────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                          │ Function Name │ Invalidations │ Invalidations % │
│                                                    │               │               │     (xᵢ/∑x)     │
├────────────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ ChainRulesCore/7MWx2/src/tangent_arithmetic.jl:102 │       -       │       2       │        0        │
└────────────────────────────────────────────────────┴───────────────┴───────────────┴─────────────────┘
```


src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -0,0 +1,276 @@
+has_lazy_inverse(P_alg) = isnothing(P_alg) || isnothing(solver_algorithm(P_alg))
+
+lazy_preconditioner(::Nothing, A) = one(A)
+
+function preconditioner_cache(P_alg, A, b)
```
simonbyrne, line 5: Can we document some of these functions? it's not clear what they're doing.

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -0,0 +1,276 @@
...
+    has_lazy_inverse(P_alg) && return lazy_mul(lazy_inv(P), lazy_b)
+    @. P_cache.b = lazy_b
+    alg = solver_algorithm(P_alg)
+    field_matrix_solve!(alg, P_cache.cache, P_cache.x, P, P_cache.b)
+    return P_cache.x
+end
+
+################################################################################
+
+abstract type PreconditionerAlgorithm end
```
simonbyrne, line 48: what is the interface a `PreconditionerAlgorithm` needs to define?

simonbyrne (changes requested):

src/MatrixFields/field_matrix_iterative_solver.jl
dennisYatunin: Each preconditioner algorithm needs to define `lazy_preconditioner`, which generates an un-materialized form of the preconditioner matrix, and `solver_algorithm`, which specifies how to invert the preconditioner matrix.

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -0,0 +1,465 @@
...
+    else
+        @. x = zero(x)
+    end
+    for iter in 1:(alg.n_iters)
+        lazy_Δb = lazy_sub(b, lazy_mul(A, x))
+        lazy_Δx = apply_preconditioner(alg.P_alg, cache.P_cache, P, lazy_Δb)
+        if alg.print_norm
+            norm_Δx = norm(concrete_field_vector(Base.materialize(lazy_Δx)))
+            @info "||x[$(iter - 1)] - x'||₂ ≈ $norm_Δx"
+        end
```
simonbyrne, line 433: Technically you're logging, not printing. A better way to control this is to do something like
```suggestion
        @debug "||x[$(iter - 1)] - x'||₂ ≈ $(norm(concrete_field_vector(Base.materialize(lazy_Δx))))"
```
Then you can switch it on by setting the environment variable `JULIA_DEBUG=MatrixFields`.

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -0,0 +1,465 @@
...
+    end
+end
+
+"""
+    ApproximateBlockArrowheadIterativeSolve(names₁...; [P_alg₁], [alg₁], [alg₂], [kwargs...])
+
+Shorthand for constructing a `BlockLUDecompositionSolve` that wraps a
+`StationaryIterativeSolve` with a `BlockArrowheadSchurComplementPreconditioner`.
+
+TODO: Add some motivation and references.
```
simonbyrne, line 453: ???

dennisYatunin: Yeah, I'm currently trying to figure out how to document this clearly. This is also called a "Schur complement solve" in the literature, and it looks like this is a "standard" way to approximately solve advection-diffusion problems in CFD. [Here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7365595/) is a paper that somewhat describes this, though I'm looking for a better one to cite in this docstring.

src/MatrixFields/field_matrix_iterative_solver.jl
dennisYatunin: With this strategy, is there any way to control the norm debugging and the radius debugging individually? Computing the error norm is very cheap, so we'll probably end up printing that somewhat frequently. The spectral radius takes a very long time to compute, though, so we'll only need to check it occasionally.

simonbyrne: JULIA_DEBUG accepts filenames, modules or groups:
```
➜  ~ JULIA_DEBUG=mygroup julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.4 (2023-11-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @debug "hello"

julia> @debug "hello" _group=:mygroup
┌ Debug: hello
└ @ Main REPL[4]:1
```

So you could use a custom group for the different things you want to log?

dennisYatunin:
@simonbyrne I've improved the docs further and switched to using `@debug`. The latter caused type inference to crash for the last three ClimaAtmos-based unit tests (`Internal error: stack overflow in type inference of ...`), but I think this will be fixed when I add some additional changes from #1506 (those changes will replace some recursively unrolled functions with generatively unrolled functions, avoiding stack overflows during type inference). Could you please take another look?

simonbyrne (approved):

PR #1555, opened 2023-11-21
Bump ClimaCore to 0.11.0

Sbozzolo:
New minor version because it breaks some of the subpackages


sriharshakandala:
Can we also update dependencies. Otherwise looks good to me!

sriharshakandala (approved):

lib/ClimaCoreMakie/Project.toml
```diff
@@ -1,10 +1,10 @@
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
 name = "ClimaCoreMakie"
 uuid = "908f55d8-4145-4867-9c14-5dad1a479e4d"
-version = "0.4.0"
+version = "0.4.1"
 
 [compat]
-ClimaCore = "0.7, 0.8, 0.9, 0.10"
+ClimaCore = "0.7, 0.8, 0.9, 0.10, 0.11"
```
simonbyrne, line 7: I actually don't think these subpackages are all backward compatible.

Sbozzolo: I'll just keep `0.11` then.

sriharshakandala:
We may not need `UnPack` anymore!

Sbozzolo:
> We may not need `UnPack` anymore!

Do you want to add a commit to this PR that removes it?

Sbozzolo:
> > We may not need `UnPack` anymore!
> 
> Do you want to add a commit to this PR that removes it?

Removed!

PR #1560, opened 2023-11-27
Remove Requires, use Julia Extensions

Sbozzolo:


simonbyrne:
Weakdeps support was only added in Julia 1.9. There are some [hacks for backward compatibility](https://pkgdocs.julialang.org/v1/creating-packages/#Backwards-compatibility), but it might be easier to just bump the Julia version requirement.

simonbyrne (approved):

PR #1561, opened 2023-11-28
Add DeviceSide

Sbozzolo:
This PR implements a new `Device` that is used to identify structs that live on the GPU. 

`DeviceSideDevice` (and the associated `DeviceSideContext`) are singleton that are passed through `Adapt.adapt_structure` to those structs that are on the GPU.

With this change, I also added support for `mapreduce_cuda` for point spaces.

### DIFF
--- a/src/devices.jl
+++ b/src/devices.jl
@@ -1,0 +1,29 @@
+import ClimaComms
+
+# Sometimes, it is convenient to know whether something is running on a device or on its
+# host. We define new AbstractDevices and AbstractCommsContext to identify the case of a
+# function running on the device with data on the device.
+
+"""
+    DeviceSideDevice()
+
+This device represents data defined on the device side of an accelerator.
+
+The most common example is data defined on a GPU. DeviceSideDevice() is used for
+operations within the accelerator.
+"""
+struct DeviceSideDevice <: ClimaComms.AbstractDevice end
+
+
+"""
+    DeviceSideContext()
+
+Context associated to data defined on the device side of an accelerator.
+
+The most common example is data defined on a GPU. DeviceSideContext() is used for
+operations within the accelerator.
+"""
+struct DeviceSideContext <: ClimaComms.AbstractCommsContext end
+
+ClimaComms.context(::DeviceSideDevice) = DeviceSideContext()
+ClimaComms.device(::DeviceSideContext) = DeviceSideDevice()
--- a/test/Spaces/spaces.jl
+++ b/test/Spaces/spaces.jl
@@ -1,10 +1,23 @@
 using Test
 using ClimaComms
 using StaticArrays, IntervalSets, LinearAlgebra
+using Adapt
 
 import ClimaCore:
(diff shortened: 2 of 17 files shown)


Sbozzolo:
@simonbyrne do you have any initial thoughts on this?

src/devices.jl
```diff
@@ -0,0 +1,37 @@
...
+    DeviceSideContext()
+
+Context associated to data defined on the device side of an accelerator.
+
+The most common example is data defined on a GPU. DeviceSideContext() is used for
+operations within the accelerator.
+"""
+struct DeviceSideContext <: ClimaComms.AbstractCommsContext end
+
+# Some functions use context.device
```
simonbyrne, line 28: can we just change those?

Sbozzolo: > can we just change those?

What do you mean?

Sbozzolo: https://github.com/CliMA/ClimaCore.jl/blob/a04027449080a29dcaa4e390a56231e6dd3fbded/src/Topologies/topology2d.jl#L127

This is where we define `device(space)` essentially

Sbozzolo: I added the `getproperty` mostly to mimic the interface that `context` has, but if everything is wired correctly, we shouldn't strictly need it. So, I am happy to remove it if you don't like it.

Sbozzolo: Okay, I missed that ClimaComms already defines 
```
device(ctx::SingletonCommsContext) = ctx.device
```

src/devices.jl
```diff
@@ -0,0 +1,29 @@
...
+
+"""
+    DeviceSideDevice()
+
+This device represents data defined on the device side of an accelerator.
+
+The most common example is data defined on a GPU. DeviceSideDevice() is used for
+operations within the accelerator.
+"""
+struct DeviceSideDevice <: ClimaComms.AbstractDevice end
```
simonbyrne, line 15: I've thought a bit more about this: why don't we put this into ClimaComms?

Sbozzolo: This would be a device that behaves differently compared to the other ones in ClimaComms. It doesn't have any associated features, and it cannot be "selected". It is just a tag at the end of the day. 

But I don't have a very strong opinion about it.

simonbyrne (approved):

Sbozzolo:
@simonbyrne I added a commit to implement `mapreduce_cuda` for `DataF` (which is needed because the reductions for `PointSpace`s now use methods for GPUs when they are defined on the GPU). Please, have a look at it when you have a chance.

test/Spaces/spaces.jl
```diff
@@ -58,7 +75,7 @@ import ClimaCore.DataLayouts: IJFH, VF
           Spaces.column(coord_data, 1, 1)[]
 end
 
-@testset "extruded (2d 1×3) finite difference space" begin
+on_gpu || @testset "extruded (2d 1×3) finite difference space" begin
```
charleskawczynski, line 78: What fails when this is on the GPU?

PR #1567, opened 2023-12-01
add convenience constructors for grids

simonbyrne:
cc @Sbozzolo 

### DIFF
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -35,6 +35,10 @@ struct IntervalDomain{CT, B} <: AbstractDomain where {
     boundary_names::B
 end
 
+const XIntervalDomain{FT, B} = IntervalDomain{Geometry.XPoint{FT}, B}
+const YIntervalDomain{FT, B} = IntervalDomain{Geometry.YPoint{FT}, B}
+const ZIntervalDomain{FT, B} = IntervalDomain{Geometry.ZPoint{FT}, B}
+
 isperiodic(domain::IntervalDomain) = isnothing(domain.boundary_names)
 boundary_names(domain::IntervalDomain) =
     isperiodic(domain) ? () : unique(domain.boundary_names)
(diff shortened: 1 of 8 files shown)


simonbyrne:
The first question is: what should we name the aliases? Currently I've gone with the simplest possible names I could come up with:
- `ColumnGrid` for a single column in the z direction 
  - requires renaming an existing object, but that wasn't exported
- `PlaneGrid` for a XZ plane
- `BoxGrid` for a XYZ box
- `ExtrudedCubedSphereGrid` for a cubed sphere
  - any thoughts on a simpler name?

One question is whether we should be a bit more specific on the names:
 - should we include the coordinate in the names (say `XZPlaneGrid` instead of `PlaneGrid`?)
 - should we include the discretization type (`FDColumnGrid`, `FDSEPlaneGrid`?)

simonbyrne:
> * `PlaneGrid` for a XZ plane

@OsKnoth suggested calling a `SliceGrid`

simonbyrne:
The other interface questions:
- how to handle stretching: currently I have a `z_stretch` argument (not `dz_X` args like in ClimaAtmos). Default is `Uniform`
- topography is not yet handled, but will probably be a single argument (`topography=Flat()` default)

src/Grids/spectralelement.jl
```diff
@@ -140,7 +141,8 @@ Note: This is accurate only for cubed-spheres of the [`Meshes.EquiangularCubedSp
 @memoize WeakValueDict function SpectralElementGrid2D(
     topology,
     quadrature_style;
-    enable_bubble = false,
+    bubble = enable_bubble,
```
charleskawczynski, line 144: Will this result in a breaking release, then?

simonbyrne: I kept the old one for backward compatibility.

simonbyrne:
I am a bit reluctant to simply call it a `CubedSphereGrid`, as a [sphere is defined as a 2D manifold](https://en.wikipedia.org/wiki/Sphere), not a 3D object. The domain we're operating in is usually called a [spherical shell](https://en.wikipedia.org/wiki/Spherical_shell).

Some options:
- `CubedSphere3DGrid`
- `CubedSphereShellGrid`

OsKnoth:
Cubed Sphere Grid is used in the community, it is a discrete manifold, which approximates the sphere, also extruded grid is used normally, in all cases we start with two-dimensional manifolds and than go to the third dimension, furthermore this construction is connected to the function spaces (spectral versus piecewise constant in the third dimension, or piecewise linear for w), it makes no sense to look for new namings outside the community.

simonbyrne:
@Sbozzolo any more thoughts on this?

Sbozzolo:
> @Sbozzolo any more thoughts on this?

Can you please provide a short description on how I would go about identifying the `Grid` that is being used (e.g., for dispatch in `ClimaAtmos`)?

simonbyrne:
```julia
f(grid::Grids.PlaneGrid) = ...
f(grid::Grids.BoxGrid) = ...
f(grid::Grids.ExtrudedCubedSphereGrid) = ...
```

Sbozzolo:
> ```julia
> f(grid::Grids.PlaneGrid) = ...
> f(grid::Grids.BoxGrid) = ...
> f(grid::Grids.ExtrudedCubedSphereGrid) = ...
> ```

Are we going with `Plane` or `Slice`? 

I still think that `ExtrudedCubedSphereGrid`, while descriptive, is unnecessarily complex for an important user-facing type. (But I am fine keeping it).

As a next step, we should implement accessors: it should be trivial to get the radius from a sphere, the heigth of a column, and so on.

simonbyrne:
> I still think that `ExtrudedCubedSphereGrid`, while descriptive, is unnecessarily complex for an important user-facing type. (But I am fine keeping it).

How about `CubedSphereShellGrid`?

simonbyrne:
Okay, I've renamed them.

src/Domains/Domains.jl
```diff
@@ -66,6 +70,50 @@ end
 IntervalDomain(coords::ClosedInterval; kwargs...) =
     IntervalDomain(coords.left, coords.right; kwargs...)
 
+
+function XIntervalDomain(;
```
charleskawczynski, line 74: @Sbozzolo what do you think about adding doc strings to all of these convenience constructors, so that users know what the kwargs are? I think just copying the code itself into the doc string should be pretty self explanatory, and it'll be easy to keep up-to-date.

szy21:
@charleskawczynski What is the status of this PR?

charleskawczynski:
I need to rebase this, and make a few changes

charleskawczynski:
Superseded by #1848.

PR #1582, opened 2023-12-22
Add missing half import

Sbozzolo:
It is used in `interpolate_array.jl`


simonbyrne (approved):

simonbyrne:
Can we add a test that triggers this?

Sbozzolo:
> Can we add a test that triggers this?

Will do

PR #1585, opened 2024-01-02
Update to Julia 1.10

charleskawczynski:
This PR upgrades ClimaCore to Julia 1.10


Sbozzolo:
`1.10` is not correctly interpreted as version 1.10 in yaml files unless we escape it with backticks (leading to failures in some of the workflows)

.github/workflows/CodeCov.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Julia
       uses: julia-actions/setup-julia@latest
       with:
-        version: 1.9
+        version: 1.10
```
simonbyrne, line 18:
```suggestion
        version: '1.10'
```

.github/workflows/JuliaFormatter.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - uses: julia-actions/setup-julia@latest
       if: steps.filter.outputs.julia_file_change == 'true'
       with:
-        version: 1.9
+        version: 1.10
```
simonbyrne, line 30:
```suggestion
        version: '1.10'
```

.github/workflows/docs.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
-          version: 1.9
+          version: 1.10
```
simonbyrne, line 13:
```suggestion
          version: '1.10'
```

examples/hybrid/sphere/README.md
```diff
@@ -13,7 +13,7 @@ Here is a sbatch script template for setting up simulations on caltech central h
 #SBATCH --output=$YOUR_SIMULATION_LOG_DIR/simulation.log
 
 module purge
-module load julia/1.9.3 openmpi/4.1.1
+module load julia/1.10.0 openmpi/4.1.1
```
Sbozzolo, line 16: I think that this snippet might no longer working since we use MPITrampoline now (required for compatibility with new versions of HDF5).

Maybe it is a good time to update it or remove it if we don't want to maintain it.

Sbozzolo (approved):

examples/hybrid/sphere/README.md
charleskawczynski: Are you suggesting to change this to `module load julia/1.10.0`?

Sbozzolo: The snipped diverged from what we do in the buildkite pipeline with respect to MPI. I am worried it wouldn't work.

Sbozzolo: It is not in the scope of this PR, but it might be a good opportunity to update it if you want to.

charleskawczynski: Ok, I don't know off hand, I'll tackle that separately

Sbozzolo: Sounds good, feel free to merge

Sbozzolo:
I updated the branch protection rules to not require tests for julia 1.9 but for 1.10

PR #1586, opened 2024-01-02
Remove obsolete function

Sbozzolo:
`ClimaCore` now requires Julia 1.9, so the function `filter18` is no longer relevant.


Sbozzolo:
Didn't realize #1585 was already open

simonbyrne (approved):

PR #1587, opened 2024-01-03
Remove truncated show for Fields

### LINKED ISSUE #1237: Overloading `Base.show(::IO, ::Type{<:Fields.Field})` invalidates many methods
See this discussion: https://github.com/JuliaCollections/DataStructures.jl/pull/854. This has been known for some time, but I wanted to open an issue for tracking.

charleskawczynski:
This PR removes truncated `show` for Fields, now that Julia 1.10 has depth-limited stack trace printing (https://github.com/JuliaLang/julia/pull/49795).

This should (finally) fix the majority of method invalidations for `show(io::IO, ::Type)`.

@simonbyrne, should we leave this but put it inside a `if VERSION ≤ v"1.10"` if-block to preserve backward functionality?

Closes #1237.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -528,63 +528,65 @@ function set!(f::Function, field::Field, args = ())
     return nothing
 end
 
-#=
-This function can be used to truncate the printing
-of ClimaCore `Field` types, which can get rather
-long.
+if VERSION < v"1.10"
+    #=
+    This function can be used to truncate the printing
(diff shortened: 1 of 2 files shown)


Sbozzolo (approved):
Thank you!

I would keep it in a if block for the time being, otherwise this will "force" julia 1.10 on downstream packages, and not all the packages are ready to move to 1.10. Maybe open an issue to remind ourselves to remove it?

Issue #1589, opened 2024-01-10
Memoization of Grids via WeakValueDIct can processes to diverge


simonbyrne:
## Describe the bug

@juliasloan25 hit a problem with an MPI job with 2 processes:

1. A grid object is created on two processes (when reading a file)
2. The grid object goes out of scope
3. the garbage collector is only invoked on rank 0
  - rank 0: the grid is removed from the memoization `WeakValueDict`
  - rank 1: the grid remains in the memoization `WeakValueDict`
5. The grid constructor is called again (reading a new file)
  - rank 0: calls the underlying constructor, which calls DSS
  - rank 1: returns the grid from the memoization `WeakValueDict`

Rank 0 then hangs, waiting for a DSS operation.

Possible solutions
- Some sort of parallel GC: basically keep distributed objects in a "purgatory" state until they can be collectively destroyed?
- Switch to using an `IdDict`, and require users to manually "destroy" grids (i.e. remove them from the cache)

simonbyrne:
I decided to switch to using a plain `Dict`: parallel GC would be quite an engineering undertaking for little potential benefit.

simonbyrne:
Fixed by #1590

PR #1590, opened 2024-01-11
don't use WeakValueDicts to memoize

simonbyrne:
Possible fix to #1589 


src/Grids/Grids.jl
```diff
@@ -30,6 +29,21 @@ Grids should define the following
...
+Remove the grid object from the cache of created grids. In most cases, this
+should not need to be called, unless you are e.g. doing a sweep over grid
+paramaters.
+
+Only call this if you are completely finished with this grid configuration, and
+no other instances of it exist.
+"""
+function remove_from_cache!(grid::AbstractGrid)
+    return nothing
+end
```
Sbozzolo, line 45: Is this a draft?

src/Grids/finitedifference.jl
```diff
@@ -45,9 +45,24 @@ mutable struct FiniteDifferenceGrid{
 end
 
 
-@memoize WeakValueDict function FiniteDifferenceGrid(
-    topology::Topologies.IntervalTopology,
-)
+# we can use a WeakValueDict to memoize here, as FiniteDifferenceGrids cannot be distributed, so won't suffer from
+#   https://github.com/CliMA/ClimaCore.jl/issues/1589
+const FD_GRID_CACHE = WeakValueDict()
```
Sbozzolo, line 50: If this is the only place where we use `WeakValueDict`, would it make sense to use `Dict` instead? 
The two reasons I see to use a `Dict` are:
- We treat this grid in the same way as the other ones
- We remove a direct dependency from a package that hasn't been updated in 5 years, and might increase our surface area for problems in the future

simonbyrne: That's a good point.

src/Utilities/Utilities.jl
```diff
@@ -3,4 +3,18 @@ module Utilities
...
+
+Remove `object` from the cache of created objects (only topologies and
+grids are cached). In most cases, this should not need to be called, unless you
+are e.g. doing a sweep over grid paramaters.
+
+Only call this if you are completely finished with this object configuration, and
+no other instances of it exist.
+"""
+function remove_from_cache!(object) # fallback
+end
```
Sbozzolo, line 17: I don't want to overengineer this, but maybe it would be useful to also have a `clean_caches!` function that cleans all the caches. I am thinking that this might be useful when doing interactive development to ensure that we are not using stale objects. 

Can you please also describe this caching mechanism somewhere in the main docs?

juliasloan25:
tested from coupler here and looks to work: https://buildkite.com/clima/climacoupler-ci/builds/2374

src/Utilities/cache.jl
```diff
@@ -0,0 +1,63 @@
...
+
+ClimaCore maintains an internal cache of topology and grid objects: this ensures
+that if the constructor with the same arguments is invoked again (e.g. by
+reading from a file), the cached object will be returned (also known as
+_memoization_). This has two main advantages:
+
+1. topology and metric information can be reused, reducing memory usage.
+
+2. it is easy to check if two fields live on the same grid: we can just check if
+   the umderlying grid objects are the same (`===`), rather than checking all
```
Sbozzolo, line 12:
```suggestion
the underlying grid objects are the same (`===`), rather than checking all
```

Sbozzolo (approved):
One typo and everything else looks good.

Maybe we can tag a release for the coupler to use.

Issue #1212, opened 2023-05-02
DivergenceC2F Boundary Condition Bug


nefrathenrici:
## Describe the bug
When constructing a `DivergenceC2F` operator, if an entire field is given for the boundary condition instead of a single column, the operator will just use the first value in the field instead of failing outright.

This was found in ClimaAtmos.jl, where the following code did not return an error, but just used the first column regardless of the set `colidx`
```
        ᶜdivᵥ_ρe_tot = Operators.DivergenceF2C(
            top = Operators.SetValue(CT3(FT(0))),
            bottom = Operators.SetValue(sfc_conditions.ρ_flux_h_tot),
        )
        @. Yₜ.c.ρe_tot[colidx] -= ᶜdivᵥ_ρe_tot(
            -(
                ᶠρK_E[colidx] *
                ᶠgradᵥ((Y.c.ρe_tot[colidx] + ᶜp[colidx]) / Y.c.ρ[colidx])
            ),
        )
```
The actual operator should have been the following:
```
        ᶜdivᵥ_ρe_tot = Operators.DivergenceF2C(
            top = Operators.SetValue(CT3(FT(0))),
            bottom = Operators.SetValue(sfc_conditions.ρ_flux_h_tot[colidx]),
        )
```

@trontrytel I am happy to talk this over more, I wasn't able to reproduce this bug without running Atmos

charleskawczynski:
Agreed, we should figure out a way to prohibit surface-column field broadcasting. There's no circumstance where this is needed.

trontrytel:
👍 I wish we could throw an error when someone forgets `[colidx]`

PR #1593, opened 2024-01-17
Add support for SetDivergence to DivergenceF2C

dennisYatunin:
This PR fixes a bug in ClimaAtmos related to the advection of precipitation.

Advection is implemented using the `DivergenceF2C` operator, and, at the bottom cell center, we need to enforce a zero-divergence boundary condition for precipitation. We initially implemented this using `SetDivergence`, but this is not supported for `DivergenceF2C` and was ignored due to the following fallback method in `finitedifference.jl`:
```
boundary_width(::DivergenceF2C, ::AbstractBoundaryCondition) = 0
```

This PR adds support for `SetDivergence` to `DivergenceF2C`, as well as its operator matrix. It also includes a new unit test for these additions.

### DIFF
--- a/src/MatrixFields/operator_matrices.jl
+++ b/src/MatrixFields/operator_matrices.jl
@@ -855,6 +855,16 @@ Base.@propagate_inbounds function op_matrix_last_row(
     J⁻ = Geometry.LocalGeometry(space, idx - half, hidx).J
     return BidiagonalMatrixRow(-C3(J⁻)', C3(FT(0))') * invJ
 end
+op_matrix_first_row(
+    ::Operators.DivergenceF2C,
+    ::Operators.SetDivergence,
+    ::Type{FT},
+) where {FT} = UpperDiagonalMatrixRow(C3(FT(0))')
+op_matrix_last_row(
+    ::Operators.DivergenceF2C,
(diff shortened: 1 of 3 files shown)


dennisYatunin:
@simonbyrne Do you think it would be a good idea to get rid of the `AbstractBoundaryCondition` fallback methods for `boundary_width` in the future? They could potentially lead to more silent bugs like this if we keep them around.

simonbyrne (approved):

charleskawczynski (approved):

simonbyrne:
> @simonbyrne Do you think it would be a good idea to get rid of the `AbstractBoundaryCondition` fallback methods for `boundary_width` in the future? They could potentially lead to more silent bugs like this if we keep them around.

What's the bug?

simonbyrne:
Is it https://github.com/CliMA/ClimaCore.jl/issues/1212?

szy21:
No, it's [here](https://github.com/CliMA/ClimaAtmos.jl/blob/be0d6e72b484d0503d2bbfb26df4223fce0e075b/src/prognostic_equations/implicit/implicit_tendency.jl#L146):  This line doesn't give an error when `SetDivergence` is not defined. It just ignores the boundary condition.

PR #1600, opened 2024-01-24
Tag v0.11.9

dennisYatunin:
New patch release that contains the bugfix from #1598.


Sbozzolo:
We can always make a new minor release, but it would be nice to add the changes for hypsography (ie, in the deep atmo PR) too.

dennisYatunin:
This is just to add a performance bugfix for a precipitation PR in ClimaAtmos, which we'd like to merge in soon.

dennisYatunin:
Looks like we also need to bump the conservation error limit for the deformation flow test in order for CI to pass. Most likely it's running on a new machine and the roundoff error is slightly higher.

charleskawczynski (approved):

PR #1604, opened 2024-02-01
Add tofloat(Point)

Sbozzolo:
It is easy to go from a number to a Geometry.Point, but it is not easy to go in the other direction (unless the type of point of point is known). This PR adds new methods to Float32, Float64, and BigFloat to X/Y/Z/Lat/Long point to convert them into numbers.

This feature would be very useful in converting a Field of ZPoints into a scalar field, which is what we need for the interpolation routines

### DIFF
--- a/src/Geometry/coordinates.jl
+++ b/src/Geometry/coordinates.jl
@@ -66,10 +66,16 @@ Define a subtype `name` of `AbstractPoint` with appropriate conversion functions
 macro pointtype(name, fields...)
     if length(fields) == 1
         supertype = :(Abstract1DPoint{FT})
+        coord = fields[1]
+        tofloat_expr = quote
+            tofloat(p::$name) = p.$coord
+        end
     elseif length(fields) == 2
         supertype = :(Abstract2DPoint{FT})
+        tofloat_expr = :()
     elseif length(fields) == 3
         supertype = :(Abstract3DPoint{FT})
+        tofloat_expr = :()
     end
     esc(
         quote
@@ -90,6 +96,7 @@ macro pointtype(name, fields...)
                 ::Type{$name{FT1}},
                 ::Type{$name{FT2}},
             ) where {FT1, FT2} = $name{promote_type(FT1, FT2)}
+            $tofloat_expr
         end,
     )
 end
--- a/test/Geometry/geometry.jl
+++ b/test/Geometry/geometry.jl
@@ -26,6 +26,7 @@ using LinearAlgebra, StaticArrays
             @test pt < pt * FT(2)
             @test pt <= pt
             @test pt <= pt * FT(2)
+            @test Geometry.tofloat(pt) == one(FT)
         end
     end
 


dennisYatunin (commented):
Overloading the constructors for primitive types is a bit risky. Is there a downside to defining a function that does this, like `Geometry.first_coordinate(::Point)`?

Sbozzolo:
> Overloading the constructors for primitive types is a bit risky. Is there a downside to defining a function that does this, like `Geometry.first_coordinate(::Point)`?

This function also converts to the specified float type. This is not strictly needed, but it is nice to have. `Geometry.coordinate` already exists but doesn't return a float, it returns a `Point`, so maybe I could define `tofloat`

dennisYatunin:
> > Overloading the constructors for primitive types is a bit risky. Is there a downside to defining a function that does this, like `Geometry.first_coordinate(::Point)`?
> 
> This function also converts to the specified float type. This is not strictly needed, but it is nice to have. `Geometry.coordinate` already exists but doesn't return a float, it returns a `Point`, so maybe I could define `tofloat`

Yes, a more "standard" way to implement this would be something like `Geometry.tofloat(::Type{T}, ::Point)`. Adding constructors to the float types implies that our points are "equivalent" to floats at a low level, which is not quite the case.

Sbozzolo:
> > > Overloading the constructors for primitive types is a bit risky. Is there a downside to defining a function that does this, like `Geometry.first_coordinate(::Point)`?
> > 
> > 
> > This function also converts to the specified float type. This is not strictly needed, but it is nice to have. `Geometry.coordinate` already exists but doesn't return a float, it returns a `Point`, so maybe I could define `tofloat`
> 
> Yes, a more "standard" way to implement this would be something like `Geometry.tofloat(::Type{T}, ::Point)`, with a default value for `T` based on the `Point` type. Adding constructors to the float types implies that our points are "equivalent" to floats at a low level, which is not quite the case.

I added a function `tofloat`. It returns the same type as the Point, which is fine. If we need more, we can add it.

src/Geometry/coordinates.jl
```diff
@@ -64,8 +64,13 @@ Base.show(io::IO, point::Abstract3DPoint) = print(
 Define a subtype `name` of `AbstractPoint` with appropriate conversion functions.
 """
 macro pointtype(name, fields...)
+    tofloat = :()
```
charleskawczynski, line 67: I know it's an extra line, but can we move this into the elseif-blocks below? And, for clarity, rename the expression to something other than the method. E.g., `tofloat_expr = :()`.

I'm fine with this PR after these two changes.

charleskawczynski (approved):
With the two changes I suggested, I'm fine with this PR.

src/Geometry/coordinates.jl
Sbozzolo: Changed!

charleskawczynski (approved):

PR #1610, opened 2024-02-08
Rewrite non-conservative interpolation

Sbozzolo:
This PR largely rewrites the non-conservative remapping module to improve its performance. CPU performance was improved by a factor of 6, GPU performance by orders of magnitude. The new code is completely non-allocating (except for the final result of the interpolation).

I re-wrote the entire thing because the previous version of the code was mostly designed to work for one point at the time, there was work that had to be done over and over. This new version computes all the relevant weights and allocates all the memory only once, and uses them over and over. Unfortunately, there is a lot of redundancy (as there was in the original version of the code). This redundancy is needed to tackle horizontal/extruded spaces, and CPU/GPU while keeping top-notch performance. 

This is not a simple piece of code, but I tried to comment it extensively.

This is a breaking change because it removes the `physical_z` option. I decided to remove that option and always perform interpolation in reference space because it became clear with SLEVE that supporting that would be a large burden. Instead, the transformation can be done in post (I will implement in ClimaAnalysis).

I verified that the new code reproduces the same values as the old one, up to roundoff precision.

GPU flamegraph before:
image

GPU flamegraph after:
image

Now interpolation is < 1 % of total execution for a diagnostic-heavy job. Almost all of that < 1 % is data movement and dealing with views (which are needed for MPI runs).


src/Remapping/distributed_remapping.jl
```diff
@@ -144,24 +215,42 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+    ξs_combined =
+        Meshes.reference_coordinates.(
+            Ref(horizontal_mesh),
+            helems,
+            local_target_hcoords,
+        )
+    num_hdims = length(ξs_combined[1])
+    # ξs is a Vector of SVector{1, Float64} or SVector{2, Float64}
+    # Here we split the two dimensions because we want to compute the two interpolation matrices.
+    ξs_split = Tuple([ξ[i] for ξ in ξs_combined] for i in 1:num_hdims)
```
charleskawczynski, line 233:
```suggestion
    ξs_split = ntuple(i->map(ξ->ξ[i], ξs_combined), num_hdims)
```
This will preserve types

src/Remapping/distributed_remapping.jl
```diff
@@ -170,39 +259,59 @@ function Remapper(target_hcoords, target_zcoords, space)
         end
 
         local_indices = map(local_target_hcoords) do hcoord
-            helem = Meshes.containing_element(horz_mesh, hcoord)
+            helem = Meshes.containing_element(horizontal_mesh, hcoord)
             return global_elem_lidx[horizontal_topology.orderindex[helem]]
         end
     end
 
+    field_values_size = Tuple(Nq for _ in 1:num_hdims)
```
charleskawczynski, line 267:
```suggestion
    field_values_size = ntuple(_-> Nq, num_hdims)
```

src/Remapping/remapping_utils.jl
```diff
@@ -0,0 +1,268 @@
...
+    is_cell_center = space isa Spaces.CenterExtrudedFiniteDifferenceSpace
+
+    ξ3s =
+        map(zcoords) do zcoord
+            velem = Meshes.containing_element(vert_mesh, zcoord)
+            ξ3, = Meshes.reference_coordinates(vert_mesh, velem, zcoord)
+            # For cell centered spaces, shift ξ3 so that we can use it for linear interpolation
+            is_cell_center && (ξ3 = ξ3 < 0 ? ξ3 + 1 : ξ3 - 1)
+            return ξ3
+        end |> collect
```
charleskawczynski, line 39: Why do we want to `collect` here?

src/Remapping/remapping_utils.jl
```diff
@@ -0,0 +1,268 @@
...
+            ξ3, = Meshes.reference_coordinates(vert_mesh, velem, zcoord)
+            if ξ3 < 0
+                v_lo = is_periodic ? mod1(velem - 1, Nz) : max(velem - 1, 1)
+                v_hi = velem
+            else
+                v_lo = velem
+                v_hi = is_periodic ? mod1(velem + 1, Nz) : min(velem + 1, Nz)
+            end
+            return v_lo, v_hi
+        end |> collect
```
charleskawczynski, line 89: Do we need `collect`?

src/Remapping/remapping_utils.jl
```diff
@@ -0,0 +1,268 @@
...
+end
+
+
+"""
+
+TODO
+"""
+function vertical_interpolation_weights(space, zcoords)
+    ξs = vertical_reference_coordinates(space, zcoords)
+    return [((1 - ξ) / 2, (1 + ξ) / 2) for ξ in ξs]
```
charleskawczynski, line 101:
```suggestion
    return map(ξ->((1 - ξ) / 2, (1 + ξ) / 2), ξs)
```

src/Remapping/remapping_utils.jl
```diff
@@ -0,0 +1,268 @@
...
+        remapper._field_values,
+        remapper.local_indices,
+        remapper.interpolation_matrices,
+        remapper.vert_interpolation_weights,
+        remapper.vert_bounding_indices,
+    )
+end
+
+function set_interpolated_values!(
+    out,
```
charleskawczynski, line 122: What is `out` here?

src/Remapping/distributed_remapping.jl
Sbozzolo: I find the original version easier to understand.

src/Remapping/remapping_utils.jl
Sbozzolo: The preallocated area of memory where we save the interpolated values. This is a private field in the remapper, used to assemble the result at the end.

src/Quadratures/Quadratures.jl
```diff
@@ -188,6 +188,29 @@ function interpolation_matrix(
     return SMatrix(M)
 end
 
+function interpolation_matrix(points_to::Vector, points_from)
```
charleskawczynski, line 191: How is this different from the one above?

Sbozzolo: Good question. Initially, I just converted from SVectors as you are suggesting, but found that it would take minutes to allocate all the SVectors needed. So, I just skipped all the SVector allocations because at the end I just wanted (Cu/)Arrays.

I changed the code a little bit in the meantime and maybe I reduced the number of SVector allocations needed. I will check again if we need a separate method that doesn't go through SVectors.

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+end
+
+"""
+    _collect_interpolated_values!(remapper::Remapper)
+
+Perform an MPI call to aggregate the interpolated points from all the MPI processes and save
+the result in the local state of the `remapper`. Only the root process will return the
+interpolated data.
+"""
+function _collect_interpolated_values!(remapper::Remapper)
```
charleskawczynski, line 667: Can we just manually inline this function? It's type-unstable, and I think the code will be clearer with it manually inlined.

src/Quadratures/Quadratures.jl
charleskawczynski: `SVectors` should generally be more performant, and they're GPU-compatible, so I'd like to understand why the ordinary array version is needed

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+
+                for j in 1:Nq, i in 1:Nq
+                    tmp +=
+                        I1[out_index, i] *
+                        I2[out_index, j] *
+                        scratch_field_values[i, j]
+                end
+                out[out_index, vindex] = tmp
+            end
+        end
```
charleskawczynski, line 406: Can we put this inside a separate cpu-implementation function?

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+    return Remapper(space, target_hcoords, target_zcoords)
+end
 
 """
-   interpolate(remapper, field; physical_z = false)
+    _set_interpolated_values!(remapper, field)
 
-Interpolate the given `field` as prescribed by `remapper`.
+Change the local state of `remapper` by performing interpolation of `Fields` on the vertical
+and horizontal points.
```
charleskawczynski, line 336: Can we expand on what this is doing mathematically?

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
     end
 
+    local_horiz_indices = ArrayType(local_horiz_indices)
+
+    field_values_size = ntuple(_ -> Nq, num_hdims)
+    field_values = ArrayType(zeros(FT, field_values_size))
+
     # We represent interpolation onto an horizontal slab as an empty list of zcoords
-    isnothing(target_zcoords) && (target_zcoords = [])
+    if isnothing(target_zcoords) || target_zcoords == []
```
charleskawczynski, line 277:
```suggestion
    if isnothing(target_zcoords) || isempty(target_zcoords)
```

src/Remapping/distributed_remapping.jl
charleskawczynski: Ok, but it results in two steps: first create a vector, then convert it to a tuple. It's more efficient to generate the ntuple from the start.

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+
+    # Compute the interpolation matrices
+    quad = Spaces.quadrature_style(space)
+    quad_points, _ = Quadratures.quadrature_points(FT, quad)
+    Nq = Quadratures.degrees_of_freedom(quad)
+
+    local_horiz_interpolation_weights = Tuple(
+        ArrayType(Quadratures.interpolation_matrix(ξs, quad_points)) for
+        ξs in ξs_split
+    )
```
charleskawczynski, line 247:
```suggestion
    local_horiz_interpolation_weights = map(ξs-> ArrayType(Quadratures.interpolation_matrix(ξs, quad_points)), ξs_split)
    )
```
Another advantage of the type-preserving pattern is that everything follows as a user expects: if a user passes in `ξs_split::Vector`, they'll get vector containers back. If it's a Tuple, they'll get a `Tuple` back (and, depending on the situation, can result in zero allocations).

src/Remapping/distributed_remapping.jl
```diff
@@ -77,43 +112,77 @@ function target_hcoords_pid_bitmask(target_hcoords, topology, pid)
...
+    vert_bounding_indices::VERT_IND
+
+    # Scratch space where we save the process-local interpolated values. We keep overwriting this
+    # to avoid extra allocations.
+    _local_interpolated_values::T8
+
+    # Scratch space where we save the process-local field value. We keep overwriting this to
+    # avoid extra allocations. Ideally, we wouldn't need this and we would use views for
+    # everything.
+    _field_values::T9
```
charleskawczynski, line 177: This is some form of array, correct? If so, what's the dims of this array?

src/Remapping/distributed_remapping.jl
```diff
@@ -77,43 +112,77 @@ function target_hcoords_pid_bitmask(target_hcoords, topology, pid)
...
+
+    # Storage area where the interpolated values are saved. This is meaningful only for the
+    # root process and gets filled by a interpolate call.
+    _interpolated_values::T10
```
charleskawczynski, line 181: This is some form of array, correct? If so, what's the dims of this array? (let's add the dims to the docs)

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+    end
+end
+
+function set_interpolated_values_kernel!(
+    out::AbstractArray,
+    (I1, I2)::NTuple{2},
+    local_horiz_indices,
+    vert_interpolation_weights,
+    vert_bounding_indices,
+    fv,
```
charleskawczynski, line 416: What's a verbose name for `fv`?

src/Remapping/distributed_remapping.jl
```diff
@@ -144,65 +219,482 @@ function Remapper(target_hcoords, target_zcoords, space)
...
+                        fv[i, j, nothing, nothing, h]
+                end
+            elseif hdims == 1
+                for i in 1:Nq
+                    out[out_index] +=
+                        local_horiz_interpolation_weights[1][out_index, i] *
+                        fv[i, nothing, nothing, nothing, h]
+                end
+            else
+                error("Cannot handle $hdims horizontal dimensions")
```
charleskawczynski, line 574: Can we check for and throw this error earlier? (so that the GPU path is also protected from this edge case)

src/Quadratures/Quadratures.jl
Sbozzolo: It is needed because allocating lots/large `SVector`s takes time. It takes 26 seconds to prepare a Remapper with 101x101x101 target points. 0.03 seconds if I skip the `SVector`.

src/Remapping/remapping_utils.jl
```diff
@@ -0,0 +1,107 @@
...
+negative, it is remapped to (0,1), if positive to (-1, 0)). This is best used alongside with
+`vertical_bounding_indices`.
+
+`zcoords` is interpreted as "reference z coordinates".
+"""
+function vertical_reference_coordinates(space, zcoords)
+    vert_topology = Spaces.vertical_topology(space)
+    vert_mesh = vert_topology.mesh
+
+    is_cell_center = space isa Spaces.CenterExtrudedFiniteDifferenceSpace
```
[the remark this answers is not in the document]
Sbozzolo: Yes, `CenterExtrudedFiniteDifferenceSpace` is the correct one.

src/Remapping/distributed_remapping.jl
Sbozzolo: Done!

src/Remapping/distributed_remapping.jl
Sbozzolo: Moved it.

src/Remapping/distributed_remapping.jl
Sbozzolo: Here, we don't want to preserve type: we want a Tuple. We use NTuple{1} vs NTuple{2} to distinguish the number of horizontal dimensions in the kernels.

src/Remapping/distributed_remapping.jl
Sbozzolo: Added.

src/Remapping/distributed_remapping.jl
```diff
@@ -77,43 +112,77 @@ function target_hcoords_pid_bitmask(target_hcoords, topology, pid)
...
+    vert_interpolation_weights::VERT_W
+
+    # Given the target_zcoords, vert_bounding_indices contain the vertical indices of the
+    # neighboring elements that are required for vertical interpolation.
+    # Array of tuples or Nothing.
+    vert_bounding_indices::VERT_IND
+
+    # Scratch space where we save the process-local interpolated values. We keep overwriting this
+    # to avoid extra allocations.
+    _local_interpolated_values::T8
```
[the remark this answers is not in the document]
Sbozzolo: Added.

src/Remapping/distributed_remapping.jl
Sbozzolo: Added.

src/Quadratures/Quadratures.jl
charleskawczynski: I understand that, but why would we need a large number of points? The idea is that this function is used point/element-wise. Doing so with SVectors allows it to be efficient and stack allocated.

> 101x101x101 target points

What are the dims supposed to be here?

src/Remapping/distributed_remapping.jl
charleskawczynski: That's fine, then let's go with `local_horiz_interpolation_weights = map(ξs-> ArrayType(Quadratures.interpolation_matrix(ξs, quad_points)), ξs_split))` (`ξs_split` is a Tuple).

src/Remapping/distributed_remapping.jl
charleskawczynski: Should we convert that comment to a doc string?

src/Quadratures/Quadratures.jl
Sbozzolo: That's the target (lat, lon, z) grid (`z` does't matter because this is only for the horizontal interpolation.) The function `interpolation_matrix` is not element-wise, that's why you have a `points_to`, `points_from`. I mean, I could use it 101x101 times with `points_to` one at the time, but I don't think that's the intent.

src/Remapping/distributed_remapping.jl
Sbozzolo: Ok.

src/Remapping/remapping_utils.jl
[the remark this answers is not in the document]
charleskawczynski: Ok, why is this (no objection, I'd just like to better understand the logic)

src/Quadratures/Quadratures.jl
charleskawczynski: > I mean, I could use it 101x101 times with `points_to` one at the time, but I don't think that's the intent.

I could be wrong, but I think that's how this function is used elsewhere. Would it be difficult to change to use it the same way?

Sbozzolo: Besides interpolation, this function is used only to interpolate from a quadrature to a different quadrature (so, with multiple N points in both directions). It is true that the old interpolation system used this function pointwise, but this was because the entire interpolation routines were written to be pointwise. 

I want to handle all the interpolation points at the same time, instead of one point at the time. This is one of the key ingredients needed to achieve the performance improvement. So, I need a data structure that holds all the interpolation matrices. If I have to treat each point separately, I would probably have to launch a seperate GPU kernel per point (~ 10^4 kernel launches) or pass ~10^4 SVectors to a single kernel. Instead, I put everything into a CuArray and pass it.

src/Remapping/distributed_remapping.jl
Sbozzolo: I added references. The commend is too longwinded to be a useful docstring.

src/Remapping/remapping_utils.jl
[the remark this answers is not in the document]
Sbozzolo: The entire interpolation scheme is designd to only work with extruded (spectral) spaces, so `space` will be a `CenterExtrudedFiniteDifferenceSpace`. 

But I do see your point if you look at the function in isolation. The function will probably work with `CenterFiniteDifferenceSpace`s too.

src/Quadratures/Quadratures.jl
charleskawczynski: We discussed this offline.

The current design is an improvement over what we have. In the long run, it'd probably be better to fuse these interpolations into a single kernel (and just do the interpolation matrix on the fly), but we can leave this for the future.

src/Remapping/distributed_remapping.jl
charleskawczynski: `ntuple(i->map(ξ->ξ[i], a), 2)` is fewer characters 🙃

src/Remapping/distributed_remapping.jl
charleskawczynski: We discussed this a bit too. using `reduce!` would be type stable.

@Sbozzolo, since this is kind of internals, I'm fine with leaving this improvement for a later iteration.

charleskawczynski (approved):

Sbozzolo:
@szy21 @akshaysridhar 

This PR removes support for physical_z in ClimaCore. The reason for this is that it is quite painful to support with the recent changes in the hypsography, and it won't be straightforward to support SLEVE. Supporting this also leads to a bunch of extra checks and functions in Atmos to understand what to do for the various cases.

My idea is to only support reference z, and move the transformation to physical z to ClimaAnalysis. This is also more or less in line with what the [CF conventions recommend](https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-hybrid-height-coordinate).

What do you think? We can still support physical_z if we want to (with a little bit of work).

src/Remapping/distributed_remapping.jl
Sbozzolo: `collected_interpolated_values!` now uses `reduce!`

akshaysridhar:
> @szy21 @akshaysridhar
> 
> This PR removes support for physical_z in ClimaCore. The reason for this is that it is quite painful to support with the recent changes in the hypsography, and it won't be straightforward to support SLEVE. Supporting this also leads to a bunch of extra checks and functions in Atmos to understand what to do for the various cases.
> 
> My idea is to only support reference z, and move the transformation to physical z to ClimaAnalysis. This is also more or less in line with what the [CF conventions recommend](https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-hybrid-height-coordinate).
> 
> What do you think? We can still support physical_z if we want to (with a little bit of work).

@Sbozzolo I think removing support for the `physical_z` within ClimaCore seems reasonable given the CF convention recommendations -  My understanding is that with this change, users can still define their transform functions from `ref -> physical_z` within `ClimaAnalysis` scripts, so the post-processing functionality (including custom functions for sigma coordinates etc.) is still available.

szy21:
Wouldn't we need physical_z in a couple of places in ClimaAtmos?

akshaysridhar:
> Wouldn't we need physical_z in a couple of places in ClimaAtmos?

Even with these changes (which iiuc affect the writer / remapper steps) - the `coordinate_field(Y.c.x).z` would still contain the "true" `physical_z` coordinates once a "warp" has been applied - can we use this in ClimaAtmos where necessary?

szy21:
Ok, I think it is fine as long as `Fields.coordinate_field(Y.c).z` is the physical_z - is this the case?

Sbozzolo:
`physical_z` in this context is only for the NetCDF writer in Atmos.

akshaysridhar:
> `physical_z` in this context is only for the NetCDF writer in Atmos.

Thanks! LGTM!

Issue #1465, opened 2023-09-19
FieldVector broadcast inference failure


charleskawczynski:
Reproducer:

```julia
import ClimaCore
import ClimaCore.Spaces as Spaces
import ClimaCore.Fields as Fields
using StaticArrays
include(
    joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"),
)
import .TestUtilities as TU

function foo!(args)
    (; residual, temp, dt,a_imp, Ui) = args
    i=1
    @. residual = temp + dt * a_imp[i, i] * residual - Ui
    return nothing
end

FT = Float32
cspace = TU.CenterExtrudedFiniteDifferenceSpace(FT);
Yc = fill((;a=FT(0)),cspace);
args = (;
    residual=Fields.FieldVector(;c=Yc),
    temp=Fields.FieldVector(;c=Yc),
    a_imp=SMatrix{1,1}(rand(1,1)),
    Ui=Fields.FieldVector(;c=Yc),
    dt = 1.0,
);
foo!(args)

using JET
JET.@test_opt foo!(args)

```

charleskawczynski:
I think this is blocking us from using the latest ClimaTimeSteppers (which is OOMing)

charleskawczynski:
Strangely

```julia
using JET; v = Int[1,7,3,4,1]; JET.@test_opt sort!(v)
using Revise; include("../cc_bc_inf.jl")
```
(where `cc_bc_inf.jl` is the MWE above) passes. Something very strange is going on with `sort!(::Vector{Int})`.. cc @simonbyrne

charleskawczynski:
```julia
julia> versioninfo()
Julia Version 1.9.3
Commit bed2cd540a1 (2023-08-24 14:43 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
```

charleskawczynski:
Boiled down to:
```julia
using BlockArrays
shape1 = (BlockArrays._BlockedUnitRange((2,)),);
shape2 = (BlockArrays._BlockedUnitRange((2,)),);
Base.Broadcast._bcs(shape1, shape2)
[...]

PR #1615, opened 2024-02-20
Specialize on diagonal fieldvector broadcasts

charleskawczynski:
This PR specializes on "diagonal" broadcast expressions with fieldvectors. Specifically, if a fieldvector broadcast expression (one with a `FieldVectorStyle`) has one and only one fieldvector type, then we don't check the broadcast axes (the combined axes is the same). The benefit of skipping this code path is that checking the axes is type-unstable (https://github.com/JuliaArrays/BlockArrays.jl/issues/310). From what I can tell, there is not a simple path to fixing https://github.com/JuliaArrays/BlockArrays.jl/issues/310 because:

 - We _want_ to use `Tuple`s, so that this is stack-allocated
 - Combining axes for `Tuple`s is type unstable because the result depends on the _values_ in the tuples.

So, I think that the simplest solution is to specialize when we know that the axes will not change.

This PR adds a test to show that `is_diagonal_bc` will return true/false when expected _at_ _compile_ _time_ (`@code_typed is_diagonal_bc(bc)` returns `true` or `false`), and tests that (what was breaking) fieldvector broadcast expressions with multiple fields does not allocate.

Closes https://github.com/CliMA/ClimaCore.jl/issues/1465.

One other benefit of this solution is that we can at some point reuse `is_diagonal_bc` to specialize on these broadcast expressions in `copyto!`, so that we perform a single cuda kernel launch (as opposed to however many fields are in the fieldvector).

The only thing remaining is that I'd like to run ClimaAtmos CI on this branch to make sure that this doesn't blow up compile-times. Update: good news, it looks fine xref: https://github.com/CliMA/ClimaAtmos.jl/pull/2700#issuecomment-1954497852

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -203,6 +203,93 @@ end
 @inline transform_broadcasted(fv::FieldVector, symb, axes) =
     parent(getfield(_values(fv), symb))
 @inline transform_broadcasted(x, symb, axes) = x
+
+@inline function first_fieldvector_in_bc(args::Tuple, rargs...)
+    x1 = first_fieldvector_in_bc(args[1], rargs...)
+    x1 isa FieldVector && return x1
+    return first_fieldvector_in_bc(Base.tail(args), rargs...)
+end
+
+@inline first_fieldvector_in_bc(args::Tuple{Any}, rargs...) =
+    first_fieldvector_in_bc(args[1], rargs...)
+@inline first_fieldvector_in_bc(args::Tuple{}, rargs...) = nothing
+@inline first_fieldvector_in_bc(x) = nothing
+@inline first_fieldvector_in_bc(x::FieldVector) = x
+
+@inline first_fieldvector_in_bc(
+    bc::Base.Broadcast.Broadcasted{FieldVectorStyle},
+) = first_fieldvector_in_bc(bc.args)
+
+@inline _is_diagonal_bc_args(
+    truesofar,
+    ::Type{TStart},
+    args::Tuple,
+    rargs...,
+) where {TStart} =
+    truesofar &&
+    _is_diagonal_bc(truesofar, TStart, args[1], rargs...) &&
+    _is_diagonal_bc_args(truesofar, TStart, Base.tail(args), rargs...)
+
+@inline _is_diagonal_bc_args(
+    truesofar,
+    ::Type{TStart},
+    args::Tuple{Any},
+    rargs...,
+) where {TStart} =
+    truesofar && _is_diagonal_bc(truesofar, TStart, args[1], rargs...)
+@inline _is_diagonal_bc_args(
+    truesofar,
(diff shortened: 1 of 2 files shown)


Sbozzolo:
I think @dennisYatunin is in a better position to review this, so I'll leave this to him.

src/Fields/fieldvector.jl
```diff
@@ -203,6 +203,93 @@ end
...
+    if bc.axes isa Nothing # Not done via dispatch to make it easier to extend instantiate(::Broadcasted{Style})
+        axes = Base.Broadcast.combine_axes(bc.args...)
+    else
+        axes = bc.axes
+        # Base.Broadcast.check_broadcast_axes is type-unstable
+        # for broadcast expressions with multiple fieldvectors.
+        # So, let's statically elide this when we have "diagonal"
+        # broadcast expressions:
+        if !is_diagonal_bc(bc)
+            Base.Broadcast.check_broadcast_axes(axes, bc.args...)
```
dennisYatunin, line 287: Can we just throw an error when this isn't true? The example for this in the unit tests can't be materialized, because materializing it would result in a `Field` broadcast error for mismatched spaces. Why not throw an error for this at the `FieldVector` broadcast level?

dennisYatunin (approved):
Overall, I think this looks good. You could simplify `is_diagonal_bc` by rewriting it in terms of `Utilities.UnrolledFunctions.unrolled_any`, but what you have is fine. There is one change I'd like to see for catching mismatched spaces errors; see my comment.

src/Fields/fieldvector.jl
charleskawczynski: It's a good question, I suppose that we could (I have no objection). The only reason I didn't was because, in theory, we could support broadcasting over different field vectors (we would obviously need to define rules). That said, it's likely not going to ever happen. Thoughts @Sbozzolo / @sriharshakandala ?

dennisYatunin: I guess we might want to allow broadcasting over `FieldVector`s with the same spaces and data structures, but with different basetypes? Like adding a `FieldVector` of `Float32`s to a `FieldVector` of `Dual`s. But that would require dealing with promotion rules, and I haven't seen any need for it yet. We could also just implement it as an extension of `is_diagonal_bc`, with more complicated type checking.

charleskawczynski: True. We can always add the error and remove it if we find a need.

charleskawczynski: I've added an error for when it's not diagonal

charleskawczynski:
@dennisYatunin, looks like we can't error: https://buildkite.com/clima/climacore-ci/builds/3150#018dc877-7cd8-4de7-928d-de25532c360d. I'm going to merge this now, and we can follow up later if there's some improvement we can make.

PR #1616, opened 2024-02-20
Add timings to matrix field broadcasting

charleskawczynski:
This PR adds timings to the matrix field broadcasting. The first simple thing we could do is break this into a utilities function, and split up the tests into separate jobs. This would help, but these tests should probably be rewritten. They're allocating a bunch, it's not clear what's passing/broken/allocating from the output. The closures are probably not helping, and funny things can happen with JET tests inside `@testset`. The first thing we can do is time each one, and see if there's a major outlier.


Sbozzolo (approved):

charleskawczynski:
Hm, I think a better solution might be to split these tests out in our suite. That could reduce the time, allow the long ones to finish, and reveal which ones are really problematic.

charleskawczynski:
Superseded by #1672

PR #1617, opened 2024-02-20
Fix inference errors when not debugging

charleskawczynski:
This PR wraps some `@debug` calls in `if get_debug(alg)`, so that we don't trigger compiler inference failures in code paths that we'll never reach.

This should fix a lot of the inference errors we're seeing, due to KrylovKit.jl.

The diff looks a bit odd, but I'm only indenting.

This should allow us to see what remaining inference failures that we have in ClimaAtmos.

### DIFF
--- a/src/MatrixFields/field_matrix_iterative_solver.jl
+++ b/src/MatrixFields/field_matrix_iterative_solver.jl
@@ -1,3 +1,5 @@
+import Base.CoreLogging
+
 """
     PreconditionerAlgorithm
 
@@ -362,9 +364,13 @@ There are 4 values that can be included in `kwargs...`:
 - `eigsolve_kwargs = (;)`: keyword arguments for the `eigsolve` function that
   can be used to tune its accuracy and speed (only applicable when debugging the
   spectral radius)
+- `debug = nothing`: keyword argument for printing debug information to `@debug`.
+  By default, `debug` is set to true if `"error_norm"` or `"spectral_radius"` is in
+  `ENV["JULIA_DEBUG"]`, which must be set by users.
 """
 struct StationaryIterativeSolve{
     correlated_solves,
+    debug,
     P <: Union{Nothing, PreconditionerAlgorithm},
     K <: NamedTuple,
 } <: LazyFieldMatrixSolverAlgorithm
@@ -375,9 +381,19 @@ end
 function StationaryIterativeSolve(;
     P_alg = nothing,
     n_iters = 1,
+    debug = nothing,
     correlated_solves = false,
     eigsolve_kwargs = (;),
 )
+    B =
+        CoreLogging.min_enabled_level(CoreLogging.current_logger()) <=
+        CoreLogging.Debug
+    _debug = if isnothing(debug)
+        B
+    else
+        B || @warn "`ENV[\"JULIA_DEBUG\"]` must be set to use `debug = true`."
+        debug
+    end
     # Since Field operations can be much slower than typical Array operations,
     # the default values of krylovdim and maxiter specified in KrylovKit.jl
     # should be replaced with smaller values, and the default value of tol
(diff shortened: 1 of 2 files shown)


src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -375,6 +376,7 @@ end
 function StationaryIterativeSolve(;
     P_alg = nothing,
     n_iters = 1,
+    debug = false,
```
Sbozzolo, line 379: Can you add a dosctring explaning what `debug` does and when one would use it?

charleskawczynski: Good idea, done!

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -411,23 +415,29 @@ check_field_matrix_solver(alg::StationaryIterativeSolve, cache, A, b) =
...
+                concrete_field_vector(@. e_view - lazy_Δe) # (I - inv(P) * A) * e
+            end
+            if info.converged == 0
+                (; tol, maxiter) = alg.eigsolve_kwargs
+                "Unable to approximate ρ(I - inv(P) * A) to within a tolerance \
+                of $(100 * tol) % in $maxiter or fewer iterations"
+            else
+                "ρ(I - inv(P) * A) ≈ $(abs(λs[1]))"
+            end
+        end _group = :spectral_radius
```
dennisYatunin, line 439:
```suggestion
        end
```
If we're going to toggle debugging at the `struct` level, let's remove toggling at the logger level. It'll be a bit of a hassle to remember to toggle both the struct flag and the value of `ENV["JULIA_DEBUG"]`. Can we get rid of the `_group` restrictions and change the `@debug`s to `@info`s?

dennisYatunin: Alternatively, can we set the default value of `debug` based on whether `ENV["JULIA_DEBUG"]` contains the substrings `"error_norm"` or `"spectral_radius"`? That's probably a bit more generalizable.

charleskawczynski: Sounds good to me. I'll make `debug` depend on `ENV["JULIA_DEBUG"]`

dennisYatunin (approved):

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -375,19 +377,29 @@ end
 function StationaryIterativeSolve(;
     P_alg = nothing,
     n_iters = 1,
+    debug::Bool = nothing,
```
dennisYatunin, line 380:
```suggestion
    debug = nothing,
```

charleskawczynski: Going to use
```julia
    B = any(("error_norm", "spectral_radius")) do x
        occursin(x, get(ENV, "JULIA_DEBUG", ""))
    end
    _debug = if isnothing(debug)
        B
    else
        B || @warn "`ENV[\"JULIA_DEBUG\"]` must be set to use `debug = true`."
        debug
    end
```

src/MatrixFields/field_matrix_iterative_solver.jl
```diff
@@ -375,19 +379,31 @@ end
 function StationaryIterativeSolve(;
     P_alg = nothing,
     n_iters = 1,
+    debug = nothing,
     correlated_solves = false,
     eigsolve_kwargs = (;),
 )
+    B = any(("error_norm", "spectral_radius")) do x
+        occursin(x, get(ENV, "JULIA_DEBUG", ""))
+    end
```
dennisYatunin, line 388:
```suggestion
    B = Logging.min_enabled_level(Logging.current_logger()) <= Logging.Debug
```
If you want to avoid needing to modify `ENV` for the unit test, you can do something like this. There's also the more specific `Logging.shouldlog` function that can check whether to log the `error_norm` and `spectral_radius` groups, but that's already called internally by the logger, and there's no benefit to doing it twice.

PR #1618, opened 2024-02-22
Update documentation

akshaysridhar:
Docs + refs updated in Hypsgraphy.

This is a step towards @Sbozzolo 's effort to summarise artifact sources and preprocessing information for external data.


Sbozzolo:
In atmos, we also smooth the profile with a Gaussian kernel before passing it to ClimaCore. Should we be doing both?

akshaysridhar:
This is not strictly necessary; currently we're doing this as a pre-processing step from the original ETOPO1 datasets - the raw data file size is intractable as an artifact so we use a truncated, smoothed dataset in ClimaAtmos.

Issue #1296, opened 2023-06-06
Move space and field from GPU to CPU


simonbyrne:
It would be helpful to have a function that can move a space and field from the GPU to CPU (or vice versa), e.g. for plotting.

closed by #1619

**Describe the solution you'd like**
The question is what the interface should look like:
- `todevice(device, field/space)`?

**Additional context**
- we could use a similar approach for switching from `MPICommsContext` to `SingletonCommsContext`.

simonbyrne:
cc @charleskawczynski @sriharshakandala thoughts?

sriharshakandala:
For very small simulations/lightweight postprocessing, yes. 
It is generally best to writeout data and do postprocessing separately, IMO!

charleskawczynski:
I like the idea. This should be somewhat straightforward with adapt, right? Are there any challenges you see?

simonbyrne:
Possibly? I'm trying some things out at the moment

charleskawczynski:
Closed by #2114

Issue #1290, opened 2023-06-05
Reorganize test suite: "expensive unit tests" are not running on the gpu CUDA


charleskawczynski:
## Describe the bug
Any tests put into our "expensive tests" group are not running on the gpu, because there's not a buildkite job for it.

This is actually very confusing because a lot of the code in the expensive tests appear to be written for cpu + gpu, and identifying what is and is not run on the gpu requires looking at the higher level scripts.

The other issue with this is that extending the existing tests is not intuitive. We should give this a bit more thought and reorganize the test suite.

My first instinct is to just make a cpu+gpu test per module, which will mean increasing the number of buildkite jobs by quite a bit but there are several advantages:
 - Extending it is easier / simpler
 - It will scale well
 - It may help cut down CI time

Sbozzolo:
An added issue (given #1689) is that it is pretty much impossible to run a comprehensive test suite for GPUs outside of buildkite

PR #1619, opened 2024-02-22
add function to convert fields, spaces, grids between CPU and GPU

juliasloan25:
Adds a function `todevice`, which basically provides an interface wrapping Adapt.jl's `adapt_structure` function. `adapt_structure` has already been extended for (I think) all fields, spaces, and grids.

closes https://github.com/CliMA/ClimaCore.jl/issues/1296

## Content
- [x] add `todevice` methods for Fields, Spaces, Grids
- [ ] add tests for Fields, Spaces, Grids
- [ ] add to documentation (API)

## To Do
- [ ] Spaces (in Fields test): `SpectralElement1D` space constructed on GPU doesn't have CuArray as parent - figure out why
  - `SpectralElement1D` not created on GPU - remove from tests
- [ ] add methods to use context instead of array type
- [ ] Spaces src/: change context of spaces, not just backing array types
- [ ] Spaces: add meaningful checks in tests - figure out how to test spaces most generally (or do I just have to write individual tests for each space type?)
  - array_type, look at grid test with cpu/gpu conversion of arrays
- [ ] Grids: add meaningful checks in tests

## QA

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -61,7 +61,7 @@ if !Sys.iswindows()
     @safetestset "Spectral elem - sphere diffusion vec" begin @time include("Operators/spectralelement/sphere_diffusion_vec.jl") end
     @safetestset "Spectral elem - sphere hyperdiffusion" begin @time include("Operators/spectralelement/sphere_hyperdiffusion.jl") end
     @safetestset "Spectral elem - sphere hyperdiffusion vec" begin @time include("Operators/spectralelement/sphere_hyperdiffusion_vec.jl") end
-    
+
     @safetestset "FD ops - column" begin @time include("Operators/finitedifference/column.jl") end
     @safetestset "FD ops - opt" begin @time include("Operators/finitedifference/opt.jl") end
     @safetestset "FD ops - wfact" begin @time include("Operators/finitedifference/wfact.jl") end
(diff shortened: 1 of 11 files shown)


test/runtests.jl
```diff
@@ -111,6 +111,7 @@ end
 if "CUDA" in ARGS
     @safetestset "GPU - cuda" begin @time include("gpu/cuda.jl") end
     @safetestset "GPU - data" begin @time include("DataLayouts/cuda.jl") end
+    @safetestset "GPU - grids" begin @time include("Grids/grids.jl") end
```
charleskawczynski, line 114: We don't run `runtests.jl` on buildkite (https://github.com/CliMA/ClimaCore.jl/issues/1290), so we need to add a buildkite job to exercise this.

juliasloan25:
This task has proved to be more work than anticipated, so I'll stop working on it for now. The biggest barriers currently are:
- `adapt_structure` seems not to be recursing all the way for some spaces/grids, so that the `local_geometry` field remains on the original device (seen for DataF and DataVF types)
- `adapt_structure` can be used to convert backing arrays/internal structures, but also need to think about converting context (or just create new space/grid with desired context - not sure if this would require hardcoding SingletonCommsContext vs MPICommsContext when moving to CPU)

juliasloan25:
@dennisYatunin

PR #1620, opened 2024-02-23
Start moving to new-central

Sbozzolo:
This PR moves the pipeline to new central (but still using the `any` partition).


charleskawczynski:
Looks like this finally went through, I'll go ahead and merge since we still have a backfill.

PR #1623, opened 2024-02-23
Do not test field mat solvers logs on gpu

charleskawczynski:
This PR fixes CI, I must have only tested this on the CPU locally (I don't think this was previously working, but the debug config changed how it was being run in the test suite).

### DIFF
--- a/test/MatrixFields/field_matrix_solvers.jl
+++ b/test/MatrixFields/field_matrix_solvers.jl
@@ -298,7 +298,10 @@ end
     end
 
     @testset "approximate iterative solve with debugging" begin
-        Logging.with_logger(Logging.SimpleLogger(stderr, Logging.Debug)) do
+        logger =
+            using_cuda ? Logging.NullLogger() :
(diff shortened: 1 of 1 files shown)


test/MatrixFields/field_matrix_solvers.jl
```diff
@@ -298,7 +298,10 @@ end
     end
 
     @testset "approximate iterative solve with debugging" begin
-        Logging.with_logger(Logging.SimpleLogger(stderr, Logging.Debug)) do
+        logger =
+            using_cuda ? Logging.NullLogger() :
+            Logging.SimpleLogger(stderr, Logging.Debug)
+        Logging.with_logger(logger) do
```
dennisYatunin, line 304: Do you need to set the logger here? I think `@test_logs` takes care of setting the logger internally.

charleskawczynski: Yeah, this is how I got this to run

dennisYatunin (approved):

PR #1625, opened 2024-02-25
Add changelog

charleskawczynski:
This PR adds a changelog.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,29 @@
+ClimaCore.jl Release Notes
+========================
+
+v0.12.1
+-------
+- Started changelog
+- Fixed matrix field iterative solver tests.
+- ![][badge-🚀performance] Specialize on diagonal fieldvector broadcasts to skip uninferred `check_broadcast_axes` PR [#1615](https://github.com/CliMA/ClimaCore.jl/pull/1615), Issue [#1465](https://github.com/CliMA/ClimaCore.jl/issues/1465).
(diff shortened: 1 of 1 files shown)


Sbozzolo (commented):
While I like the badges and the resulting neatly organized document, I think this is a little bit too many possible badges. I am worried that just thinking about what badge to use will add friction that discourage people from keeping the changelog properly updated. This is the same as our GitHub labels, we have so many that it is unclear which ones should be used.

I think that what is the most important to flag is interface breaking and physics altering changes.

juliasloan25:
I agree with Gabriele's comment - maybe we could limit the options to:
- behavioral change
- precision change
- breaking
- bugfix
- feature/enhancement

charleskawczynski:
Sounds good, I agree. I thought I'd try adding a big list figuring that we might want to trim it down

Sbozzolo:
Okay. I'd say, let's keep 
```
[badge-🔥behavioralΔ]: https://img.shields.io/badge/🔥behavioralΔ-red.svg
[badge-🤖precisionΔ]: https://img.shields.io/badge/🤖precisionΔ-black.svg
[badge-💥breaking]: https://img.shields.io/badge/💥BREAKING-red.svg
```
because people reading the changelog will need to informed about this.

Can also you please also add a one line description of what the badges are meant to identify so that everyone is on the same page?

sriharshakandala:
I am a bit ambiguous on the `precision change` badge. Otherwise looks good to me!

juliasloan25 (approved):
This looks good! The badge descriptions are helpful and it will be nice to have a changelog.

Maybe we can implement in other repos next, using this as a template

charleskawczynski:
> I am a bit ambiguous on the precision change badge. Otherwise looks good to me!

I've put a description of what it means in the file. I think it's useful for people who are upgrading packages-- they will have better expectations on how their results may change if they only see precision change as opposed to behavioral change.

PR #1628, opened 2024-02-26
Improve type broadcasting

### LINKED ISSUE #1602: Inference failure when inlining
I need to dig a bit more into why this happens, however, what I've seen is that if `PhasePartition` (in Thermodynamics.jl) in this reproducer is inlined, then the `broadcasted` anonymous function in `ClimaCore.jl/Fields/broadcast.jl`:

```julia
# types aren't isbits
Base.Broadcast.broadcasted(
    fs::AbstractFieldStyle,
    ::Type{T},
    args...,
) where {T} = Base.Broadcast.broadcasted(fs, (x...) -> T(x...), args...)
```

is not inferred. Here's the reproducer:

```julia
#=
julia --project=perf
include("../perf/climacore_type_bc_inf.jl")
=#
using Revise
import ClimaCore;
import Thermodynamics as TD
import Thermodynamics.Parameters as TDP
import CloudMicrophysics as CM
import ClimaCore.Fields as Fields
import ClimaCore.Operators as Operators
import ClimaCore.Spaces as Spaces
import ClimaCore.Geometry as Geometry
import ClimaComms
@isdefined(TU) || include(joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"));
import .TestUtilities as TU;
FT = Float64;
zelem=25
cspace = TU.CenterExtrudedFiniteDifferenceSpace(FT;zelem, helem=10);
fspace = Spaces.FaceExtrudedFiniteDifferenceSpace(cspace);
space = cspace;
@show ClimaComms.device(cspace)

# From ClimaAtmos
using ClimaParams
import ClimaParams as CP
aliases = string.(fieldnames(TD.Parameters.ThermodynamicsParameters));
toml_dict = CP.create_toml_dict(FT);
const thermo_params = TD.Parameters.ThermodynamicsParameters(toml_dict)
const cm_params = CM.Parameters.Parameters0M(FT)

x = (;
    ᶜts       = Fields.Field(TD.PhaseEquil{FT}, space),
    ᶜS_ρq_tot = Fields.Field(FT, space),
    ρ         = Fields.Field(FT, space),
)

total_specific_humidity(tp, ts) = TD.PhasePartition(tp, ts).tot
function compute_precipitation_cache!(x)
    (; ᶜts, ρ, ᶜS_ρq_tot) = x
    @. ᶜS_ρq_tot =
        ρ * CM.Microphysics0M.remove_precipitation(
            cm_params,
            TD.PhasePartition(thermo_params, ᶜts),
        )
    # @. ᶜS_ρq_tot = ρ * TD.PhasePartition(thermo_params, ᶜts).tot # allocates!
    # @. ᶜS_ρq_tot = ρ * total_specific_humidity(thermo_params, ᶜts) # allocation-free
    return nothing
end

compute_precipitation_cache!(x)

import JET
JET.@test_opt compute_precipitation_cache!(x)

function do_work!(x)
    for i in 1:200
        compute_precipitation_cache!(x)
    end
end

import Profile, ProfileCanvas
@info "Compiling first..."
do_work!(x) # compile first
@info "Collecting profile..."
Profile.clear()
prof = Profile.@profile do_work!(x)
results = Profile.fetch()
Profile.clear()
@info "Generating html..."
ProfileCanvas.html_file("flame.html", results)

@info "Benchmarking..."
import BenchmarkTools
trial = BenchmarkTools.@benchmark compute_precipitation_cache!($x)
show(stdout, MIME("text/plain"), trial)
```

This function is ~100x slower as a result. cc @glwagner and @dennisYatunin (just FYI). It's possible that we're hitting a heuristic (`PhasePartition` calls `PhasePartition_equil` with a handful of arguments), or maybe it's the use of `phase_type`, which is somewhat of an edge case in Thermodynamics.

Whatever the issue is, this is a useful reproducer because one other thing I noticed is that `@. ᶜS_ρq_tot = ρ * TD.PhasePartition(thermo_params, ᶜts).tot` allocates, making it slower. It's important that we track and document these issues, and spread best-practice patterns to avoid shortcomings with our broadcast machinery, or improve its robustness.

In addition, JET seems to miss this inference failure, however, it is caught in the flame graph (and the timing difference is observed through BenchmarkTools).

charleskawczynski:
This PR should help improve how types are broadcasted, especially on the GPU.

Closes #1602.

We may want to also remove

```julia
Base.Broadcast.broadcasted(
    fs::Base.Broadcast.DefaultArrayStyle{0},
    ::Type{T},
    args...,
) where {T <: Geometry.AxisTensor} =
    Base.Broadcast.broadcasted(fs, (x...) -> T(x...), args...)
```
I'll try that out separately.


Sbozzolo:
Do you want to get this and 1636 in before releasing 0.13?

charleskawczynski:
Yeah, I think that'd be nice. #1636 supersedes this, so I think we only need #1636. I'm double checking now that it for sure closes out #1602.

charleskawczynski:
Superseded by #1636.

PR #1629, opened 2024-02-27
Add flop counting tests back in

charleskawczynski:
This PR adds the flop counting tests back in.


perf/Project.toml
```diff
@@ -8,6 +8,7 @@ ClimaCoreTempestRemap = "d934ef94-cdd4-4710-83d6-720549644b70"
 ClimaCoreVTK = "c8b6d40d-e815-466f-95ae-c48aefa668fa"
 ClimaTimeSteppers = "595c0a79-7f3d-439a-bc5a-b232dc3bde79"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+CountFlops = "1db9610d-79e1-487a-8d40-77f3295c7593"
```
Sbozzolo, line 11: Shouldn't this be in the test project as well, givne that it is required by a script in the test filder?

charleskawczynski: Ah, yes, thank you

PR #1638, opened 2024-03-04
Fix ClimaCoreMakie tests

charleskawczynski:
Missed in https://github.com/CliMA/ClimaCore.jl/pull/1632.


Sbozzolo:
What's the difference, except for the imports?

charleskawczynski:
> What's the difference, except for the imports?

I had used `Geometry.ZPoint`, assuming `Geometry` was in scope (but it wasn't), like most of the other tests.

Sbozzolo:
Got it!

Sbozzolo (approved):

PR #1641, opened 2024-03-11
Add support for MultiBroadcastFusion

charleskawczynski:
This PR adds support for the use of [MultiBroadcastFusion.jl](https://github.com/CliMA/MultiBroadcastFusion.jl), in order to allow users to fuse multiple broadcast expressions into a single kernel launch.

We'll be able to decorate multiple broadcasts with `@fused`, e.g.,:

```julia
@fused begin
    @. y1 = x1+x2+x3
    @. y2 = x1*x2*x3
end
```
Which will result in the compiler being able to hoist global memory reads, improving performance.

I'll open an issue on expanding this for a few cases:

 - Support for more memory layouts
 - Support for more types of broadcast styles (right now, only point-wise kernels, with compatible broadcast types)

But I'd like to start with this simple case first, since there should be some low hanging fruit that we can leverage (and also see how things work in production).

Once https://github.com/JuliaRegistries/General/pull/102559 is merged, I'll update the PR to add the correct dependency.

This is a step towards https://github.com/CliMA/ClimaAtmos.jl/issues/2632.

Right now, we are restricted to the following limitations:

 - Only similar field types can be fused, and no combinations of non-similar types are allowed in the left or right-hand side of the expression (they should error). For example, for following are not allowed
 - Vertical interpolation and finite difference operators are not (yet) supported.
 - Horizontal operators are not (yet) supported.

Some examples summarizing this are below:

The following will error:
```julia
@fused begin
  @. center_field = 1
  @. face_field = 1
end
```
The following will error:
```julia
@fused begin
  @. center_sphere = 1
  @. center_column = 1
end
```

The following will error:
```julia
@fused begin
  @. center_field_1 = 1 * ᶜinterp(face_field)
  @. center_field_2 = 2 *  ᶜinterp(face_field)
end
```

The following will error:
```julia
@fused begin
  @. center_field_1 = 1 * ᶜinterp(face_field)
  @. center_field_2 = 2 *  ᶜinterp(face_field)
end
```

The following will work:
```julia
@fused begin
  @. center_field_1 = 2
  @. center_field_2 = 1
end
```

Any pointwise function should work, and it's advantageous to use `@fused` when there is at least one variable that is shared across multiple broadcast expressions in any way. For example:

converting
```julia
    @. ᶜ∇²uₕʲs.:($$j) = C12(ᶜ∇²uʲs.:($$j))
    @. ᶜ∇²uᵥʲs.:($$j) = C3(ᶜ∇²uʲs.:($$j))
```
to
```julia
@fused begin
    @. ᶜ∇²uₕʲs.:($$j) = C12(ᶜ∇²uʲs.:($$j))
    @. ᶜ∇²uᵥʲs.:($$j) = C3(ᶜ∇²uʲs.:($$j))
end
```
will reduce 1 memory read of because `ᶜ∇²uʲs.:($$j)` shows up in the right-hand side of both expressions. In addition, converting
```julia
    @. ᶜts = ts_gs(thermo_args..., ᶜspecific, ᶜK, ᶜΦ, Y.c.ρ)
    @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
```
to
```julia
@fused begin
    @. ᶜts = ts_gs(thermo_args..., ᶜspecific, ᶜK, ᶜΦ, Y.c.ρ)
    @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
end
```
Will result in the reuse of `ᶜts`, resulting in 1 fewer read, as the value computed in the first line can be reused in the second line.

### DIFF
--- a/src/DataLayouts/broadcast.jl
+++ b/src/DataLayouts/broadcast.jl
@@ -1,3 +1,11 @@
+import MultiBroadcastFusion as MBF
+import MultiBroadcastFusion: fused_direct
+
+# Make a MultiBroadcastFusion type, `FusedMultiBroadcast`, and macro, `@fused`:
+# via https://github.com/CliMA/MultiBroadcastFusion.jl
+MBF.@make_type FusedMultiBroadcast
+MBF.@make_fused fused_direct FusedMultiBroadcast fused_direct
+
 # Broadcasting of AbstractData objects
 # https://docs.julialang.org/en/v1/manual/interfaces/#Broadcast-Styles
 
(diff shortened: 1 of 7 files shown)


Sbozzolo:
Can you please add a short usage guide for developers? (Ie, when should we use `@fused`?)

src/DataLayouts/broadcast.jl
```diff
@@ -1,3 +1,9 @@
+import MultiBroadcastFusion as MBF
+
+# Make a MultiBroadcastFusion type, `FusedMultiBroadcast`, and macro, `@fused`:
+# via https://github.com/CliMA/MultiBroadcastFusion.jl
+MBF.@make_fused FusedMultiBroadcast fused
```
dennisYatunin, line 5: Would you mind leaving the original definition of this struct here, without the macro annotation? I think it's fine to define iteration utilities and frontend macros in `MultiBroadcastFusion.jl`, but this sort of core struct definition should probably be explicitly written out in ClimaCore, as it was in previous commits to this PR. Also, we could potentially use other frontends to this struct; i.e., with more fusion capability than MBF.

charleskawczynski: Sure, did you have something specific in mind? Is it alright if I leave it as a comment for now? I just figured that it would be nice to not require users to define macros and deal with hygiene. Also, since it defines the struct, we could make different types and `@fused` macros.

The "core" function of MBF is `fused_pairs`, but I think we could add more options (https://github.com/CliMA/MultiBroadcastFusion.jl/issues/5).

charleskawczynski: cc @dennisYatunin?

dennisYatunin: Sure, you can leave this a comment for now to get some performance measurements. We can remove the `make_fused` definition and replace it with one that supports fusion across center+face spaces in a future PR.

charleskawczynski: So, `make_fused` is technically compatible with fusion across center+face spaces, those checks/that functionality have to happen on the ClimaCore side. `make_fused` is currently limited in that it can't contain loops, (generally) function calls, aliases, and if-else statements. I agree that we can try this in a future PR.

charleskawczynski:
I've renamed `@fused` to `@fused_direct`, so that we also specify the style of how the expressions are collected. This will help differentiate against when we add more collection methods.

@dennisYatunin, I think that this is in good enough shape to merge. Can you take a look when you have a chance?

dennisYatunin (approved):
This is a good start! Looking forward to seeing this expanded to mismatched center/face/surface spaces and to non-pointwise operators.

PR #1642, opened 2024-03-11
Fix interpolate! for MPI runs

Sbozzolo:
`interpolate!` did not allow `dest` to be `nothing`, as we would have in MPI runs (we only allocate `dest` on root).

This fix is blocking me from using it in ClimaAtmos, so I am also making a patch release.

### DIFF
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -715,17 +715,19 @@ function _collect_interpolated_values!(dest, remapper::Remapper)
 end
 
 """
-   interpolate(remapper, field)
-   interpolate!(dest, remapper, field)
+   interpolate(remapper::Remapper, field)
+   interpolate!(dest, remapper::Remapper, field)
 
 Interpolate the given `field` as prescribed by `remapper`.
 
 This call mutates the internal (private) state of the `remapper`.
 
 Horizontally, interpolation is performed with the barycentric formula in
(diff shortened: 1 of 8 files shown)


src/Remapping/distributed_remapping.jl
```diff
@@ -770,17 +772,21 @@ function interpolate(remapper::Remapper, field::T) where {T <: Fields.Field}
...
 
     axes(field) == remapper.space ||
         error("Field is defined on a different space than remapper")
 
-    size(dest) == size(remapper._interpolated_values) || error(
-        "Destination array is not compatible with remapper (size mismatch)",
-    )
+    if !isnothing(dest)
+        # !isnothing(dest) means that this is the root process, in this case, the size have
+        # to match
```
charleskawczynski, line 785: What happens below if this is not the root process?

charleskawczynski (approved):

src/Remapping/distributed_remapping.jl
Sbozzolo: Each MPI process has to run the code below. Each MPI process performs interpolation of its local points, then there is a MPI reduce! call to move everything to root and save it to `dest`.

Sbozzolo:
ClimaCore 0.13.1 running in Atmos with interpoalte!:

https://buildkite.com/clima/climaatmos-ci/builds/17355

Everything passes.

PR #1647, opened 2024-03-18
Always inline

charleskawczynski:
This PR makes all cuda kernel launches always inline, since this can reduce function call overhead. In addition, it could allow the compiler to hoist memory reads/writes.

### DIFF
--- a/.dev/Manifest.toml
+++ b/.dev/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.0"
+julia_version = "1.10.2"
 manifest_format = "2.0"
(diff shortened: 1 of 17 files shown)


Sbozzolo:
I'd be good to check that this doesn't break coupler and land too.

charleskawczynski:
Agreed, there’s still one thing to fix in atmos

charleskawczynski:
Atmos is fixed.

charleskawczynski:
I'm going to merge this, and we can fix the issues we run into.

Sbozzolo:
> I'm going to merge this, and we can fix the issues we run into.

I am [rebuilding](https://buildkite.com/clima/climaland-ci/builds/3633) your PR in Land that tests this out.

Issue #1652, opened 2024-03-21
Improve threading in gpu `run_field_matrix_solver!`


charleskawczynski:
Right now, our `single_field_solve` launches separate kernels per variable, but we can improve the performance, and scaling with respect to the number of state variables, by improving the threading pattern

<img width="1048" alt="Screen Shot 2024-03-21 at 3 09 10 PM" src="https://github.com/CliMA/ClimaCore.jl/assets/1880641/01994093-989c-4863-85fd-9252f98bc746">

I ran this on clima, with a single A100.

charleskawczynski:
Closed by #1653. I'll run the gpu pipeline soon to see the impact.

PR #1653, opened 2024-03-21
Fuse single_field_solve

charleskawczynski:
Trying out the simplest thing against #1652. I'll open a ClimaAtmos branch and try this out in the gpu pipeline.


src/MatrixFields/single_field_solver.jl
```diff
@@ -80,6 +80,57 @@ single_field_solve_kernel!(
...
+function fused_single_field_solve!(_, x, A::UniformScaling, b)
+    foreach(matrix_row_keys(keys(A))) do name
+        single_field_solve!(cache[name], x[name], A[name, name], b[name])
+    end
+end
+fused_single_field_solve!(cache, x, A::ColumnwiseBandMatrixField, b) =
+    fused_single_field_solve!(ClimaComms.device(axes(A)), cache, x, A, b)
+
+# TODO: fuse
+function fused_single_field_solve!(
```
dennisYatunin, line 93: Could we keep these as methods of `run_field_matrix_solver!` in `field_matrix_solver.jl`? This file should probably just have what's currently called `_single_field_solve!`; we can drop what's currently called `single_field_solve!` and move the GPU kernel logic into a fused method of `run_field_matrix_solver!`.

charleskawczynski: Yeah, agreed, I thought about starting a new file but thought I'd see what the perf looks like first since it's a somewhat simple backend change.

PR #1655, opened 2024-03-22
Test reapply some of 1011 with tests

charleskawczynski:
This PR applies some of #1011 with some tests to check the validity of the bug fixes. `RecursiveApply.rdiv((w⁺ ⊠ a⁺) ⊞ (w⁻ ⊠ a⁻), (w⁺ ⊞ w⁻))` has already been corrected in main, so I've skipped that change.


charleskawczynski:
I looked at this a bit more carefully, and I do believe that it's equivalent.

Also, I like `ri = rw = right_idx(space)` over `ri = rw = left_idx(space) + length(space) - 1`. So, let's merge this.

PR #1656, opened 2024-03-23
Update allocation tests

charleskawczynski:
This PR adds GPU coverage to some allocation tests.


charleskawczynski:
Ah, I think this is fixed in Julia 1.11..

charleskawczynski:
This will be fixed when we upgrade to julia 1.11

PR #1657, opened 2024-04-03
Test also on 1.9

Sbozzolo:
I am finding that something in ClimaCore fails for 1.9:

https://github.com/CliMA/ClimaDiagnostics.jl/actions/runs/8541780429/job/23401922394

We should at least have one runner on 1.9.


charleskawczynski:
I'm almost inclined to say that we should stop maintaining support for julia 1.9. There are a bunch of issues in 1.9 (e.g., printing fields), that make 1.9 really unfriendly to work with.

PR #1658, opened 2024-04-03
Dont use .style for FieldVectorStyle

charleskawczynski:
I think this should fix the failure in #1657.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
 ClimaCore.jl Release Notes
 ========================
 
+main
+-------
+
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Do you know why this is failing only in 1.9?

charleskawczynski:
> Do you know why this is failing only in 1.9?

`style` must be an internal field, that was added in 1.10. There was always an associated style type, but perhaps not a field.

Sbozzolo (approved):
Thank you!

Sbozzolo:
You can also add an entry in the NEWS if you feel like

PR #1660, opened 2024-04-04
Add TuplesOfNTuples dependency

charleskawczynski:
This PR adds TuplesOfNTuples.jl as a dependency, and is a peel off from #1653. This should help reduce rebase conflicts.


charleskawczynski:
These exact changes already passed (except for Aqua) in #1653, so I'm going to merge.

PR #1662, opened 2024-04-05
TVD slope limiters 

akshaysridhar:
Update flux correction stencils
Update column and deformation flow examples to use TVD limiters.

### DIFF
--- a/examples/column/tvd_advection.jl
+++ b/examples/column/tvd_advection.jl
@@ -1,0 +1,192 @@
+using Test
+using LinearAlgebra
+import ClimaComms
+ClimaComms.@import_required_backends
+using OrdinaryDiffEqSSPRK: ODEProblem, solve, SSPRK33
+using ClimaTimeSteppers
+
+import ClimaCore:
+    Fields,
+    Domains,
+    Topologies,
+    Meshes,
+    DataLayouts,
+    Operators,
+    Geometry,
+    Spaces
+
+
+# Advection Equation, with constant advective velocity (so advection form = flux form)
+# ∂_t y + w ∂_z y  = 0
+# the solution translates to the right at speed w,
+# so at time t, the solution is y(z - w * t)
+
+# visualization artifacts
+ENV["GKSwstype"] = "nul"
+using ClimaCorePlots, Plots
+Plots.GRBackend()
+dir = "tvd_advection"
+path = joinpath(@__DIR__, "output", dir)
+mkpath(path)
+
+
+function tendency!(yₜ, y, parameters, t)
+    (; w, Δt, limiter_method) = parameters
+    FT = Spaces.undertype(axes(y.q))
+    bcvel = pulse(-π, t, z₀, zₕ, z₁)
+    divf2c = Operators.DivergenceF2C(
+        bottom = Operators.SetValue(Geometry.WVector(FT(0))),
+        top = Operators.SetValue(Geometry.WVector(FT(0))),
+    )
+    upwind1 = Operators.UpwindBiasedProductC2F(
+        bottom = Operators.Extrapolate(),
+        top = Operators.Extrapolate(),
+    )
+    upwind3 = Operators.Upwind3rdOrderBiasedProductC2F(
+        bottom = Operators.ThirdOrderOneSided(),
+        top = Operators.ThirdOrderOneSided(),
+    )
+    FCTZalesak = Operators.FCTZalesak(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
+    )
+    TVDSlopeLimited = Operators.TVDLimitedFluxC2F(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
+        method = limiter_method,
+    )
+
+    If = Operators.InterpolateC2F()
+
+    if limiter_method == "Zalesak"
+        @. yₜ.q =
+            -divf2c(
+                upwind1(w, y.q) + FCTZalesak(
+                    upwind3(w, y.q) - upwind1(w, y.q),
+                    y.q / Δt,
+                    y.q / Δt - divf2c(upwind1(w, y.q)),
+                ),
+            )
+    else
+        Δfluxₕ = @. w * If(y.q)
+        Δfluxₗ = @. upwind1(w, y.q)
+        @. yₜ.q =
+            -divf2c(
+                upwind1(w, y.q) +
+                TVDSlopeLimited(upwind3(w, y.q) - upwind1(w, y.q), y.q / Δt, w),
+            )
+    end
+end
+
+# Define a pulse wave or square wave
+
+FT = Float64
+t₀ = FT(0.0)
+t₁ = FT(6)
+z₀ = FT(0.0)
+zₕ = FT(2π)
+z₁ = FT(1.0)
+speed = FT(-1.0)
+pulse(z, t, z₀, zₕ, z₁) = abs(z - speed * t) ≤ zₕ ? z₁ : z₀
+
+n = 2 .^ 8
+elemlist = 2 .^ [3, 4, 5, 6, 7, 8, 9, 10]
+Δt = FT(0.3) * (20π / n)
+@info "Timestep Δt[s]: $(Δt)"
+
+domain = Domains.IntervalDomain(
+    Geometry.ZPoint{FT}(-10π),
+    Geometry.ZPoint{FT}(10π);
+    boundary_names = (:bottom, :top),
+)
+
+stretch_fns = [Meshes.Uniform()]
+plot_string = ["uniform"]
+
+for (i, stretch_fn) in enumerate(stretch_fns)
+    limiter_methods = (
+        Operators.RZeroLimiter(),
+        Operators.RMaxLimiter(),
+        Operators.KorenLimiter(),
+        Operators.SuperbeeLimiter(),
+        Operators.MonotonizedCentralLimiter(),
+        "Zalesak",
+    )
+    for (j, limiter_method) in enumerate(limiter_methods)
+        @info (limiter_method, stretch_fn)
+        mesh = Meshes.IntervalMesh(domain, stretch_fn; nelems = n)
+        cent_space = Spaces.CenterFiniteDifferenceSpace(mesh)
+        face_space = Spaces.FaceFiniteDifferenceSpace(cent_space)
+        z = Fields.coordinate_field(cent_space).z
+        O = ones(FT, face_space)
+
+        # Initial condition
+        q_init = pulse.(z, 0.0, z₀, zₕ, z₁)
+        y = Fields.FieldVector(q = q_init)
+
+        # Unitary, constant advective velocity
+        w = Geometry.WVector.(speed .* O)
+
+        # Solve the ODE
+        parameters = (; w, Δt, limiter_method)
+        prob = ODEProblem(
+            ClimaODEFunction(; T_exp! = tendency!),
+            y,
+            (t₀, t₁),
+            parameters,
+        )
+        sol = solve(
+            prob,
+            ExplicitAlgorithm(SSP33ShuOsher()),
+            dt = Δt,
+            saveat = Δt,
+        )
+
+        q_final = sol.u[end].q
+        q_analytic = pulse.(z, t₁, z₀, zₕ, z₁)
+
+        err = norm(q_final .- q_analytic)
+        rel_mass_err = norm((sum(q_final) - sum(q_init)) / sum(q_init))
+
+        if j == 1
+            fig = Plots.plot(q_analytic; label = "Exact", color = :red)
+        end
+        linstyl = [:dash, :dot, :dashdot, :dashdotdot, :dash, :solid]
+        clrs = [:orange, :gray, :green, :maroon, :pink, :blue]
+        if limiter_method == "Zalesak"
+            fig = plot!(
+                q_final;
+                label = "Zalesak",
+                linestyle = linstyl[j],
+                color = clrs[j],
+                dpi = 400,
+                xlim = (-0.5, 1.1),
+                ylim = (-15, 10),
+            )
+        else
+            fig = plot!(
+                q_final;
+                label = "$(typeof(limiter_method))"[21:end],
+                linestyle = linstyl[j],
+                color = clrs[j],
+                dpi = 400,
+                xlim = (-0.5, 1.1),
+                ylim = (-15, 10),
+            )
+        end
+        fig = plot!(legend = :outerbottom, legendcolumns = 2)
+        if j == length(limiter_methods)
+            Plots.png(
+                fig,
+                joinpath(
+                    path,
+                    "SlopeLimitedFluxSolution_" *
+                    "$(typeof(limiter_method))"[21:end] *
+                    ".png",
+                ),
+            )
+        end
+        @test err ≤ 0.25
+        @test rel_mass_err ≤ 10eps()
+    end
+end
--- a/examples/column/vanleer_advection.jl
+++ b/examples/column/vanleer_advection.jl
@@ -1,0 +1,168 @@
+using Test
+using LinearAlgebra
+import ClimaComms
+ClimaComms.@import_required_backends
+using OrdinaryDiffEqSSPRK: ODEProblem, solve, SSPRK33
+using ClimaTimeSteppers
+
+import ClimaCore:
+    Fields,
+    Domains,
+    Topologies,
+    Meshes,
+    DataLayouts,
+    Operators,
+    Geometry,
+    Spaces
+
+
+# Advection Equation, with constant advective velocity (so advection form = flux form)
+# ∂_t y + w ∂_z y  = 0
+# the solution translates to the right at speed w,
+# so at time t, the solution is y(z - w * t)
+
+# visualization artifacts
+ENV["GKSwstype"] = "nul"
+using ClimaCorePlots, Plots
+Plots.GRBackend()
+dir = "vanleer_advection"
+path = joinpath(@__DIR__, "output", dir)
+mkpath(path)
+
+
+function tendency!(yₜ, y, parameters, t)
+    (; w, Δt, limiter_method) = parameters
+    FT = Spaces.undertype(axes(y.q))
+    bcvel = pulse(-π, t, z₀, zₕ, z₁)
+    divf2c = Operators.DivergenceF2C(
+        bottom = Operators.SetValue(Geometry.WVector(FT(0))),
+        top = Operators.SetValue(Geometry.WVector(FT(0))),
+    )
+    VanLeerMethod = Operators.LinVanLeerC2F(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
+        constraint = limiter_method,
+    )
+
+    If = Operators.InterpolateC2F()
+
+    @. yₜ.q = -divf2c(VanLeerMethod(w, y.q, Δt))
+end
+
+# Define a pulse wave or square wave
+
+FT = Float64
+t₀ = FT(0.0)
+t₁ = FT(6)
+z₀ = FT(0.0)
+zₕ = FT(2π)
+z₁ = FT(1.0)
+speed = FT(-1.0)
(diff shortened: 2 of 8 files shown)


.buildkite/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.5"
+julia_version = "1.11.0"
```
akshaysridhar, line 3: @charleskawczynski I'll revert this (local Julia is at 1.11.0)

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1794,185 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+@inline function compute_limiter_coeff(r, ::VanLeerLimiter)
+    return (r + abs(r)) / (1 + abs(r) + eps(eltype(r)))
+end
+
+# ??? Do we want to allow flux method types to be determined here? 
+struct TVDSlopeLimitedFlux{BCS} <: AdvectionOperator
+    bcs::BCS
+end
+
+TVDSlopeLimitedFlux(; kwargs...) = TVDSlopeLimitedFlux(NamedTuple(kwargs))
```
charleskawczynski, line 1875:
```suggestion
TVDSlopeLimitedFlux(; method, kwargs...) = TVDSlopeLimitedFlux((;method, kwargs...))
```

charleskawczynski: This way users cannot make this without the `method` keyword (which would later result in a non-existent key later)

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1794,185 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+(4) MinModLimiter
+(5) KorenLimiter
+(6) SuperbeeLimiter
+(7) MonotonizedCentralLimiter
+(8) VanLeerLimiter
+
+"""
+### Here we define abstract types and specific implementations 
+### TVD limiters (see, for instance, Durran's notes)
+abstract type AbstractTVDSlopeLimiter end
```
charleskawczynski, line 1826: Let's add this to the API, and move / delete the comments above so that the doc strict correctly attaches to `AbstractTVDSlopeLimiter`

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1794,185 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+struct RZeroLimiter <: AbstractTVDSlopeLimiter end
+struct RHalfLimiter <: AbstractTVDSlopeLimiter end
+struct RMaxLimiter <: AbstractTVDSlopeLimiter end
+struct MinModLimiter <: AbstractTVDSlopeLimiter end
+struct KorenLimiter <: AbstractTVDSlopeLimiter end
+struct SuperbeeLimiter <: AbstractTVDSlopeLimiter end
+struct MonotonizedCentralLimiter <: AbstractTVDSlopeLimiter end
+struct VanLeerLimiter <: AbstractTVDSlopeLimiter end
```
charleskawczynski, line 1834: Should we document these?

src/Operators/finitedifference.jl
akshaysridhar: Updated.

src/Operators/finitedifference.jl
akshaysridhar: Complete forms of these limiters are available in the Durran reference (these are common in literature so perhaps the reference is sufficient)

examples/column/tvd_advection.jl
```diff
@@ -0,0 +1,156 @@
...
+        rel_mass_err = norm((sum(q_final) - sum(q_init)) / sum(q_init))
+
+
+        plot(q_final)
+        Plots.png(
+            Plots.plot!(q_analytic, title = "$(typeof(limiter_method))"),
+            joinpath(
+                path,
+                "exact_and_computed_advected_square_wave_TVDSlopeLimitedFlux_" *
+                "$(typeof(limiter_method))_" *
```
charleskawczynski, line 148:
```suggestion
                "$(nameof(typeof(limiter_method)))_" *
```

charleskawczynski (commented):
A lot of these limiters don't seem to prevent negative values:

Screenshot 2024-11-19 at 3 50 48 PM

Is this behavior expected? Should we add more granular unit tests?

The RZeroLimiter seems to work well at removing negative values:

Screenshot 2024-11-19 at 3 51 45 PM

examples/column/tvd_advection.jl
charleskawczynski: This will help keep the names a bit shorter

akshaysridhar:
> A lot of these limiters don't seem to prevent negative values:
> Screenshot 2024-11-19 at 3 50 48 PM
> 
> Is this behavior expected? Should we add more granular unit tests?
> 
> The RZeroLimiter seems to work well at removing negative values:
> Screenshot 2024-11-19 at 3 51 45 PM

These are from a class of slope limiting fluxes which combine low-order (monotone) and high-order (not necessarily monotone) fluxes : the `RZero` here then corresponds to the low order (first order upwind, likely too dissipative to be useful in our simulations) option, and the `RMax` returns the `third order upwind` method (which is the current choice of the high-order flux in this method. (Zalesak, similarly, is a combination of such low / high order fluxes , with flux corrections through an "anti-diffusive" term)

examples/column/tvd_advection.jl
```diff
@@ -0,0 +1,156 @@
...
+mkpath(path)
+
+function linkfig(figpath, alt = "")
+    # buildkite-agent upload figpath
+    # link figure in logs if we are running on CI
+    if get(ENV, "BUILDKITE", "") == "true"
+        artifact_url = "artifact://$figpath"
+        print("\033]1338;url='$(artifact_url)';alt='$(alt)'\a\n")
+    end
+end
```
charleskawczynski, line 39:
```suggestion
```
This doesn't look like it's called

charleskawczynski:
> @charleskawczynski These are from a class of slope limiting fluxes which combine low-order (monotone) and high-order (not necessarily monotone) fluxes : the `RZero` here then corresponds to the low order (first order upwind, likely too dissipative to be useful in our simulations) option, and the `RMax` returns the `third order upwind` method (which is the current choice of the high-order flux in this method. `RHalf` equally weights the low / high order methods). (Zalesak, similarly, is a combination of such low / high order fluxes , with flux corrections through an "anti-diffusive" term). This PR introduces the alternative to flux-corrected transport through total-variation diminishing limiters.

Ok, could we maybe also plot what the solution looks like without a limiter? That way we can see how each one improves over the baseline.

charleskawczynski:
> Ok, could we maybe also plot what the solution looks like without a limiter? That way we can see how each one improves over the baseline.

Actually, maybe that is unstable? Perhaps this is fine? I just want to make sure that this is working, and I'd like to have some sort of test to confirm it. The plots look suspiciously ineffective.

szy21:
I'm not the best person to review this. @charleskawczynski @dennisYatunin Could any of you take a look at this PR ASAP? We hope to merge this before the end of tomorrow, thanks!

src/Operators/finitedifference.jl
```diff
@@ -1320,6 +1320,204 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+
+The diffusion implied by these methods is proportional to the local upwind CFL number.
+The `mismatch` Δ𝜙 = 0 returns the first-order upwind method. Special cases (discussed in Lin et al (1994)) include setting the 𝜙_min = 0 or 𝜙_max = saturation mixing ratio for water vapor
+are not considered here in favour of the generalized local extrema in equation (5a, 5b).
+"""
+abstract type LimiterConstraint end
+struct AlgebraicMean <: LimiterConstraint end
+struct PosDef <: LimiterConstraint end
+struct Mono4 <: LimiterConstraint end
+struct Mono5 <: LimiterConstraint end
```
dennisYatunin, line 1344: `Mono4` and `Mono5` aren't very clear names. Maybe something like `HarmonicMean` and `LocallyMonotonic` would be better? Also you can probably write out `PositiveDefinite` instead of `PosDef`; it's okay to be more verbose in these names.

src/Operators/finitedifference.jl
```diff
@@ -1320,6 +1320,204 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+    min𝜙 = min(a⁻, a⁰, a⁺)
+    max𝜙 = max(a⁻, a⁰, a⁺)
+    𝛼 = min(abs(Δ𝜙_avg), 2 * (a⁰ - min𝜙), 2 * (max𝜙 - a⁰))
+    Δ𝛼 = sign(Δ𝜙_avg) * 𝛼 * (1 - sign(v) * v * dt)
+end
+
+function compute_Δ𝛼_linvanleer(a⁻, a⁰, a⁺, v, dt, ::Mono4)
+    Δ𝜙_avg = ((a⁰ - a⁻) + (a⁺ - a⁰)) / 2
+    c = sign(v) * v * dt
+    if sign(a⁰ - a⁻) == sign(a⁺ - a⁰) && Δ𝜙_avg != eltype(v)(0)
```
dennisYatunin, line 1369: The conversion to `eltype(v)` on this line is unnecessary (it will be inserted automatically by the compiler) and makes the code a bit harder to read. The only places where we need such conversions are isolated return statements (e.g., `return eltype(v)(0)` a few lines down). You can remove all such conversions from conditional statements like this one.

src/Operators/finitedifference.jl
```diff
@@ -1320,6 +1320,204 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+    arg_space::AllCenterFiniteDifferenceSpace,
+    dt,
+) = velocity_space
+
+function compute_Δ𝛼_linvanleer(a⁻, a⁰, a⁺, v, dt, ::Mono5)
+    Δ𝜙_avg = ((a⁰ - a⁻) + (a⁺ - a⁰)) / 2
+    min𝜙 = min(a⁻, a⁰, a⁺)
+    max𝜙 = max(a⁻, a⁰, a⁺)
+    𝛼 = min(abs(Δ𝜙_avg), 2 * (a⁰ - min𝜙), 2 * (max𝜙 - a⁰))
+    Δ𝛼 = sign(Δ𝜙_avg) * 𝛼 * (1 - sign(v) * v * dt)
```
dennisYatunin, line 1363: Shouldn't all of this be done with `RecursiveApply`? The `slope_limited_product` product function seems to support `Tuple`/`NamedTuple` arguments, but the support seems to disappear in these `compute_Δ𝛼_linvanleer` methods.

src/Operators/finitedifference.jl
```diff
@@ -1320,6 +1320,204 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+        Δ𝛼 = compute_Δ𝛼_linvanleer(a⁻, a⁺, a⁺⁺, v, dt, method)
+        return v ⊠ (a⁺ ⊟ RecursiveApply.rdiv(Δ𝛼, 2))
+    end
+end
+
+stencil_interior_width(::LinVanLeerC2F, velocity, arg, dt) =
+    ((0, 0), (-half - 1, half + 1), (0, 0))
+
+Base.@propagate_inbounds function stencil_interior(
+    ℱ::LinVanLeerC2F,
```
dennisYatunin: I think we've decided to avoid script characters like `ℱ`. Maybe try sticking to ASCII, Greek, and super/subscripts to be consistent with the rest of our code.

src/Operators/finitedifference.jl
```diff
@@ -1320,6 +1320,204 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+[Lin1994](@cite), four limiter constraint options are provided
+- AlgebraicMean: Algebraic mean, this guarantees neither positivity nor monotonicity (eq 2)
+- PosDef: Positive-definite with implicit diffusion based on local stencil extrema (eq 3b, 3c, 5a, 5b)
+- Mono4: Monotonicity preserving harmonic mean, this implies a strong monotonicity constraint (eq 4)
+- Mono5: Monotonicity preserving, with extrema bounded by the edge cells in the stencil (eq 5)
+
+The diffusion implied by these methods is proportional to the local upwind CFL number.
+The `mismatch` Δ𝜙 = 0 returns the first-order upwind method. Special cases (discussed in Lin et al (1994)) include setting the 𝜙_min = 0 or 𝜙_max = saturation mixing ratio for water vapor
+are not considered here in favour of the generalized local extrema in equation (5a, 5b).
+"""
```
dennisYatunin, line 1339: Can this docstring be attached to `LinVanLeerC2F` instead of `LimiterConstraint`, to match the way we document other operators? The constraint (method?) should also probably be a field of the `LinVanLeerC2F` struct, rather than a separate object passed through the boundary conditions.

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1989,182 @@ Base.@propagate_inbounds function stencil_right_boundary(
     return Geometry.Contravariant3Vector(zero(eltype(eltype(A_field))))
 end
 
+"""
```
dennisYatunin, line 1992: Can this docstring also be attached to the relevant operator?

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1989,182 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+
+@inline function compute_limiter_coeff(r, ::SuperbeeLimiter)
+    return max(zero(eltype(r)), min(one(eltype(r)), r), min(2, r))
+end
+
+@inline function compute_limiter_coeff(r, ::MonotonizedCentralLimiter)
+    return max(zero(eltype(r)), min(2r, (1 + r) / 2, 2))
+end
+
+struct TVDSlopeLimitedFlux{BCS} <: AdvectionOperator
```
dennisYatunin, line 2053: This should probably be called something like `TVDLimitedFluxC2F` to match the other operator names.

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1989,182 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+    TVDSlopeLimitedFlux((; method, kwargs...))
+
+return_eltype(::TVDSlopeLimitedFlux, A, Φ, 𝓊) =
+    Geometry.Contravariant3Vector{eltype(eltype(A))}
+
+return_space(
+    ::TVDSlopeLimitedFlux,
+    A_space::AllFaceFiniteDifferenceSpace,
+    Φ_space::AllCenterFiniteDifferenceSpace,
+    𝓊_space::AllFaceFiniteDifferenceSpace,
```
dennisYatunin, line 2067: Could you use a regular `u` instead of `𝓊`? We generally want to stick with English, Greek, and superscript/subscript characters.

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1989,182 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+) = A_space
+
+function tvd_limited_flux(Aⱼ₋₁₂, Aⱼ₊₁₂, ϕⱼ₋₁, ϕⱼ, ϕⱼ₊₁, ϕⱼ₊₂, rⱼ₊₁₂, method)
```
dennisYatunin, line 2070: This function should also probably be compatible with `RecursiveApply`.

src/Operators/finitedifference.jl
```diff
@@ -1787,7 +1989,182 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
+
+Base.@propagate_inbounds function stencil_interior(
+    op::TVDSlopeLimitedFlux,
+    loc,
+    space,
+    idx,
+    hidx,
+    A_field,
+    Φ_field,
+    𝓊_field,
```
dennisYatunin, line 2090: We might want to modify this so that we can pass `dt` separately, as in the case of the VanLeer limiter. I think that would be somewhat clearer, but we can keep the arguments as they are now if it's more standard.

examples/column/vanleer_advection.jl
```diff
@@ -0,0 +1,158 @@
...
+        )
+
+        q_final = sol.u[end].q
+
+        if limiter_method != Operators.AlgebraicMean()
+            @info "Extrema with $(limiter_method): $(extrema(q_final))"
+            @assert maximum(q_final) <= FT(1)
+            @assert isapprox(maximum(q_final .- 1), FT(0), atol = FT(0.05))
+            @assert isapprox(minimum(q_final .- 0), FT(0), atol = FT(0.05))
+        end
```
dennisYatunin, line 121: This looks good! I'd expect a lower error than 5%, but maybe if `Δt` is big for this particular problem, and if there are several stages that each introduce a bit of error, then this result makes sense. Could you also add a similar monotonicity check for the `RZeroLimiter` in the TVD advection test, and any other limiters that should preserve monotonicity?

dennisYatunin: Does this error decrease if you lower the value of `Δt`? It would be good to have a unit test that demonstrates monotonicity preservation up to machine epsilon precision, if that's possible for a sufficiently small CFL number in this test case.

src/Operators/finitedifference.jl
akshaysridhar: Yes - currently this choice of args is also true in the `Zalesak` limiter - I can switch this in a future PR to assign `dt` as a separate argument.

src/Operators/finitedifference.jl
akshaysridhar: These names currently match the terminology in the reference, but I agree verbose names would be helpful - with a docstring reference to the paper terminology.

charleskawczynski:
Hi @akshaysridhar, just a heads up: I'm going to rebase and push some changes

src/Operators/finitedifference.jl
akshaysridhar: Fixed.

src/Operators/finitedifference.jl
akshaysridhar: Fixed.

akshaysridhar:
> Hi @akshaysridhar, just a heads up: I'm going to rebase and push some changes

Noted; I've added some commits addressed the unit test changes (I'll wait for the rebase update to push). Thanks @charleskawczynski

examples/column/vanleer_advection.jl
akshaysridhar: @dennisYatunin Updated with reduced cfl and info statements.

akshaysridhar:
> > Hi @akshaysridhar, just a heads up: I'm going to rebase and push some changes
> 
> Noted; I've added some commits addressing the unit test comments (I'll wait for the rebase update to push). Thanks @charleskawczynski

I've pushed these changes @charleskawczynski

src/Operators/finitedifference.jl
charleskawczynski: I discussed this with @dennisYatunin and @akshaysridhar. There are some ergonomic issues with using conditions and RecursiveApply for algorithms. RecursiveApply is nice in that it allows us to operate on NamedTuples, but conditions on NamedTuples becomes more complicated (and entangled with the algorithm). Furthermore, from a parallelization point of view, RecursiveApply couples variables (e.g., a given thread must read multiple variables, even if the algorithm can be performed in parallel across variables).

I think relaxing the use of RecursiveApply here is okay, until we determine if we really need support for this.

src/Operators/finitedifference.jl
charleskawczynski: See https://github.com/CliMA/ClimaCore.jl/pull/1662#discussion_r1893066699

src/Operators/finitedifference.jl
charleskawczynski: Noted, fixed the important ones.

src/Operators/finitedifference.jl
charleskawczynski: It's a bit duplicate, but I added a short stub so that users can navigate to the general documentation.

src/Operators/finitedifference.jl
charleskawczynski: For the sake of getting this PR merged, let's follow this up later.

charleskawczynski: Opened https://github.com/CliMA/ClimaCore.jl/issues/2107 to track.

charleskawczynski:
Alright, it looks like this worked. The only failure is unrelated, it seems like the fieldvector benchmarks have a 137 error. I'll look into that separately.

PR #1663, opened 2024-04-05
Define and use `auto_launch!`

charleskawczynski:
This PR was updated to only define and use `auto_launch!` with suggested threads and blocks. This will allow us to easily and experimentally optimize the number of threads and blocks kernel.

This PR is a no-op.

Next, we can leverages CUDA.jl's [`launch_configuration`](https://cuda.juliagpu.org/dev/lib/driver/#CUDA.launch_configuration) function, to auto-tune the number of blocks and threads, for most kernel launches. I've excluded those where we specify shmem (spectral element kernels), because it's not yet clear to me how we can use it there.

This will allow us to remove the hard-coded `_max_threads_cuda() = 256`.

I'm going to try this out in ClimaAtmos, and run the target pipeline with this.

### DIFF
--- a/ext/ClimaCoreCUDAExt.jl
+++ b/ext/ClimaCoreCUDAExt.jl
@@ -13,6 +13,7 @@ import ClimaCore.Utilities: half
 import ClimaCore.RecursiveApply:
     ⊠, ⊞, ⊟, radd, rmul, rsub, rdiv, rmap, rzero, rmin, rmax
 
+include(joinpath("cuda", "cuda_utils.jl"))
 include(joinpath("cuda", "data_layouts.jl"))
 include(joinpath("cuda", "fields.jl"))
 include(joinpath("cuda", "topologies_dss.jl"))
@@ -23,8 +24,8 @@ include(joinpath("cuda", "remapping_interpolate_array.jl"))
 include(joinpath("cuda", "limiters.jl"))
 include(joinpath("cuda", "operators_sem_shmem.jl"))
 include(joinpath("cuda", "operators_thomas_algorithm.jl"))
+include(joinpath("cuda", "matrix_fields_single_field_solve.jl"))
 include(joinpath("cuda", "matrix_fields_multiple_field_solve.jl"))
 include(joinpath("cuda", "operators_spectral_element.jl"))
-include(joinpath("cuda", "matrix_fields_single_field_solve.jl"))
 
 end
--- a/ext/cuda/cuda_utils.jl
+++ b/ext/cuda/cuda_utils.jl
@@ -1,0 +1,94 @@
+import CUDA
+import ClimaCore.Fields
+import ClimaCore.DataLayouts
+
+get_n_items(field::Fields.Field) =
+    prod(size(parent(Fields.field_values(field))))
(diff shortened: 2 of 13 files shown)


charleskawczynski:
This pattern seems to necessitate linearizing all of the indexing (we are currently specifying threads whose tuple elements span the DataLayout's backed array). A pattern that I was aiming for is:

```julia
import CUDA

function _configure_threadblock(max_threads, nitems)
    nthreads = min(max_threads, nitems)
    nblocks = cld(nitems, nthreads)
    return (nthreads, nblocks)
end

function auto_launch!(f!::F!, args, N) where {F!}
    kernel = CUDA.@cuda launch = false f!(args...)
    config = CUDA.launch_configuration(kernel.fun)
    threads = min(N, config.threads)
    blocks = cld(N, threads)
    @cuda always_inline = true threads = threads blocks = blocks f!(args...)
end
using BenchmarkTools

function run1(data, n)
    nthreads, nblocks = _configure_threadblock(256, prod(n))
    CUDA.@cuda threads = nthreads blocks = nblocks kernel!(data, n)
end

function run2(data, n)
    auto_launch!(kernel!, (data,n), length(data))
end

function main()
    n = (20, 5, 5, 3, 1000)
    data = CUDA.rand(n...)
    println("My pattern")
    nthreads, nblocks = _configure_threadblock(256, prod(n))
    @show nblocks, nthreads
    trial = BenchmarkTools.@benchmark CUDA.@sync run1($data, $n)
    show(stdout, MIME("text/plain"), trial)
    println()

    kernel = CUDA.@cuda launch = false kernel!(data, n)
    config = CUDA.launch_configuration(kernel.fun)
    @show config

    println("CUDA.jl's launch_configuration")
    trial = BenchmarkTools.@benchmark CUDA.@sync run2($data, $n)
    show(stdout, MIME("text/plain"), trial)
    println()
end
function get_indexes(n::Tuple)
    tidx = (CUDA.blockIdx().x - 1) * CUDA.blockDim().x + CUDA.threadIdx().x
    inds = if 1 ≤ tidx ≤ prod(n)
        CartesianIndices(map(x->Base.OneTo(x), n))[tidx].I
    else
        ntuple(x-> -1, length(n))
    end
    return inds
end
valid_range(inds, n::Tuple) = all(i -> 1 ≤ inds[i] ≤ n[i], 1:length(n))

function valid_range(n::Tuple)
    inds = get_indexes(n)
    return all(i -> 1 ≤ inds[i] ≤ n[i], 1:length(n))
end
function kernel!(data, n)
    inds = get_indexes(n)
    if valid_range(inds, n)
        data[inds...] += 1.0 # do work
    end
    return nothing
end

main()
```

@vchuravy, thoughts on this PR?

Sbozzolo:
This looks like a very neat feature. Maybe we can start with the most commonly launched kernels (and "manually" linearize the index there)?

sriharshakandala:
Generally speaking, `_max_threads_cuda() = 256` is a fairly reasonable number in most cases. CUDA automatically schedules multiple `threadblocks` on the same SM if the SM is under utilized and there are no resource constraints (shmem, registers, etc. used by the kernel). It would be good to look at performance comparisons at our target resolutions between these implementations!

PR #1665, opened 2024-04-08
Add broken GPU test for implicit prognostic edmf

charleskawczynski:
This PR adds a broken GPU test for `ᶠgradᵥ_matrix ⋅ BidiagonalMatrixRow`, which is used in implicit prognostic edmf. This test works on the CPU and fails on the GPU. I'm assuming that the GPUCompiler cannot prune the else branch in `eltype(arg) <: BandMatrixRow` of `multiply_matrix_at_index`.


charleskawczynski:
This has gotten stale

PR #1672, opened 2024-04-12
Parallelize matrix field broadcasting tests

charleskawczynski:
Supersedes #1616. This should hopefully shine light on what's going on.


charleskawczynski:
@dennisYatunin, I think I may have run into this in the past: move variables or a function definition from inside `@testset` to outside, and JET gives different results. IIRC, function definitions outside of `@testset` more closely match what you see in applications.

Maybe I'm missing something, but it may be that these passing JET tests are actually a false positive pass because too many things are inside `@testset`. There have been issues with JET not detecting issues (https://github.com/aviatesk/JET.jl/issues/608).

I'll need to dig into this more to confirm.

charleskawczynski:
Ahhh, the last build is not reaching the job that actually fails, so these inference tests are never reached. They're actually just failing on main. I'll make the 5 failing jobs soft fail, so that we at least hook in the tests that are being reached.

charleskawczynski:
@dennisYatunin, I'm going to move forward with merging this, please let me know if you want to follow up with any changes.

PR #1674, opened 2024-04-15
Fix ClimaTimesteppers warning in example

charleskawczynski:
This PR fixes a warning we're seeing in ClimaTimesteppers ([here](https://buildkite.com/clima/climacore-ci/builds/3366#018ee1e1-73ff-4bc0-b970-a60ba427387d)):

```
┌ Warning: Both T_exp! and T_lim! are not nothing, please use T_exp_T_lim!
│ instead.
└ @ ClimaTimeSteppers /central/scratch/esm/slurm-buildkite/shared_depot/packages/ClimaTimeSteppers/OK4Dd/src/functions.jl:33
```

### DIFF
--- a/examples/hybrid/sphere/deformation_flow.jl
+++ b/examples/hybrid/sphere/deformation_flow.jl
@@ -184,6 +184,11 @@ function vertical_tendency!(Yₜ, Y, cache, t)
     end
 end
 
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Not sure what's going on here, and not sure this is worth digging into.

PR #1675, opened 2024-04-15
Fix failed inference for mat field broadcasting

charleskawczynski:
This PR uses UnrolledUtilities to fix failed inference in some matrix field broadcasting tests.


dennisYatunin (approved):
It's actually a bit surprising that we need to unroll here, since matrix field entries should have uniform types, and there are definitely fewer than 32 entries. Maybe this is a case of the deep recursion heuristic? If so, disabling the recursion limits on `multiply_matrix_at_index` and some related functions could help with the remaining inference failures.

charleskawczynski:
> Maybe this is a case of the deep recursion heuristic

I think so, #1678 seems to fix one case at least. We may need a hybrid solution. I'm going to tackle this incrementally

charleskawczynski:
Superseded by #1683

PR #1684, opened 2024-04-18
fixes for land implicit solver

juliasloan25:
A couple of fixes are needed for GPU compatibility of the land implicit solver using the updated interface: https://github.com/CliMA/ClimaLand.jl/pull/542

passing build in ClimaAtmos using this branch: https://buildkite.com/clima/climaatmos-ci/builds/18472#_

## Content
- [x] extend `adapt_structure` for all operator and boundary condition types
- [x] use `unrolled_map` in `multiply_matrix_at_index` to avoid the recursive inference limit when compiling nested matrix operations

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -634,7 +634,6 @@ steps:
       - label: "Unit: matrix multiplication recursion example (GPU)"
         key: gpu_matrix_multiplication_recursion
         command: "julia --color=yes --project=.buildkite test/MatrixFields/matrix_multiplication_recursion.jl"
-        soft_fail: true
         agents:
           slurm_gpus: 1
 
@@ -862,6 +861,7 @@ steps:
       - label: "Unit: matrix field broadcasting (GPU)"
         key: unit_matrix_field_broadcasting_gpu_non_scalar_3
         command: "julia --color=yes --check-bounds=yes --project=.buildkite test/MatrixFields/matrix_fields_broadcasting/test_non_scalar_3.jl"
+        soft_fail: true
         agents:
           slurm_gpus: 1
           slurm_mem: 10GB
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -1,4 +1,4 @@
-import ..Utilities: PlusHalf, half
+import ..Utilities: PlusHalf, half, UnrolledFunctions
 
 const AllFiniteDifferenceSpace =
     Union{Spaces.FiniteDifferenceSpace, Spaces.ExtrudedFiniteDifferenceSpace}
@@ -239,12 +239,6 @@ Adapt.adapt_structure(to, sbc::StencilBroadcasted{Style}) where {Style} =
         Adapt.adapt(to, sbc.args),
         Adapt.adapt(to, sbc.axes),
     )
-
-function Adapt.adapt_structure(to, op::FiniteDifferenceOperator)
-    op
-end
-
-
 
 function Base.Broadcast.instantiate(sbc::StencilBroadcasted)
     op = sbc.op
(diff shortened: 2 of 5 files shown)


src/Operators/finitedifference.jl
```diff
@@ -2612,19 +2606,30 @@ Base.@propagate_inbounds function stencil_right_boundary(
...
-
-function Adapt.adapt_structure(to, bc::SetValue)
-    SetValue(Adapt.adapt_structure(to, bc.val))
+# Extend `adapt_structure` for all boundary conditions containing a `val` field.
+function Adapt.adapt_structure(to, bc::AbstractBoundaryCondition)
+    if hasfield(typeof(bc), :val)
+        return typeof(bc).name.wrapper(Adapt.adapt_structure(to, bc.val))
+    else
+        return bc
+    end
```
charleskawczynski, line 2615: I think `name.wrapper` is a Julia internal, and it's probably best that we not use it. I know that we already use it in `replace_basetype`, but I don't think we really need it here.

Maybe instead we could write:
```julia
for bct in subtypes(AbstractBoundaryCondition)
    @eval adapted_bc(val, ::$(bct)) = $bct(val)
end
# Extend `adapt_structure` for all boundary conditions containing a `val` field.
function Adapt.adapt_structure(to, bc::AbstractBoundaryCondition)
    if hasfield(typeof(bc), :val)
        return adapted_bc(Adapt.adapt_structure(to, bc.val), bc)
    else
        return bc
    end
end
```

We can use the same pattern below

juliasloan25: How is ` @eval $bct(val)` etc different from `typeof(bc)(Adapt.adapt_structure(to, bc.val)`?

charleskawczynski: It's equivalent to `typeof(bc).name.wrapper` since `subtypes(AbstractBoundaryCondition)` yield UnionAll types, which do not have type information:

```julia
abstract type AFoo end

struct Foo{T} <: AFoo
    val::T
end
struct Bar{T} <: AFoo
    val::T
end

for BCT in subtypes(AFoo)
    @show BCT
    @eval make_new(val, ::$BCT) = $BCT(val)
end

make_new(1, Foo(2.0))
```
yields:
```julia
julia> for BCT in subtypes(AFoo)
           @show BCT
           @eval make_new(val, ::$BCT) = $BCT(val)
       end
BCT = Bar
BCT = Foo

julia> make_new(1, Foo(2.0))
Foo{Int64}(1)
```

dennisYatunin: I don't think the `eval` pattern is a good idea for the BCs and operators, for two reasons:
- Some operators are sub-sub-types of `FiniteDifferenceOperator`, so we'd need to implement some sort of `recursive_subtypes` function.
- As the number of subtypes increases, so will the method table size. This will hurt compilation time and dynamic dispatch.

dennisYatunin: Also I think the `@eval` version is a bit harder to read. If we want to make the original version a bit clearer, we can move the `.name.wrapper` functionality into a utility function (`unionall_type(::Type{T}) where {T} = T.name.wrapper`) with clear documentation.

charleskawczynski: > Some operators are sub-sub-types of FiniteDifferenceOperator, so we'd need to implement some sort of recursive_subtypes function.

That's a fair point. But, if that's the case, then wouldn't we need to update this code given that we check `hasfield(bc, ::val)`? (this won't correctly adapt apply `bc.op.val` if it's nested). I suppose we could dispatch on the more nested composed definitions of `FiniteDifferenceOperator`

> As the number of subtypes increases, so will the method table size. This will hurt compilation time and dynamic dispatch.

I don't think that this is that big of a problem:

```julia
12-element Vector{Any}:
 ClimaCore.MatrixFields.BottomRightMatrixCorner
 ClimaCore.MatrixFields.TopLeftMatrixCorner
 ClimaCore.Operators.Extrapolate
 ClimaCore.Operators.FirstOrderOneSided
 ClimaCore.Operators.LeftStencilBoundary
 ClimaCore.Operators.NullBoundaryCondition
 ClimaCore.Operators.RightStencilBoundary
 ClimaCore.Operators.SetCurl
 ClimaCore.Operators.SetDivergence
 ClimaCore.Operators.SetGradient
 ClimaCore.Operators.SetValue
 ClimaCore.Operators.ThirdOrderOneSided
```

Regarding readability, I disagree (this implementation doesn't require looking into Julia Base). That said, I'll compromise with moving `unionall_type` out.

I worry about future compatibility, so I'd like to get a second opinion and ask Julia folks about using `.name.wrapper`

dennisYatunin: The `FiniteDifferenceOperator` method checks `hasfield(op, :bcs)`, which should work for all nested subtypes. But yeah, the effect on compilation time here is probably negligible; I was just thinking it could become an issue in the future if we use the `eval` pattern a lot.

charleskawczynski: > The `FiniteDifferenceOperator` method checks `hasfield(op, :bcs)`, which should work for all nested subtypes.

Yes, but shouldn't we, for example, define
```julia
Adapt.adapt_structure(to, op::FDOperatorMatrix) = Adapt.adapt_structure(to, op.op)
```
?

dennisYatunin: Ah, right, we definitely need that as well.

charleskawczynski: cc @juliasloan25

charleskawczynski: I'm going to resolve conversation and approve. Before merging

 - I'm fine with reverting to a documented `unionall_type(::Type{T}) where {T} = T.name.wrapper`, but we should get some feedback from the julia community in terms of whether this is sustainable (it's okay to merge before getting feedback).
 - Let's make sure to define `Adapt.adapt_structure(to, op::FDOperatorMatrix) = Adapt.adapt_structure(to, op.op)`
 - Let's test this branch in ClimaAtmos

charleskawczynski (approved):
One item to resolve, but LGTM after that

src/Operators/finitedifference.jl
juliasloan25: Thank you both for your feedback! Why do we have to handle `FDOperatorMatrix` specially? It looks like it's a subtype of `FiniteDifferenceOperator` but also contains an `op <: FiniteDifferenceOperator`. Is it the only operator with that structure?

dennisYatunin: Each operator matrix is implemented as a wrapper for the original operator, and it needs to account for the operator's boundary conditions. And yes, this is the only place where we have such an operator-within-operator structure.

.buildkite/pipeline.yml
```diff
@@ -629,12 +629,11 @@ steps:
 
       - label: "Unit: matrix multiplication recursion example (CPU)"
         key: cpu_matrix_multiplication_recursion
-        command: "julia --color=yes --check-bounds=yes --project=test test/MatrixFields/matrix_multiplication_recursion.jl"
+        command: "julia --color=yes --check-bounds=yes --project=.buildkite test/MatrixFields/matrix_multiplication_recursion.jl"
```
charleskawczynski, line 632: Ah, sorry, I must have missed this environment when merging

juliasloan25: No problem!

charleskawczynski:
https://buildkite.com/clima/climacore-ci/builds/3525 already passed, so I'll just go ahead and manually merge

PR #1685, opened 2024-04-19
Update gpu prog edmf repro

charleskawczynski:
This PR updates the inference failure reproducer. Perhaps fortunately, this kernel doesn't even compile on the cpu (and it never has). So I'll aim to fix that first.


charleskawczynski:
I've updated the test so that it demonstrates what works and what does not. Currently, to our best knowledge, we're hitting a compiler complexity heuristic, that results in failed inference with:

```julia
    @. ᶠtridiagonal_matrix_c3 =
        -(ᶠgradᵥ_matrix()) ⋅ ifelse(
            ᶜu₃ʲ.components.data.:1 > 0,
            convert(BidiagonalMatrixRow{FT}, ᶜleft_bias_matrix()),
            convert(BidiagonalMatrixRow{FT}, ᶜright_bias_matrix()),
        )
```
However, we can decompose this into simpler broadcast expressions by hoisting out the parts:

```julia
    @. bdmr_l = convert(BidiagonalMatrixRow{FT}, ᶜleft_bias_matrix())
    @. bdmr_r = convert(BidiagonalMatrixRow{FT}, ᶜright_bias_matrix())
    @. bdmr = ifelse(ᶜu₃ʲ.components.data.:1 > 0, bdmr_l, bdmr_r)
    @. ᶠtridiagonal_matrix_c3 = -(ᶠgradᵥ_matrix()) ⋅ bdmr
```

While this works, it moves us in the opposite direction for performance: we've increased the number of cuda kernels, and reads/writes. We need to, at some point do one of the following:
 - Fix inference on our side (if possible)
 - Figure out a way to adjust `Core.Compiler.InferenceParams`, so that the compiler can successfully infer these complex kernels
 - Wait to see if /hope that newer versions of Julia fix this

Issue #1667, opened 2024-04-10
Move support for CUDA to extension


Sbozzolo:
It would be nice for ClimaCore not to depend directly on CUDA.

Right now, packages that depend on ClimaCore have to download and instantiate CUDA even if they are not able to use it (e.g., on Github Action runners). The CUDA runtime is not exactly tiny, which adds to the initialization costs. 

Moving to an extension would also set a precedent for supporting different backends. 

(In many ways, this seems to be precisely the use case extensions were designed for.)

PR #1687, opened 2024-04-22
Make CUDA an extension

charleskawczynski:
This PR makes CUDA an extension.

This should allow us to more easily test on newer versions, if cuda is a bit late to catching up. Also, it should help with code loading times when we're only running on the cpu.

Closes: #1667


charleskawczynski:
A couple quarks I applied are, so that we can overload them with cuda:

 - `cuda_synchronize(device::ClimaComms.AbstractDevice; kwargs...) = nothing`
 - `allow_scalar_func(::ClimaComms.AbstractDevice, f, args) = f(args...)` & `allow_scalar_func(::ClimaComms.CUDADevice, f, args) = CUDA.@allowscalar f(args...)`
 - `is_CuArray_type(::Type{T}) where {T <: CUDA.CuArray} = true`

I had to move a few remapping interpolate field functions, so that we can dispatch between device types.

PR #1340, opened 2023-06-20
New runtests sketch

charleskawczynski:
This is a proposal for the new `test/runtests.jl`.

Right now, there is a mismatch between what is run in buildkite vs `test/runtests.jl` vs `test_runtests_expensive.jl`. The latter two are run on GHA, but they don't run with the same environment variables / or CL arguments (xref: https://github.com/CliMA/ClimaCore.jl/pull/1321#issuecomment-1593728957).

This PR:
 - Defines a few helper structs in the test directory
 - Collects all unit tests from the buildkite yaml
 - Tabulates them
 - Runs them with the same environment variables / CL arguments (we can modify these appropriately if we, e.g.,  don't have the same resources)

Here's an output of what the [table](https://gist.github.com/charleskawczynski/4f3ce7590fb236a2e67d6ef50160dee3) looks like (which is printed before running tests in `test/runtests.jl`, for transparency). Since all of the tests and configs are collected into a list, we could even use Distributed.jl and `pmap` over the tests (perhaps the non-distributed tests, at least).

Remaining TODO in this PR, if people like this direction:
 - Sketch out how `test_job(:: BuildkiteJob)` should work

PR #1688, opened 2024-04-22
Fix broken tests on GPU

Sbozzolo:
Running `runtests.jl` on a GPU node leads to series of failures to the mismatch between array types. This PR fixes them.


charleskawczynski:
Running `runtests.jl` on buildkite is redundant -- our tests on buildkite basically run everything in `runtests.jl` in parallel to cut down on CI time. Also, `runtests.jl` runs many (but not all) tests on buildkite (see https://github.com/CliMA/ClimaCore.jl/issues/1290 and https://github.com/CliMA/ClimaCore.jl/pull/1340)

juliasloan25:
When this PR was opened we only had 2 unit tests running on GPU in buildkite, but now we have 88. Since we individually run each test we want to check on GPU, it would be pretty repetitive to also run all tests on GPU, so I'll close this PR

PR #1693, opened 2024-04-24
3 component dss transform

akshaysridhar:
Currently, the dss_transform function treats the 3rd component of a 
Covariant123Vector datatype as a scalar. This PR updates the metric index getter methods and the transform functions to apply the 3 component vector dss. 

- [x] Update tests
- [x] Update docstrings

### DIFF
--- a/src/Topologies/dss.jl
+++ b/src/Topologies/dss.jl
@@ -29,6 +29,10 @@ struct DSSBuffer{S, G, D, A, B, VI}
     "field id for all contravariant12vector fields stored in the `data` array"
     contravariant12fidx::VI
     "internal local elements (lidx)"
+    covariant123fidx::VI
+    "field id for all contravariant123vector fields stored in the `data` array"
+    contravariant123fidx::VI
+    "internal local elements (lidx)"
     internal_elems::VI
     "local elements (lidx) located on process boundary"
     perimeter_elems::VI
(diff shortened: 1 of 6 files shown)


src/Topologies/dss.jl
```diff
@@ -134,13 +134,13 @@ function create_dss_buffer(
                     @show supportedvectortypes
                 end
                 @assert fieldtype <: supportedvectortypes
-                if fieldtype <: Geometry.Covariant12Vector
```
sriharshakandala, line 137: Don't we still need to support DSS for `Covariant12Vector`?

akshaysridhar: Yes; I'll update this to retain the `Covariant12Vector` support

akshaysridhar: Added tests in the `terrain_warp` test file + cleaned up the stranded `return nothing` statement.

akshaysridhar:
cc @sriharshakandala there's an issue when running a GPU example (e.g. sphere baroclinic wave rhoe) with the updates here when tracking this branch in ClimaAtmos; the CPU cases run without issues.

akshaysridhar:
@dennisYatunin There are some soft-fails in `MatrixFields` test sets on buildkite - can you please verify that this PR doesn't affect behaviour of those components ?

dennisYatunin:
> @dennisYatunin There are some soft-fails in `MatrixFields` test sets on buildkite - can you please verify that this PR doesn't affect behaviour of those components ?

Those are compilation-intensive unit tests that have been soft-failing for a while. The buildkite run looks good to me.

PR #1696, opened 2024-04-29
Update environments

charleskawczynski:
This PR combines the environments into a buildkite env. Closes #1692.

PR #1697, opened 2024-04-29
Update ClimaComms

charleskawczynski:
This PR updates us to the latest ClimaComms.


Sbozzolo:
It doesn't matter too much, but the Project.toml for the test environment should increase the compat for ClimaComms to 0.6

charleskawczynski:
I think we actually need to delete `test/Project.toml` (I forgot to delete it in #1696)

PR #1698, opened 2024-04-29
Add back depot

Sbozzolo:
This got removed somewhere, but there's no strong reason not to have it to cut down instantiation time.


charleskawczynski (approved):
I think Simon preferred CI to be stateless. I don't have strong feelings either way. It's kind of nice when instantiation takes less than a minute

Sbozzolo:
> I think Simon preferred CI to be stateless. I don't have strong feelings either way. It's kind of nice when instantiation takes less than a minute

I agree that it would be nice (and often depots lead to failures), but I think that there's a lot of time and frustration saved in a faster turnaround time. Init will soon be ~15 minutes, and it could be reduced to ~2

PR #1699, opened 2024-04-29
add matrix multiplication recursion reproducer

juliasloan25:
add a reproducer for #1684 
trying to recreate failure from land implicit solver

- without extending `adapt_structure` for all operators and BCs, we get an error that something is not isbits because the space is not removed as an operator is applied ([here](https://buildkite.com/clima/climacore-ci/builds/3497#_))
- after adding in the `adapt_structure` methods from #1684 for testing, we see this error:
```julia
ERROR: LoadError: InvalidIRError: compiling MethodInstance for ClimaCoreCUDAExt.copyto_stencil_kernel!(::ClimaCore.Fields.Field{…}, ::ClimaCore.Operators.StencilBroadcasted{…}, ::ClimaCore.Spaces.FiniteDifferenceSpace{…}, ::NTuple{…}, ::Int64, ::Int64, ::Int64) resulted in invalid LLVM IR
Reason: unsupported call to an unknown function (call to julia.new_gc_frame)
Reason: unsupported call to an unknown function (call to julia.push_gc_frame)
Reason: unsupported call to an unknown function (call to julia.get_gc_frame_slot)
Reason: unsupported dynamic function invocation (call to map(f, t::Tuple{Any, Any}) @ Base tuple.jl:292)
Stacktrace:
 [1] multiply_matrix_at_index
   @ ~/ClimaCore.jl/src/MatrixFields/matrix_multiplication.jl:378
 [2] stencil_left_boundary
   @ ~/ClimaCore.jl/src/MatrixFields/matrix_multiplication.jl:449
 [3] getidx
   @ ~/ClimaCore.jl/src/Operators/finitedifference.jl:3059
 [4] copyto_stencil_kernel!
   @ ~/ClimaCore.jl/ext/cuda/operators_finite_difference.jl:67
```

This is the error we aim to fix in #1684 , so the reproducer is correct

### DIFF
--- a/test/MatrixFields/matrix_multiplication_recursion.jl
+++ b/test/MatrixFields/matrix_multiplication_recursion.jl
@@ -1,0 +1,83 @@
+import ClimaCore
+if !(@isdefined(TU))
+    include(
+        joinpath(
+            pkgdir(ClimaCore),
+            "test",
+            "TestUtilities",
+            "TestUtilities.jl",
+        ),
+    )
+end
+import .TestUtilities as TU
+import ClimaCore: Spaces, Geometry, Operators, Fields, MatrixFields
+import ClimaCore.MatrixFields: ⋅
+
+# Set up operators
+const FT = Float64
+# Create divergence operator
+const divf2c_op = Operators.DivergenceF2C()
+const divf2c_matrix = MatrixFields.operator_matrix(divf2c_op)
+# Create gradient operator, and set gradient at boundaries to 0
+const gradc2f_op = Operators.GradientC2F(
+    top = Operators.SetGradient(Geometry.WVector(FT(0))),
+    bottom = Operators.SetGradient(Geometry.WVector(FT(0))),
+)
+const gradc2f_matrix = MatrixFields.operator_matrix(gradc2f_op)
+# Create interpolation operator
+const interpc2f_op = Operators.InterpolateC2F(
+    bottom = Operators.Extrapolate(),
+    top = Operators.Extrapolate(),
+)
+
+# Construct 3D space (column)
+boundary_names = (:bottom, :top)
+zlim = (FT(-1.5), FT(0))
+column = ClimaCore.Domains.IntervalDomain(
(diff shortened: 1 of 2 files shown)


test/MatrixFields/matrix_multiplication_recursion.jl
```diff
@@ -0,0 +1,79 @@
...
+    )
+end
+import .TestUtilities as TU
+import ClimaCore: Spaces, Geometry, Operators, Fields, MatrixFields
+import ClimaCore.MatrixFields: ⋅
+
+# Set up operators
+FT = Float64
+# Create divergence operator
+divf2c_op = Operators.DivergenceF2C()
```
charleskawczynski, line 19:
```suggestion
const divf2c_op = Operators.DivergenceF2C()
```
or use func

test/MatrixFields/matrix_multiplication_recursion.jl
```diff
@@ -0,0 +1,79 @@
...
+#dthetaresdtheta :: MatrixFields.TridiagonalMatrixRow{FT}-valued Field
+K = Fields.zeros(subsurface_space)
+dψdϑ_res = Fields.zeros(subsurface_space) .+ FT(115.901)
+# try value instead of field
+
+tridiag_type = MatrixFields.TridiagonalMatrixRow{FT}
+dest_field = Fields.Field(tridiag_type, subsurface_space)
+fill!(parent(dest_field), NaN)
+
+@. dest_field =
```
charleskawczynski, line 72: or, e.g.,
```julia
function foo!(X)
    (; K, ....) = X
    @. dest_field =
        divf2c_matrix() ⋅ (
            MatrixFields.DiagonalMatrixRow(interpc2f_op(-K)) ⋅ gradc2f_matrix() ⋅
            MatrixFields.DiagonalMatrixRow(dψdϑ_res) +
            MatrixFields.LowerDiagonalMatrixRow(
                topBC_op(Geometry.Covariant3Vector(zero(interpc2f_op(K)))),
            )
        )
end
```

PR #1709, opened 2024-05-02
Remove test Project.toml

charleskawczynski:
Missed in #1696


Sbozzolo:
I don't think we should remove the Project, we have additional deps there (e.g., Tests, Safetests). We could define them as extras in the main Project.toml

charleskawczynski:
> I don't think we should remove the Project, we have additional deps there (e.g., Tests, Safetests). We could define them as extras in the main Project.toml

Ah, I thought I did that in #1696, I must have peeled that off, and that's why I didn't remove the test/Project.toml.

Sbozzolo (approved):

PR #1712, opened 2024-05-02
Always inline in multiple_field_solve

### LINKED ISSUE #1694: Improve performance of `multiple_field_solve_kernel`
Right now, it's launched without `always_inline`.

We should see if we can fix the inference failure when we use `always_inline`, this very well may be the issue.

charleskawczynski:
Closes #1694.


charleskawczynski:
This didn't work, and nothing I tried with UnrolledUtilities worked either.

PR #1720, opened 2024-05-08
Add to cuda mapreduce test

charleskawczynski:
This PR updates and adds to the cuda mapreduce test


charleskawczynski:
We can merge this (I'll rebase soon), but I do think that we need to widen the tests to multiple field variables, thats where there's been confusion.

PR #1728, opened 2024-05-14
Revert multiple field solve

charleskawczynski:
This PR reverts the multiple field solve, to compare with the sequential single field solves, as the multiple field solve may have an issue with branch divergence (threads within the same warp having different instructions).


charleskawczynski:
There may still be a way to avoid branch divergence, so long all threads in a warp have similar instructions. Looking into this now..

charleskawczynski:
Superseded by #1732.

PR #1731, opened 2024-05-15
Reduce global memory reads

charleskawczynski:
This PR reduces memory reads in our `band_matrix_solve!`, by reusing register data from previous loop iterations.


charleskawczynski:
I tried this out on the prognostic edmf on the A100, and `ldiv!` went from 291.34 ms to 271.5 ms (93% of what it used to cost).

PR #1736, opened 2024-05-20
Covariant derivatives on GPU

akshaysridhar:
Extends the `spectralelement` operators on `gpu` to work identically compared with their cpu implementations. Towards Christoffel symbol computations and consistent tensor divergence calculations. 

Removes two `broken` tags against [cuda spectral element op tests](https://buildkite.com/clima/climacore-ci/builds/4106#01910f18-7c8a-4600-89df-5472c1756cd9/205-245) following additions to `ext/cuda/operators_spectral_element.jl`.

### DIFF
--- a/ext/cuda/operators_spectral_element.jl
+++ b/ext/cuda/operators_spectral_element.jl
@@ -1,7 +1,6 @@
 import ClimaCore: Spaces, Quadratures, Topologies
 import ClimaCore: Operators, Geometry, Quadratures, RecursiveApply
 import ClimaComms
-using CUDA: @cuda
 using CUDA
 import ClimaCore.Operators: AbstractSpectralStyle, strip_space
 import ClimaCore.Operators: SpectralBroadcasted, set_node!, get_node
@@ -256,7 +255,7 @@ end
 
 Base.@propagate_inbounds function operator_evaluate(
     op::Gradient{(1, 2)},
-    (input,),
+    input,
     space,
     ij,
     slabidx,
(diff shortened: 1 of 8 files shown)


ext/cuda/operators_spectral_element.jl
```diff
@@ -269,14 +268,53 @@ Base.@propagate_inbounds function operator_evaluate(
...
+        )
+    else
+        v₁, v₂, v₃ = input
+        ∂f₁∂ξ₁ = D[i, 1] ⊠ v₁[1, j, vt]
+        ∂f₁∂ξ₂ = D[j, 1] ⊠ v₁[i, 1, vt]
+        ∂f₂∂ξ₁ = D[i, 1] ⊠ v₂[1, j, vt]
+        ∂f₂∂ξ₂ = D[j, 1] ⊠ v₂[i, 1, vt]
+        ∂f₃∂ξ₁ = D[i, 1] ⊠ v₃[1, j, vt]
+        ∂f₃∂ξ₂ = D[j, 1] ⊠ v₃[i, 1, vt]
+        @simd for k in 2:Nq
```
charleskawczynski, line 305:
```suggestion
        @inbounds @simd for k in 2:Nq
```

ext/cuda/operators_spectral_element.jl
```diff
@@ -269,14 +268,53 @@ Base.@propagate_inbounds function operator_evaluate(
...
+        end
+        return Geometry.Covariant12Vector(∂f∂ξ₁, ∂f∂ξ₂)
+    elseif length(input) == 2
+        # Update `shmem`
+        v₁, v₂ = input
+        ∂f₁∂ξ₁ = D[i, 1] ⊠ v₁[1, j, vt]
+        ∂f₁∂ξ₂ = D[j, 1] ⊠ v₁[i, 1, vt]
+        ∂f₂∂ξ₁ = D[i, 1] ⊠ v₂[1, j, vt]
+        ∂f₂∂ξ₂ = D[j, 1] ⊠ v₂[i, 1, vt]
+        @simd for k in 2:Nq
```
charleskawczynski, line 287:
```suggestion
        @inbounds @simd for k in 2:Nq
```

ext/cuda/operators_spectral_element.jl
```diff
@@ -269,14 +268,53 @@ Base.@propagate_inbounds function operator_evaluate(
...
-    ∂f∂ξ₂ = D[j, 1] * input[i, 1, vt]
-
-    for k in 2:Nq
-        ∂f∂ξ₁ += D[i, k] * input[k, j, vt]
-        ∂f∂ξ₂ += D[j, k] * input[i, k, vt]
+    if length(input) == 1 # check types
+        (v₁,) = input
+        ∂f∂ξ₁ = D[i, 1] ⊠ v₁[1, j, vt]
+        ∂f∂ξ₂ = D[j, 1] ⊠ v₁[i, 1, vt]
+        for k in 2:Nq
```
charleskawczynski, line 275:
```suggestion
        @inbounds for k in 2:Nq
```

charleskawczynski: Actually, maybe we can just wrap the whole block in `@inbounds` since the initial values computed can use them too

akshaysridhar: Updated to wrap blocks in `@inbounds` ; not sure why the formatter now fails since there are no changes to the `DataLayouts` tests in this PR

ext/cuda/operators_spectral_element.jl
```diff
@@ -269,14 +268,59 @@ Base.@propagate_inbounds function operator_evaluate(
...
+                ∂f₂∂ξ₁ = ∂f₂∂ξ₁ ⊞ D[i, k] ⊠ v₂[k, j, vt]
+                ∂f₂∂ξ₂ = ∂f₂∂ξ₂ ⊞ D[j, k] ⊠ v₂[i, k, vt]
+                ∂f₃∂ξ₁ = ∂f₃∂ξ₁ ⊞ D[i, k] ⊠ v₃[k, j, vt]
+                ∂f₃∂ξ₂ = ∂f₃∂ξ₂ ⊞ D[j, k] ⊠ v₃[i, k, vt]
+            end
+        end
+        return Geometry.AxisTensor(
+            (Geometry.Covariant12Axis(), Geometry.UVWAxis()),
+            (∂f₁∂ξ₁, ∂f₁∂ξ₂, ∂f₂∂ξ₁, ∂f₂∂ξ₂, ∂f₃∂ξ₁, ∂f₃∂ξ₂),
+        )
```
charleskawczynski, line 322: Just to note, I may have done this as

```julia
        e1 = length(input) ≥ 1 ? (mapreduce(k -> D[i, k] ⊠ input[1][k, j, vt], ⊞, 1:Nq), mapreduce(k -> D[j, k] ⊠ input[1][i, k, vt], ⊞, 1:Nq)) : ()
        e2 = length(input) ≥ 2 ? (mapreduce(k -> D[i, k] ⊠ input[2][k, j, vt], ⊞, 1:Nq), mapreduce(k -> D[j, k] ⊠ input[2][i, k, vt], ⊞, 1:Nq)) : ()
        e3 = length(input) ≥ 3 ? (mapreduce(k -> D[i, k] ⊠ input[3][k, j, vt], ⊞, 1:Nq), mapreduce(k -> D[j, k] ⊠ input[3][i, k, vt], ⊞, 1:Nq)) : ()
        t = (e1...,e2...,e3...)

        if length(input) == 1 # check types
            return Geometry.Covariant12Vector(t...)
        elseif length(input) == 2
            return Geometry.AxisTensor((Geometry.Covariant12Axis(), Geometry.UVAxis()), t)
        else
            return Geometry.AxisTensor((Geometry.Covariant12Axis(), Geometry.UVWAxis()), t)
        end
```
but I think the `@simd` may actually be faster (when available), it's also a bit simpler, so I think that what you have is good.

charleskawczynski (approved):
This looks great! I'm glad to see that the CPU/GPU are more unified with their results.

ext/cuda/operators_spectral_element.jl
akshaysridhar: Noted, thanks! Perhaps in a future pass we can update these patterns (here and in the `Curl` operators as well).

akshaysridhar:
> This looks great! I'm glad to see that the CPU/GPU are more unified with their results.

Thanks! I'll rebase + format and merge shortly.

Issue #1749, opened 2024-05-22
single_field_solve! / bycolumn bug


juliasloan25:
## Describe the bug
In ClimaLand's RichardsModel implicit solver, I encountered a bug in the MatrixFields solver upon upgrading to ClimaCore v0.14. See [this build](https://buildkite.com/clima/climaland-ci/builds/3943#018f7dfd-3c79-4d10-af01-cfa2b359aef9) for a stacktrace.

This was introduced in https://github.com/CliMA/ClimaCore.jl/pull/1653, perhaps in [this line](https://github.com/CliMA/ClimaCore.jl/pull/1653/files#diff-fd85ec06a7d1ce92773b2c873ef104edc2471ecf15908935906af32125bb38a1R92-R99). It is not an issue when using the commit just before this PR (ClimaCore#033e5b0), and is an issue in all subsequent commits (starting with ClimaCore#31e69ef).

PR to add unit test for reproducer: https://github.com/CliMA/ClimaCore.jl/pull/1750

## To Reproduce
```julia
import ClimaComms
import ClimaCore: Spaces, MatrixFields, Fields, Domains, Meshes, Topologies, Geometry
import ClimaCore.MatrixFields: @name

FT = Float32
zmax = FT(0)
zmin = FT(-0.35)
nelems = 5

# Make FiniteDifferenceSpace
context = ClimaComms.context()
z_domain = Domains.IntervalDomain(
    Geometry.ZPoint(zmin),
    Geometry.ZPoint(zmax);
    boundary_names = (:bottom, :top),
)
z_mesh = Meshes.IntervalMesh(z_domain, nelems = nelems)
z_topology = Topologies.IntervalTopology(context, z_mesh)

space = Spaces.CenterFiniteDifferenceSpace(z_topology)

tridiag_type = MatrixFields.TridiagonalMatrixRow{FT}
# Create a field containing a `TridiagonalMatrixRow` at each point
tridiag_field = Fields.Field(tridiag_type, space)

A = MatrixFields.FieldMatrix((@name(_), @name(_)) => tridiag_field)
field = Fields.ones(space)
b = Fields.FieldVector(; _ = field)
x = similar(b)

solver = MatrixFields.FieldMatrixSolver(MatrixFields.BlockDiagonalSolve(), A, b)

# this passes with ClimaCore#033e5b0 and fails with ClimaCore#31e69ef
MatrixFields.field_matrix_solve!(solver, x, A, b)
```

[...]

PR #1750, opened 2024-05-22
add field_matrix_solve! reproducer

juliasloan25:
Adds a unit test implementing a reproducer for https://github.com/CliMA/ClimaCore.jl/issues/1749. I'm not sure where the best place to add this test is, so I'm happy to move it to a different file if preferred!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,16 @@ ClimaCore.jl Release Notes
 main
 -------
 
+v0.14.5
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Alright, I pushed the fix, and it seems to fix the unit test. I've also updated the news and bumped the patch version. I'll merge once this passes CI and make a patch release after merging.

Issue #1746, opened 2024-05-21
Use shared/local memory in FD stencils kernels


charleskawczynski:
We should implement shared memory for our FD stencil kernels, as this will lower the global memory traffic, reducing the memory bandwidth of the kernels and improve the performance.

A count of different operators in ClimaAtmos:

 - ᶠinterp: 139
 - ᶠwinterp: 14
 - ᶠgradᵥ: 35
 - ᶠcurlᵥ: 1
 - ᶠupwind1: 17 + more
 - ᶠupwind3: 17 + more
 - ᶠfct_boris_book: 1
 - ᶠfct_zalesak: 1
 - ᶠlin_vanleer: 1
 - ᶜinterp: 65
 - ᶜdivᵥ: 51
 - ᶜgradᵥ: ~28
 - ᶜadvdivᵥ: 49
 - ᶜsubdivᵥ: 9
 - ᶜprecipdivᵥ: 17
 - ᶠright_bias: 15
 - ᶜleft_bias: 12
 - ᶜright_bias: 6

charleskawczynski:
This may depend on https://github.com/CliMA/ClimaCore.jl/issues/1754 being fixed.

charleskawczynski:
We can use the same high-level design as the shared memory for the spectral element kernels:

 - Add `work` to the stencil broadcasted object
 - When inside the stencil kernel, we reconstruct the stencil broadcasted object, populating the shared memory for each operator
 - recursively fill the shmem for all operators in the broadcasted object
 - call `getidx` on the broadcasted object, which grabs the data from the shmem instead of global memory

charleskawczynski:
> I still don't understand this.
> 
> Here are some questions:
> 
> * What's `work`?

`work` is the shared memory. We could rename it, but that was the name used in the SEM broadcasted object (I was trying to keep things somewhat unified).

> * Is shared memory allocated once or multiple times per broadcasted expression?

It's allocated once per broadcasted object.

> * What are you recursiving over and what are you filling the shared memory with?

We are recursing over the broadcasted object (it is a recursive object). We fill the shared memory with (for `DivergenceF2C`) `Ju³` at each face. Different operators may store different things, since different operators have different arguments (and sometimes a different number of arguments). Concretely, the `eltype` of the shared memory is the `operator_eltype`.

> * Where is the result of the operation stored?

The result of which operation?

[...]

PR #1735, opened 2024-05-20
Use local memory in band matrix solve

charleskawczynski:
This PR adds the use of local memory in the band matrix solve.

Closes https://github.com/CliMA/ClimaCore.jl/issues/1738.

PR #1763, opened 2024-05-31
Transform FD broadcast objs to use MArrays

charleskawczynski:
Similar to #1735, this PR transforms our broadcast expressions by replacing array-backed DataLayouts to MArray-backed DataLayouts in each column of the stencil kernels.

The main benefit here is that this will result in at most 1 heap memory read for whatever fields we transform. One downside is that those reads become unconditional. I'll update #1746 with more details on these trade-offs, and what design choices we can make to ensure that we only see performance gains without regressions.

Closes https://github.com/CliMA/ClimaCore.jl/issues/1746.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -921,12 +921,14 @@ rebuild(data::VF{S, Nv}, array::AbstractArray{T, 2}) where {S, Nv, T} =
     Nv = nlevels(data)
     Nf = ncomponents(data)
     FT = eltype(parent(data))
+    # @show Nf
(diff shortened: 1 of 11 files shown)


charleskawczynski:
To recap on why this stalled: transforming the broadcasted object into an MArray-backed data structure _forces_ reads of _all_ field variables in the broadcasted object. That means that, even if a variable is not used in the broadcast expression, it still incurs a read, which will tank performance.

If there was some way that we could determine which fields are read, and only transform those variables, then this would be feasible, but I'm not sure if that's possible.

charleskawczynski:
Superseded by #2184.

charleskawczynski:
This is basically superseded by `columnwise!`, as `columnwise!` can apply local memory to both the state and the local geometry.

PR #1768, opened 2024-05-31
Define array2field and field2array

charleskawczynski:
This PR moves some of the `array2field` and `field2array` methods into ClimaCore (from ClimaAtmos). I also added some simple tests.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -1026,6 +1026,8 @@ rebuild(
     array::AbstractArray{T, 5},
 ) where {S, Nv, Nij, T} = VIJFH{S, Nv, Nij}(array)
 
+nlevels(::VIJFH{S, Nv}) where {S, Nv} = Nv
+
 function replace_basetype(
     data::VIJFH{S, Nv, Nij},
     ::Type{T},
@@ -1192,6 +1194,8 @@ end
 
 rebuild(data::VIFH{S, Nv, Ni}, array::A) where {S, Nv, Ni, A <: AbstractArray} =
     VIFH{S, Nv, Ni}(array)
+
+nlevels(::VIFH{S, Nv}) where {S, Nv} = Nv
 
 function replace_basetype(data::VIFH{S, Nv, Ni}, ::Type{T}) where {S, Nv, Ni, T}
     array = parent(data)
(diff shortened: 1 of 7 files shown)


src/Fields/Fields.jl
```diff
@@ -582,4 +580,52 @@ if VERSION < v"1.10"
...
+```
+
+The dimensions of the resulting array are `([number of vertical nodes], number
+of horizontal nodes)`. Also, `field` must be a `Field` of scalars, so that the
+element type of the array is the same as the struct type of `field`.
+"""
+array2field(array, space) = Field(
+    DataLayouts.array2data(array, Spaces.local_geometry_data(space)),
+    space,
+)
```
Sbozzolo, line 601: This docstring is mismatched with the actual function

src/DataLayouts/DataLayouts.jl
```diff
@@ -1481,4 +1485,52 @@ get_Nij(::IFH{S, Nij}) where {S, Nij} = Nij
...
+
+The dimensions of the resulting array are
+ - `([number of vertical nodes], number of horizontal nodes)`.
+
+Also, this assumes that `eltype(data) <: Real`.
+"""
+function data2array end
+
+data2array(data::Union{IF, IFH}) = reshape(parent(data), :)
+data2array(data::Union{IJF, IJFH}) = reshape(parent(data), :)
```
Sbozzolo, line 1503: Are these two functions being tested?

src/Fields/Fields.jl
```diff
@@ -582,4 +580,52 @@ if VERSION < v"1.10"
...
+`array`. If this restriction were removed, one would also need to pass the
+desired `Field` struct type as an argument to `array2field`, which would then
+need to permute the dimensions of `array` to match the target `DataLayout`.
+"""
+function field2array(field::Field)
+    if sizeof(eltype(field)) != sizeof(eltype(parent(field)))
+        f_axis_size = sizeof(eltype(parent(field))) ÷ sizeof(eltype(field))
+        error("unable to use field2array because each Field element is \
+               represented by $f_axis_size array elements (must be 1)")
+    end
```
Sbozzolo, line 627: Would it be easy to add a test for this error?

Sbozzolo (approved):
Thanks, I think there's a couple of functions that are not tested and two docstrings that are swapped, but everything else looks good to me.

src/Fields/Fields.jl
charleskawczynski: Ah, yeah, bummer, I’ll fix it

src/DataLayouts/DataLayouts.jl
charleskawczynski: I believe so, IIRC

charleskawczynski: Ah, actually, no. I'll add some (there were none in ClimaAtmos)

src/Fields/Fields.jl
charleskawczynski: Yes, but I ran out of time this evening.

PR #1775, opened 2024-06-03
Add downstream tests

### LINKED ISSUE #1776: Add downstream tests
Running the test suite for pacakges that depend on ClimaCore can help us prevent problems that might be otherwise hard to track down.

Sbozzolo:
Lots of packages depend on ClimaCore and it is hard to predict the ripple effects of small changes. This commit adds tests for the main downstream repositories that depend on ClimaCore. Passing these tests are not required to merge PRs, but can be a valuable piece of information.

Closes #1776

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -1,0 +1,40 @@
+name: Downstream
+on:
+  push:
+    branches:
+      - main
+    tags: '*'
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: downstream ${{ matrix.package }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - 'ClimaAtmos.jl'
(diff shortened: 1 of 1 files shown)


Sbozzolo:
The failure for ClimaLand is beyond my understanding
```
Soil horizontal operators unit tests, FT = Float64: Test Failed at /home/runner/work/ClimaCore.jl/ClimaCore.jl/ClimaLand.jl/test/standalone/Soil/soil_test_3d.jl:232
  Expression: maximum(abs.(dYX .- expected)) < 10 ^ 11 * eps(FT)
   Evaluated: 1.2806473799198415e-6 < 2.699512151593808e-7
```

This implies that `eps(FT)` is not what we expect, which can happen only if FT is not what we expect or if FT for this architecture is really weird. The same test passes in the CI for ClimaLand.

Sbozzolo:
The ClimaTimeSteppers test currently fails because of a compat in the test project for ClimaTimeSteppers. I have fixed that in my SSPKnoth branch that is ready for review, so I think we can go ahead and merge this even with that test failing.

(None of these tests are/will be required to merge)

charleskawczynski (approved):
Looks good, thanks!

PR #1778, opened 2024-06-03
Fix dss type param

### LINKED ISSUE #1772: With ClimaCore 0.14.6, the remapping module in ClimaCoupler produces incorrect results
The ClimaCoupler test suite passes with ClimaCore 0.14.5, but when updating to 0.14.6, the remapping tests now fail. See example https://github.com/CliMA/ClimaCoupler.jl/actions/runs/9325823164/job/25675049295

I don't know if this is an issue with ClimaCoupler doing something that it shouldn't, or something that was missed in recent changes in ClimaCore. The only change that could have affected this is using Nv, I think (eg https://github.com/CliMA/ClimaCore.jl/commit/1b84475b2405b8e0d2014e32fda5a7914cf9794e)

charleskawczynski:
Closes #1772.

I've also removed type parameters from some of the other methods, since they are unused.

I'm actually a bit flabbergasted that this didn't require MSE changes (https://github.com/CliMA/ClimaAtmos.jl/pull/3058).

My thoughts on this are one of the following must be true:
 - This somehow (miraculously) worked due to `Nq`/`Nv` (though I doubt this)
 - ClimaAtmos' regression tests are insufficient or no longer working as expected (possible, though seems less likely)
 - The code that seems like it should break badly (below) is not actually executed/exercised (certainly possible)

Code that looks like it should break if `Nv` was sent in as `Nq`
```julia
Base.IteratorEltype(::Type{Perimeter2D{Nq}}) where {Nq} = Base.HasEltype()
Base.eltype(::Type{Perimeter2D{Nq}}) where {Nq} = Tuple{Int, Int}

Base.IteratorSize(::Type{Perimeter2D{Nq}}) where {Nq} = Base.HasLength()
Base.length(::Perimeter2D{Nq}) where {Nq} = 4 + (Nq - 2) * 4

function Base.iterate(perimeter::Perimeter2D{Nq}, loc = 1) where {Nq}
    if loc <= 4
        return (vertex_node_index(loc, Nq), loc + 1)
    elseif loc ≤ length(perimeter)
        f = cld(loc - 4, Nq - 2)
        n = mod(loc - 4, Nq - 2) == 0 ? (Nq - 2) : mod(loc - 4, Nq - 2)
        return (face_node_index(f, Nq, 1 + n), loc + 1)
    else
        return nothing
    end
end

function Base.getindex(perimeter::Perimeter2D{Nq}, loc = 1) where {Nq}
    if loc < 1 || loc > length(perimeter)
        return (-1, -1)
    elseif loc <= 4
        return vertex_node_index(loc, Nq)
    else
        f = cld(loc - 4, Nq - 2)
        n = mod(loc - 4, Nq - 2) == 0 ? (Nq - 2) : mod(loc - 4, Nq - 2)
        return face_node_index(f, Nq, 1 + n)
    end
end
```

Regardless, it would be nice if we somehow added tests for `Perimeter2D`. That said, the example that failed is a bit difficult to test precisely because we simply sent in the wrong integer into `Perimeter2D`'s type space. I don't think we should overcomplicate `Perimeter2D` for the sake of testing, but we should be able to somehow test `dss!` to take this into account. cc @sriharshakandala

### DIFF
--- a/src/Topologies/dss_transform.jl
+++ b/src/Topologies/dss_transform.jl
@@ -316,7 +316,7 @@ recv_buffer(ghost::GhostBuffer) = ghost.recv_data
 create_ghost_buffer(data, topology::Topologies.AbstractTopology) = nothing
 
 function create_ghost_buffer(
-    data::Union{DataLayouts.IJFH{S, Nij}, DataLayouts.VIJFH{S, Nij}},
+    data::Union{DataLayouts.IJFH{S, Nij}, DataLayouts.VIJFH{S, <:Any, Nij}},
     topology::Topologies.Topology2D,
 ) where {S, Nij}
     if data isa DataLayouts.IJFH
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.14.6"
+version = "0.14.7"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/ext/cuda/topologies_dss.jl
+++ b/ext/cuda/topologies_dss.jl
@@ -17,9 +17,9 @@ _configure_threadblock(nitems) =
 function Topologies.dss_load_perimeter_data!(
     ::ClimaComms.CUDADevice,
     dss_buffer::Topologies.DSSBuffer,
-    data::Union{DataLayouts.IJFH{S, Nij}, DataLayouts.VIJFH{S, Nij}},
+    data::Union{DataLayouts.IJFH, DataLayouts.VIJFH},
     perimeter::Topologies.Perimeter2D,
-) where {S, Nij}
+)
     pperimeter_data = parent(dss_buffer.perimeter_data)
     pdata = parent(data)
     (nlevels, nperimeter, nfid, nelems) = size(pperimeter_data)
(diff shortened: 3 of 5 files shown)


Sbozzolo:
Thanks for tracking this down! It would be good to understand why this worked in ClimaAtmos or if we should be concerned.

src/Topologies/dss_transform.jl
```diff
@@ -316,7 +316,7 @@ recv_buffer(ghost::GhostBuffer) = ghost.recv_data
 create_ghost_buffer(data, topology::Topologies.AbstractTopology) = nothing
 
 function create_ghost_buffer(
-    data::Union{DataLayouts.IJFH{S, Nij}, DataLayouts.VIJFH{S, Nij}},
+    data::Union{DataLayouts.IJFH{S, Nij}, DataLayouts.VIJFH{S, <:Any, Nij}},
```
Sbozzolo, line 319: Why `Any`?

charleskawczynski: I did this to avoid specializing on `Nv`, we could write it as `NV ... where {..., NV, ...}`, but `Nv` is not used anywhere, so it doesn't really contribute anything useful (aside from maybe following a more consistent pattern name?). The `Any` is because type parameters are ["values"](https://docs.julialang.org/en/v1/manual/types/#%22Value-types%22):

```julia
julia> foo(::Val{<:Int}) = 1
foo (generic function with 1 method)

julia> foo(Val(2))
ERROR: MethodError: no method matching foo(::Val{2})

Closest candidates are:
  foo(::Val{<:Int64})
   @ Main REPL[1]:1

Stacktrace:
 [1] top-level scope
   @ REPL[2]:1

julia> foo(Val(Int))
1
```

PR #1788, opened 2024-06-07
Add wrapper for FieldMatrix and FieldMatrixSolver

dennisYatunin:
This PR adds the `InvertibleFieldMatrix` struct, which is a simple wrapper for a `FieldMatrix` and a corresponding `FieldMatrixSolver`. The wrapper has a constructor that specifies a default `FieldMatrixSolverAlgorithm`, and it also has methods for `similar`, `ldiv!`, and `mul!`. This should allow us to remove a lot of boilerplate code that is currently needed to set up implicit solvers (see [this comment](https://github.com/CliMA/ClimaTimeSteppers.jl/pull/292#discussion_r1630312851) in ClimaTimeSteppers).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@ ClimaCore.jl Release Notes
 main
 -------
 
+v0.14.8
+-------
+
+- ![][badge-✨feature/enhancement] Added `FieldMatrixWithSolver`, a wrapper that helps defining implicit Jacobians. PR [#1788](https://github.com/CliMA/ClimaCore.jl/pull/1788)
+
+
 v0.14.6
 -------
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.14.7"
+version = "0.14.8"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/docs/src/matrix_fields.md
+++ b/docs/src/matrix_fields.md
@@ -31,6 +31,7 @@ operator_matrix
 ```@docs
 FieldMatrixSolverAlgorithm
 FieldMatrixSolver
+FieldMatrixWithSolver
 field_matrix_solve!
 BlockDiagonalSolve
 BlockLowerTriangularSolve
--- a/src/MatrixFields/MatrixFields.jl
+++ b/src/MatrixFields/MatrixFields.jl
@@ -43,7 +43,8 @@ multiples of `LinearAlgebra.I`. This comes with the following functionality:
 """
 module MatrixFields
 
-import LinearAlgebra: I, UniformScaling, Adjoint, AdjointAbsVec, mul!, inv, norm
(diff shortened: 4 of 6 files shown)


Sbozzolo (approved):

Sbozzolo:
Looks good, please fix the tests, and maybe let's make another release so that we can use it in ClimaTimeSteppers.

src/MatrixFields/invertible_field_matrix.jl
```diff
@@ -0,0 +1,48 @@
...
+A wrapper that combines a `FieldMatrix` `A` with a `FieldMatrixSolver` that can
+be used to solve the equation `A * x = b` for `x`, where `x` and `b` are both
+`FieldVector`s. Similar to a `LinearAlgebra.Factorization`, this wrapper can be
+passed to `ldiv!`, whereas a regular `FieldMatrix` cannot be passed to `ldiv!`.
+
+By default, the `FieldMatrixSolverAlgorithm` `alg` is set to a
+[`BlockDiagonalSolve`](@ref), so a custom `alg` must be specified when `A` is
+not a block diagonal matrix.
+"""
+struct InvertibleFieldMatrix{M, S} <: AbstractDict{FieldNamePair, Any}
```
charleskawczynski, line 13: I'm not sure I like this name, the ability to invert a matrix depends on the type of matrix but may also depend on the _values_ in the matrix. So it's not clear that this type will hold up to its name (unless we check upon construction, which sounds expensive).

Is there a different name that we can use that might be more suitable?

charleskawczynski (approved):
Overall I think that this is an improvement, but I think that some names can be improved. Let's let #1787 get through and rebase this, first.

src/MatrixFields/invertible_field_matrix.jl
Sbozzolo: cc: @dennisYatunin

dennisYatunin: The closest analogy to this in Julia is a `LinearAlgebra.Factorization`, but I don't think we should use a name like that because we're not planning to do any factorization here. Something like `LDivableFieldMatrix` would probably be accurate, but it sounds awful.

We currently do one simple check for invertibility---making sure that there are no missing (zero) blocks along the diagonal. This check only relies on type information, so it's very cheap. We could conceivably add a debug mode to this struct as well, which would compute the full matrix determinant and check that it isn't too close to zero. This would help us catch issues in the Jacobian before they cause the solution to blow up, so we might want it for debugging implicit prognostic EDMF.

Since we already do one check for invertibility and we could potentially add more expensive checks later, maybe `InvertibleFieldMatrix` is fine? I can also modify the docstring to include this information.

charleskawczynski: @dennisYatunin, how about `SquareFieldMatrix`?

charleskawczynski: Or even just `FieldMatrix`?

charleskawczynski: I personally like `SquareFIeldMatrix`, since it's necessary but not sufficient for invertibility

Sbozzolo: Going with `FieldMatrixWithSolver`.

PR #1798, opened 2024-06-12
Refactor terrain warp tests

charleskawczynski:
This PR refactors the terrain warp unit tests, which take over 10 minutes to run:

```julia
julia> using Revise; include(joinpath("test", "Spaces", "terrain_warp.jl"))
Test Summary:                    | Pass  Total     Time
2D Extruded Terrain Warped Space |  600    600  1m18.3s
Test Summary:                           | Pass  Total   Time
2D Extruded Terrain Laplacian Smoothing |  144    144  39.2s
Test Summary:                       | Pass  Total  Time
2D Warped Mesh RHS Integration Test |    2      2  1.6s
Test Summary:                    | Pass  Total     Time
3D Extruded Terrain Warped Space |  600    600  5m35.7s
Test Summary:                           | Pass  Total     Time
3D Extruded Terrain Laplacian Smoothing |   72     72  6m31.1s
```

Refactoring items include:

 - We pass `device` through, since it's type unstable (it depends on `ENV`)
 - Reduces the number of combinations that the tests are run for (some are currently testing 150 combinations of npoly, helem, velem, 300 including Float32+Float64)
 - Removed Float64 tests (those should converge if the Float64 converge)
 - Split `generate_base_spaces` into two type-stable methods
 - Combine the `test_throws` test with the one before it (to avoid recreating the same objects)

I think that there's more that can be done, but perhaps #1797 is a culprit here.

With these changes, the runtime looks more reasonable:

```julia
julia> using Revise; include(joinpath("test", "Spaces", "terrain_warp.jl"))
Test Summary:                    | Pass  Total   Time
2D Extruded Terrain Warped Space |   36     36  14.4s
Test Summary:                           | Pass  Total  Time
2D Extruded Terrain Laplacian Smoothing |   54     54  2.4s
Test Summary:                       | Pass  Total  Time
2D Warped Mesh RHS Integration Test |    2      2  4.2s
Test Summary:                    | Pass  Total   Time
3D Extruded Terrain Warped Space |   36     36  22.7s
Test Summary:                           | Pass  Total     Time
3D Extruded Terrain Laplacian Smoothing |   27     27  1m13.7s
Test Summary:                    | Pass  Total  Time
Interior Mesh `Adaption` ηₕ Test |    1      1  2.8s
Test Summary:                             | Pass  Total  Time
Interior Mesh `Adaption` (ηₕ=1, s=1) Test |   12     12  0.3s
```

PR #1800, opened 2024-06-12
Iterate over unit tests, tabulate sorted timings

charleskawczynski:
This PR is an alternative to #1340.

One thing I found from this is that we are currently running `test/Spaces/spaces.jl`, and one other test, twice.

Iterating over all the tests has a few nice advantages:
 - We can ensure we're not doing duplicate work
 - We can tabulate the results, to make it easier to see which tests are expensive
 - We could search for all tests in `test/` starting with `unit_` to see if we're missing any

Another thing I did was, I added a flag: `prevent_leaky_tests`, which allows us to toggle on and off (effectively) using `@safetestset`. We could even make this `true` on GHA, and `false` locally, so that running unit tests locally is much faster, but CI could still catch leaky test situations (which will happen anyway on buildkite).

I think this might be a bit simpler/declarative than #1340, so I'm inclined to favor this PR over #1340. Thoughts?


test/runtests.jl
```diff
@@ -1,128 +1,153 @@
...
-    @safetestset "MatrixFields - BandMatrixRow" begin @time include("MatrixFields/band_matrix_row.jl") end
-    @safetestset "MatrixFields - rmul_with_projection" begin @time include("MatrixFields/rmul_with_projection.jl") end
-    @safetestset "MatrixFields - field2arrays" begin @time include("MatrixFields/field2arrays.jl") end
-    @safetestset "MatrixFields - matrix multiplication at boundaries" begin @time include("MatrixFields/matrix_multiplication_at_boundaries.jl") end
-    @safetestset "MatrixFields - field names" begin @time include("MatrixFields/field_names.jl") end
-    # now part of buildkite
-    # @safetestset "MatrixFields - matrix field broadcasting" begin @time include("MatrixFields/matrix_field_broadcasting.jl") end
-    # @safetestset "MatrixFields - operator matrices" begin @time include("MatrixFields/operator_matrices.jl") end
-    # @safetestset "MatrixFields - field matrix solvers" begin @time include("MatrixFields/field_matrix_solvers.jl") end
+mutable struct UnitTest
```
Sbozzolo, line 7: Why does it have to be mutable? Is it only for the `elapsed`?

Sbozzolo (commented):
I also prefer this. This might also allow us to parallelize tests distributing the for loop.

I wonder if we could make this even simpler:

For example, `gpu_only` could be a property of the test file itself: tests that are gpu_only would be skipped if the device is not CUDADevice. This would have the benefit that the test would be skipped also when run with `julia --project=test test/Fields/reduction_cuda.jl` and not just when run with `tests/runtests.jl`. This would also remove a special case in handling tests (essentialy, CUDA tests would become no-op when CUDA is not available).

You can also remove SafeTestsSet from the dependencies now.

test/runtests.jl
```diff
@@ -1,128 +1,153 @@
...
-    @safetestset "Operators - extruded sphere space operators CUDA" begin @time include("Operators/hybrid/extruded_sphere_cuda.jl") end
-    @safetestset "Operators - extruded 3dbox space operators CUDA" begin @time include("Operators/hybrid/extruded_3dbox_cuda.jl") end
-    @safetestset "Fields - CUDA mapreduce" begin @time include("Fields/reduction_cuda.jl") end
+prevent_leaky_tests = false
+# for test in [unit_tests[1], unit_tests[2]] # for debugging this file
+for test in unit_tests
+    if test.gpu_only && !(ClimaComms.device() isa ClimaComms.CUDADevice)
+        continue
+    end
+    @info "--- About to test $(test.filename)"
```
nefrathenrici, line 144:
```suggestion
    @info "About to test $(test.filename)"
```
Are the dashes used for anything?

test/runtests.jl
```diff
@@ -1,128 +1,153 @@
...
-    @safetestset "GPU - data" begin @time include("DataLayouts/cuda.jl") end
-    @safetestset "GPU - spaces" begin @time include("Spaces/spaces.jl") end
-    @safetestset "Spaces - serial CUDA DSS" begin @time include("Spaces/ddss1.jl") end
-    @safetestset "Spaces - serial CUDA DSS on CubedSphere" begin @time include("Spaces/ddss1_cs.jl") end
-    @safetestset "Operators - spectral element CUDA" begin @time include("Operators/spectralelement/rectilinear_cuda.jl") end
-    @safetestset "Operators - finite difference CUDA" begin @time include("Operators/hybrid/unit_cuda.jl") end
-    @safetestset "Operators - extruded sphere space operators CUDA" begin @time include("Operators/hybrid/extruded_sphere_cuda.jl") end
-    @safetestset "Operators - extruded 3dbox space operators CUDA" begin @time include("Operators/hybrid/extruded_3dbox_cuda.jl") end
-    @safetestset "Fields - CUDA mapreduce" begin @time include("Fields/reduction_cuda.jl") end
+prevent_leaky_tests = false
```
nefrathenrici, line 138: We could set this via a check if the file is running in GHA

nefrathenrici (commented):
I think this PR cleans up the unit tests nicely. It's simpler than #1340 and doesn't build from the buildkite yaml.

Does this cause issues with testing on Windows? I don't see any equivalents to the `Sys.iswindows()` used before.

test/runtests.jl
charleskawczynski: Yeah, just for `elapsed`

test/runtests.jl
charleskawczynski: Yeah, I had thought about that

charleskawczynski:
> Does this cause issues with testing on Windows? I don't see any equivalents to the `Sys.iswindows()` used before.

Before we were skipping windows by checking if `Sys.iswindows()`, but that issue very well may be resolved. This PR doesn't actually do anything to fix that issue, though.

charleskawczynski:
> I also prefer this. This might also allow us to parallelize tests distributing the for loop.

True, we might not be able to do that on GHA, but it's definitely an option locally.

> I wonder if we could make this even simpler: For example, `gpu_only` could be a property of the test file itself...

My hope is that we can unify the gpu-only tests and not have to make this distinction at all, but this seemed like the quickest thing to do for now. Maybe we can punt on that and see how far we can go with unifying the cpu-gpu tests?

> You can also remove SafeTestsSet from the dependencies now.

Good point. I should be able to add that to this PR. I was actually going to try to use `SafeTestsSet`, but there doesn't seem to be a way to interpolate the test name, so I resorted to eval-ing.

test/runtests.jl
charleskawczynski: I'll remove it, I think it's actually a bit too noisy

test/runtests.jl
charleskawczynski: I don't think it's a big deal, and it's type-stable. We can always iterate on it.

test/runtests.jl
charleskawczynski: I'll set it to `!isnothing(get(ENV, "CI", nothing)`

charleskawczynski:
Bummer, the tests are timing out. I'm going to comment out the ones that seem to be hanging (this PR adds a bunch of new ones in), and see if we can fix them.

test/runtests.jl
charleskawczynski: I'm actually going to leave them commented, just for debugging

charleskawczynski:
Ok, I think that this PR is ready, if people like the idea of moving in this direction. One nice benefit is that we get a table of sorted timings at the end, so we can quickly see which tests are taking the longest.

I tried updating the tabulation logic (and I added docs) in such a way that we might split it out into a separate package.

charleskawczynski:
Wow, the terrain test appears to hang, even after #1798

test/tabulated_tests.jl
```diff
@@ -0,0 +1,225 @@
...
+function tabulate_tests(
+    unit_tests::Vector{UnitTest};
+    include_timings::Bool = true,
+    time_format::Symbol = :second,
+)
+    title =
+        include_timings ? "Tests results" :
+        "Running the following unit tests..."
+    if include_timings
+        sort!(unit_tests; by = x -> x.elapsed)
```
nefrathenrici, line 93: This sorts time by ascending value, I think descending would be better for quickly checking long-running tests.

test/tabulated_tests.jl
```diff
@@ -0,0 +1,225 @@
...
+        header = ["Name", "Filename"]
+        data =
+            hcat(map(x -> x.name, unit_tests), map(x -> x.filename, unit_tests))
+    end
+    PrettyTables.pretty_table(
+        data;
+        title,
+        header,
+        alignment = :l,
+        crop = :horizontal,
```
nefrathenrici, line 125: Does the table need to be cropped horizontally?

nefrathenrici (commented):
This will make it easier to keep track of all of the unit tests. I just left a few small comments.

I also noticed this `1 column omitted` [line in the GHA log](https://github.com/CliMA/ClimaCore.jl/actions/runs/9503250602/job/26193253097?pr=1800#step:6:716), I'm not sure if that is an issue since all of the columns appear to be displayed.

test/tabulated_tests.jl
charleskawczynski: It's pretty jumbled on GHA when we don't. We could change the behavior in CI vs interactive, but the last column is the filename and the test name usually fits on the page. I mostly put `crop = :horizontal` to ensure that it's not cropped in the vertical (I think the default is `crop = :both`).

charleskawczynski:
> I also noticed this `1 column omitted` [line in the GHA log](https://github.com/CliMA/ClimaCore.jl/actions/runs/9503250602/job/26193253097?pr=1800#step:6:716), I'm not sure if that is an issue since all of the columns appear to be displayed.

Yeah, those are the filenames it's pretty jumbled on GHA when if we don't crop. We could change the behavior in CI vs interactive, but the last column is the filename and the test name usually fits on the page.

test/tabulated_tests.jl
charleskawczynski: Sure, I don't mind whether it's ascending/descending

charleskawczynski:
This PR increases our coverage by 4.93%: https://app.codecov.io/gh/CliMA/ClimaCore.jl/pull/1800/indirect-changes

test/runtests.jl
```diff
@@ -1,128 +1,123 @@
...
+    test -> !(
+        test.meta == :gpu_only &&
+        !(ClimaComms.device() isa ClimaComms.CUDADevice)
+    ),
+    unit_tests,
+)
+# Note: for `fail_fast = false`, the tests are all wrapped in `@testset "Unit tests"`
+#       so output is suppressed until all tests are complete.
+fail_fast = true
+prevent_leaky_tests = !isnothing(get(ENV, "CI", nothing)) # prevent leaky tests on CI)
```
Sbozzolo, line 107: I am not a big fan of using env variables like this, especially when they are so generic. On people's shells, `CI` could be defined for reasons that are unrelated to ClimaCore (in the same way `HOME`, `EDITOR`, `TERM` and several other variables are typicallly defined). The value is also not checked, only its existance. I don't think tests should silently beheve differently depending on this. 

If you want to use env variable, I suggest you change to it `CLIMACORE_TESTS_PREVENT_LEAKY_TESTS` and define this in the GitHub runner by adding it to the workflow step.

Sbozzolo (approved):
Thanks, looks much cleaner!

test/runtests.jl
charleskawczynski: Yeah, I tend to prefer more specific ones also, but I figured that this was relatively standard (and somewhat easy/portable) since GHA and buildkite already define `ENV["CI"]` 🤷🏻.

charleskawczynski: I'll merge for now, but I'm happy to update it if anyone runs into problems.

test/tabulated_tests.jl
charleskawczynski: Ah, I forgot to reverse the table, I’ll follow up on this

PR #1802, opened 2024-06-12
Fix broken GPU dispatching

charleskawczynski:
Sadly, #1787 was 2 characters away (`<:`) from being correct, but the result was that it introduced a bug. GPU dispatching is broken on the main branch, this PR fixes the issue, and adds and fixes the test (which requires views to catch the `SubArray` case).

I've also defined `rebuild(data::IF{S, Nij}` since I found it was missing, and I've included `unit_ndims.jl` into the buildkite pipeline.

cc @trontrytel, @AlexisRenchon, @szy21


Sbozzolo (approved):
Thank you!

PR #1804, opened 2024-06-13
Add DataLayouts `copyto!` unit tests

charleskawczynski:
This PR adds unit tests for the DataLayouts `copyto!`. This PR also adds CUDA `copyto!` support for `IFH`, `IJF`, `IF` and `VIFH` datalayouts (which were previously missing).

I've also defined some aliases: `BroadcastedUnionIJFH`, `BroadcastedUnionIFH` and alike, to simplify some of the method definitions (we can use them in at least one other place).

I did also switch over to using `auto_launch!`'s `auto` (I already did this for `fill!`), but we should probably benchmark the difference. Perhaps we can break this PR up into the tests (some of which will be broken) and benchmarks, so that we can compare. Thoughts on this? I do really like the simplicity of unifying and using the cuda utils.


Sbozzolo (approved):

src/DataLayouts/DataLayouts.jl
```diff
@@ -46,6 +46,13 @@ abstract type AbstractData{S} end
 
 Base.size(data::AbstractData, i::Integer) = size(data)[i]
 
+"""
+    universal_size(::AbstractData)
+
+Return sizes `(Ni, Nj, Nf, Nv, Nh)`, for all
+data types. If a dimension does not exist,
+then return 1 for that dimension.
```
Sbozzolo, line 54: I know that this is used often in the code, but could you please define the various `(Ni, Nj, Nf, Nv, Nh)` in this docstring too?

Sbozzolo:
Looks good to me!

src/DataLayouts/DataLayouts.jl
charleskawczynski: Yes, will do

charleskawczynski:
I decided to not replace some of the existing `copyto!` kernels, as they seem to slightly outperform the ones with CUDA's `launch_configuration`, however, I did add support for some of the missing methods (cpu and gpu), and I left the missing ones commented in the code with a comment. I think that this is objectively better in that there's now more support, and more code is tested.

PR #1805, opened 2024-06-13
Define rebuild for DataF, and some type aliases

charleskawczynski:
This PR defines `rebuild` for `DataF` and some type aliases, for `Union{<:AbstractData, Broadcasted{<:DataStyle}}`.


Sbozzolo (approved):
Can you please add a test if it is not already there?

charleskawczynski:
> Can you please add a test if it is not already there?

Yes, I will do it separately if #1804 doesn't work out (this is somewhat of a peel off from that PR)

PR #1807, opened 2024-06-14
Split fd column tests

charleskawczynski:
This PR splits up the unit and convergence column tests, and renames the column benchmark files


charleskawczynski:
CI passed, except for the 2-process DSS (allocations in that test seem to have a bit of noise). I bumped the limit, merging now.

Issue #1797, opened 2024-06-12
Improve inference in space constructors


charleskawczynski:
In particular, I'm seeing a bunch of jet failures in `warpedspace` of `test/Spaces/terrain_warp.jl`. We should probably add more systematic tests, as this can improve compile times.

charleskawczynski:
It turns out that a big culprit here is the `OBJECT_CACHE`, which is inherently type stable since it takes a type-unbounded dict.

IMO, this is not really the correct way to test if two spaces are compatible (and won't generally work, even for the coupler). Removing the `get!`-`OBJECT_CACHE` logic from all the constructors fixes a huge number of type instabilities, ironically speeding up the construction of the space. Subsequent calls can be faster, but I'm wondering if this is worth it

PR #1810, opened 2024-06-14
Add test and improve inference of monotonic_check

charleskawczynski:
This PR adds unit tests for `monotonic_check`, and inference tests (a step towards #1797). Previously, the closure `z` was getting captured inside `any`, and `map` was returning an object.


Sbozzolo (approved):
Thank you!

charleskawczynski:
Unfortunately, I'm getting slightly different inference results locally compared to central. I'll try to match the central numbers. Hopefully we can eventually drive this down to 0 on both.

charleskawczynski:
In the case that it's flakey, I'm going to pick conservative numbers, and rely on `@test_opt` for decision making.

PR #1821, opened 2024-06-17
Remove `IntervalTopology(::Mesh)` method

charleskawczynski:
This PR removes the `IntervalTopology(::Mesh)` method, as it may be incompatible with the device based on the latest `ENV["CLIMACOMMS_DEVICE"]` value. Instead, users are required to pass `device` (or the context) into this method.

This is breaking, but the fix is easy. Might as well rip the bandaid off.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ ClimaCore.jl Release Notes
 main
 -------
 
+v0.15.0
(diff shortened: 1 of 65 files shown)


Sbozzolo (approved):
Thanks for fixing this!

PR #1828, opened 2024-06-18
Use a Symmetric matrix in the Covariant Axis2Tensor

charleskawczynski:
This PR adds the use of symmetric matrices for the Covariant Axis2Tensors.

I think that this should be all we need, and this may result in a nice improvement in our kernels that use these tensors.

I first tried to use `LinearAlgebra.Symmetric`, however, it has a `uplo::Char` field, and this isn't a float type and as a result does not pass the `check_basetype(T, S)` test in our DataLayouts.  So, I grabbed most of those definitions from LinearAlgebra and simply removed `uplo`.

I did have to widen the allowable types in `AxisTensor`.


charleskawczynski:
It looks like we'll need to generalize the `rmul_with_projection` logic to handle more generic AxisTensors (i.e., ones backed by `SimpleSymmetric` matrices).

Issue #1467, opened 2023-09-21
Compiling functions that take tuples of Fields with Spaces inside is unreasonably expensive


Sbozzolo:
Below is shown a simple script that proves that there is a significant performance issue introduced by using (named) tuples of fields that have space information.

The script sets up a space, and a constant field `Y` defined on this space. Then, it defines a trivial function `mytest(p) = p`. For `p`, I test four difference choices:
- `p_named_tuple`, which is a `NamedTuple` with 44 references to `Y` (`p = (; a0 = Y, a1 = Y, ....)`)
- `p_tuple`, a tuple with 44 references to `Y`
- `p_array`, an array with 44 references to `Y`
- `p_struct`, an instance of a struct that contains 44 fields. In this case, all references to `Y`

The script shows that compile time is significantly worse when using tuples:
```
Named tuple:   1.202479 seconds (343 allocations: 24.969 KiB, 100.00% compilation time)
Tuple:   1.190572 seconds (339 allocations: 24.438 KiB, 100.00% compilation time)
Array:   0.002351 seconds (339 allocations: 24.438 KiB, 99.33% compilation time)
Struct:   0.003234 seconds (339 allocations: 24.438 KiB, 99.45% compilation time)
```

Following intuition from @simonbyrne, I checked what would happen if we didn't have space information. So, I created a new `Y_array`, where we don't have space information. In this case:
```
Tuple with Y array:   0.003157 seconds (339 allocations: 24.438 KiB, 99.46% compilation time)
```

The increase in compile time is non-linear. If we double the number of entries, the compile time for tuples grows to
```
Named tuple:   3.747541 seconds (346 allocations: 25.016 KiB, 100.00% compilation time)
Tuple:   3.768719 seconds (342 allocations: 24.484 KiB, 100.00% compilation time)
```

This problem has profound implications for latency in `ClimaAtmos` since the cache is a massive named tuple. Every function that calls the cache, no matter how trivial it is, can introduce several seconds of unnecessary latency.

I have a rough test implementation where I make the cache a struct instead of a named tuple.

For a simple moist case without radiation, that following code compiles 60% faster on my branch:
```julia
import ClimaAtmos as CA
config = CA.AtmosConfig(parsed_args = Dict("config_file" => XXXX)
CA.get_integrator(config)
```

Given that the cache is a complex and rich object, and recreating its structure in a struct is not an easy task, my branch currently fails when solving the equations.

[...]

PR #1487, opened 2023-10-04
Add Grids, make them mutable objects

simonbyrne:
Fixes #1120 and #1467.

I've gone with option 2 in https://github.com/CliMA/ClimaCore.jl/issues/1120#issuecomment-1737903200.

In the following benchmark:
```julia
using ClimaComms

import ClimaCore:
    ClimaCore,
    slab,
    Spaces,
    Domains,
    Meshes,
    Geometry,
    Topologies,
    Spaces,
    Fields,
    Operators


FT = Float64
vertdomain = Domains.IntervalDomain(
    Geometry.ZPoint{FT}(0.0),
    Geometry.ZPoint{FT}(1.0);
    boundary_tags = (:bottom, :top),
)
vertmesh = Meshes.IntervalMesh(vertdomain, nelems = 10)
vert_center_space = Spaces.CenterFiniteDifferenceSpace(vertmesh)

horzdomain = Domains.SphereDomain(30.0)
horzmesh = Meshes.EquiangularCubedSphere(horzdomain, 4)
horztopology = Topologies.Topology2D(
    ClimaComms.SingletonCommsContext(ClimaComms.CPUSingleThreaded()),
    horzmesh,
)
quad = Spaces.Quadratures.GLL{3 + 1}()
horzspace = Spaces.SpectralElementSpace2D(horztopology, quad)

hv_center_space =
    Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_center_space)

coords = Fields.coordinate_field(hv_center_space)
x = Geometry.UVWVector.(cosd.(coords.lat), 0.0, 0.0)


t = ntuple(i -> x, 80);
f(t) = t
@time f(t);
```

On main:
```julia
julia> @time f(t);
  1.455831 seconds (343 allocations: 24.953 KiB, 100.00% compilation time)
```
on this branch
```julia
julia> @time f(t);
  0.002401 seconds (343 allocations: 24.953 KiB, 99.12% compilation time)
```

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

Issue #1120, opened 2023-02-06
Guarantee uniqueness of topologies and spaces


simonbyrne:
**Is your feature request related to a problem? Please describe.**

It's very easy to get the mismatched spaces warning, especially when loading data from multiple files. It will also become an issue 

Related issues
- #639 was the original issue, and was closed by #640 which turned it into a warning
- #960 referred to this problem in the context of time series, but it is more general

**Describe the solution you'd like**

The basic idea would be to keep track of existing objects (spaces and below) in some sort of internal cache: if a constructor is called with the same arguments, we instead use the cached object.

The major challenge would be making sure that objects get GCed even if they're not referenced anywhere outside the cache. Julia provides `WeakRef`s for this purpose. See also [`WeakKeyDict`](https://github.com/JuliaLang/julia/blob/46032c56c505b62a5aae1eefef537e8d4e339856/base/weakkeydict.jl) (unfortunately we can't use a `WeakKeyDict` directly, as not all our objects are mutable).

We may want to change some objects to be `mutable struct`s as that may make the table lookup easier.

**Describe alternatives you've considered**
The alternative is to allow more operations on duplicate objects, but this will be less efficient in terms of memory usage (as we will end up with duplicate objects) and performance (as we will need to spend more time checking objects are actually the same).

**Additional context**

simonbyrne:
One place where this could cause problems is when we extract levels or columns of extruded spaces. Will need to think about it.

simonbyrne:
There are is a keen design choice on how to handle staggered grids.

Currently we have a field `staggering` which details whether the field values will live on cell centers or cell faces, and we can construct one from the other by just creating a new object and switching that field. This will no longer work, as it will create new objects, and we will have no way to know if fields match up or not.

So some possible ways to do this:

1. We define one field as a wrapper around the other. Since most values are on cell centers, it would make sense to define the center field as the base case, and the face one as a wrapper:
    ```julia
    mutable struct CenterFiniteDifferenceSpace
     # all the information
    end
    
    # wrapper
    struct FaceFiniteDifferenceSpace{C<:CenterFiniteDifferenceSpace}
       center_space::C
    end
    ```
    and then we can switch between them by wrapping/unwrapping.

2. Define some intermediate object (say a "grid" for want of a better name), and make both spaces wrappers around this:
    ```julia
    mutable struct FiniteDifferenceGrid
     # all the information
    end

[...]

PR #1848, opened 2024-06-26
Add convenience constructors for grids

charleskawczynski:
Supersedes #1567.

I've made _all_ arguments optional, so that users can call, for example,  `ExtrudedCubedSphereGrid()` for an extruded cubed sphere grid.

### DIFF
--- a/src/CommonGrids/CommonGrids.jl
+++ b/src/CommonGrids/CommonGrids.jl
@@ -1,0 +1,646 @@
+"""
+    CommonGrids
+
+CommonGrids contains convenience constructors for common
+grids. Constructors in this module are sometimes dynamically
+created. You may want to use a different constructor if you're
+making the object in a performance-critical section, and if
+you know the type parameters at compile time.
+
+If no convenience constructor exists, then you may need to
+create a custom grid using our low-level compose-able API.
+
+
+# Transitioning to using CommonGrids
+
+You may have constructed a grid in the following way:
+
+```julia
+using ClimaComms
+using ClimaCore: DataLayouts, Geometry, Topologies, Quadratures, Domains, Meshes, Grids
+FT = Float64
+z_elem = 63
+z_min = FT(0)
+z_max = FT(1)
+radius = FT(6.371229e6)
+h_elem = 15
+n_quad_points = 4
+device = ClimaComms.device()
+context = ClimaComms.context(device)
+hypsography = Grids.Flat()
+global_geometry = Geometry.ShallowSphericalGlobalGeometry(radius)
+quad = Quadratures.GLL{n_quad_points}()
+h_mesh = Meshes.EquiangularCubedSphere(Domains.SphereDomain{FT}(radius), h_elem)
+h_topology = Topologies.Topology2D(context, h_mesh)
+z_boundary_names = (:bottom, :top)
+z_domain = Domains.IntervalDomain(
+    Geometry.ZPoint{FT}(z_min),
+    Geometry.ZPoint{FT}(z_max);
+    boundary_names = z_boundary_names,
+)
+z_mesh = Meshes.IntervalMesh(z_domain; nelems = z_elem)
+h_grid = Grids.SpectralElementGrid2D(h_topology, quad)
+z_topology = Topologies.IntervalTopology(context, z_mesh)
+z_grid = Grids.FiniteDifferenceGrid(z_topology)
+grid = Grids.ExtrudedFiniteDifferenceGrid(
+    h_grid,
+    z_grid,
+    hypsography,
+    global_geometry,
+)
+```
+
+You may re-write this as:
+
+```julia
+using ClimaCore.CommonGrids: ExtrudedCubedSphereGrid
+grid = ExtrudedCubedSphereGrid(;
+    z_elem = 63,
+    z_min = 0,
+    z_max = 1,
+    radius = 6.371229e6,
+    h_elem = 15,
+    n_quad_points = 4,
+)
+```
+"""
+module CommonGrids
+
+export ExtrudedCubedSphereGrid,
+    CubedSphereGrid, ColumnGrid, Box3DGrid, SliceXZGrid, RectangleXYGrid
+
+import ClimaComms
+import ..DataLayouts,
+    ..Meshes, ..Topologies, ..Geometry, ..Domains, ..Quadratures, ..Grids
+
+include("Helpers.jl")
+import .Helpers.DefaultSliceXMesh
+import .Helpers.DefaultZMesh
+import .Helpers.DefaultRectangleXYMesh
+
+#####
+##### Grids
+#####
+
+"""
+    ExtrudedCubedSphereGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
+        z_elem::Integer,
+        z_min::Real,
+        z_max::Real,
+        radius::Real,
+        h_elem::Integer,
+        n_quad_points::Integer,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+        stretch::Meshes.StretchingRule = Meshes.Uniform(),
+        hypsography_fun = (h_grid, z_grid) -> Grids.Flat(),
+        global_geometry::Geometry.AbstractGlobalGeometry = Geometry.ShallowSphericalGlobalGeometry(radius),
+        quad::Quadratures.QuadratureStyle = Quadratures.GLL{n_quad_points}(),
+        h_mesh = Meshes.EquiangularCubedSphere(Domains.SphereDomain{FT}(radius), h_elem),
+        h_topology::Topologies.AbstractDistributedTopology = Topologies.Topology2D(context, h_mesh),
+        horizontal_layout_type = DataLayouts.IJFH,
+        z_mesh::Meshes.IntervalMesh = DefaultZMesh(FT; z_min, z_max, z_elem, stretch),
+        enable_bubble::Bool = false
+    )
+
+A convenience constructor, which builds an
+[`Grids.ExtrudedFiniteDifferenceGrid`](@ref), given:
+
+ - `FT` the floating-point type (defaults to `Float64`) [`Float32`, `Float64`]
+ - `z_elem` the number of z-points
+ - `z_min` the domain minimum along the z-direction.
+ - `z_max` the domain maximum along the z-direction.
+ - `radius` the radius of the cubed sphere
+ - `h_elem` the number of horizontal elements per side of every panel (6 panels in total)
+ - `n_quad_points` the number of quadrature points per horizontal element
+ - `device` the `ClimaComms.device`
+ - `context` the `ClimaComms.context`
+ - `stretch` the mesh `Meshes.StretchingRule` (defaults to [`Meshes.Uniform`](@ref))
+ - `hypsography_fun` a function or callable object (`hypsography_fun(h_grid, z_grid) -> hypsography`) for constructing the hypsography model.
+ - `global_geometry` the global geometry (defaults to [`Geometry.CartesianGlobalGeometry`](@ref))
+ - `quad` the quadrature style (defaults to `Quadratures.GLL{n_quad_points}`)
+ - `h_mesh` the horizontal mesh (defaults to `Meshes.EquiangularCubedSphere`)
+ - `h_topology` the horizontal topology (defaults to `Topologies.Topology2D`)
+ - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`). This parameter describes how data is arranged in memory. See [`Grids.SpectralElementGrid2D`](@ref) for its use.
+ - `z_mesh` the vertical mesh, defaults to an `Meshes.IntervalMesh` along `z` with given `stretch`
+ - `enable_bubble` enables the "bubble correction" for more accurate element areas when computing the spectral element space. See [`Grids.SpectralElementGrid2D`](@ref) for more information.
+
+# Example usage
+
+```julia
+using ClimaCore.CommonGrids
+grid = ExtrudedCubedSphereGrid(;
+    z_elem = 10,
+    z_min = 0,
+    z_max = 1,
+    radius = 10,
+    h_elem = 10,
+    n_quad_points = 4,
+)
+```
+"""
+ExtrudedCubedSphereGrid(; kwargs...) =
+    ExtrudedCubedSphereGrid(Float64; kwargs...)
+
+function ExtrudedCubedSphereGrid(
+    ::Type{FT};
+    z_elem::Integer,
+    z_min::Real,
+    z_max::Real,
+    radius::Real,
+    h_elem::Integer,
+    n_quad_points::Integer,
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+    stretch::Meshes.StretchingRule = Meshes.Uniform(),
+    hypsography_fun = (h_grid, z_grid) -> Grids.Flat(),
+    global_geometry::Geometry.AbstractGlobalGeometry = Geometry.ShallowSphericalGlobalGeometry(
+        radius,
+    ),
+    quad::Quadratures.QuadratureStyle = Quadratures.GLL{n_quad_points}(),
+    h_mesh = Meshes.EquiangularCubedSphere(
+        Domains.SphereDomain{FT}(radius),
+        h_elem,
+    ),
+    h_topology::Topologies.AbstractDistributedTopology = Topologies.Topology2D(
+        context,
+        h_mesh,
+    ),
+    horizontal_layout_type = DataLayouts.IJFH,
+    z_mesh::Meshes.IntervalMesh = DefaultZMesh(
+        FT;
+        z_min,
+        z_max,
+        z_elem,
+        stretch,
+    ),
+    enable_bubble::Bool = false,
+) where {FT}
+    @assert horizontal_layout_type <: DataLayouts.AbstractData
+    @assert ClimaComms.device(context) == device "The given device and context device do not match."
+
+    z_boundary_names = (:bottom, :top)
+    h_grid = Grids.SpectralElementGrid2D(
+        h_topology,
+        quad;
+        horizontal_layout_type,
+        enable_bubble,
+    )
+    z_topology = Topologies.IntervalTopology(context, z_mesh)
+    z_grid = Grids.FiniteDifferenceGrid(z_topology)
+    return Grids.ExtrudedFiniteDifferenceGrid(
+        h_grid,
+        z_grid,
+        hypsography_fun(h_grid, z_grid),
+        global_geometry,
+    )
+end
+
+"""
+    CubedSphereGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
+        radius::Real,
+        h_elem::Integer,
+        n_quad_points::Integer,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+        quad::Quadratures.QuadratureStyle = Quadratures.GLL{n_quad_points}(),
+        h_mesh = Meshes.EquiangularCubedSphere(Domains.SphereDomain{FT}(radius), h_elem),
+        h_topology::Topologies.AbstractDistributedTopology = Topologies.Topology2D(context, h_mesh),
+        horizontal_layout_type = DataLayouts.IJFH,
+    )
+
+A convenience constructor, which builds a
+[`Grids.SpectralElementGrid2D`](@ref) given:
+
+ - `FT` the floating-point type (defaults to `Float64`) [`Float32`, `Float64`]
+ - `radius` the radius of the cubed sphere
+ - `h_elem` the number of horizontal elements per side of every panel (6 panels in total)
+ - `n_quad_points` the number of quadrature points per horizontal element
+ - `device` the `ClimaComms.device`
+ - `context` the `ClimaComms.context`
+ - `quad` the quadrature style (defaults to `Quadratures.GLL{n_quad_points}`)
+ - `h_mesh` the horizontal mesh (defaults to `Meshes.EquiangularCubedSphere`)
+ - `h_topology` the horizontal topology (defaults to `Topologies.Topology2D`)
+ - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`). This parameter describes how data is arranged in memory. See [`Grids.SpectralElementGrid2D`](@ref) for its use.
+
+# Example usage
+
+```julia
+using ClimaCore.CommonGrids
+grid = CubedSphereGrid(; radius = 10, n_quad_points = 4, h_elem = 10)
+```
+"""
+CubedSphereGrid(; kwargs...) = CubedSphereGrid(Float64; kwargs...)
+function CubedSphereGrid(
+    ::Type{FT};
+    radius::Real,
+    h_elem::Integer,
+    n_quad_points::Integer,
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+    quad::Quadratures.QuadratureStyle = Quadratures.GLL{n_quad_points}(),
+    h_mesh = Meshes.EquiangularCubedSphere(
+        Domains.SphereDomain{FT}(radius),
+        h_elem,
+    ),
+    h_topology::Topologies.AbstractDistributedTopology = Topologies.Topology2D(
+        context,
+        h_mesh,
+    ),
+    horizontal_layout_type = DataLayouts.IJFH,
+) where {FT}
+    @assert horizontal_layout_type <: DataLayouts.AbstractData
+    @assert ClimaComms.device(context) == device "The given device and context device do not match."
+    return Grids.SpectralElementGrid2D(h_topology, quad; horizontal_layout_type)
+end
+
+"""
+    ColumnGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
+        z_elem::Integer,
+        z_min::Real,
+        z_max::Real,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+        stretch::Meshes.StretchingRule = Meshes.Uniform(),
+        z_mesh::Meshes.IntervalMesh = DefaultZMesh(FT; z_min, z_max, z_elem, stretch),
+    )
+
+A convenience constructor, which builds a
+[`Grids.FiniteDifferenceGrid`](@ref).
+
+# Example usage
+
+```julia
+using ClimaCore.CommonGrids
+grid = ColumnGrid(; z_elem = 10, z_min = 0, z_max = 10)
+```
+"""
+ColumnGrid(; kwargs...) = ColumnGrid(Float64; kwargs...)
+function ColumnGrid(
+    ::Type{FT};
+    z_elem::Integer,
+    z_min::Real,
+    z_max::Real,
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+    stretch::Meshes.StretchingRule = Meshes.Uniform(),
+    z_mesh::Meshes.IntervalMesh = DefaultZMesh(
+        FT;
+        z_min,
+        z_max,
+        z_elem,
+        stretch,
+    ),
+) where {FT}
+    @assert ClimaComms.device(context) == device "The given device and context device do not match."
+    z_topology = Topologies.IntervalTopology(context, z_mesh)
+    return Grids.FiniteDifferenceGrid(z_topology)
+end
+
+"""
+    Box3DGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
+        z_elem::Integer,
+        x_min::Real,
+        x_max::Real,
+        y_min::Real,
+        y_max::Real,
+        z_min::Real,
+        z_max::Real,
+        periodic_x::Bool,
+        periodic_y::Bool,
+        n_quad_points::Integer,
+        x_elem::Integer,
+        y_elem::Integer,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+        stretch::Meshes.StretchingRule = Meshes.Uniform(),
+        hypsography_fun = (h_grid, z_grid) -> Grids.Flat(),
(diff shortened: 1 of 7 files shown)


charleskawczynski:
I don't like that `ColumnGrid` returns a `FiniteDifferenceGrid`, maybe we should bike-shed that name a bit more.

Sbozzolo:
There are things that look odd to me. For example, I am looking at the sphere, the vertical resolution is hidden, while much less useful arguments like `boundary_names` are immediately customizable.

Maybe we should tackle this by collecting requirements and coming up with a proposal before implementing it, while also asking for opinions by users/scientists too. 

In this proposal, we should indentify our goals with the convenience constructors and understand what is important to be immadiately expose and settable and what is not. ClimaLand already implements something similar ([Domains](https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/Domains.jl)), can we come up with something that Land could use instead of reimplmenting it? 

The broader problem I would like to solve is that the abstractions provided by ClimaCore for computational grids are too low level for pretty much all practical applications. Maybe we don't really need convenience constructors, but another layer of abstraction that connects more with real applications (while leaving the lower level abstraction still accessible). So, science applications like Atmos and Land would use this level of abstraction instead of dealing with the more primitives objects.

Here was my attempt to identify what matters to scientists in ClimaAtmos:

https://github.com/CliMA/ClimaAtmos.jl/pull/2147

For spheres, all atmos uses is this set of parameters and options:
```
            nh_poly
            h_elem
            z_elem
            radius
            z_max
            z_stretch
            dz_bottom
            dz_top
            float_type
            enable_bubble
            topography
```
As we can see, there is no real reference to meshes, domains, topology, quadrature, and so on, which we could use as a guide to target our implementation.

(I realize it is not easy to strike the correct balance and design the best solution)

charleskawczynski:
> There are things that look odd to me. For example, I am looking at the sphere, the vertical resolution is hidden while much less useful arguments like `boundary_names` are immediately customizable.

How is vertical resolution hidden? `z_elem::Integer = 10` is a kwarg, just like `boundary_names`.

> Maybe we should tackle this by collecting requirements and coming up with a proposal before implementing it, while also asking for opinions by users/scientists too.

> In this proposal, we should indentify our goals with the convenience constructors and understand what is important to be immadiately expose and settable and what is not. ClimaLand already implements something similar ([Domains](https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/Domains.jl)), can we come up with something that Land could use instead of reimplmenting it?

I don't think that this is something that needs requirements, these are convenience constructors, not new data structures. I don't see what https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/Domains.jl is solving. I'd probably suggest that those data structures be removed (by effectively inlining all of that code). Doing so would probably greatly simplify the land code, its ClimaCore interface, and potentially performance (since there are a lot of abstract container types). This is the same exercise I performed in TurbulenceConvection (TC), and the only reason that I didn't remove TC's `Grid` abstraction was because `Nv` was not in the type space, and TC's grid basically helped prevent loads of type instabilities by storing `Nv` in the type space. Now that [``](https://github.com/CliMA/ClimaCore.jl/issues/11) is closed, TC's grid abstraction is no longer needed (and should probably be removed).

> The broader problem I would like to solve is that the abstractions provided by ClimaCore for computational grids are too low level for pretty much all practical applications. Maybe we don't really need convenience constructors, but another layer of abstraction that connects more with real applications (while leaving the lower level abstraction still accessible). So, science applications like Atmos and Land would use this level of abstraction instead of dealing with the more primitives objects. ....

I don't think that defining higher level abstractions than what ClimCore currently provides is well defined at the layer of ClimaCore. 

This looks like a really separate matter, and I'd rather keep the scope of this from creeping. So, I think the first question is, do we want to support convenience constructors for grids or not?

We can discuss https://github.com/CliMA/ClimaAtmos.jl/pull/2147 separately, but I think that is a separate discussion.

Sbozzolo:
> > There are things that look odd to me. For example, I am looking at the sphere, the vertical resolution is hidden while much less useful arguments like `boundary_names` are immediately customizable.
> 
> How is vertical resolution hidden? `z_elem::Integer = 10` is a kwarg, just like `boundary_names`.

Ah, yes, you are right. I was thinking about resolution in terms of `dz`, so this didn't register to me.

> > Maybe we should tackle this by collecting requirements and coming up with a proposal before implementing it, while also asking for opinions by users/scientists too.
> 
> > In this proposal, we should indentify our goals with the convenience constructors and understand what is important to be immadiately expose and settable and what is not. ClimaLand already implements something similar ([Domains](https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/Domains.jl)), can we come up with something that Land could use instead of reimplmenting it?
> 
> I don't think that this is something that needs requirements, these are convenience constructors, not new data structures. I don't see what https://github.com/CliMA/ClimaLand.jl/blob/main/src/shared_utilities/Domains.jl is solving. I'd probably suggest that those data structures be removed (by effectively inlining all of that code). Doing so would probably greatly simplify the land code, its ClimaCore interface, and potentially performance (since there are a lot of abstract container types). This is the same exercise I performed in TurbulenceConvection (TC), and the only reason that I didn't remove TC's `Grid` abstraction was because `Nv` was not in the type space, and TC's grid basically helped prevent loads of type instabilities by storing `Nv` in the type space. Now that [``](https://github.com/CliMA/ClimaCore.jl/issues/11) is closed, TC's grid abstraction is no longer needed (and should probably be removed).

This is the problem. What ClimaCore provides is impractical to use, so every application (from test suites to other packages) had to define additional abstractions/functions to work with ClimaCore.

> > The broader problem I would like to solve is that the abstractions provided by ClimaCore for computational grids are too low level for pretty much all practical applications. Maybe we don't really need convenience constructors, but another layer of abstraction that connects more with real applications (while leaving the lower level abstraction still accessible). So, science applications like Atmos and Land would use this level of abstraction instead of dealing with the more primitives objects. ....
> 
> I don't think that defining higher level abstractions than what ClimCore currently provides is well defined at the layer of ClimaCore.
> 
> This looks like a really separate matter, and I'd rather keep the scope of this from creeping. So, I think the first question is, do we want to support convenience constructors for grids or not?

I think the problem we are trying to solve with this PR is simplifying creating computational domains. 

I think that this is well within in the scope of ClimaCore and I think this was one of Simon's goals with introducing `Grids` (besides the technical point of making `Spaces` mutable).

Several numerical libraries have an abstraction for the computational grid. E.g.,
- [trixi](https://github.com/trixi-framework/Trixi.jl/blob/961f2e7423bd14c6d64170c296ea8238a26a2995/examples/special_elixirs/elixir_euler_ad.jl#L8-L9)
- [oceananingans](https://github.com/CliMA/Oceananigans.jl/blob/fbf05fb72624c4436c854705dc59f80a9f62a4d2/examples/internal_wave.jl#L23)
- [FEniCS](https://github.com/SciML/FEniCS.jl/blob/78541ecc8b24fb31786bd8e654342d1878138907/examples/heat_equation.jl#L15)

charleskawczynski:
> Ah, yes, you are right. I was thinking about resolution in terms of `dz`, so this didn't register to me.

👍🏻 

> This is the problem. What ClimaCore provides is impractical to use, so every application (from test suites to other packages) had to define additional abstractions/functions to work with ClimaCore.

I'm not sure I agree-- our spaces are the highest level abstraction, and I think they're practical. The ClimaCore test suite does define some helpers, but that's mostly for setting up examples and reusing code between tests. I think Atmos uses the ClimaCore data structures just fine. Regardless, let's keep the introduction of new data structures/abstractions to a separate task.

> > > The broader problem I would like to solve is that the abstractions provided by ClimaCore for computational grids are too low level for pretty much all practical applications. Maybe we don't really need convenience constructors, but another layer of abstraction that connects more with real applications (while leaving the lower level abstraction still accessible). So, science applications like Atmos and Land would use this level of abstraction instead of dealing with the more primitives objects. ....

There's a few things to unpack here. First, low level code does have advantages (e.g., less overhead, and simpler/more direct behavior). Also, it can be pretty painful to backpedal from (i.e., remove support for) higher level abstractions to make breaking changes after users have adopted them.

User-friendliness is, IMHO, a bit subjective. Regarding `computational grids are too low level for pretty much all practical applications`, if grids/spaces are too low level, what _information_ is missing from these data structures? If it's just the matter of the boilerplate needed to _make_ spaces in the first place, then the convenience constructors should help. But, if more data structures are needed, I'd first like to know what information is missing that users want, before defining new data structures that reorganize already existing information in spaces.

> I think the problem we are trying to solve with this PR is simplifying creating computational domains.

If that's the case, then this PR should help.

> I think that this is well within in the scope of ClimaCore and I think this was one of Simon's goals with introducing `Grids` (besides the technical point of making `Spaces` mutable).

This is Simon's PR that added grids: https://github.com/CliMA/ClimaCore.jl/pull/1487, it was implemented to fix https://github.com/CliMA/ClimaCore.jl/issues/1467 and https://github.com/CliMA/ClimaCore.jl/issues/1120.

> Several numerical libraries have an abstraction for the computational grid. E.g.,
> 
> * [trixi](https://github.com/trixi-framework/Trixi.jl/blob/961f2e7423bd14c6d64170c296ea8238a26a2995/examples/special_elixirs/elixir_euler_ad.jl#L8-L9)
> * [oceananingans](https://github.com/CliMA/Oceananigans.jl/blob/fbf05fb72624c4436c854705dc59f80a9f62a4d2/examples/internal_wave.jl#L23)
> * [FEniCS](https://github.com/SciML/FEniCS.jl/blob/78541ecc8b24fb31786bd8e654342d1878138907/examples/heat_equation.jl#L15)
> 
> And it could be that the solution is simply defining convenience constructors.

Yes, and we've had similar abstractions for a while-- our mesh is similar to the FEniCS and trixi links and our mesh/space is similar to the oceananigans one (as that grid also includes topology).

This goes back to my earlier question: what information is missing? All the links above are from examples, which tend to use convenience constructors for pedagogical/demonstrative reasons. Oceananigans still have relatively complex looking grids that are probably not called from examples (https://github.com/CliMA/Oceananigans.jl/blob/fbf05fb72624c4436c854705dc59f80a9f62a4d2/src/Grids/latitude_longitude_grid.jl#L3-L70 😵‍💫)

Reading through https://github.com/CliMA/ClimaAtmos.jl/pull/2147#issue-1909542051, I think that one thing that we could do is add more high-level getter-helpers that grab things from internals-- and I think that this is a perfectly fine/safe thing to do. For example, we could define `z_max(space::AbstractSpace) = space.vertical_topology.mesh.domain.coord_max.z`.

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,379 @@
...
+    context = ClimaComms.context(device),
+    stretch = Meshes.Uniform(),
+    hypsography = Flat(),
+    radius::Real = FT(6.371229e6),
+    global_geometry = Geometry.ShallowSphericalGlobalGeometry(radius),
+    h_elem::Integer = 4,
+    Nq::Integer = 4,
+    quad::Quadratures.QuadratureStyle = Quadratures.GLL{Nq}(),
+    h_domain = Domains.SphereDomain(radius),
+    h_mesh = Meshes.EquiangularCubedSphere(h_domain, h_elem),
```
charleskawczynski, line 52:
```suggestion
    h_mesh = Meshes.EquiangularCubedSphere(Domains.SphereDomain(radius), h_elem),
```
?

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,379 @@
...
+) where {FT}
+
```
charleskawczynski, line 54:
```suggestion
    @assert ClimaComms.device(context) == device "The given device and context device do not match."
```

charleskawczynski:
@Sbozzolo, I think that this may be ready for another round of review

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,366 @@
+#=
+While we recommend users to use our composable
+methods, this file contains some convenience
+constructors, which are meant to improve
+developer experience.
+=#
```
Sbozzolo, line 6: I think we should recommend using the compasable constructors only if one needs them, and try using these constructors when possible

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,366 @@
...
+    ::Type{FT};
+    z_elem::Integer,
+    x_min::Real,
+    x_max::Real,
+    y_min::Real,
+    y_max::Real,
+    z_min::Real,
+    z_max::Real,
+    x1periodic::Bool,
+    x2periodic::Bool,
```
Sbozzolo, line 184:
```suggestion
    periodic_x::Bool,
    periodic_y::Bool,
```
seems a little clearer to me

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,366 @@
...
+function SliceXZGrid(
+    ::Type{FT};
+    z_elem::Integer,
+    x_min::Real,
+    x_max::Real,
+    z_min::Real,
+    z_max::Real,
+    x1periodic::Bool,
+    Nq::Integer,
+    n1::Integer,
```
Sbozzolo, line 267:
```suggestion
    x_elem::Integer,
```
maybe? (For consistency)

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,366 @@
...
+    ::Type{FT};
+    x_min::Real,
+    x_max::Real,
+    y_min::Real,
+    y_max::Real,
+    x1periodic::Bool,
+    x2periodic::Bool,
+    Nq::Integer,
+    n1::Integer, # number of horizontal elements
+    n2::Integer, # number of horizontal elements
```
Sbozzolo, line 338:
```suggestion
    x_elem::Integer, # number of horizontal elements
    y_elem::Integer, # number of horizontal elements
```
so that it follows `z_elem`?

Sbozzolo (commented):
Thank you. I think this looks good and is very useful. 

The next step here would be to add documentation and examples so that is clear how to use these new functions. Given that these are user-facing, I think we should target a level of documentation and explanations useful to someone who is not very familiar with ClimaCore (e.g., what does `h_elem` mean?).

src/Grids/convenience_constructors.jl
charleskawczynski: Sorry it's been some time. Done!

src/Grids/convenience_constructors.jl
charleskawczynski: Done!

src/Grids/convenience_constructors.jl
charleskawczynski: Done!

charleskawczynski:
I don't really like the name `Nq`, I think I prefer `N_poly` or `N_quad_points`. Thoughts?

Sbozzolo:
> I don't really like the name `Nq`, I think I prefer `N_poly` or `N_quad_points`. Thoughts?

I agree

src/Grids/convenience_constructors.jl
charleskawczynski: I've updated the doc string for this, I think it's an improvement. Can you take another look?

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,386 @@
...
+        z_elem::Integer,
+        z_min::Real,
+        z_max::Real,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        context::ClimaComms.AbstractCommsContext = ClimaComms.context(device),
+        stretch::Meshes.StretchingRule = Meshes.Uniform(),
+    )
+
+A convenience constructor, which builds a
+`FiniteDifferenceGrid` given.
```
Sbozzolo, line 133: This sentence is incomplete

Sbozzolo (commented):
These constructors look good to me and I am looking forward seeing them used.

I think it's very important that we add examples and start converting code to using these constructors as much as possible so that people learn that this is how they can build computational grids. 

So, this PR should also:
- Add some examples on how to use the constructors.
- Add a NEWS entry that also quickly highlight how to use this (vs the old way)

I think that these functions are very useful and also very user-facing, so I would like to see the docstrings expanded to explain the various terms in a way that is accessible to someone who is not familiar to ClimaCore.

src/Grids/convenience_constructors.jl
Sbozzolo: Update looks good

charleskawczynski:
> I think it's very important that we add examples and start converting code to using these constructors as much as possible so that people learn that this is how they can build computational grids.

Why do we want to spread their use? One concern I have with that is that all of these constructors are coupled to _all_ of the components they build on-top of. So, changes to any layer beneath it requires changes for these constructors. This is not dissimilar to dependencies, lighter dependencies tend to be easier to maintain. Speaking of which, I just realized that we missed one of the arguments: `Topology2D` optionally takes `elemorder` (for space filling curve), `elempid`, and `orderindex`. Also, `h_topology` could be `Topologies.Topology2D` or `Topologies.DistributedTopology2D`, so those need to optional keyword arguments.

> * Add some examples on how to use the constructors.

Sounds good, where do you think we should add them?

> * Add a NEWS entry that also quickly highlight how to use this (vs the old way)

Sure, I'll add hyperlinks to the constructor and their doc strings.

> I think that these functions are very useful and also very user-facing, so I would like to see the docstrings expanded to explain the various terms in a way that is accessible to someone who is not familiar to ClimaCore.

Sounds good.

src/Grids/convenience_constructors.jl
charleskawczynski: I'm going to try and finish the interface before working on the docs

charleskawczynski:
My only other thought is: should we put this inside a module? It might be nice if users could do tab completion on `ClimaCore.Grids.ConvenienceConstructors.<tab>` Otherwise the list will be pretty big

charleskawczynski:
I've updated the news, but I'll update it again once the PR is merged so that I can use permanent links

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+
+A convenience constructor, which builds an
+`ExtrudedFiniteDifferenceGrid`, given:
+
+ - `FT` the float type (defaults to `Float64`) [`Float32`, `Float64`]
+ - `z_elem` the number of z-points
+ - `z_min` the domain minimum along the z-direction.
+ - `z_max` the domain maximum along the z-direction.
+ - `radius` the radius of the cubed sphere
+ - `h_elem` the number of elements per side of every panel (6 panels in total)
```
Sbozzolo, line 158:
```suggestion
 - `h_elem` the number of horizontal elements per side of every panel (6 panels in total)
```

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `n_quad_points` the number of quadrature points
+ - `device` the [`ClimaComms.device`](@ref)
+ - `context` the [`ClimaComms.context`](@ref)
```
Sbozzolo, line 161: We should add links to the documentation for ClimaComms here

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `stretch` the mesh [`Meshes.StretchingRule`](@ref) (defaults to `Uniform`) []
```
Sbozzolo, line 162:
```suggestion
 - `stretch` the mesh [`Meshes.StretchingRule`](@ref) (defaults to `Uniform`)
```

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `hypsography` the hypsography
```
Sbozzolo, line 163: Can you expand upon this?

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `global_geometry` the global geometry (defaults to `Geometry.CartesianGlobalGeometry`)
+ - `quad` the quadrature style (defaults to `Quadratures.GLL{n_quad_points}`)
+ - `h_mesh` the horizontal mesh (defaults to `Meshes.EquiangularCubedSphere`)
+ - `h_topology` the horizontal topology (defaults to `Topologies.Topology2D`)
+ - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`)
+ - `z_mesh` the z-mesh, defaults to an `Meshes.IntervalMesh` along `z` with given `stretch`
+ - `enable_bubble` enables the "bubble correction" for more accurate element areas.
```
Sbozzolo, line 170: Can you add a link to documentation explaining this?

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `device` the [`ClimaComms.device`](@ref)
+ - `context` the [`ClimaComms.context`](@ref)
+ - `stretch` the mesh [`Meshes.StretchingRule`](@ref) (defaults to `Uniform`) []
+ - `hypsography` the hypsography
+ - `global_geometry` the global geometry (defaults to `Geometry.CartesianGlobalGeometry`)
+ - `quad` the quadrature style (defaults to `Quadratures.GLL{n_quad_points}`)
+ - `h_mesh` the horizontal mesh (defaults to `Meshes.EquiangularCubedSphere`)
+ - `h_topology` the horizontal topology (defaults to `Topologies.Topology2D`)
+ - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`)
+ - `z_mesh` the z-mesh, defaults to an `Meshes.IntervalMesh` along `z` with given `stretch`
```
Sbozzolo, line 169:
```suggestion
 - `z_mesh` the vertical mesh, defaults to an `Meshes.IntervalMesh` along `z` with given `stretch`
```

src/Grids/convenience_constructors.jl
```diff
@@ -0,0 +1,680 @@
...
+ - `enable_bubble` enables the "bubble correction" for more accurate element areas.
+
```
Sbozzolo, line 171: Can you add a `See also` section where you add `ref`s and links to the ClimaCore concepts used in this function, such as `Mesh`, `Domain`, `DataLayout`, `GlobalGeometry`, et cetera.

I think it would be useful to also add a few words to describe what they encode in the list of parameters, e.g.
```
 - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`). This parameter describes how data is arranged in memory and it relevant for performance.
```

src/Grids/convenience_constructors.jl
charleskawczynski: Ah, yeah, I missed that one

charleskawczynski:
> I don't have strong opinions about putting the constructors in a module. My only thought is that I wouldn't want people to think that they shouldn't use them. To me, the name `ConvenienceConstructors` has some implication that these are not first-class citizens and I should be using something else.

But, they are convenience constructors. All they do is call the other constructors and provide a "flattened constructor" for users, so that they don't have to think about compose-ability. That said, these constructors make (hopefully useful) assumptions about what the user likely wants (e.g., no periodicity in the z-direction).

I don't care too much about the name, I'm also fine with calling the module (assuming we like the idea of putting them into a module) "CommonGrids".

> Can you please also add examples where these are used for something? I don't think that the current examples in the docstrings are enough to tell people how to use them.

Sure

> It would also be very useful to have a transition guide to distribute to current users. Something like "You are used to doing this, now you can do this".

I'm not sure where this should go, maybe if it's in the module, we could put it an example like this in the module docs? I don't want to put it in individual docs, because there is already a lot of documentation duplication, which I'm not really happy about.

> I encourage you to take this opportunity to also add a page in the documentation describing the computational grids and how to set them up and use them.

Let's leave this for a subsequent PR. I don't want to hold people up from being able to use this. I'll open an issue.

Sbozzolo:
> > I don't have strong opinions about putting the constructors in a module. My only thought is that I wouldn't want people to think that they shouldn't use them. To me, the name `ConvenienceConstructors` has some implication that these are not first-class citizens and I should be using something else.
> 
> But, they are convenience constructors. All they do is call the other constructors and provide a "flattened constructor" for users, so that they don't have to think about compose-ability. That said, these constructors make (hopefully useful) assumptions about what the user likely wants (e.g., no periodicity in the z-direction).
> 
> I don't care too much about the name, I'm also fine with calling the module (assuming we like the idea of putting them into a module) "CommonGrids".
> 

Yes, I agree with you! I just don't want the name to send the wrong message. I like `CommonGrids`.

> I'm not sure where this should go, maybe if it's in the module, we could put it an example like this in the module docs? I don't want to put it in individual docs, because there is already a lot of documentation duplication, which I'm not really happy about.

You could add it in a documentation page and add a `@ref` to the docstrings. I would also add it to the NEWS so that those who follow the changelog can immediately understand what they need to do.

charleskawczynski:
Unfortunately, I'm running issues into making this work with different Hypsography options (the default works). In particular, some hypsography constructors accept _fields_, which means that the space must have already been made. Also, we can't make an artificial one first, since it must match the horizontal grid (which is made inside the convenience constructors). In other words, we'd get an (appropriate) assertion error: `AssertionError: Spaces.grid(axes(adaption.surface)) == horizontal_grid`. So, I don't think that these convenience constructors can allow sufficient flexibility at the layer of grids. I think we can try to make these convenience constructors for spaces, instead.

I think most of the constructors will be the same, except for the FD ones, which will require `Grids.CellCenter` / `Grids.CellFace` as an additional argument.

One downside of this is that users will be required to send "instructions" on how to make the hypsography. It's a bit easier to manage this when it's done incrementally (as we do now with the compose-able interface).

src/Grids/convenience_constructors.jl
charleskawczynski: I updated the doc string and wrote to see `SpectralElement2DGrid` for more information. A lot of these arguments are forwarded, and I think we need a way to reduce documentation duplication. It'll exist here, in the space convenience constructors, and in the compose-able space construction function, where it's finally used.

charleskawczynski:
Alright, the way around this was to allow users to inject a function for computing the hypsography, and this function should basically have access to all of the input arguments.

I know that the documentation can be improved, but my preference would be to iterate on this. I think what this PR has now is an improvement over the main branch.

Can we merge this and iterate on docs?

Next, I'll open a PR that has a thin layer that introduces a `CommonSpaces` module. Thoughts?

src/Grids/convenience_constructors.jl
charleskawczynski: Something wasn't working with that, we can iterate on this, though.

src/CommonGrids/CommonGrids.jl
```diff
@@ -0,0 +1,647 @@
+"""
+    CommonGrids
+
+CommonGrids contains convenience constructors for common
+grids. Constructors in this module accept integer
+keyword inputs, so they are dynamically created. You may
```
Sbozzolo, line 6: What do you mean with "integer keyword inputs"?

src/CommonGrids/CommonGrids.jl
```diff
@@ -0,0 +1,647 @@
...
+```
+
+You may re-write this as:
+
+```julia
+using ClimaCore.CommonGrids: ExtrudedCubedSphereGrid
+grid = ExtrudedCubedSphereGrid(;
+    z_elem = 63,
+    z_min = 0,
+    z_max = 1,
```
Sbozzolo, line 61: Will this work? I often found that Integer values don't work with our spaces and we really need a float

src/CommonGrids/CommonGrids.jl
```diff
@@ -0,0 +1,647 @@
...
+ - `FT` the floating-point type (defaults to `Float64`) [`Float32`, `Float64`]
+ - `radius` the radius of the cubed sphere
+ - `h_elem` the number of horizontal elements per side of every panel (6 panels in total)
+ - `n_quad_points` the number of quadrature points per horizontal element
+ - `device` the `ClimaComms.device`
+ - `context` the `ClimaComms.context`
+ - `quad` the quadrature style (defaults to `Quadratures.GLL{n_quad_points}`)
+ - `h_mesh` the horizontal mesh (defaults to `Meshes.EquiangularCubedSphere`)
+ - `h_topology` the horizontal topology (defaults to `Topologies.Topology2D`)
+ - `horizontal_layout_type` the horizontal DataLayout type (defaults to `DataLayouts.IJFH`). This parameter describes how data is arranged in memory. See [`Grids.SpectralElementGrid2D`](@ref) for its use.
```
Sbozzolo, line 227: I think I would have preferred `h_layout_type` to be consistent with the other ones, but I think we also use `horizontal_layout_type`, so it impossible to be consistent everywhere.

Sbozzolo (approved):
Thank you! I am fine with starting with this, and we can do another round of documentation when we add the spaces. Critically, what is missing is the link with what people are doing now and we need to explain how people should use these constructors.

src/CommonGrids/CommonGrids.jl
charleskawczynski: Meaning `n_quad_points` in

```julia
grid = ExtrudedCubedSphereGrid(;
    z_elem = 10,
    z_min = 0,
    z_max = 1,
    radius = 10,
    h_elem = 10,
    n_quad_points = 4,
)
```
It's passed as an `Int`, so that's all the compiler knows. Which means, `quad = Quadratures.GLL{n_quad_points}()` inside `ExtrudedCubedSphereGrid` is a dynamically constructed type (i.e., a type instability).

It's not a performance issue for us, but if a user did this in a performance-critical section, then it could be for them (I doubt that would be the case, so I think it's okay).

The only way we could make this type stable is if we changed the keyword to positional argument and caught the argument through the type space, for example:

```julia
grid = ExtrudedCubedSphereGrid(n_quad_points; ....)
function ExtrudedCubedSphereGrid(::Type{n_quad_points}; ....) where {n_quad_points}
    ...
end
```
But this isn't really user-friendly in the way I think we want these constructors to be.

src/CommonGrids/CommonGrids.jl
charleskawczynski: Yeah, I mean, we could use a shorter name here and change it internally, but then we'd have two different names. I'm indifferent. Do you have a preference?

Sbozzolo: No, up to you.

src/CommonGrids/CommonGrids.jl
Sbozzolo: Thank, I understand now. I think the phrasing is unclear, so I'd maybe add an explanation like this somewhere in the comments (or you could add a subsection in the module docstring).

src/CommonGrids/CommonGrids.jl
charleskawczynski: I'm going to leave this option verbose since it's more of an internal detail, rather than a grid parameter.

src/CommonGrids/CommonGrids.jl
charleskawczynski: Yes, we type-cast everything inside to `FT` (which defaults to `Float64`), so this interface allows us to pass in `Int`s.

src/CommonGrids/CommonGrids.jl
charleskawczynski: I'll re-phrase/simplify it a bit.

PR #1850, opened 2024-06-28
Fix type instability in `copyto_stencil_kernel!`

charleskawczynski:
This PR removes the `window` variable in `copyto_stencil_kernel!`, and instead, uses different windows depending on `gid`.

I'm actually not sure how the gpu even allows this block of code:

```julia
            (v, i, j, h) = cart_ind((Nv, Nq, Nq, Nh), gid).I
            hidx = (i, j, h)
            idx = v - 1 + li
            window =
                idx < lw ?
                LeftBoundaryWindow{Spaces.left_boundary_name(space)}() :
                (
                    idx > rw ?
                    RightBoundaryWindow{Spaces.right_boundary_name(space)}() :
                    Interior()
                )
            setidx!(
                space,
                out,
                idx,
                hidx,
                Operators.getidx(space, bc, window, idx, hidx),
            )
```
because

 - the type of `window` depends on the value of `idx`
 - the value of `idx` depends on the value of `v`
 - the value of `v` depends on the value of `gid`
 - the value of `gid` depends on the individual thread, which is most definitely non-deterministic

The only way I could see this being type stable is if the compiler can determine that all three branches result in the same return type, and that, then, it doesn't matter what the value of `window` will be within the scope of `copyto_stencil_kernel!`.

This is surprising to me. Any insights @vchuravy?

### DIFF
--- a/ext/cuda/operators_finite_difference.jl
+++ b/ext/cuda/operators_finite_difference.jl
@@ -55,21 +55,18 @@ function copyto_stencil_kernel!(out, bc, space, bds, us)
             (v, i, j, h) = cart_ind((Nv, Nq, Nq, Nh), gid).I
             hidx = (i, j, h)
             idx = v - 1 + li
-            window =
-                idx < lw ?
-                LeftBoundaryWindow{Spaces.left_boundary_name(space)}() :
-                (
-                    idx > rw ?
-                    RightBoundaryWindow{Spaces.right_boundary_name(space)}() :
-                    Interior()
-                )
-            setidx!(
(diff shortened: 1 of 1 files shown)


vchuravy:
The compiler may actual perform this transformation for you? We can probably union-split this.

> I'm actually not sure how the gpu even allows this block of code:

Then this is likely happening

charleskawczynski:
Alright, I refactored this a bit, and this is now fewer lines of code and type stable. So I'm happy to move forward with it.

PR #1852, opened 2024-06-28
Update compats and github workflows

Sbozzolo:
Closes #1849, #1734, #1717, #1705, #1706


examples/bickleyjet/bickleyjet_cg_invariant_hypervisc.jl
```diff
@@ -14,7 +14,7 @@ import ClimaCore:
 using OrdinaryDiffEq: ODEProblem, solve, SSPRK33
 
 using Logging
-ClimaComms.@import_required_backends
+@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
```
charleskawczynski, line 17: I don't know that it's worth adding these conditionals -- to do this properly / thoroughly, we would also need to add similar conditionals around every new feature since ClimaComms version 0.6 (e.g., `ClimaComms.allowscalar` and more), could we just require 0.6 for this version instead? Also, to be sure that it was done correctly, we should probably get #1781 working.

In the same breadth, if this is for some reason a headache for you for testing purposes, I'm fine with temporarily adding it.

charleskawczynski (approved):
I support the compat/GHA related changes, but I'm a bit hesitant on the `@static pkgversion(ClimaComms) >= v"0.6" &&` additions. I left a more detailed message about that. I'd prefer we remove them, but please have a look at the more detailed message.

examples/bickleyjet/bickleyjet_cg_invariant_hypervisc.jl
Sbozzolo: Okay, then we should bump the compat to `0.6`

Sbozzolo:
> I support the compat/GHA related changes, but I'm a bit hesitant on the `@static pkgversion(ClimaComms) >= v"0.6" &&` additions. I left a more detailed message about that. I'd prefer we remove them, but please have a look at the more detailed message.

Yes, I think we are over the update to 0.6, so I don't have problems with just bumping the compat.

examples/bickleyjet/bickleyjet_cg_invariant_hypervisc.jl
Sbozzolo: I removed the conditional and bumped the compat to 0.6

Sbozzolo:
https://buildkite.com/clima/climacore-ci/builds/3953#01906f2f-ff58-4d80-80d1-7c4b5e2d4403

```
Number of JET failures: Test Failed at /central/scratch/esm/slurm-buildkite/climacore-ci/3953/climacore-ci/test/Spaces/opt_spaces.jl:19
  Expression: n_found ≤ n_allowed
   Evaluated: 295 ≤ 287
```

For `test_n_failures(287,  TU.SpectralElementSpace2D, context)`

I am thinking that the only possible explanation is that now we use a newer version of `BlockArrays`?

charleskawczynski:
> https://buildkite.com/clima/climacore-ci/builds/3953#01906f2f-ff58-4d80-80d1-7c4b5e2d4403
> 
> ```
> Number of JET failures: Test Failed at /central/scratch/esm/slurm-buildkite/climacore-ci/3953/climacore-ci/test/Spaces/opt_spaces.jl:19
>   Expression: n_found ≤ n_allowed
>    Evaluated: 295 ≤ 287
> ```
> 
> For `test_n_failures(287, TU.SpectralElementSpace2D, context)`
> 
> I am thinking that the only possible explanation is that now we use a newer version of `BlockArrays`?

Yeah, let's just bump the limit and upgrade our dependencies. The type instability from `get!(Cache.OBJECT_CACHE)` complicates how many inference failures are reported, so it's possible that upgrading is not even an issue.

PR #1853, opened 2024-06-28
Break multiply_matrix_at_index into smaller functions

charleskawczynski:
This PR breaks `multiply_matrix_at_index` into smaller functions.


charleskawczynski:
Ah, `test_scalar_15` is now taking an hour after this PR. That said, maybe it's an argument that it's making it farther before returning a failure status? 🤷🏻

PR #1861, opened 2024-07-01
Fix gpu dispatching on ReshapedArrays

charleskawczynski:
This PR fixes gpu dispatching when the parent array fits into one of the following categories:

```julia
const cu_array = CUDA.CuArray
const TSubArray{T} = SubArray{<:Any, <:Any, <:T}
const TReshapedArray{T} = Base.ReshapedArray{<:Any, <:Any, <:T}

const CuArrayBackedTypes = Union{
    cu_array,
    TSubArray{cu_array},
    TSubArray{TReshapedArray{cu_array}},
    TSubArray{TReshapedArray{TSubArray{cu_array}}},
    TReshapedArray{cu_array},
    TReshapedArray{TSubArray{cu_array}},
}
```

After using `launch_configuration` for `VIJFH` datalayouts, this should fix #1854 (xref: https://github.com/CliMA/ClimaCore.jl/issues/1854#issuecomment-2200116511).

This PR also changes to using CUDA's `launch_configuration` for the VF datalayouts, I'll take a look at the stencil benchmarks to see if/how the performance changes.

I've added a test only for DataLayouts `fill!`, as we use `CuArrayBackedTypes` for the `copyto!` kernels as well. We should probably add correctness tests for these other cases, as these composite arrays (and presumably their indexing) is, I imagine, quite complex.

The type declaration for `CuArrayBackedTypes` is rather complicated, and it seems impossible to define this for all possible types. I think we might need to switch to a dispatch solution, by recursing through the parent array until we either find `Array` or `CuArray`, and then dispatch on a 3rd argument.

PR #1863, opened 2024-07-01
CPU-GPU specialization -> dispatching

charleskawczynski:
This PR is an alternative to #1861, and I'm feeling more inclined to move forward with this design. This PR switches from using specialization on `CuArrayBackedTypes`, which required naming specific types:

```julia
const cu_array = CUDA.CuArray
const TSubArray{T} = SubArray{<:Any, <:Any, <:T}
const TReshapedArray{T} = Base.ReshapedArray{<:Any, <:Any, <:T}

const CuArrayBackedTypes = Union{
    cu_array,
    TSubArray{cu_array},
    TSubArray{TReshapedArray{cu_array}},
    TSubArray{TReshapedArray{TSubArray{cu_array}}},
    TReshapedArray{cu_array},
    TReshapedArray{TSubArray{cu_array}},
}
```
that could be passed in, to using dispatch on either `Array` or `CuArray`. The advantage of dispatching is that we no longer have to account for every possible way that these types are composed (e.g., the above definition will break on `TSubArray{TSubArray{cu_array}}` and `TSubArray{TSubArray{TSubArray{cu_array}}}`). The way that this PR achieves this is by having the top-level methods (`Base.copyto!` and `Base.fill!`) call (for example) `Base.copyto!(dest, bc, backed_array(dest, bc))`. `backed_array` does some checking, and then calls an internal `_backed_array` which is recursively defined for all array types that we support:

```julia
_backed_array(x::CUDA.CuArray) = x # in extension
_backed_array(x::Array) = x
_backed_array(x::SubArray) = _backed_array(parent(x))
_backed_array(x::Base.ReshapedArray) = _backed_array(parent(x))
_backed_array(x::AbstractData) = _backed_array(parent(x))
```

This way, every possible combination of CPU/GPU-backed `SubArray` and `ReshapedArray` are now supported and should "just work".

The price we pay is that now we're dispatching on 3 arguments (e.g., `fill!(::VF, val, ::Array)`), which is not as pretty, but will not involve the existing whack-a-mole situation.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -40,6 +40,11 @@ import Adapt
 import ..slab, ..slab_args, ..column, ..column_args, ..level
 export slab, column, level, IJFH, IJF, IFH, IF, VF, VIJFH, VIFH, DataF
 
+# Internal types for managing CPU/GPU dispatching
+abstract type AbstractDispatchToDevice end
+struct ToCPU <: AbstractDispatchToDevice end
+struct ToCUDA <: AbstractDispatchToDevice end
+
 include("struct.jl")
 
 abstract type AbstractData{S} end
@@ -325,7 +330,7 @@ function Base.size(data::IJFH{S, Nij}) where {S, Nij}
     (Nij, Nij, 1, Nv, Nh)
 end
 
-function Base.fill!(data::IJFH, val)
+function Base.fill!(data::IJFH, val, ::ToCPU)
     (_, _, _, _, Nh) = size(data)
     @inbounds for h in 1:Nh
         fill!(slab(data, h), val)
@@ -494,7 +499,7 @@ function Base.size(data::IFH{S, Ni}) where {S, Ni}
     (Ni, 1, 1, Nv, Nh)
 end
 
-function Base.fill!(data::IFH, val)
+function Base.fill!(data::IFH, val, ::ToCPU)
     (_, _, _, _, Nh) = size(data)
     @inbounds for h in 1:Nh
         fill!(slab(data, h), val)
@@ -618,7 +623,7 @@ function DataF(x::T) where {T}
 end
 
 
-function Base.fill!(data::DataF, val)
+function Base.fill!(data::DataF, val, ::ToCPU)
     @inbounds data[] = val
     return data
 end
@@ -746,7 +751,7 @@ end
 function Base.size(data::IJF{S, Nij}) where {S, Nij}
     return (Nij, Nij, 1, 1, 1)
 end
-function Base.fill!(data::IJF{S, Nij}, val) where {S, Nij}
+function Base.fill!(data::IJF{S, Nij}, val, ::ToCPU) where {S, Nij}
     @inbounds for j in 1:Nij, i in 1:Nij
         data[i, j] = val
     end
@@ -884,7 +889,7 @@ function replace_basetype(data::IF{S, Ni}, ::Type{T}) where {S, Ni, T}
     return IF{S′, Ni}(similar(array, T))
 end
 
-function Base.fill!(data::IF{S, Ni}, val) where {S, Ni}
+function Base.fill!(data::IF{S, Ni}, val, ::ToCPU) where {S, Ni}
     @inbounds for i in 1:Ni
         data[i] = val
     end
@@ -998,7 +1003,7 @@ Base.size(data::VF{S, Nv}) where {S, Nv} = (1, 1, 1, Nv, 1)
 
 nlevels(::VF{S, Nv}) where {S, Nv} = Nv
 
-function Base.fill!(data::VF, val)
+function Base.fill!(data::VF, val, ::ToCPU)
     Nv = nlevels(data)
     @inbounds for v in 1:Nv
         data[v] = val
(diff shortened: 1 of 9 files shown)


charleskawczynski:
This PR also removes

```julia

function Base.copyto!(
    dest::VF{S, Nv},
    bc::DataLayouts.BroadcastedUnionVF{S, Nv, A},
) where {S, Nv, A <: CuArrayBackedTypes}
    _, _, _, _, Nh = size(dest)
    if Nv > 0 && Nh > 0
        auto_launch!(
            knl_copyto!,
            (dest, bc),
            dest;
            threads_s = (1, 1),
            blocks_s = (Nh, Nv),
        )
    end
    return dest
end
```
as this doesn't seem to perform better than the simpler and flat implementation.

Sbozzolo:
Do think it is possible to shomehow also provide a two-argument `fill` and `copyto`? I think it would be very nice to do so because we would follow the same style as `Base`

charleskawczynski:
> Do think it is possible to shomehow also provide a two-argument `fill` and `copyto`? I think it would be very nice to do so because we would follow the same style as `Base`

The top-level methods are two-argument based:
```
Base.fill!(dest::AbstractData, val) = Base.fill!(dest, val, backed_array(dest))
Base.copyto!(
    dest::AbstractData,
    bc::Union{AbstractData, Base.Broadcast.Broadcasted},
) = Base.copyto!(dest, bc, backed_array(dest))
```
but they immediately call the 3-argument methods in order to properly dispatch.

Sbozzolo:
> > Do think it is possible to shomehow also provide a two-argument `fill` and `copyto`? I think it would be very nice to do so because we would follow the same style as `Base`
> 
> The top-level methods are two-argument based:
> 
> ```
> Base.fill!(dest::AbstractData, val) = Base.fill!(dest, val, backed_array(dest))
> Base.copyto!(
>     dest::AbstractData,
>     bc::Union{AbstractData, Base.Broadcast.Broadcasted},
> ) = Base.copyto!(dest, bc, backed_array(dest))
> ```
> 
> but they immediately call the 3-argument methods in order to properly dispatch.

Perfect! I missed this!

charleskawczynski:
Ah, some of the `backed_array` calls lead to `MArray`s. I can't think of a good name for `Union{Array, MArray}`, so instead I've defined new types:

```julia
abstract type AbstractDispatchToDevice end
struct ToCPU <: AbstractDispatchToDevice end
struct ToCUDA <: AbstractDispatchToDevice end
```
and renamed `backed_array` to `dispatch_device`:

```julia
device_dispatch(dest::AbstractData) = _device_dispatch(dest)

_device_dispatch(x::Array) = ToCPU()
_device_dispatch(x::SubArray) = _device_dispatch(parent(x))
_device_dispatch(x::Base.ReshapedArray) = _device_dispatch(parent(x))
_device_dispatch(x::AbstractData) = _device_dispatch(parent(x))
_device_dispatch(x::SArray) = ToCPU()
_device_dispatch(x::MArray) = ToCPU()
```

And catch CPU/GPU code paths accordingly.

This is perhaps just a symptom of not having the ClimaComms context/device not stored in the datalayouts. But, I think doing that would be pretty breaking, so let's just improve the existing design for now, and we can tackle the comms context part separately if we deem it important enough.

src/DataLayouts/DataLayouts.jl
```diff
@@ -40,6 +40,10 @@ import Adapt
 import ..slab, ..slab_args, ..column, ..column_args, ..level
 export slab, column, level, IJFH, IJF, IFH, IF, VF, VIJFH, VIFH, DataF
 
+abstract type AbstractDispatchToDevice end
+struct ToCPU <: AbstractDispatchToDevice end
+struct ToCUDA <: AbstractDispatchToDevice end
```
charleskawczynski, line 45: I'll add a comment here:

`# Types for managing CPU/GPU dispatching`

src/DataLayouts/DataLayouts.jl
```diff
@@ -1610,4 +1614,22 @@ array2data(
     ::VIJFH{<:Any, Nv, Nij},
 ) where {T, Nv, Nij} = VIJFH{T, Nv, Nij}(reshape(array, Nv, Nij, Nij, 1, :))
 
+"""
+    device_dispatch(data::AbstractData)
+
+Returns an `Array` or a `CUDA.CuArray` depending
+on how `data` is backed.
```
charleskawczynski, line 1621:
```suggestion
Returns an `ToCPU` or a `ToCUDA` for CPU
and CUDA-backed arrays accordingly.
```

Sbozzolo (approved):

charleskawczynski:
Ah, interesting, there _is_ a catastrophic case. On main:

```julia
function Base.copyto!(
    dest::VIFH{S, Nv, Ni},
    bc::BroadcastedUnionVIFH{S, Nv, Ni},
) where {S, Nv, Ni}
    (_, _, _, _, Nh) = size(bc)
    # copy contiguous columns
    @inbounds for h in 1:Nh, i in 1:Ni
        col_dest = column(dest, i, h)
        col_bc = column(bc, i, h)
        copyto!(col_dest, col_bc)
    end
    return dest
end
```
is executed for when running on the CPU _and_ the GPU. It works because, on the CPU, we launch `Nh`*`Ni` kernel launches (`column` is defined on CPU and GPU for broadcasted objects), 1 per column!

We likely never run into this, but if we do / were to, this fix will be a _major_ performance improvement.

PR #1886, opened 2024-07-16
Define rcompare and rprint_diff for FieldVectors

charleskawczynski:
This PR defines two user-facing pieces:

 - A function, `rcompare(...)::Bool`, indicating that two fieldvectors recursively satisfy `==`
 - A macro, `rprint_diff`, which prints differences between two fieldvectors (only when `==` fails)

These  are meant to be used for debugging and testing.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -361,3 +361,80 @@ import ClimaComms
 ClimaComms.array_type(x::FieldVector) = promote_type(
     UnrolledFunctions.unrolled_map(ClimaComms.array_type, _values(x))...,
 )
+
+function __rprint_diff(
+    io::IO,
+    x::T,
+    y::T;
+    pc,
+    xname,
+    yname,
+) where {T <: Union{FieldVector, Field, DataLayouts.AbstractData, NamedTuple}}
+    for pn in propertynames(x)
+        pc_full = (pc..., ".", pn)
+        xi = getproperty(x, pn)
+        yi = getproperty(y, pn)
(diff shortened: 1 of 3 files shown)


src/Fields/fieldvector.jl
```diff
@@ -361,3 +361,80 @@ import ClimaComms
...
+# Recursively compare contents of similar fieldvectors
+_rcompare(pass, x::T, y::T) where {T <: Field} =
+    pass && _rcompare(pass, field_values(x), field_values(y))
+_rcompare(pass, x::T, y::T) where {T <: DataLayouts.AbstractData} =
+    pass && (parent(x) == parent(y))
+_rcompare(pass, x::T, y::T) where {T} = pass && (x == y)
+
+function _rcompare(pass, x::T, y::T) where {T <: FieldVector}
+    for pn in propertynames(x)
+        pass &= _rcompare(pass, getproperty(x, pn), getproperty(y, pn))
```
szy21, line 429: Why do we need `pass` as an input for `_rcompare` here?

szy21 (approved):

src/Fields/fieldvector.jl
charleskawczynski: Good question. I think we might be able to just remove `pass` from all `_compare` definitions except this one. That said, this implementation looks kind of safe in that `pass` is always joined with the condition.

charleskawczynski: I'm going to merge for now, but I'll see if I can simplify later

Issue #1888, opened 2024-07-18
Forward `fill!` to parent


charleskawczynski:
Our `fill!` method follows our `copyto!` pattern, which uses `CartesianIndex`ing, which is suboptimal especially in the context of `fill!`, so, we should just forward all of our `fill!` methods to `fill!(parent(data), val)`. Here's a local benchmark on an A100 at our target resolution:

```julia
julia> @benchmark CUDA.@sync fill!(X.x1, 1)
BenchmarkTools.Trial: 2911 samples with 1 evaluation.
 Range (min … max):  52.190 μs … 2.701 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      2.440 ms             ┊ GC (median):    0.00%
 Time  (mean ± σ):    1.715 ms ± 1.091 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █     ▁▃ ▁                                           ▇█▇▅▃ ▁
  █▁▁▁▁▁██▇█▆█▅▇█▆▅▄▄▅▇▃▃▃▅▆▃▁▁▃▁▃▃▅▄▁▇▁▃▁▁▁▁▁▁▁▃▃▇▄▅▅▆█████ █
  52.2 μs      Histogram: log(frequency) by time     2.62 ms <

 Memory estimate: 592 bytes, allocs estimate: 19.

julia> @benchmark CUDA.@sync fill!(parent(X.x1), 1)
BenchmarkTools.Trial: 4535 samples with 1 evaluation.
 Range (min … max):  20.560 μs … 2.682 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     25.200 μs             ┊ GC (median):    0.00%
 Time  (mean ± σ):    1.101 ms ± 1.185 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █     ▁▂                                             ▅▅▆▄▂ ▁
  █▁▁▁▄▁██▇█▆▆▇▇▇▇▅▁▅▃▇▁▃▄▄▅▁▃▄▅▃▁▁▁▁▄▇▁▃▃▁▁▁▃▁▃▁▃▇▆▅▄▄█████ █
  20.6 μs      Histogram: log(frequency) by time     2.59 ms <

 Memory estimate: 128 bytes, allocs estimate: 7.
```

charleskawczynski:
We really just need to fix #1910, since the implementation in Base/CUDA is likely using linear indexing.

PR #1890, opened 2024-07-18
Forward fill to parent array

charleskawczynski:
Closes #1888.

PR #1894, opened 2024-07-22
Add Nh to type parameter space

### LINKED ISSUE #11: Add size to parameter space of data layout

charleskawczynski:
This PR adds `Nh` to the type parameter space. This seems to be needed for GPU performance, as indicated by the recently added performance benchmark script results:

```julia
at_dot_call!($X_array, $Y_array):
     6 milliseconds, 775 microseconds
at_dot_call!($X_vector, $Y_vector):
     2 milliseconds, 834 microseconds
custom_sol_kernel!($X_vector, $Y_vector, $(Val(N))):
     2 milliseconds, 547 microseconds
custom_kernel_bc!($X_vector, $Y_vector, $us; printtb = false):
     2 milliseconds, 561 microseconds
custom_kernel_bc!($X_array, $Y_array, $us; printtb = false, use_pw = false):
     4 milliseconds, 160 microseconds
custom_kernel_bc!($X_array, $Y_array, $us; printtb = false, use_pw = true):
     2 milliseconds, 584 microseconds
custom_kernel_bc!($X_vector, $Y_vector, $uss; printtb = false):
     2 milliseconds, 540 microseconds
custom_kernel_bc!($X_array, $Y_array, $uss; printtb = false, use_pw = false):
     2 milliseconds, 715 microseconds
custom_kernel_bc!($X_array, $Y_array, $uss; printtb = false, use_pw = true):
     2 milliseconds, 547 microseconds
```

Where `us` and `uss` are the "universal sizes" and "static universal sizes" respectively. The only difference is how `Nh` is stored (dynamically vs statically). It's clear from this benchmark that simply moving `Nh` into the type domain, (some) of our kernels can improve.

Not all of them will because, for example, our `copyto!` kernel is using static ranges since it only accesses `size(dest, 4)`:

```julia
function knl_copyto!(dest, src)

    i = CUDA.threadIdx().x
    j = CUDA.threadIdx().y

    h = CUDA.blockIdx().x
    v = CUDA.blockDim().z * (CUDA.blockIdx().y - 1) + CUDA.threadIdx().z

    if v <= size(dest, 4)
        I = CartesianIndex((i, j, 1, v, h))
        @inbounds dest[I] = src[I]
    end
    return nothing
end
```

This is likely why the "flat" implementation:

```julia
function knl_copyto_flat!(dest::AbstractData, bc)
    @inbounds begin
        n = size(dest)
        tidx = thread_index()
        if valid_range(tidx, prod(n))
            I = kernel_indexes(tidx, n)
            dest[I] = bc[I]
        end
    end
    return nothing
end
```
was being outperformed.

Also, I noticed in our `single_field_solve_kernel!` can improve on this:

```julia
function single_field_solve_kernel!(device, cache, x, A, b)
    idx = CUDA.threadIdx().x + (CUDA.blockIdx().x - 1) * CUDA.blockDim().x
    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
    if idx <= Ni * Nj * Nh
        (i, j, h) = CartesianIndices((1:Ni, 1:Nj, 1:Nh))[idx].I

        _single_field_solve!(
            device,
            Spaces.column(cache, i, j, h),
            Spaces.column(x, i, j, h),
            Spaces.column(A, i, j, h),
            Spaces.column(b, i, j, h),
        )
    end
    return nothing
end
```

Rather than using `Val` everywhere (which will result in uglier interfaces and many creations of runtime types on the CPU), I think it's worth just moving this into the type space, so that we get this performance improvement by default.

Closes #11 for good.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -186,6 +186,10 @@ walltime = @elapsed sol = OrdinaryDiffEq.solve!(integrator)
 any(isnan, sol.u[end]) && error("NaNs found in result.")
 
 if is_distributed # replace sol.u on the root processor with the global sol.u
+    global_Y_c_1 =
+        DataLayouts.gather(comms_ctx, Fields.field_values(sol.u[1].c))
+    global_Y_f_1 =
+        DataLayouts.gather(comms_ctx, Fields.field_values(sol.u[1].f))
     if ClimaComms.iamroot(comms_ctx)
         global_h_space = make_horizontal_space(
             horizontal_mesh,
@@ -194,14 +198,10 @@ if is_distributed # replace sol.u on the root processor with the global sol.u
         )
         global_center_space, global_face_space =
             make_hybrid_spaces(global_h_space, z_max, z_elem; z_stretch)
-        global_Y_c_type = Fields.Field{
-            typeof(Fields.field_values(Y.c)),
-            typeof(global_center_space),
-        }
-        global_Y_f_type = Fields.Field{
-            typeof(Fields.field_values(Y.f)),
-            typeof(global_face_space),
-        }
+        global_Y_c_type =
+            Fields.Field{typeof(global_Y_c_1), typeof(global_center_space)}
+        global_Y_f_type =
+            Fields.Field{typeof(global_Y_f_1), typeof(global_face_space)}
         global_Y_type = Fields.FieldVector{
             FT,
             NamedTuple{(:c, :f), Tuple{global_Y_c_type, global_Y_f_type}},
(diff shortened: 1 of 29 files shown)


charleskawczynski:
Ahh, fortunately, our stencils should see an improvement:

```julia
    ::Val{Nv},
    ::Val{Nq},
    Nh,
) where {Nv, Nq}
    @inbounds begin
        gid = threadIdx().x + (blockIdx().x - 1) * blockDim().x
        if gid ≤ Nv * Nq * Nq * Nh
            (li, lw, rw, ri) = bds
            (v, i, j, h) = cart_ind((Nv, Nq, Nq, Nh), gid).I
```

charleskawczynski:
Closes #1888, supersedes #1890.

PR #1903, opened 2024-07-25
Add column_accumulate and generalize column_reduce

dennisYatunin:
This PR adds a `column_accumulate!` function that can `accumulate` values along each column of a `Field`, a pointwise `Broadcasted` object, or a columnwise `StencilBroadcasted` object. It also generalizes `column_mapreduce!`, which could only accept `Field`s, to `column_reduce!`, which can accept `Broadcasted` and `StencilBroadcasted` objects as well.

In addition, the `column_integral_definite!` and `column_integral_indefinite!` functions have been redefined in terms of `column_reduce!` and `column_accumulate!`, respectively. This API will generally allow us to define arbitrary "vertical sweep" operations without the need for custom kernel functions or external caches, which should allow us to simplify the non-orographic gravity wave parametrization in ClimaAtmos.

This PR also updates ClimaComms to 0.6.4 in all Manifest files to make use of the new device-flexible `assert` macro.

### DIFF
--- a/src/Operators/integrals.jl
+++ b/src/Operators/integrals.jl
@@ -1,315 +1,333 @@
-import ..RecursiveApply: rzero, ⊠, ⊞, radd, rmul
-import ..Utilities
-import ..DataLayouts
+import ..RecursiveApply: rzero, ⊠, ⊞
 import RootSolvers
 import ClimaComms
 
 """
-    column_integral_definite!(∫field::Field, ᶜfield::Field)
+    column_integral_definite!(∫field, ᶜfield, [init])
 
-Sets `∫field```{}= \\int_0^{z_{max}}\\,```ᶜfield```(z)\\,dz``, where ``z_{max}``
-is the value of `z` at the top of the domain. The input `ᶜfield` must lie on a
-cell-center space, and the output `∫field` must lie on the corresponding
-horizontal space.
+Sets `∫field```{}= \\int_{z_{min}}^{z_{max}}\\,```ᶜfield```(z)\\,dz +{}```init`,
+where ``z_{min}`` and ``z_{max}`` are the values of `z` at the bottom and top of
+the domain, respectively. The input `ᶜfield` must lie on a cell-center space,
+and the output `∫field` must lie on the corresponding horizontal space. The
+default value of `init` is 0.
 """
-column_integral_definite!(∫field::Fields.Field, ᶜfield::Fields.Field) =
-    column_integral_definite!(ClimaComms.device(axes(ᶜfield)), ∫field, ᶜfield)
-
-function column_integral_definite_kernel!(
-    ∫field,
-    ᶜfield::Fields.CenterFiniteDifferenceField,
-)
-    _column_integral_definite!(∫field, ᶜfield)
-    return nothing
-end
-
-column_integral_definite!(
-    ::ClimaComms.AbstractCPUDevice,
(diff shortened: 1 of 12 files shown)


src/Operators/integrals.jl
```diff
@@ -1,315 +1,371 @@
...
+    ) do (ϕ_prev, ϕ_prev_implicit), (z, Δz)
+        ϕ_new_implicit_residual(ϕ_new_implicit) =
+            (ϕ_new_implicit - ϕ_prev_implicit) / Δz -
+            ∂ϕ∂z(average(ϕ_prev_implicit, ϕ_new_implicit), z)
+        ϕ_new_implicit =
+            RootSolvers.find_zero(
+                ϕ_new_implicit_residual,
+                RootSolvers.NewtonsMethodAD(ϕ_prev_implicit),
+                RootSolvers.CompactSolution(),
+            ).root
```
charleskawczynski, line 143: Should we do error handling here? (what if `find_zero` does not converge?)

src/Operators/integrals.jl
```diff
@@ -1,315 +1,371 @@
...
+        NamedTuple{keys(op.bcs)}(
+            unrolled_map(extract_column(colidx), values(op.bcs)),
+        ),
+    )
+extract_column(bc::PointwiseOrColumnwiseBroadcasted, colidx) =
+    Base.Broadcast.broadcasted(
+        bc isa Operators.StencilBroadcasted ? extract_column(bc.op, colidx) :
+        bc.f,
+        unrolled_map(extract_column(colidx), bc.args)...,
+    )
```
charleskawczynski, line 29: How is this different from just calling `column`?

src/Operators/integrals.jl
```diff
@@ -1,315 +1,371 @@
...
+    input,
+    init,
+) where {F, T}
+    first_level = left_boundary_idx(axes(input))
+    last_level = right_boundary_idx(axes(input))
+    @inbounds if init == UnspecifiedInit()
+        reduced_value = column_value_at_level(input, first_level)
+        level_range = (first_level + 1):last_level
+    else
+        reduced_value = init
```
charleskawczynski, line 237:
```suggestion
        reduced_value::typeof(init) = init
```
It might turn out to be a good idea to this assertion here, so that the compiler doesn't need to infer through the entire `level_range` (since we're not doing `reduce_value .= ...`)

charleskawczynski: (and same above)

src/Operators/integrals.jl
charleskawczynski: If this is needed, can we add some unit tests for it?

charleskawczynski (commented):
Looks good to me, I left some comments, my main comment is: can we add some more granular unit tests?

dennisYatunin:
I've enabled GPU support and added more unit tests. Adding the error handling for `find_zero` actually revealed that one of the profiles in ClimaAtmos was not converging for the default absolute tolerance of 0.001, so I changed it to a relative tolerance of 0.001. I'm going to move the device-specific `@assert` to ClimaComms, and then this will be ready to merge in. @charleskawczynski, could you please take another look?

src/Operators/integrals.jl
dennisYatunin: There aren't any type instabilities in the GPU tests at the moment, so it doesn't look like the additional typecast makes a difference here.

PR #1920, opened 2024-08-09
Remove multiple integer indexing in DataLayouts

charleskawczynski:
This PR removes multiple integer indexing in DataLayouts, which basically duplicates logic of the `CartesianIndex`ing. We will need to fix wherever we call this style indexing, but it also will allow us to use a single integer for linear indexing. This is technically breaking, but I don't think anyone is (or should be) using this.


charleskawczynski:
This somehow breaks CI on Julia 1.11.

We can easily add back this functionality all within DataLayouts. So let's merge and figure out 1.11 when we need, and we can revert in DataLayouts if it comes to that. I'll open an issue to track.

PR #1919, opened 2024-08-09
Add linear index support in get_struct

charleskawczynski:
Add support for linear indexing.

PR #1922, opened 2024-08-09
Add support for linear indexing for pointwise kernels

charleskawczynski:
Extends / supersedes #1919. Depends on #1920.

### DIFF
--- a/ext/cuda/data_layouts.jl
+++ b/ext/cuda/data_layouts.jl
@@ -54,3 +54,16 @@ function Adapt.adapt_structure(
         end,
     )
 end
+
+import Adapt
+import CUDA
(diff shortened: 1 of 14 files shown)


charleskawczynski:
Just to leave some breadcrumbs here:

This is basically what Tim resorted to in https://github.com/JuliaGPU/GPUArrays.jl/pull/520. We can't really do this unless we either drop the field dimension, or move the field dimension to the end (e.g., `VIJHF`), which will allow us to have simple stride.

I think that moving the field dimension to the end is probably the most ideal solution, but it is a breaking change.

charleskawczynski:
Superseded by #2055.

PR #1930, opened 2024-08-13
add hyperbolic tangent stretching

### LINKED ISSUE #1927: add hyperbolic tangent stretching
Following Appendix A in the dycore paper.

szy21:
Closes #1927. Also fixes a bug in stretching for non-zero z_bottom.

- [ ] Code is exercised in an integration test OR N/A. (This will be tested in ClimaAtmos)

### DIFF
--- a/src/Meshes/interval.jl
+++ b/src/Meshes/interval.jl
@@ -16,6 +16,7 @@ Constuct a 1D mesh on `domain` with `nelems` elements, using `stretching`. Possi
 - [`Uniform()`](@ref)
 - [`ExponentialStretching(H)`](@ref)
 - [`GeneralizedExponentialStretching(dz_bottom, dz_top)`](@ref)
+- [`HyperbolicTangentStretching(dz_bottom)`](@ref)
 """
 struct IntervalMesh{I <: IntervalDomain, V <: AbstractVector} <: AbstractMesh1D
     domain::I
@@ -165,6 +166,8 @@ model configurations).
 Then, the user can define a stretched mesh via
 
     ClimaCore.Meshes.IntervalMesh(interval_domain, ExponentialStretching(H); nelems::Int, reverse_mode = false)
+
+`faces` contain reference z without any warping.
 """
 struct ExponentialStretching{FT} <: StretchingRule
     H::FT
@@ -213,6 +216,8 @@ For land configurations, use `reverse_mode` = `true` (default value `false`).
 Then, the user can define a generalized stretched mesh via
 
     ClimaCore.Meshes.IntervalMesh(interval_domain, GeneralizedExponentialStretching(dz_bottom, dz_top); nelems::Int, reverse_mode = false)
+
+`faces` contain reference z without any warping.
 """
 struct GeneralizedExponentialStretching{FT} <: StretchingRule
     dz_bottom::FT
@@ -241,9 +246,9 @@ function IntervalMesh(
         throw(ArgumentError("dz_top must be ≤ dz_bottom"))
     end
 
-    # bottom coord height value, always min, for both atmos and land, since z-axis does not change
+    # bottom coord height value is always min and top coord height value is always max
+    # since the vertical coordinate is positive upward
     z_bottom = Geometry.component(domain.coord_min, 1)
-    # top coord height value, always max, for both atmos and land, since z-axis does not change
     z_top = Geometry.component(domain.coord_max, 1)
     # but in case of reverse_mode, we temporarily swap them together with dz_bottom and dz_top
     # so that the following root solve algorithm does not need to change
@@ -256,7 +261,7 @@ function IntervalMesh(
     # define the inverse σ⁻¹ exponential stretching function
     exp_stretch(ζ, h) = ζ == 1 ? ζ : -h * log(1 - (1 - exp(-1 / h)) * ζ)
 
-    # nondimensional vertical coordinate (]0.0, 1.0])
+    # nondimensional vertical coordinate ([0.0, 1.0])
     ζ_n = LinRange(one(FT_solve), nelems, nelems) / nelems
 
     # find bottom height variation
@@ -292,7 +297,7 @@ function IntervalMesh(
         find_top,
         RootSolvers.SecantMethod(guess₋, guess₊),
         RootSolvers.CompactSolution(),
-        RootSolvers.ResidualTolerance(FT_solve(1e-3)),
+        RootSolvers.ResidualTolerance(FT_solve(tol)),
     )
     if h_top_sol.converged !== true
         error(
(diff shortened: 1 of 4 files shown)


src/Meshes/interval.jl
```diff
@@ -318,6 +319,91 @@ function IntervalMesh(
...
+"""
+    HyperbolicTangentStretching(dz_surface::FT)
+
+Apply a hyperbolic tangent stretching to the domain when constructing elements.
+`dz_surface` is the target element grid spacing at the surface. In typical atmosphere
+configuration, it is the grid spacing at the bottom of the
+vertical column domain (m). On the other hand, for typical land configurations,
+it is the grid spacing at the top of the vertical column domain.
+
+For an interval ``[z_0,z_1]``, this makes the elements uniformally spaced in
```
tapios, line 331:
```suggestion
For an interval ``[z_0,z_1]``, this makes the elements uniformly spaced in
```

src/Meshes/interval.jl
```diff
@@ -318,6 +319,91 @@ function IntervalMesh(
...
+``\\zeta``, where
+```math
+\\eta = 1 - \\frac{tanh[\\gamma(1-\\zeta)]}{tanh(\\gamma)},
+```
+where ``\\eta = \\frac{z - z_0}{z_1-z_0}``. The stretching parameter ``\\gamma``
```
tapios, line 336: \eta does not need to be a linear terrain following coordinate (Gal-Chen and Sommerville) but can be a general terrain-following coordinate. Please keep this general here, and have a separate function for the terrain-following transformation, as in the paper draft.

src/Meshes/interval.jl
```diff
@@ -318,6 +319,91 @@ function IntervalMesh(
...
+is chosen to achieve a given resolution `dz_surface` at the surface. 
+
+Then, the user can define a stretched mesh via
+
+    ClimaCore.Meshes.IntervalMesh(interval_domain, HyperbolicTangentStretching(dz_surface); nelems::Int, reverse_mode)
+
+`reverse_mode` is default to false for atmosphere configurations. For land configurations,
```
tapios, line 343:
```suggestion
`reverse_mode` defaults to false for atmosphere configurations. For land configurations, 
```

src/Meshes/interval.jl
```diff
@@ -318,6 +319,91 @@ function IntervalMesh(
...
+    tol = 1e-5,
+    reverse_mode::Bool = false,
+) where {CT <: Geometry.Abstract1DPoint{FT}} where {FT}
+    if nelems ≤ 1
+        throw(ArgumentError("`nelems` must be ≥ 2"))
+    end
+
+    dz_surface = FT_solve(stretch.dz_surface)
+
+    # bottom coord height value, always min, for both atmos and land, since z-axis does not change
```
tapios, line 364: I'm confused by this comment. If this is bottom, for land, why would it be the min? Perhaps you can clarify what this comment is supposed to mean.

src/Meshes/interval.jl
```diff
@@ -318,6 +319,91 @@ function IntervalMesh(
...
+        RootSolvers.CompactSolution(),
+        RootSolvers.ResidualTolerance(FT_solve(tol)),
+    )
+    if γ_sol.converged !== true
+        error(
+            "gamma root failed to converge for dz_surface: $dz_surface on domain ($z_bottom, $z_top)",
+        )
+    end
+
+    faces = (z_bottom + (z_top - z_bottom)) * tanh_stretch.(ζ_n, γ_sol.root)
```
tapios, line 395: We shouldn't hard code linear terrain following coordinates but leaves this as a general, separate transformation.

tapios (commented):
We should keep the terrain-following and stretching transformations cleanly separated.

src/Meshes/interval.jl
```diff
@@ -292,7 +293,7 @@ function IntervalMesh(
         find_top,
         RootSolvers.SecantMethod(guess₋, guess₊),
         RootSolvers.CompactSolution(),
-        RootSolvers.ResidualTolerance(FT_solve(1e-3)),
+        RootSolvers.ResidualTolerance(FT_solve(tol)),
```
charleskawczynski, line 296: Thanks for updating this!

test/Meshes/interval.jl
```diff
@@ -288,6 +288,50 @@ end
...
+        nelems = 1,
+    )
+    # test a gcm like configuration
+    dom, mesh = unit_intervalmesh(
+        stretching = Meshes.HyperbolicTangentStretching(0.03 / 75.0),
+        nelems = 75, # 76 face levels
+    )
+    # test the mesh coordinates are eqv to dz_bottom
+    @test Meshes.coordinates(mesh, 1, 1) == Geometry.ZPoint(0.0)
+    @test Meshes.coordinates(mesh, 1, 2) ≈ Geometry.ZPoint(0.03 / 75.0)
```
charleskawczynski, line 308: Can we add one more test for `@test Meshes.coordinates(mesh, 1, nelems) ≈ ...`?

charleskawczynski (commented):
Code changes look good overall, I agree. I left some comments, it would be good to generalize the coordinate warping, since that will change the interface.

src/Meshes/interval.jl
szy21: I copied this comment from the other stretching method. But I think it is because for land z is also positive upward (?), so the bottom z has the minimum (most negative) value.

tapios: Maybe just explain it better then.

szy21:
To clarify, we are not hard coding linear terrain-following coordinates here. `IntervalMesh` deals with stretching, and the linear function is for the cases with non-zero z_bottom. Warping is done in Hyposgraphy, which has generalized terrain-following coordinates.

The code can be cleaned up though, I agree. We will refactor it in separate PRs.

test/Meshes/interval.jl
szy21: Added.

src/Meshes/interval.jl
```diff
@@ -305,7 +310,7 @@ function IntervalMesh(
     h =
         h_bottom .+
         (ζ_n .- ζ_n[1]) * (h_top - h_bottom) / (ζ_n[end - 1] - ζ_n[1])
-    faces = (z_bottom + (z_top - z_bottom)) * exp_stretch.(ζ_n, h)
+    faces = z_bottom .+ (z_top - z_bottom) * exp_stretch.(ζ_n, h)
```
szy21, line 313: @charleskawczynski Here is a bug fix. Would this be a breaking change? This technically changes the behavior of the vertical mesh, but since we always use z_bottom = 0 in ClimaAtmos, it wouldn't change the results. Just want to make sure if the next climacore release should be a minor or patch release.

PR #1932, opened 2024-08-14
Refactor DataLayouts

charleskawczynski:
This PR refactors some of the internals of DataLayouts. Some of the goals are:
 - use `parent` less
 - use `universal_size` more consistently
 - Reduce code duplication (e.g., `replace_basetype` logic)

I think that there's more that we can do with helpers like `type_params` and `union_all` (added here), in the context of `getproperty`, and perhaps even `slab`/`column` etc.


Sbozzolo (approved):
Thank you! Maybe you could add a docstring to `union_all` because it is not immediately clear how/when to use it.

PR #1934, opened 2024-08-14
Reduce use of DataLayouts internals

charleskawczynski:
This is basically a peel off from #1929, and extends #1932. This PR:
 - Reduces the use of `parent(data)` and especially `size(parent(data))`
 - eliminates `get_n_items`, and instead only integers are passed (when needed)
 - Uses `UniversalSize` to leverage static parameters more
 - Introduces `DataLayouts.array_size` and `DataLayouts.farray_size`, which return static sizes of the datalayouts (`array_size` uses a field dimension of 1, which is often helpful)

This does increase some inference failures because I've passed some additional variables through to gpu kernels using `Val`, which should result in statically known offsets, potentially improving performance of some kernels (likely by no more than 30%).

### DIFF
--- a/ext/cuda/limiters.jl
+++ b/ext/cuda/limiters.jl
@@ -21,23 +21,15 @@ function compute_element_bounds!(
     ρ,
     ::ClimaComms.CUDADevice,
 )
-    S = size(Fields.field_values(ρ))
-    (Ni, Nj, _, Nv, Nh) = S
+    ρ_values = Fields.field_values(Operators.strip_space(ρ, axes(ρ)))
+    ρq_values = Fields.field_values(Operators.strip_space(ρq, axes(ρq)))
+    (_, _, _, Nv, Nh) = DataLayouts.universal_size(ρ_values)
     nthreads, nblocks = config_threadblock(Nv, Nh)
 
-    args = (
-        limiter,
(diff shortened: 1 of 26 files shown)


ext/cuda/limiters.jl
```diff
@@ -21,39 +21,26 @@ function compute_element_bounds!(
     ρ,
     ::ClimaComms.CUDADevice,
 )
-    S = size(Fields.field_values(ρ))
-    (Ni, Nj, _, Nv, Nh) = S
+    (_, _, Nv, _, Nh) = DataLayouts.universal_size(ρ)
```
charleskawczynski, line 24: This appears to be a bug fix.

charleskawczynski: Ah, no, all is fine. I copied incorrect code over, that turned out to never be used. That is now fixed in this PR also (in DataLayouts.)

charleskawczynski:
I'll bump the allocation limits, but it looks like CI is otherwise passing.

Sbozzolo (approved):
Look good to me, just a question

charleskawczynski:
> just a question

?

Sbozzolo:
> > just a question
> 
> ?

I thought I left a question in the review, but maybe I didn't click "submit". Upon re writing the question, I found the answer myself :)

PR #1955, opened 2024-08-26
Refactor column, level, and slab

charleskawczynski:
This PR removes some code duplication in `column`, `level` and `slab`.

One change that this PR does is that it changes all view to directly create `SubArrays`, one benefit that I see with this is that the parameters entered into `Base.OneTo` are statically known, as opposed to using `:` (`Colon`) which requires a dynamic size lookup. It's probably not very important, but nice that this makes all of the implementations do the same thing. Also, the bounds checking are now a bit more rigorous in that all of the indices are checked for all view types.

I skipped `IV1JH2` and `IH1JH2` because they are a bit special cases.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -100,9 +100,23 @@ Statically returns `Nv`.
 """
     get_Nij(::UniversalSize)
 
-Statically returns `Nij`.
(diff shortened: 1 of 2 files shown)


charleskawczynski:
This got too messy and abstract. I'm going to leave the remaining methods as they are even though there's still a bit of code duplication.

PR #1957, opened 2024-08-27
Reduce code duplication in DataLayouts

charleskawczynski:
This PR reduces code duplication in DataLayouts, and defines `get_Ni` and `get_Nj` to allow for distinguishing from `get_Nij`.

This is basically a peel off from #1955.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -363,11 +363,7 @@ end
 
 Base.length(data::IJFH) = get_Nh(data)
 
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Sounds good, yes, I can split up the commits.

charleskawczynski:
I think some of the errors actually were related to `get_Ni `/`get_Nj`, so I removed that. Also, I removed the `≤`-related changes for now. I'll take a look at Fira Code. So, this PR now only changes some of the method definitions in DataLayouts to forward to equivalent methods.

Issue #1910, opened 2024-07-31
Fix uncoalesced memory reads


charleskawczynski:
Broadcast operations on our most simple expressions result in uncoalesced memory reads for `VIJFH` datalayouts, which was observed via nsight compute.

Issues are being analyzed/outlined in the following benchmark scripts:
 - `benchmarks/scripts/index_swapping.jl`
 - `benchmarks/scripts/indexing_and_static_ndranges.jl`
 - `benchmarks/scripts/thermo_bench.jl` (needs updated)

charleskawczynski:
This is an attempt to summarize the information gathered from benchmarks, and experimental development branches. Here is a high-level summary of the numbers. The `bw` column is the percentage of peak bandwidth reached for `Float32` and `Float64` data on NVIDIA A100s. The details of the kernels are discussed below.

### indexing_and_static_ndranges.jl
```
@. y1 = foo(x1, x2, x3); foo(x1, x2, x3) = zero(x1) # kernel (0 reads, 1 write)
[ Info: ArrayType = CuArray
Problem size: (63, 4, 4, 1, 5400), device_bandwidth_GBs=2039
┌───────────────────────────────────────────────────────────────────────────┬─────────┬─────────┐
│ funcs                                                                     │ bw% F32 │ bw% F64 │
├───────────────────────────────────────────────────────────────────────────┼─────────┼─────────┤
│ BSR.at_dot_call!(X_array, Y_array; nreps=1000, bm)                        │ 14.4882 │ 28.8217 │
│ BSR.at_dot_call!(X_vector, Y_vector; nreps=1000, bm)                      │ 72.1366 │ 70.4848 │
│ BSR.custom_sol_kernel!(X_vector, Y_vector, Val(N); nreps=1000, bm)        │ 76.943  │ 78.1221 │
│ BSR.custom_kernel_bc!(X_vector, Y_vector, uss; nreps=1000, bm)            │ 76.9247 │ 78.1975 │
│ BSR.custom_kernel_bc!(X_array, Y_array, uss; use_pw=false, nreps=1000, bm)│ 37.3141 │ 73.3654 │ # ~ClimaCore kernels
│ BSR.custom_kernel_bc!(X_array, Y_array, uss; use_pw=true, nreps=1000, bm) │ 76.9613 │ 78.1095 │ K1
└───────────────────────────────────────────────────────────────────────────┴─────────┴─────────┘
```

### thermo_bench_bw.jl
```
[...]

PR #1958, opened 2024-08-27
Refactor DSS to better leverage DataLayouts

charleskawczynski:
This PR refactors DSS to avoid using DataLayout internals, and reduces code duplication. We can apply a similar pattern to the cuda code (wip).

This should make fixing #1910 easier.

What this PR does:
 - We pass the datalayouts themselves into the dss function, instead of the underlying array.
 - Swap the field and level order of the loops, and indexing into the datalayouts using the universal spatial index `(I, J, F, V, H)`, which recursively indexes into the field variables via `get_struct`/`set_struct`.
 - I'm reusing the already existing `dss_transform` to transform the data.
 - The only unexpected thing I ran into was the need to define `drop_vert_dim` in `Geometry.UVWVector` due to a comment in `dss.jl`: `For DSS of Covariant123Vector, the third component is treated like a scalar and is not transformed`. I'm not 100% sure if what I did was correct here
 
The last bullet point still confuses me because it seems like `N` variables come in and `N - M` are defined. Can someone confirm that this is correct?

### DIFF
--- a/src/Topologies/dss.jl
+++ b/src/Topologies/dss.jl
@@ -22,16 +22,6 @@ struct DSSBuffer{S, G, D, A, B, VI}
     send_buf_idx::B
     "indexing array for loading (and summing) data from recv buffer to `perimeter_data`"
     recv_buf_idx::B
-    "field id for all scalar fields stored in the `data` array"
-    scalarfidx::VI
-    "field id for all covariant12vector fields stored in the `data` array"
-    covariant12fidx::VI
-    "field id for all contravariant12vector fields stored in the `data` array"
-    contravariant12fidx::VI
-    "internal local elements (lidx)"
-    covariant123fidx::VI
-    "field id for all contravariant123vector fields stored in the `data` array"
-    contravariant123fidx::VI
     "internal local elements (lidx)"
     internal_elems::VI
     "local elements (lidx) located on process boundary"
@@ -69,14 +59,15 @@ function create_dss_buffer(
     Nf = DataLayouts.ncomponents(data)
     nfacedof = Nij - 2
     T = eltype(parent(data))
-    TS = _transformed_type(data, local_geometry, local_weights, DA) # extract transformed type
     # Add TS for Covariant123Vector
     # For DSS of Covariant123Vector, the third component is treated like a scalar
     # and is not transformed
-    if eltype(data) <: Geometry.Covariant123Vector
-        TS = Geometry.UVWVector{T}
+    TS = if eltype(data) <: Geometry.Covariant123Vector
+        Geometry.UVWVector{T}
     elseif eltype(data) <: Geometry.Contravariant123Vector
-        TS = Geometry.UVWVector{T}
+        Geometry.UVWVector{T}
+    else
+        _transformed_type(data, local_geometry, local_weights, DA) # extract transformed type
     end
     perimeter_data =
         DataLayouts.VIFH{TS, Nv, Np, Nh}(DA{T}(undef, Nv, Np, Nf, Nh))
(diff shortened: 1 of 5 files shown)


src/Topologies/dss.jl
```diff
@@ -380,166 +212,82 @@ function dss_transform!(
...
+    CI = CartesianIndex
+    (; ∂ξ∂x) = local_geometry
     @inbounds for elem in localelems
         for (p, (ip, jp)) in enumerate(perimeter)
-            pw = pweight[linear_ind(sizet_wt, (ip, jp, 1, elem))]
+            for level in 1:nlevels
+                loc = CI(ip, jp, 1, level, elem)
+                src = dss_transform(
+                    data[loc],
+                    local_geometry[CI(ip, jp, 1, level, elem)],
```
charleskawczynski, line 229:
```suggestion
                    local_geometry[loc],
```

charleskawczynski:
One benefit of this change, aside from deleting hundreds of lines of code, is that it also strictly and explicitly follows conversion rules (which already exist). Our previous implementation bypasses these checks since it accesses the parent arrays.

akshaysridhar:
@charleskawczynski Even with the modifications to use the existing `transform` functions, it would be good to have the ability to `convert` `C123 -> UVW -> apply dss op -> untransform to retain 3 component vectors` (I'll look over the `drop_vert_dim` functions), i.e. avoid treating the last component as a scalar quantity. I'm unsure about the consequences of this PR on overall performance. Generally I'm in favour of simplifying these code blocks (especially since it's clear we have some major duplication between the `cpu` and `cuda ext` files in our code right now). It would be good if @sriharshakandala also chimes in here!

charleskawczynski:
> @charleskawczynski Even with the modifications to use the existing `transform` functions, it would be good to have the ability to `convert` `C123 -> UVW -> apply dss op -> untransform to retain 3 component vectors` (I'll look over the `drop_vert_dim` functions), i.e. avoid treating the last component as a scalar quantity.

So, this turned out to be needed because `perimeter_data` doesn't have storage for the 3-component vectors (this is why `drop_vert_dim` was needed in the first place). In fact, this PR may have identified a bug in the existing implementation.

I've reached out to others on slack for additional review.

> I'm unsure about the consequences of this PR on overall performance. Generally I'm in favour of simplifying these code blocks (especially since it's clear we have substantial duplication, e.g. of the conversion operations, between the `cpu` and `cuda ext` files in our code right now). It would be good if @sriharshakandala also chimes in here!

I think we should be able to get good performance from this.

charleskawczynski:
Hi @sriharshakandala, I ran the GPU target pipeline, and these are the preliminary results:

### main ([build](https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/329#019104b4-024c-437b-9874-382af1923ac9))
```
solve!:: 88.144602 seconds (20.75 M allocations: 474.717 MiB, 0.72% compilation time)
[ Info: sypd: 2.687084644368064
[ Info: wall_time_per_timestep: 91 milliseconds, 763 microseconds
```

### This PR ([build](https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/335#0191bd3d-3bf3-4020-933a-4bd5de0ebf7e))
```
solve!:: 70.939605 seconds (20.45 M allocations: 472.976 MiB, 1.01% compilation time)
[ Info: sypd: 3.3397802737331332
[ Info: wall_time_per_timestep: 73 milliseconds, 829 microseconds
```

I don't think that this implementation is necessarily responsible for the speedup, but I also don't think that it's slowed anything down. Also, I think that we can actually make it faster by avoiding the `div` calls in the kernels, and just rely on tuned cartesian indexing (like the `copyto!` and stencil kernels).

At this point, I think I'm ready to merge if people are happy with the changes.

charleskawczynski:
Hi @akshaysridhar, I've looked into this a bit more, and it seems that DSS does have storage for the vertical component, so the 3-component case should work just fine.

The behavior that I didn't expect was coming from: https://github.com/CliMA/ClimaCore.jl/blob/f36251cf09c49851e47c1530380902f44441a0df/src/Topologies/dss_transform.jl#L90-L125

which transforms a `Covariant12Vector` into a `UVWVector`. So, either that method needs to be updated to preserve the dimensionality, or just keep `drop_vert_dim`. For now, I think it's probably easiest to just keep `drop_vert_dim`.

sriharshakandala:
It looks good to me overall. I am not sure if `drop_dims` is required. The recent PR treats vector transformations in an normal way (no special treatment for vertical velocity component). I am assuming there are no performance penalties from using the `AxisTensors` conversion framework, which we had when we originally put this code!

sriharshakandala:
We should also remove the comments indicating that we are treating the vertical component in a special way, since this is no longer true!

sriharshakandala (approved):

charleskawczynski:
Alright, moving forward with this.

Issue #1780, opened 2024-06-04
Reduce code duplication


charleskawczynski:
There are several places where we can reduce code duplication:

PR #1961, opened 2024-08-29
Refactor cubed sphere dss test

charleskawczynski:
This PR refactors the cubed sphere test by adding some utility functions that can be used for the cpu and gpu. I've also split out (what I hope to be) the unit test and cpu-gpu comparison test. Since it right now only does a cpu-gpu comparison, it runs the risk of both being wrong and matching. Ideally, we test the result against a simply computed result in the test suite, and that test can run both on the cpu and gpu.

The refactored test also tests covariant and contravariant transformations, which are not exercised in `ddss1.jl`, so this PR also adds this as a CPU test to buildkite and the unit test suite.

This is a step towards #1780.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -278,17 +278,21 @@ steps:
         key: unit_ddss1
         command:
           - "julia --project=.buildkite -e 'using CUDA; CUDA.versioninfo()'"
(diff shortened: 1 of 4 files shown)


Sbozzolo (approved):
Thank you!

PR #1962, opened 2024-08-29
Add pre-push hook to prevent pushing directly to the main branch

charleskawczynski:
This PR adds a pre-push hook to prevent pushing commits directly to the main branch-- regardless of admin privileges.

### DIFF
--- a/.github/hooks/pre-push.sh
+++ b/.github/hooks/pre-push.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# This script disables users from pushing directly to the main branch
+# Refs:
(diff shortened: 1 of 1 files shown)


Sbozzolo (commented):
I think we should keep the option turned off in the GitHub settings instead of adding a script that people have to know to install to their ClimaCore repos.

charleskawczynski:
> I think we should keep the option turned off in the GitHub settings instead of adding a script that people have to know to install to their ClimaCore repos.

Yeah, I think I'd also prefer avoiding installation of additional things. I'll close it.

PR #1964, opened 2024-08-29
Remove unused dependency

Sbozzolo:
`DataStructures` is never used.


Sbozzolo:
Just kidding, it is used in SortedSet.

PR #1972, opened 2024-09-06
Bump compat for CubedSphere to 0.3

Sbozzolo:
CubedSphere 0.3 does not introduce any new feature, it moves some accessory features out of the package. In this, most dependencies are dropped.

Now, ClimaCore no longer depends on FFTW and the various other dependencies that are pulled in with FFTW.


charleskawczynski (approved):
Nice! I remembered there was a related issue with FFTW, but it turns out that this was in ClimateMachine (https://github.com/CliMA/ClimateMachine.jl/issues/1478). So, that's a nice dependency to remove.

charleskawczynski:
Hm, the error is a bit odd. I'll try pushing a commit to update the manifests.

charleskawczynski:
It looks like OrdinaryDiffEq broke into many subpackages? Maybe we can now adjust our dependencies on it in the test/examples environment.

charleskawczynski:
Weird, the versions got downgraded During develop or instantiate. I’m gonna try clearing the depot.

charleskawczynski:
Looks like we'll need to update this, limit OrdinaryDiffEq, or figure out how to eliminate this dependency. @dennisYatunin, was `update_W!` part of the public API?

Sbozzolo:
Many of the places where OrdindaryDiffEq is currently used should be changed with SciMLBase.

charleskawczynski:
> Many of the places where OrdindaryDiffEq is currently used should be changed with SciMLBase.

Yeah, I tried removing it a while back, but we need some of the algorithms, which don't exist in SciMLBase. Maybe we can use the subpackages now that it's split

PR #1979, opened 2024-09-09
Remove OrdinaryDiffEq dependence in all environments

charleskawczynski:
This PR removes our dependence on OrdinaryDiffEq. It looks like all of the algorithms in the driver now point to those in ClimaTimesteppers, so I think we can simply remove all of the code to handle the OrdinaryDiffEq cases.


charleskawczynski:
I think that this will allow us to merge #1972.

Sbozzolo (approved):

PR #1987, opened 2024-09-12
Add LevelPlaceholderSpace

dennisYatunin:
This PR adds a `LevelPlaceholderSpace`, which is needed to pass mixtures of extruded 3D and level 2D fields onto the GPU. Specifically, this will enable us to perform the column accumulation operation needed for the non-orographic gravity wave parametrization in ClimaAtmos (CliMA/ClimaAtmos.jl#3267).

### DIFF
--- a/src/Operators/common.jl
+++ b/src/Operators/common.jl
@@ -69,13 +69,21 @@ end
 
 # Functions for CUDASpectralStyle
 struct PlaceholderSpace <: Spaces.AbstractSpace end
+struct LevelPlaceholderSpace <: Spaces.AbstractSpace end
 struct CenterPlaceholderSpace <: Spaces.AbstractSpace end
 struct FacePlaceholderSpace <: Spaces.AbstractSpace end
 
-
+placeholder_space(current_space, parent_space) = current_space
 placeholder_space(current_space::T, parent_space::T) where {T} =
     PlaceholderSpace()
-placeholder_space(current_space, parent_space) = current_space
+placeholder_space(
+    current_space::Spaces.AbstractPointSpace,
+    parent_space::Spaces.AbstractFiniteDifferenceSpace,
+) = LevelPlaceholderSpace()
+placeholder_space(
+    current_space::Spaces.AbstractSpectralElementSpace,
+    parent_space::Spaces.ExtrudedFiniteDifferenceSpace,
+) = LevelPlaceholderSpace()
 placeholder_space(
     current_space::Spaces.CenterFiniteDifferenceSpace,
     parent_space::Spaces.FaceFiniteDifferenceSpace,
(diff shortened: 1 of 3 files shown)


Sbozzolo (approved):

src/Operators/common.jl
```diff
@@ -69,13 +69,21 @@ end
...
     PlaceholderSpace()
-placeholder_space(current_space, parent_space) = current_space
+placeholder_space(
+    current_space::Spaces.AbstractPointSpace,
+    parent_space::Spaces.AbstractFiniteDifferenceSpace,
+) = LevelPlaceholderSpace()
+placeholder_space(
+    current_space::Spaces.AbstractSpectralElementSpace,
+    parent_space::Spaces.ExtrudedFiniteDifferenceSpace,
+) = LevelPlaceholderSpace()
```
charleskawczynski, line 86: Why should these return level spaces?

dennisYatunin: Suppose that, as in the case of the nonorographic gravity wave parametrization, we have a broadcast over extruded/vertical fields and horizontal/point fields that needs to be evaluated on the GPU. The parent_space in this case is the space of the extruded/vertical output field, and the current_space is either the same as the parent_space or just one level of the parent_space. In the first case we get a PlaceholderSpace, and in the second case we get a LevelPlaceholderSpace.

charleskawczynski: That may makes sense for the second method, but the first method looks to me like it should return the parent space.

charleskawczynski:
@dennisYatunin, please make sure to add tests so that this new code is exercised.

src/Operators/common.jl
charleskawczynski: That’s after all how we support point-sphere broadcasting.

dennisYatunin: If the first method returned `parent_space` instead of `LevelPlaceholderSpace`, there would be multiple non-placeholder spaces passed to the GPU. My impression is that we want only one non-placeholder space (the output of the broadcast) passed to the GPU, and all other spaces can be reconstructed from that non-placeholder space when they are needed.

With the current design, if a `PointField` is used in a broadcast expression with a `FiniteDifferenceField` as the output, a `LevelPlaceholderSpace` will be returned by `placeholder_space` and a `PointSpace` will be returned by `reconstruct_placeholder_space`. If the first method returned `parent_space` instead of `LevelPlaceholderSpace`, `reconstruct_placeholder_space` would return a `FiniteDifferenceSpace` for a `PointField`, which makes no sense to me.

charleskawczynski: Yeah, sorry, I meant `..., but the first method looks to me like it should a type similar to the parent space.`.

> if a PointField is used in a broadcast expression with a FiniteDifferenceField as the output

Given this condition, I agree with you, but this condition is not enforced in `placeholder_space`, which is why this change looks a bit odd to me.

PR #1999, opened 2024-09-19
Fix InputOutput when enable_bubble is true

Sbozzolo:
It turns out that we were not saving and restoring a piece of information that is required to fully reconstruct a `2DSpectralSpace`: the bubble correction.

When enabled, the bubble correction recomputes the Jacobian to improve their accuracy. `InputOutput` was not saving this piece of information into the HDF5 file, so spaces read from file always assumed that it was false. This led to discrepancies in the space, and these discrepancies led to problems in the evolution.

Interestingly, for the low-resolution cases we consider in the Atmos CI, `enable_bubble` has very small effects on the value of the Jacobian, with a relative difference compared to the case without bubble smaller than the floating point precision for Float32. In spite of this, it affected the evolution in more noticeable way, in particular in vectors and with hyperdiffusion.

Now, `bubble` is saved as attribute so that it can be read back.

This commit also bumps the version of ClimaCore to 0.14.6

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,22 @@ ClimaCore.jl Release Notes
 main
 -------
 
+v0.14.16
+-------
+
+### ![][badge-🐛bugfix] Fix restarting simulations from `Space`s with `enable_bubble = true`
+
+Prior to this change, the `ClimaCore.InputOutput` module did not save whether a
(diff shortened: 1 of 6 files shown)


charleskawczynski (approved):
Good catch, thanks for fixing this!

PR #2000, opened 2024-09-20
extend weighted_dss for FieldVector

juliasloan25:
Extend `weighted_dss!` to work for FieldVectors in addition to Fields.

Needed for https://github.com/CliMA/ClimaCoupler.jl/pull/959, and will also be useful for ClimaAtmos.

### To Do
- [x] add method
- [x] add tests
- [x] update NEWS.md

### Code quality

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -184,6 +184,41 @@ end
     return dest
 end
 
+"""
+    Spaces.create_dss_buffer(fv::FieldVector)
+
+Create a NamedTuple of buffers for communicating neighbour information of
+each Field in `fv`. In this NamedTuple, the name of each field is mapped
+to the buffer.
+"""
+function Spaces.create_dss_buffer(fv::FieldVector)
+    NamedTuple{propertynames(fv)}(
+        map(
+            key -> Spaces.create_dss_buffer(getproperty(fv, key)),
+            propertynames(fv),
+        ),
+    )
+end
+
+"""
+    Spaces.weighted_dss!(fv::FieldVector, dss_buffer = Spaces.create_dss_buffer(fv))
+
+Apply weighted direct stiffness summation (DSS) to each field in `fv`.
+If a `dss_buffer` object is not provided, a buffer will be created for each
+field in `fv`.
+Note that using the `Pair` interface here parallelizes the `weighted_dss!` calls.
+"""
+function Spaces.weighted_dss!(
+    fv::FieldVector,
+    dss_buffer = Spaces.create_dss_buffer(fv),
+)
+    pairs = map(propertynames(fv)) do key
+        Pair(getproperty(fv, key), getproperty(dss_buffer, key))
+    end
+    Spaces.weighted_dss!(pairs...)
+end
+
+
 # Recursively call transform_bc_args() on broadcast arguments in a way that is statically reducible by the optimizer
 # see Base.Broadcast.preprocess_args
 @inline transform_bc_args(args::Tuple, inds...) = (
--- a/test/Fields/field_opt.jl
+++ b/test/Fields/field_opt.jl
@@ -393,4 +393,49 @@ using JET
     @test_opt ifelsekernel!(S, ρ)
 end
 
+@testset "dss of FieldVectors" begin
+    function field_vec(center_space, face_space)
+        Y = Fields.FieldVector(
+            c = map(Fields.coordinate_field(center_space)) do coord
+                FT = Spaces.undertype(center_space)
+                (;
+                    ρ = FT(coord.lat + coord.long),
+                    uₕ = Geometry.Covariant12Vector(
+                        FT(coord.lat),
+                        FT(coord.long),
+                    ),
+                )
+            end,
+            f = map(Fields.coordinate_field(face_space)) do coord
+                FT = Spaces.undertype(face_space)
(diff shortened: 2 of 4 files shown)


src/Fields/fieldvector.jl
```diff
@@ -184,6 +185,43 @@ end
...
+        ),
+    )
+end
+
+"""
+    Spaces.weighted_dss!(fv::FieldVector, dss_buffer = Spaces.create_dss_buffer(fv))
+
+Apply weighted direct stiffness summation (DSS) to each field in `fv`.
+Reuse the same `dss_buffer` for all fields in `fv`, which is constructed using
+the first field in `fv` if it isn't passed explicitly.
```
dennisYatunin, line 209: I think these two lines are out-of-date.

src/Fields/fieldvector.jl
```diff
@@ -184,6 +185,43 @@ end
...
+"""
+# TODO distribute `propertynames(fv)` over processes to parallelize `unrolled_foreach`
```
dennisYatunin, line 211: Could you move this comment to the end of the function? Putting it before the function will probably mess up the docstring.

test/Fields/field_opt.jl
```diff
@@ -393,4 +393,41 @@ using JET
...
+    @test getproperty(dss_buffer, :c) isa Topologies.DSSBuffer
+    @test getproperty(dss_buffer, :f) isa Topologies.DSSBuffer
+
+    c_copy = copy(getproperty(fv, :c))
+    f_copy = copy(getproperty(fv, :f))
+
+    # Test weighted_dss! with and without preallocated buffer
+    p = @allocated Spaces.weighted_dss!(fv, dss_buffer) # DSS2
+    @test getproperty(fv, :c) ≈ c_copy
+    @test getproperty(fv, :f) ≈ f_copy
```
dennisYatunin, line 426: It doesn't look like these tests are actually checking for consistency of DSS; they are only checking that the values remain approximately unchanged after DSS. If you'd like to do a quantitative check here, it would be better to fill `fv` with random data and then check that applying `weighted_dss!` to it gives the same results as applying `weighted_dss!` to `fv.c` and `fv.f`.

test/Fields/field_opt.jl
```diff
@@ -393,4 +393,41 @@ using JET
...
+    end
+
+    fv = field_vec(toy_sphere(Float64)...)
+
+    # Test that dss_buffer is created and has the correct keys and buffer types
+    dss_buffer = Spaces.create_dss_buffer(fv)
+    @test haskey(dss_buffer, :c)
+    @test haskey(dss_buffer, :f)
+    @test getproperty(dss_buffer, :c) isa Topologies.DSSBuffer
+    @test getproperty(dss_buffer, :f) isa Topologies.DSSBuffer
```
dennisYatunin, line 418: Unit tests like this aren't really necessary, as they're just checking internal implementation details instead of overall correctness, but you can leave them in if you prefer.

dennisYatunin (approved):

src/Fields/fieldvector.jl
```diff
@@ -184,6 +185,44 @@ end
...
+)
+    unrolled_foreach(
+        key -> Spaces.weighted_dss!(
+            getproperty(fv, key),
+            getproperty(dss_buffer, key),
+        ),
+        propertynames(fv),
+    )
+end
+# TODO distribute `propertynames(fv)` over processes to parallelize `unrolled_foreach`
```
charleskawczynski, line 223:
```suggestion
function Spaces.weighted_dss!(
    fv::FieldVector,
    dss_buffer = Spaces.create_dss_buffer(fv),
)
    pairs = map(propertynames(fv)) do pn
        Pair(getproperty(fv, key), getproperty(dss_buffer, key))
    end
    Spaces.weighted_dss!(pairs...)
end
```
Better yet, let's just call the `Pair` interface, which will parallelize the weighted dss calls (this addresses the TODO).

src/Fields/fieldvector.jl
```diff
@@ -184,6 +185,44 @@ end
...
+each Field in `fv`. In this NamedTuple, the name of each field is mapped
+to the buffer.
+"""
+function Spaces.create_dss_buffer(fv::FieldVector)
+    NamedTuple{propertynames(fv)}(
+        unrolled_map(
+            key -> Spaces.create_dss_buffer(getproperty(fv, key)),
+            propertynames(fv),
+        ),
+    )
```
charleskawczynski, line 201: Is `unrolled_map` actually needed here? My preference would be to use `map` until we deem `unrolled_map` needed.

juliasloan25: In ClimaLand, our state is a FieldVector with Fields defined on different spaces - would this count as a nonuniform collection that we'd want to use `unrolled_map` with?

charleskawczynski: `unrolled_map` is just a optimized form of `Base`'s `map` to overcome limitations with compiler inference. So, `map` may result in inference failures, but it should work just fine. Using `unrolled_map` is not critical here in the sense that `create_dss_buffer` is only called during initialization, and it's not called inside a gpu kernel, so the only penalty is that `map` _could_ lead to runtime dispatch.

If we found that this fixes inference failures in this function, then `unrolled_map` is fine, but it's not strictly needed (as it is in other places).

juliasloan25: Thanks for the explanation! I'll change it to `map`

PR #2004, opened 2024-09-25
Fix type instability in slab_index and vindex

charleskawczynski:
Found in https://buildkite.com/clima/climaatmos-ci/builds/20805#01922a42-bcbe-49b0-ae2d-2247cb3b2a49


Sbozzolo (approved):
Thank you!

PR #2005, opened 2024-09-25
Make Nh a dynamic parameter

charleskawczynski:
This PR reverts https://github.com/CliMA/ClimaCore.jl/pull/1894 as we are running into extremely large compile times.

*Please note that this may cause runtime slowdowns, because `Nh` will be (again) no longer statically known.*

Perhaps we can revert this PR once https://github.com/JuliaLang/julia/issues/55807 is resolved.

PR #2002, opened 2024-09-23
Define a linear partition, and use in FD stencils

charleskawczynski:
A performance regression was found in #1969. Concretely,

##### Main
```julia
Problem size: (4, 4, 1, 63, 1536), N-reps: 1,  Float_type = Float64, Device_bandwidth_GBs=2039
┌───────────────────────────────────────────────────────────────┬───────────────────────────────────┬───────────┬─────────────┬────────────────┐
│ funcs                                                         │ time per call                     │ bw %      │ achieved bw │ N reads-writes │
├───────────────────────────────────────────────────────────────┼───────────────────────────────────┼───────────┼─────────────┼────────────────┤
│ (op_GradientF2C!, :none)                                      │ 31 microseconds, 280 nanoseconds  │ 36.1744   │ 737.597     │ 2              │
│ (op_GradientF2C!, :SetValue, :SetValue)                       │ 45 microseconds, 911 nanoseconds  │ 24.6461   │ 502.533     │ 2              │
│ (op_GradientC2F!, :SetGradient, :SetGradient)                 │ 49 microseconds, 740 nanoseconds  │ 22.7483   │ 463.838     │ 2              │
│ (op_GradientC2F!, :SetValue, :SetValue)                       │ 41 microseconds, 890 nanoseconds  │ 27.0119   │ 550.772     │ 2              │
│ (op_DivergenceF2C!, :none)                                    │ 2 milliseconds, 101 microseconds  │ 0.807571  │ 16.4664     │ 3              │
│ (op_DivergenceF2C!, :Extrapolate, :Extrapolate)               │ 496 microseconds, 567 nanoseconds │ 3.41798   │ 69.6925     │ 3              │
│ (op_DivergenceC2F!, :SetDivergence, :SetDivergence)           │ 77 microseconds, 619 nanoseconds  │ 21.8664   │ 445.856     │ 3              │
│ (op_InterpolateF2C!, :none)                                   │ 31 microseconds, 831 nanoseconds  │ 35.5482   │ 724.828     │ 2              │
│ (op_InterpolateC2F!, :SetValue, :SetValue)                    │ 43 microseconds, 859 nanoseconds  │ 25.7986   │ 526.033     │ 2              │
│ (op_InterpolateC2F!, :Extrapolate, :Extrapolate)              │ 39 microseconds, 699 nanoseconds  │ 28.502    │ 581.155     │ 2              │
[...]

PR #2006, opened 2024-09-26
Hoist UniversalSize computation outside of kernels

charleskawczynski:
This PR simply hoists computing the `UniversalSize` computation outside of kernels and adds `us::UniversalSize` objects as an argument to some of the universal index functions, so that they're more flexible. This seems to be needed with both #2005 and #2002, so I'm peeling it off.


charleskawczynski:
I think that the case that broke is unrelated to this PR. Merging.

PR #2007, opened 2024-09-26
Bump patch version

charleskawczynski:
This PR bumps the patch version, so that we can propagate the fix applied in #2004.


juliasloan25:
btw, this PR should have updated the NEWS.md too

Issue #1983, opened 2024-09-11
Non-scalar Matrix Field bugs


dennisYatunin:
## Describe the bug
In the process of working the RosSSP timestepper, @Sbozzolo has uncovered 3 bugs in the Field/FieldMatrix broadcasting machinery:
- Division (not multiplication) of a Field of BandMatrixRows of non-scalars by a scalar throws an error
- Multiplciation/division of a FieldMatrix by a scalar throws an error
- `Base.one` of a FieldMatrix with non-scalar subblocks generates a FieldMatrix with scalar subblocks, which then throws an error when combined with a FieldMatrix similar to the original

## Steps to Reproduce
Using `W` to denote the `ImplicitEquationJacobian` used in ClimaAtmos, the bugs can be reproduced as follows:
- For the first bug, running `W.matrix[@name(f.u₃), @name(f.u₃)] ./ 2` reproduces the error
- For the second bug, running `W.matrix .* 2` reproduces the error
- For the third bug, running `one(W.matrix)` generates the incorrect result, and running `W.matrix .+ one(W.matrix)` reproduces the error

PR #2014, opened 2024-09-27
Remove unnecessary method specializations for Numbers

dennisYatunin:
There appear to be inference issues arising from unnecessary specializations on `Number` types in several `RecursiveApply` methods. Specifically, the specializations for `rdiv` (but not for `rmul`) are the source of the first bug described in #1983, and removing them fixes that error. Removing these specializations may also fix other inference problems for operations on complex fields (e.g., matrix fields of non-scalar values).

Several specializations of `rmuladd` are not removed in this PR because `rmap` does not currently support 3 input arguments (only 1 or 2). This can be fixed in the future, especially if we end up rewriting `rmap` in terms of `unrolled_map`.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -895,6 +895,11 @@ steps:
         key: unit_matrix_field_broadcasting_cpu_non_scalar_4
         command: "julia --color=yes --check-bounds=yes --project=.buildkite test/MatrixFields/matrix_fields_broadcasting/test_non_scalar_4.jl"
 
+      - label: "Unit: matrix field broadcasting (CPU)"
+        key: unit_matrix_field_broadcasting_cpu_non_scalar_5
+        command: "julia --color=yes --check-bounds=yes --project=.buildkite test/MatrixFields/matrix_fields_broadcasting/test_non_scalar_5.jl"
+        soft_fail: true
+
   - group: "Unit: MatrixFields - broadcasting (GPU)"
     steps:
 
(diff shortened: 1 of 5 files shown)


charleskawczynski (approved):
This is my favorite kind of PR, better yet favorite way to fix inference failures: delete code. Fantastic.

charleskawczynski:
Actually, @dennisYatunin, would you mind adding a JET test for the case that is breaking on the main branch?

dennisYatunin:
I just added a new `test_non_scalar_5` unit test, which divides a matrix field of `Covariant12Vector`s by a scalar. On my machine, this test fails with the `Number` specializations and passes without them. If the CPU and GPU tests on CI pass, I think this PR is good to merge in. Unfortunately, it doesn't appear to fix any of the soft fails we currently have in our CI, but it definitely fixes the first part of #1983.

charleskawczynski:
Excellent, thanks!

dennisYatunin:
Wow, looks like the CPU test on CI allocates 16 bytes for the in-place broadcast expression, even though the version I ran locally doesn't allocate anything. The only difference I can see is that I'm running Julia 1.10.4, while CI is running Julia 1.10.5. @charleskawczynski, I think this might be indicating another compiler regression?

Although the CPU test allocates, both it and the GPU test are able to run without crashing. So, this PR is definitely improving inference, and I'll merge it in with the CPU test as a soft-fail for now. We can look into the allocations issue in a separate PR.

PR #2018, opened 2024-09-30
Fix Julia 1.11

### LINKED ISSUE #1923: #1920 broke CI on Julia 1.11
#1920 broke CI on Julia 1.11.

For some reason, `Base.Broadcast` in Julia 1.10 does not seem to require `getindex(::AbstractData, ::Integer)` to be defined, but it does seem to be required on Julia 1.11.

We can revert #1920 all within the DataLayouts module, if we need to. But it may be better to figure out why this behavior changed across Julia versions.

charleskawczynski:
This PR effectively reverts part of #1920, and should (hopefully) fix CI on Julia 1.11. This should close #1923.

I don't think that this should actually be needed, but it seems that there have been changes to Julia Base's broadcasting that require multi-dimensional indexing.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -1181,6 +1181,40 @@ end
     )
 end
 
+### --------------- Support for multi-dimensional indexing
+# TODO: can we remove this? It's not needed for Julia 1.10,
+#       but seems needed in Julia 1.11.
+@inline Base.getindex(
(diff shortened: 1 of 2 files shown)


charleskawczynski:
Let's merge this, I think seeing upstream issues early will be useful, despite that I think this shouldn't be needed. At least support for multi-dimensional indexing is localized to this small section.

PR #2021, opened 2024-10-01
Add information about deep grid in restart files

Sbozzolo:
This commit adds some information that was lost in saving to a checkpoint: the `deep` option. It also adds tests for types too, making the tests more stringent.

This also fixes an issue with `rcompare` for different types of FieldVectors.

Closes #2022

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -487,10 +487,18 @@ end
 
 Recursively compare given fieldvectors via `==`.
 Returns `true` if `x == y` recursively.
+
+FieldVectors with different types are considered different.
 """
 rcompare(x::T, y::T) where {T <: Union{FieldVector, NamedTuple}} =
     _rcompare(true, x, y)
 
-# Define == to call rcompare for two fieldvectors of the same
-# exact type.
-Base.:(==)(x::T, y::T) where {T <: FieldVector} = rcompare(x, y)
+rcompare(x::T, y::T) where {T <: FieldVector} = _rcompare(true, x, y)
+
+rcompare(x::T, y::T) where {T <: NamedTuple} = _rcompare(true, x, y)
+
+# FieldVectors with different types are always different
+rcompare(x::FieldVector, y::FieldVector) = false
+
+# Define == to call rcompare for two fieldvectors
+Base.:(==)(x::FieldVector, y::FieldVector) = rcompare(x, y)
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,17 @@ Due to a bug, `==` was not recursively checking `FieldVector`s with different
 types, which resulted in false positives. This is now fixed and `FieldVector`s
 with different types are always considered different.
 
+### ![][badge-🐛bugfix] Fix restarting simulations from `Space`s with `deep = true`
+
+Prior to this change, the `ClimaCore.InputOutput` module did not save whether a
+`Space` was constructed with `deep = true`. This meant that restarting a
+simulation from a HDF5 file led to inconsistent and incorrect spaces and
+`Field`s. This affected only extruded 3D spectral spaces.
+
+We now expect `Space`s read from a file to be bitwise identical to the original
+one.
+
+
 PR [#2021](https://github.com/CliMA/ClimaCore.jl/pull/2021).
 
 v0.14.16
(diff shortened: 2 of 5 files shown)


test/InputOutput/hybrid2dbox.jl
```diff
@@ -116,4 +116,5 @@ end
     restart_Y = InputOutput.read_field(reader, "Y") # read fieldvector from hdf5 file
     close(reader)
     @test restart_Y == Y # test if restart is exact
+    @test typeof(Y) == typeof(restart_Y)
```
charleskawczynski, line 119: I think these tests are actually redundant in that `Y == restart_Y` effectively already tests for type diagonalization through `Fields.rcompare`. Should we remove them?

Sbozzolo: I found this to be necessary. The test for `Y == Y_restarted` for deep atmosphere passes even without fixing the bug. I didn't investigate why `rcompare` doesn't catch that

charleskawczynski: Hm, but the method definition for `rcompare` is

```julia
rcompare(x::T, y::T) where {T <: Union{FieldVector, NamedTuple}} =
    _rcompare(true, x, y)
```
So, if the types aren't exactly equal, it will method error. Are `restart_Y` and `Y` `FieldVector`s or something else?

Sbozzolo: Here's a reproducer
```

Comment (Sbozzolo) [reply test/InputOutput/hybrid2dbox.jl:119]: This is why I added the explicit type comparison, but I am happy to remove it if you figure out what's going wrong with `rcompare`

Comment (Sbozzolo) [reply test/InputOutput/hybrid2dbox.jl:119]: Ah, I know what's wrong:

```julia
julia> Fields.rcompare(restart_Y, Y)
ERROR: MethodError: no method matching rcompare(::ClimaCore.Fields.FieldVector{…}, ::ClimaCore.Fields.FieldVector{…})

Closest candidates are:
  rcompare(::T, ::T) where T<:Union{NamedTuple, ClimaCore.Fields.FieldVector}
   @ ClimaCore ~/repos/ClimaCore.jl/src/Fields/fieldvector.jl:456

Stacktrace:
 [1] top-level scope
   @ REPL[41]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> restart_Y == Y
true
```

Sbozzolo:
```
julia> @which (==)(restart_Y,Y)
==(A::AbstractArray, B::AbstractArray)
     @ Base abstractarray.jl:2937
```

charleskawczynski: Thanks! Let me try this out (we'd better be sure that `rcompare` is working correctly if we're using it for correctness tests!)

charleskawczynski: Ahhhh, it's _not_ calling `rcompare`! The types don't match, and so it falls back on `==` for abstract arrays. Perhaps we should catch the case when it doesn't, and throw an error? Good catch!

charleskawczynski: Calling `rcompare` directly will work, but I think we'd better just define a version that breaks when comparing two different types

src/Fields/fieldvector.jl
```diff
@@ -487,10 +487,14 @@ end
...
+FieldVectors with different types are considered different.
 """
 rcompare(x::T, y::T) where {T <: Union{FieldVector, NamedTuple}} =
     _rcompare(true, x, y)
 
-# Define == to call rcompare for two fieldvectors of the same
-# exact type.
-Base.:(==)(x::T, y::T) where {T <: FieldVector} = rcompare(x, y)
+# FieldVectors with different types are always different
+rcompare(x::T1, y::T2) where {T1 <: FieldVector, T2 <: FieldVector} = false
```
charleskawczynski, line 497:
```suggestion
rcompare(x::FieldVector, y::FieldVector) = false
```

charleskawczynski: This is the same thing, but it won't force specialization

charleskawczynski (approved):

PR #2016, opened 2024-09-27
add back uniform_vertical_levels

juliasloan25:
#1992 removed the `uniform_vertical_levels` function. This led to a GPU error downstream in ClimaLand (see error message below). This PR reverts that change and makes a patch release - this has been tested from ClimaLand (see run [here](https://buildkite.com/clima/climaland-long-runs/builds/1182)).

The error came up in ClimaLand in a PR switching the canopy temperature variable from being stepped explicitly to implicitly. We also step soil water content implicitly, but there has been no error with that case. In the soil-only case, our implicit variable is defined on a `ExtrudedFiniteDifferenceSpace`. In the new combined soil/canopy case, we have one implicit variable on a `ExtrudedFiniteDifferenceSpace`, and the other on a `SpectralElementSpace2D`. Maybe the bug comes from having two variables on two different spaces, or from the `SpectralElementSpace2D` specifically?

Here is the error (e.g. from [this failing run](https://buildkite.com/clima/climaland-long-runs/builds/1148#01922b97-b120-4fd6-b316-11c9b6b7f504)):
```
ERROR: LoadError: InvalidIRError: compiling MethodInstance for ClimaCoreCUDAExt.multiple_field_solve_kernel!(::ClimaComms.CUDADevice, ::Tuple{…}, ::Tuple{…}, ::Tuple{…}, ::Tuple{…}, ::ClimaCore.Fields.Field{…}, ::ClimaCore.DataLayouts.UniversalSize{…}, ::Val{…}) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to _single_field_solve!)
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/ClimaCore/8E0O8/ext/cuda/matrix_fields_multiple_field_solve.jl:70
 [2] generated_single_field_solve!
   @ ~/.julia/packages/ClimaCore/8E0O8/ext/cuda/matrix_fields_multiple_field_solve.jl:56
 [3] multiple_field_solve_kernel!
   @ ~/.julia/packages/ClimaCore/8E0O8/ext/cuda/matrix_fields_multiple_field_solve.jl:95
Hint: catch this exception as `err` and call `code_typed(err; interactive = true)` to introspect the erronous code with Cthulhu.jl
Stacktrace:
  [1] check_ir(job::GPUCompiler.CompilerJob{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams}, args::LLVM.Module)
[...]

PR #2024, opened 2024-10-02
Fix multiple field solve reproducer

charleskawczynski:
It turns out that #2016 effectively does fix the issue for the land, but doesn't fix the multiple field solve reproducer because #2016 results in dispatch to the `single_field_solve!` method (not the internal `_single_field_solve!` method for the gpu). So, this PR adjusts the reproducer to be more faithful to where dispatch sends flow control.

The missing implementation, which is caught in this case in `src/` is:

```julia
function single_field_solve_diag_matrix_row!(
    cache,
    x,
    A::ColumnwiseBandMatrixField,
    b,
)
    Aⱼs = unzip_tuple_field_values(Fields.field_values(A.entries))
    b_vals = Fields.field_values(b)
    x_vals = Fields.field_values(x)
    (A₀,) = Aⱼs
    @. x_vals = inv(A₀) ⊠ b_vals
end
single_field_solve!(_, x, A::UniformScaling, b) = x .= inv(A.λ) .* b
function single_field_solve!(cache, x, A::ColumnwiseBandMatrixField, b)
    if eltype(A) <: MatrixFields.DiagonalMatrixRow
        single_field_solve_diag_matrix_row!(cache, x, A, b)
    else
        single_field_solve!(ClimaComms.device(axes(A)), cache, x, A, b)
    end
end
```
And based on this, we may want to keep dispatch how it was, since `@. x_vals = inv(A₀) ⊠ b_vals` is just as parallelizeable as the UniformScaling case. So we can update that logic accordingly.

### DIFF
--- a/test/MatrixFields/multiple_field_solve_reproducer_1.jl
+++ b/test/MatrixFields/multiple_field_solve_reproducer_1.jl
@@ -104,28 +104,11 @@ explicit_blocks =
 
 A_mf = MatrixFields.FieldMatrix(implicit_blocks..., explicit_blocks...);
 A = MatrixFields.replace_name_tree(A_mf, keys(b).name_tree);
-names = MatrixFields.matrix_row_keys(keys(A))
 
 alg = MatrixFields.BlockDiagonalSolve();
 solver = MatrixFields.FieldMatrixSolver(alg, A_mf, vector);
 (; cache) = solver;
-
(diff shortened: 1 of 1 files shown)


charleskawczynski:
I ran this locally, and it works as intended, going to manually merge to avoid wasting CI cycles.

PR #2028, opened 2024-10-03
Bump patch versions/compats, update the news

charleskawczynski:
This PR bumps some patch versions / compats for new releases, and updates the news.

### DIFF
--- a/lib/ClimaCorePlots/Project.toml
+++ b/lib/ClimaCorePlots/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCorePlots"
 uuid = "cf7c7e5a-b407-4c48-9047-11a94a308626"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.2.11"
+version = "0.2.12"
 
 [deps]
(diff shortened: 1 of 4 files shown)


lib/ClimaCorePlots/Project.toml
```diff
@@ -11,7 +11,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TriplotBase = "981d1d27-644d-49a2-9326-4793e63143c3"
 
 [compat]
-ClimaCore = "0.11, 0.12, 0.13, 0.14"
+ClimaCore = "0.11, 0.12, 0.13, 0.14, 0.18"
```
Sbozzolo, line 14:
```suggestion
ClimaCore = "0.11, 0.12, 0.13, 0.14"
```

We are not at version 0.18 yet :)

Does this need ClimaCore 0.14.18?

Same below for ClimaCoreTempestRemap

charleskawczynski: You're right, whoops! We can leave it as is since it's a patch update.

charleskawczynski: I was actually scratching my head about this when I was making the changes.

Sbozzolo (approved):

PR #2029, opened 2024-10-04
Update TagBot

charleskawczynski:
I've been getting notifications about Tagbot not finding tags: [xref](https://github.com/CliMA/ClimaCore.jl/issues/186#issuecomment-2394024120), so I thought I'd try out `lookback` as discussed here: https://github.com/JuliaRegistries/TagBot?tab=readme-ov-file#lookback-period

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -4,6 +4,22 @@ on:
     types:
       - created
   workflow_dispatch:
+    inputs:
+      lookback:
+        default: "3"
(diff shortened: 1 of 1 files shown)


Sbozzolo:
It seems that the lookback of 3 is the default, so part of this change seems redundant. But I also don't know how we could approach this problem, so we could maybe start merging this (this is the same config as in the example) and see if it fixes things.

Sbozzolo (approved):

PR #2033, opened 2024-10-10
Add DataLayouts convenience constructors

charleskawczynski:
I've found myself churning a lot of code in the test suite, related to construction of datalayouts.

This PR adds new convenience constructors for datalayouts, meant to be used at the top-level (as they are not inferable).

They have a few advantages over the existing way we construct them. Concretely, they:
 - Require less code duplication
 - Infer `Nf`
 - Use keywords for clarity, and are therefore order-agnostic

So, instead of writing

```julia
array = rand(FT, Nv, Nij, Nij, 3, Nh)
data = VIJFH{S, Nv, Nij}(array)
```

We can write:

```julia
data = VIJFH{S}(ArrayType{FT}, rand; Nv, Nij, Nh)
array = parent(data) # (if you still need `array`)
```

So, this will hopefully reduce the surface area of these constructors if we use them consistently in the test suite.

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -302,6 +302,13 @@ function replace_basetype(data::AbstractData{S}, ::Type{T}) where {S, T}
     )
 end
 
+maybe_populate!(array, ::typeof(similar)) = nothing
+maybe_populate!(array, ::typeof(ones)) = fill!(array, 1)
+maybe_populate!(array, ::typeof(zeros)) = fill!(array, 0)
+function maybe_populate!(array, ::typeof(rand))
+    parent(array) .= typeof(array)(rand(eltype(array), size(array)))
+end
+
 # ==================
 # Data3D DataLayout
 # ==================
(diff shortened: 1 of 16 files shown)


src/DataLayouts/DataLayouts.jl
```diff
@@ -302,12 +302,20 @@ function replace_basetype(data::AbstractData{S}, ::Type{T}) where {S, T}
...
+
 # ==================
 # Data3D DataLayout
 # ==================
 
 """
     IJKFVH{S, Nij, Nk}(array::AbstractArray{T, 6}) <: Data3D{S, Nij, Nk}
+    IJKFVH{S}(ArrayType[, ones | zeros | rand]; Nij, Nk, Nv)
 
 A 3D DataLayout. TODO: Add more docs
```
Sbozzolo, line 320: I think these all look good, but can you please add more documentation? 

It took me a couple of minutes to wrap my head around what they were doing and I had to read the code to understand it.

I think the docstrings should explain what these constructors do and highlight any possible pitfall with using them.

charleskawczynski: Yeah, good point. I added some docs, can you take a second look?

Sbozzolo (approved):
Thank you! Looks good to me!

PR #2040, opened 2024-10-15
Reduce unnecessary specializations

charleskawczynski:
This PR removes some unnecessary specializations.

One thing I realized is that we were compiling quite a few `copyto!` methods because of all the `slab` / `column` calls, which forwarded to additional `copyto!` methods that then needed to be compiled. I don't think that this actually buys us anything.

I used https://timholy.github.io/SnoopCompile.jl/dev/tutorials/pgdsgui/ to apply some de-specializations. Here's one result:

```julia
copyto!(dest::ClimaCore.DataLayouts.AbstractData, bc::Union{Base.Broadcast.Broadcasted, ClimaCore.DataLayouts.AbstractData}) @ ClimaCore.DataLayouts ~/dev/CliMA/ClimaCore.jl/src/DataLayouts/copyto.jl:5
(397 specializations) # before

copyto!(dest::ClimaCore.DataLayouts.AbstractData, bc::Union{Base.Broadcast.Broadcasted, ClimaCore.DataLayouts.AbstractData}) @ ClimaCore.DataLayouts ~/dev/CliMA/ClimaCore.jl/src/DataLayouts/copyto.jl:5
(... specializations) # now (pending...)
```

I also applied `strip_space` to the broadcasted objects in the stencil kernels, since this could reduce specializations on `getindex` of the broadcasted object.

I also added a singleton type into the datalayouts, which we can use to keep our generic code while avoid overspecializing on specific datalayouts.

### DIFF
--- a/ext/cuda/data_layouts_copyto.jl
+++ b/ext/cuda/data_layouts_copyto.jl
@@ -1,4 +1,4 @@
-DataLayouts._device_dispatch(x::CUDA.CuArray) = ToCUDA()
+DataLayouts.device_dispatch(x::CUDA.CuArray) = ToCUDA()
 
 function knl_copyto!(dest, src, us)
     I = universal_index(dest)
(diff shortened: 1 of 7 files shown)


charleskawczynski:
Superseded by #2041, #2042 and #2043

Issue #1871, opened 2024-07-08
`getidx` spends most of its time in `eltype` for at least some MatrixField operations


charleskawczynski:
I was inspired by some performance helpdesk discussion over the weekend, and made a very simple reproducer for our matrix-field `getidx`. I recently had a hunch that we should hoist our `eltype` calls, but @dennisYatunin pointed out that this should simply return a compile-time constant. The benchmark shows, however, that we're spending most of the time in `eltype`:

```julia
#=
julia --project
using Revise; include(joinpath("test", "Operators", "finitedifference", "getidx.jl"))
=#
using ClimaComms
ClimaComms.@import_required_backends
import BenchmarkTools
import ClimaCore
@isdefined(TU) || include(
    joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"),
);
include(joinpath(pkgdir(ClimaCore), "test", "MatrixFields", "matrix_fields_broadcasting", "test_scalar_utils.jl"))
import .TestUtilities as TU;
using Test
using JET

import ClimaCore.MatrixFields: ⋅
import ClimaCore: Utilities, Spaces, Fields, Operators
import LazyBroadcast: @lazy

function call_getidx(space, bc, loc, idx, hidx)
    @inbounds Operators.getidx(space, bc, loc, idx, hidx)
    return nothing
end

device = ClimaComms.device()
space = TU.CenterExtrudedFiniteDifferenceSpace(
    FT;
    zelem = 30,
    helem = 4,
    context = ClimaComms.context(device),
)
bc = @lazy @. ᶠᶜmat ⋅ ᶜᶠmat ⋅
    (2 * ᶠᶜmat ⋅ ᶜᶜmat ⋅ ᶜᶠmat + ᶠᶠmat ⋅ ᶠᶠmat / 3 - (4I,)) ⋅ ᶠᶠmat ⋅
    (ᶠᶜmat ⋅ ᶜᶜmat ⋅ ᶜᶠmat * 2 - (ᶠᶠmat / 3) ⋅ ᶠᶠmat + (4I,)) ⋅ ᶠᶠmat;
loc = Operators.Interior()
idx = 10
idx = Utilities.PlusHalf(10)
hidx = (1, 1, 1)
call_getidx(space, bc, loc, idx, hidx)
BenchmarkTools.@benchmark call_getidx(space, bc, loc, idx, hidx)

import Profile, ProfileCanvas

[...]

PR #2043, opened 2024-10-15
Reduce number of specialized methods in stencil kernels

charleskawczynski:
This is another peel off from https://github.com/CliMA/ClimaCore.jl/pull/2040.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -3327,30 +3327,6 @@ function Base.similar(
     return Field(Eltype, sp)
 end
 
-function _serial_copyto!(field_out::Field, bc, Ni::Int, Nj::Int, Nh::Int)
-    space = axes(field_out)
-    bounds = window_bounds(space, bc)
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Interesting, this still fails in some relatively complex cases (e.g., our `wfact` opt examples). I think that this revealed that some of the functions in our `window_bounds` seems to either require inlining, or function barriers to pass JET tests, which reminds me of #1871.

charleskawczynski:
One thing I definitely found is that we need this definition:

```julia
strip_space_args(args::Tuple{Any}, space) =
    (strip_space(args[1], space), )
```

charleskawczynski:
Perhaps `map((arg, width) -> arg - width[1], args_idx, widths)` needs to be unrolled?

charleskawczynski:
This didn't really improve things. The primary cost seems to be due to `getidx`. We've improved it some, but it may be that we need compiler improvements to see better results.

PR #2045, opened 2024-10-16
Change FH -> HF

charleskawczynski:
WIP to explore HF datalayouts.


sriharshakandala:
I think this can happen naturally, with existing data layouts, by separating the prognostic variables. This naturally restricts the `NF` to at most `3`, except for passive scalars!

charleskawczynski:
> I think this can happen naturally, with existing data layouts, by separating the storage for prognostic variables. This naturally restricts the `NF` to at most `3` (for example for velocity), except for passive scalars! The F dimension can be used for tuning most expensive kernels occuring in our workload rather than for collecting storage for all (cell-centered or face-centered) variables

I'm not sure I follow, avoiding integer division, which is one of the primary advantages of these datalayouts, cannot be solved by restricting Nf to at most 3. And we can still tune kernels with this new datalayout.

charleskawczynski:
Superseded by #2055

charleskawczynski:
> I think this can happen naturally, with existing data layouts, by separating the storage for prognostic variables. This naturally restricts the `NF` to at most `3` (for example for velocity), except for passive scalars! The `F` dimension can be used for tuning most expensive kernels occuring in our workload rather than for collecting storage for all (cell-centered or face-centered) variables

I'm not sure I follow this comment. This PR was changing the field dimension to the end in order to allow us to specialize on linear indexing for field variables without requiring users to change their code.

PR #2047, opened 2024-10-17
Remove use of farray in dss

charleskawczynski:
This PR replaces `farray` with use of `size` and `ncomponents`. `farray` is useful in some places, but it returns the parent array size, which is kind of internal details.


charleskawczynski:
Superseded by #2049

Issue #2048, opened 2024-10-17
Use less internals in DSS


charleskawczynski:
Our CUDA DSS implementation uses lots of DataLayout internals. Making any material changes to DataLayouts results in needing to nearly rewrite all of `ext/cuda/topologies_dss.jl`. I think that we need to refactor this code and avoid use of internals.

This is the biggest piece that blocks https://github.com/CliMA/ClimaCore.jl/pull/2045.

PR #2049, opened 2024-10-17
Use less DataLayouts internals in DSS

charleskawczynski:
This is a first step towards #2048.

### DIFF
--- a/ext/cuda/data_layouts_threadblock.jl
+++ b/ext/cuda/data_layouts_threadblock.jl
@@ -170,6 +170,20 @@ end
 ##### Custom partitions
 #####
 
+##### linear partition
+@inline function linear_partition(nitems::Integer, n_max_threads::Integer)
+    threads = min(nitems, n_max_threads)
+    blocks = cld(nitems, threads)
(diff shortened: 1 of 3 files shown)


sriharshakandala:
Generally speaking, `max_threads = 256` works well. If resource constraints (shmem size, etc.) permit, CUDA will automatically schedule multiple thread blocks on the same streaming multiprocessor. Its generally up to the user to design the best threadblock configuration for their application, and I believe we should custom tune it, especially for performance critical kernels!

charleskawczynski:
> Generally speaking, `max_threads = 256` works well. If resource constraints (shmem size, register usage, etc.) permit, CUDA will automatically schedule multiple thread blocks on the same streaming multiprocessor and no additional intervention is required. Using `threads_via_occupancy` does not necessarily guarantee optimality! Its up to the user to design the best `threadblock` configuration for their application, and I believe we should custom tune it, especially for performance critical kernels!

I'm fine with reverting that for now because it's not really part of the refactor. I am curious how this differs from the occupancy API

PR #2055, opened 2024-10-22
Add HF datalayouts

charleskawczynski:
This PR adds support for new datalayouts:

 - `VIJHF`
 - `IJHF`
 - `IHF`
 - `VIHF`

For these new datalayouts, and other datalayouts that end with the field index, we'll be able to add fast paths for linear indexing. This will simply involve reviving https://github.com/CliMA/ClimaCore.jl/pull/1922, and specializing on all datalayouts with the field index at the end.

Finally, we can compare this with our `VIJFH` datalayout performance.

To make this work with our existing implementation, users can now make horizontal spaces with a new keyword argument:
`HorizontalLayout = DataLayouts.IJFH`, which users may change to `HorizontalLayout = DataLayouts.IJHF` to try the new datalayout.


charleskawczynski:
cc @sriharshakandala, @dennisYatunin, @Sbozzolo, this PR may not be completely ready as I'm ironing out some issues, but can you please review the extension of our datalayouts and the design of how I've added the ability to make HF spaces?

examples/common_spaces.jl
```diff
@@ -35,14 +35,15 @@ function make_horizontal_space(
     mesh,
     npoly,
     context::ClimaComms.SingletonCommsContext,
+    HorizontalLayout = DataLayouts.IJFH,
```
dennisYatunin, line 38: Could the `horizontal_layout` keyword argument use camel case instead? That would be a little more standard.

charleskawczynski: I did think about this, my main motivation for using TitleCase was that it is a Type, I don't have strong preferences though. @sriharshakandala, what do you think?

src/DataLayouts/non_extruded_broadcasted.jl
```diff
@@ -0,0 +1,139 @@
...
+# To handle scalar cases, let's just switch back to
+# Base.Broadcast.Broadcasted and allow cartesian indexing:
+Base.@propagate_inbounds Base.getindex(bc::NonExtrudedBroadcasted) = to_broadcasted(bc)[CartesianIndex(())]
+
+import StaticArrays
+to_tuple(::Tuple{}) = ()
+to_tuple(t::Tuple) = t
+to_tuple(t::NTuple{N, <: Base.OneTo}) where {N} = map(x->x.stop, t)
+to_tuple(t::NTuple{N, <: StaticArrays.SOneTo}) where {N} = map(x->x.stop, t)
+n_dofs(bc::NonExtrudedBroadcasted) = prod(to_tuple(axes(bc)))
```
dennisYatunin, line 91:
```suggestion
n_dofs(bc::NonExtrudedBroadcasted) = prod(length, axes(bc))
```
Does this work? It's a little simpler to read.

charleskawczynski: I can try

charleskawczynski: Hm, I _think_ that should be fine, but it does assume that `axes` of `bc` is a `Base.OneTo` or `StaticArrays.SOneTo`. If it returned a tuple of `Int`s, then `length(4) # 1` will not error but give a wrong number for `n_dofs`. Ok, yeah, it looks like `axes` is always supposed to return a range, I can change it.

charleskawczynski: It turns out that this didn’t work for some reason, so I reverted it.

dennisYatunin: Hmm, that's strange. What's the example it didn't work for, and are we sure it's correct as is? I would expect the total number of degrees of freedom to be the product of the number of degrees of freedom along each axis, where the latter should correspond to the number of elements along the axis.

charleskawczynski: Here is the build: https://buildkite.com/clima/climacore-ci/builds/4612, looks like a lot of jobs broke, actually

charleskawczynski: Ah, I got the edge case wrong, I had `n_dofs(bc::NonExtrudedBroadcasted) = prod(length, axes(bc); init=0)` and it should apparently be `n_dofs(bc::NonExtrudedBroadcasted) = prod(length, axes(bc); init=1)`-- I didn't expect this behavior from `prod`:

```julia
julia> prod(())
1
```

Back to your question of is it correct as is: I think it's mostly correct, but I'll add some unit tests to give us more confidence.

charleskawczynski: I think I fixed the remaining cases that I could think of.

PR #2071, opened 2024-11-01
Fix GPU inference (2065)

### LINKED ISSUE #2065: inference failure when broadcasting over multiple structs and a Field
## Describe the bug
Reproducer added in #2062 
Initially found in https://github.com/CliMA/ClimaLand.jl/pull/739

charleskawczynski:
This PR fixes the GPU inference issue described in #2065. Closes #2065.

### DIFF
--- a/src/DataLayouts/non_extruded_broadcasted.jl
+++ b/src/DataLayouts/non_extruded_broadcasted.jl
@@ -140,8 +140,21 @@ Base.@propagate_inbounds function _broadcast_getindex(
 end
 @inline _broadcast_getindex_evalf(f::Tf, args::Vararg{Any, N}) where {Tf, N} =
     f(args...)  # not propagate_inbounds
-Base.@propagate_inbounds _getindex(args::Tuple, I) =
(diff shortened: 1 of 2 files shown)


charleskawczynski:
On downside of this, I realized, is that this solution only works for data layouts that end with the field index, and only works on Julia 1.11.

An ideal solution would be to use an immutable version of`Ref`. We’ve had issues with `Ref` in the past, which is why we switched to using tuples.

We tried to recreate this issue with ordinary cuarrays, but we weren’t able to reproduce the error. Which suggests that there is something else different. Perhaps it’s worth taking a second look at this, since it could lead to a fix on our end, or a fix in GPUCompiler/julia inference.

charleskawczynski:
I did manage to make a reproducer in https://github.com/JuliaGPU/CUDA.jl/issues/2623. I think the proper path forward is to further simplify the reproducer in that issue and resolve the compiler limitation (whether that be in Base or in CUDA.jl), so that this issue is properly fixed. I'm going to close this PR, since it's really only a monkey patch that is not really widely applicable.

PR #2070, opened 2024-11-01
Add FieldVector microbenchmarks and improve fieldvector broadcast performance

charleskawczynski:
Closes #2067. This turned out to be pretty easy: any fieldvector operation is embarrassingly parallel. We can leverage `NonExtrudedBroadcasted`, forward everything to the backing arrays, and just linearly index everywhere. This puts us in a pretty good shape for fieldvector operations, and it'll be even better at lower resolution since we parallelize across field variables.

#### main
```julia
N reads-writes: 5,  Float_type = Float64, Device_bandwidth_GBs=2039
┌─────────────┬───────────────────────────────────┬─────────┬─────────────┬──────────────┬────────┐
│ funcs       │ time per call                     │ bw %    │ achieved bw │ problem size │ n-reps │
├─────────────┼───────────────────────────────────┼─────────┼─────────────┼──────────────┼────────┤
│ FieldVector │ 1 millisecond, 104 microseconds   │ 54.161  │ 1104.34     │ (32745600,)  │ 4475   │
│ FieldVector │ 307 microseconds, 597 nanoseconds │ 48.6243 │ 991.45      │ (8186400,)   │ 10000  │
└─────────────┴───────────────────────────────────┴─────────┴─────────────┴──────────────┴────────┘
```
#### this PR
```julia
N reads-writes: 5,  Float_type = Float64, Device_bandwidth_GBs=2039
┌─────────────┬───────────────────────────────────┬─────────┬─────────────┬──────────────┬────────┐
│ funcs       │ time per call                     │ bw %    │ achieved bw │ problem size │ n-reps │
├─────────────┼───────────────────────────────────┼─────────┼─────────────┼──────────────┼────────┤
│ FieldVector │ 757 microseconds, 513 nanoseconds │ 78.9779 │ 1610.36     │ (32745600,)  │ 6491   │
│ FieldVector │ 202 microseconds, 849 nanoseconds │ 73.7335 │ 1503.43     │ (8186400,)   │ 10000  │
└─────────────┴───────────────────────────────────┴─────────┴─────────────┴──────────────┴────────┘
```

#### main
```julia
N reads-writes: 5,  Float_type = Float64,
┌─────────────┬──────────────────────────────────┬─────────────┬──────────────┬────────┐
│ funcs       │ time per call                    │ achieved bw │ problem size │ n-reps │
├─────────────┼──────────────────────────────────┼─────────────┼──────────────┼────────┤
[...]

PR #2072, opened 2024-11-04
Add fieldvector unit tests, update `rcompare`

charleskawczynski:
This PR adds a fieldvector unit test, which was extracted from https://github.com/CliMA/ClimaCore.jl/pull/2070, where the unit tests passed but a test in ClimaLand broke.

It's still not clear to me whether this is exercising broadcasting over different fieldvectors types with the same propertynames, or something different. Either way, we should probably test for that, plus, we could update `rcompare` to handle fieldvectors with different propertyname orders.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ ClimaCore.jl Release Notes
 main
 -------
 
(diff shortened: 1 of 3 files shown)


charleskawczynski:
I think updating `rcompare` to handle different propertyname orders is okay, since users can easily test `typeof(x)` == `typeof(y)`, but I figure we can make these a keyword option so that users can pick which they want. By default, though, the `==` will be strict, I think that's probably the safest option.

charleskawczynski:
This fails when combined with #2070, so this should help us figure out what is wrong with #2070.

PR #2075, opened 2024-11-06
Automatically split fused kernels by parameter memory limits

charleskawczynski:
This PR leverages https://github.com/CliMA/MultiBroadcastFusion.jl/pull/52, so that we can have:

```julia
@fused_assemble begin # (or many broadcast expressions using `@fused_direct`)
    for i in 1:100000
        @. x += y * z
    end
end
```
And this will automatically fuse as many kernels as permitted by the available parameter memory.


charleskawczynski:
Hm, let's split off the dependency updates

PR #2078, opened 2024-11-11
Add downgrade workflow, fix lower bounds

Sbozzolo:
This commit adds the downgrade github action and cleans up some incorrect lower bounds. I also removed a couple of packages that could be trivially removed.

Closes #1781

PR #2081, opened 2024-11-11
Qualify IntervalSets and OrderedCollections

charleskawczynski:
This PR qualifies types/methods in IntervalSets and OrderedCollections.

This is a peel off of https://github.com/CliMA/ClimaCore.jl/pull/2078.

### DIFF
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -1,7 +1,7 @@
 module Domains
 
 import ..Geometry: Geometry, float_type
-using IntervalSets
+import IntervalSets
 export RectangleDomain
 
 """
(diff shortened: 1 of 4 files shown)


Sbozzolo (commented):
The change for OrderedCollections is more subtle. I moved it to a different commit here: https://github.com/CliMA/ClimaCore.jl/pull/2078/commits/c35c57cce162ccca8eed26815203782c67edbb12

I have to remove OrderedCollecions because I could not find a lower version that was compatible.

charleskawczynski:
> The change for OrderedCollections is more subtle. I moved it to a different commit here: [c35c57c](https://github.com/CliMA/ClimaCore.jl/commit/c35c57cce162ccca8eed26815203782c67edbb12)
> 
> I have to remove OrderedCollecions because I could not find a lower version that was compatible.

Hm, that's a bit odd, DataStructures just depends on OrderedCollections

Sbozzolo:
I tried a little bit and couldn't get it to work. Given that the package is only use in tests and we already depend on DataStructures, it was simpler and quicker to just use DataStructures directly.

charleskawczynski:
> I tried a little bit and couldn't get it to work. Given that the package is only use in tests and we already depend on DataStructures, it was simpler and quicker to just use DataStructures directly.

That's a bit unfortunate, I'd prefer we use the lower level dependencies. That said, having correctly configured lower compat bounds seems like a much bigger win to me. So I'm fine with this.

Superseded by #2078.

PR #2084, opened 2024-11-14
Implement and test vertical mass borrowing limiters

### LINKED ISSUE #2083: Implement purely vertical water-borrowing limiter operators
The implementation we want is described in Appendix B, [here](https://gmd.copernicus.org/articles/11/1971/2018/gmd-11-1971-2018.pdf).

charleskawczynski:
This is a work in progress. This PR implements vertical mass borrowing limiters.

Closes #2083.

Todo:
 - [ ] How should `qmin` be computed?
 - [ ] Is `pdel` the correct interpretation?
 - [ ] Add meaningful tests
 - [ ] Add GPU implementation


charleskawczynski:
This does seem to work in that it removes negative tracer values, and seems to preserve mass (both of which are unit-tested cc @tapios):

On a single column:

```julia
    @test 0 ≤ minimum(q)
    @test isapprox(sum(q), sum_q_init; atol = 0.00000000001)
    @test isapprox(sum(q), sum_q_init; rtol = 0.01)
```

On a full cubed sphere:
```julia
    @test 0 ≤ minimum(q)
    @test isapprox(sum(q), sum_q_init; atol = 0.013)
    @test isapprox(sum(q), sum_q_init; rtol = 0.01)
```

I think `rtol = 0.01` should be sufficient on a sphere, right?

But, the transport/wavespeed doesn't look great:

Screenshot 2024-11-19 at 7 51 47 PM

I guess that's a bit orthogonal, though?

src/Limiters/vertical_mass_borrowing_limiter.jl
```diff
@@ -0,0 +1,119 @@
+import .DataLayouts as DL
+
+"""
+    VerticalMassBorrowingLimiter(f::Fields.Field, q_min)
+
+A vertical-only mass borrowing limier.
```
tapios, line 6:
```suggestion
A vertical-only mass borrowing limiter.
```

src/Limiters/vertical_mass_borrowing_limiter.jl
```diff
@@ -0,0 +1,119 @@
...
+
+The mass borrower borrows tracer mass from an adjacent layer.
```
tapios, line 8:
```suggestion
The mass borrower borrows tracer mass from an adjacent, lower layer.
```

src/Limiters/vertical_mass_borrowing_limiter.jl
```diff
@@ -0,0 +1,119 @@
...
+It conserves the mass and can avoid negative tracers.
```
tapios, line 9:
```suggestion
It conserves the total tracer mass and can avoid negative tracers.
```

charleskawczynski:
pdel is a hydrostatic approximation of rho*dz, so we need to update this.

Also, the mass conservation test at the end should be rho*q instead.

tapios (changes requested):
Mass integral from rho dz, rather than dp

charleskawczynski:
And let's visualize how this looks on the sphere.

src/Limiters/vertical_mass_borrowing_limiter.jl
```diff
@@ -0,0 +1,133 @@
...
+    @. ic = 0
+    @. bmass = 0
+    q_vals = Fields.field_values(q)
+    # top to bottom
+    for f in 1:DataLayouts.ncomponents(q_vals)
+        for v in 1:nlevels
+            CI = CartesianIndex(1, 1, f, v, 1)
+            # new mass in the current layer
+            ρΔz_v =
+                DL.getindex_field(Δz_vals, CI) * DL.getindex_field(ρ_vals, CI)
```
charleskawczynski, line 95:
```suggestion
                DL.getindex_field(ΔV_vals, CI) * DL.getindex_field(ρ_vals, CI)
```
where `ΔV_vals = WJ` `W` has horizontal terms + has an additional `/2` on boundaries, so instead use `ΔV_vals = J`. (thanks, @dennisYatunin!).

PR #2102, opened 2024-12-16
Fix reading/writing purely vertical spaces

### LINKED ISSUE #2099: Support `DataLayouts.VF` in `InputOutput.read_field()`
The [`read_field()`](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/InputOutput/readers.jl#L466) function does not support reading in `VF` fields at the moment. This is the type of data layout I get when storing values from 1D `FiniteDifferenceSpace` data, so I would like to be able to load it again.

The function should be adapted to support this type of `DataLayout`. Things that need to be fixed for sure to make this work:

- [ ] add a `VF` case to [`_scan_data_layout()`
](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/InputOutput/readers.jl#L147)
- [] [extend `h_dim()`](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/DataLayouts/DataLayouts.jl#L1722) for `VFSingleton` (needed [here](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/InputOutput/readers.jl#L486)
- [] after that, [this line](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/InputOutput/readers.jl#L495) fails for me with the following error, this needs to be fixed:

```
ERROR: MethodError: no method matching getindex(::Nothing, ::Int64)
The function `getindex` exists, but no method is defined for this combination of argument types.
```
Here I do not know what should be fixed.

For the `h_dim()` function, I was not sure what value should be put. 1? 0?

Sbozzolo:
Previously, the restart module assumed that there was an horizontal space.

Closes #2099


test/InputOutput/finitedifference.jl
```diff
@@ -0,0 +1,49 @@
+using Test
```
charleskawczynski, line 1: We need to include this to be run in the test suite somewhere. Also, can we can follow the convention and call this filename `unit_finitedifference.jl`?

Sbozzolo: All the other files in the folder also didn't have the `unit_` prefix, so I changed the names of all the files in `test/InputOutput`

(And added to `runtests`)

charleskawczynski: Ok, thanks!

PR #2103, opened 2024-12-16
Add `face/center_space`

Sbozzolo:
As I started using the `CommonSpaces`, I realized that specifying the staggering was a pain point. It required importing `Grids`, and I pretty much always wanted to build CenterSpaces anyways. In addition to this, I realized that going from center to faces and viceversa was not trivial. This required carrying around both the center and the face space when only one is really needed. In this commit, I add a default staggering to `CommonSpaces` and a function to go from center to face spaces and back.

### DIFF
--- a/src/CommonSpaces/CommonSpaces.jl
+++ b/src/CommonSpaces/CommonSpaces.jl
@@ -8,15 +8,20 @@ argument, `staggering::Staggering` to construct the desired space.
 module CommonSpaces
 
 export ExtrudedCubedSphereSpace,
-    CubedSphereSpace, ColumnSpace, Box3DSpace, SliceXZSpace, RectangleXYSpace
+    CubedSphereSpace,
+    ColumnSpace,
+    Box3DSpace,
+    SliceXZSpace,
+    RectangleXYSpace,
+    CellCenter,
+    CellFace
 
-export Grids
 import ClimaComms
 
 import ..DataLayouts,
     ..Meshes, ..Topologies, ..Geometry, ..Domains, ..Quadratures, ..Grids
 
-import ..Grids: Staggering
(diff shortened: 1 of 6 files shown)


src/CommonSpaces/CommonSpaces.jl
```diff
@@ -80,24 +79,38 @@ Note that these arguments are all the same as
...
+import Grids: FaceCenter
 space = ExtrudedCubedSphereSpace(;
     z_elem = 10,
     z_min = 0,
     z_max = 1,
     radius = 10,
     h_elem = 10,
     n_quad_points = 4,
-    staggering = Grids.CellCenter()
+    staggering = FaceCenter()
```
charleskawczynski, line 102:
```suggestion
    staggering = CellFace()
```

src/CommonSpaces/CommonSpaces.jl
```diff
@@ -335,14 +354,17 @@ space = SliceXZSpace(;
...
 ```
 """
 function SliceXZSpace end
 SliceXZSpace(; kwargs...) = SliceXZSpace(Float64; kwargs...)
-SliceXZSpace(::Type{FT}; staggering::Staggering, kwargs...) where {FT} =
+SliceXZSpace(
+    ::Type{FT};
+    staggering::Staggering = Grids.CellCenter(),
+    kwargs...,
+) where {FT} =
```
charleskawczynski, line 367: `staggering` should not have a default value, just like `z_elem` and alike. Users need to be aware of what they're making.

charleskawczynski (commented):
Overall, I like the idea of having simple convenience adaptors between cell cell center and cell face spaces, but I think adding defaults for the common space constructors should be reverted.

Issue #2091, opened 2024-11-21
adapt for spaces/fields is broken


haakon-e:
## Describe the bug
`adapt`ing a (center)space leads to malformed struct, which is realized during printing of resulting object.

As a consequence, the space cannot be used to set up a field correctly.

<details><summary> error message </summary>

```julia
julia> Fields.Adapt.adapt(array_type, center_space)
CenterExtrudedFiniteDifferenceSpace:
  context: Error showing value of type ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{ClimaCore.Grids.DeviceExtrudedFiniteDifferenceGrid{ClimaCore.Topologies.DeviceIntervalTopology{@NamedTuple{bottom::Int64, top::Int64}}, ClimaCore.Quadratures.GLL{4}, ClimaCore.Geometry.CartesianGlobalGeometry, ClimaCore.DataLayouts.VIJFH{ClimaCore.Geometry.LocalGeometry{(1, 2, 3), ClimaCore.Geometry.XYZPoint{Float64}, Float64, StaticArraysCore.SMatrix{3, 3, Float64, 9}}, 82, 4, CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}}, ClimaCore.DataLayouts.VIJFH{ClimaCore.Geometry.LocalGeometry{(1, 2, 3), ClimaCore.Geometry.XYZPoint{Float64}, Float64, StaticArraysCore.SMatrix{3, 3, Float64, 9}}, 83, 4, CUDA.CuArray{Float64, 5, CUDA.DeviceMemory}}}, ClimaCore.Grids.CellCenter}:
ERROR: type DeviceExtrudedFiniteDifferenceGrid has no field horizontal_grid
Stacktrace:
  [1] getproperty
    @ ./Base.jl:37 [inlined]
  [2] horizontal_space(full_space::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{…})
    @ ClimaCore.Spaces ~/.julia/packages/ClimaCore/rSpyb/src/Spaces/extruded.jl:151
  [3] show(io::IOContext{…}, space::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{…})
    @ ClimaCore.Spaces ~/.julia/packages/ClimaCore/rSpyb/src/Spaces/extruded.jl:136
[...]

Issue #1790, opened 2024-06-10
Can we make the LocalGeometry / IntervalMesh lazy?


charleskawczynski:
The one frequent feedback tip we give to developers is: make small structs over large structs, especially if you want to make a field of them.

Unfortunately, we've not followed that tip ourselves, and the `LocalGeometry` struct is huge for `ExtrudedFiniteDifferenceSpace`s:

```julia
julia> DataLayouts.ncomponents(Spaces.local_geometry_data(space))
42
```

The reason that this is problem is that: the bandwidth of a cuda kernel depends on the number of struct fields accessed in the kernel. The more fields available, the larger the bandwidth can grow. If there is a 0-difference in the number of accesses between the large and small struct, then difference between the large and small structs is only about 9% (perhaps due to the difference in memory strides).

Here's a demo I was experimenting with:

```julia
#=
julia --project
include(joinpath("..", "perf", "struct_bandwidth.jl"))
=#
ENV["CLIMACOMMS_DEVICE"] = "CUDA"
using Revise
import ClimaComms
ClimaComms.@import_required_backends
import ClimaCore:Spaces, Fields
import ClimaCore
import ClimaCore.DataLayouts as DataLayouts
using BenchmarkTools;
@isdefined(TU) || include(joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"));
import .TestUtilities as TU;
FT = Float64;
space = TU.CenterExtrudedFiniteDifferenceSpace(FT;zelem=63, helem=30);
abstract type AbstractStruct end
Base.@kwdef struct BigStruct{T} <: AbstractStruct
    x1::T = 0
    x2::T = 0
    x3::T = 0
    x4::T = 0
    x5::T = 0
    x6::T = 0
    x7::T = 0
    x8::T = 0
    x9::T = 0
    x10::T = 0
    x11::T = 0
    x12::T = 0
    x13::T = 0
    x14::T = 0
    x15::T = 0
    x16::T = 0
    x17::T = 0
    x18::T = 0
    x19::T = 0
    x20::T = 0
    x21::T = 0
    x22::T = 0
    x23::T = 0
    x24::T = 0
    x25::T = 0
    x26::T = 0
    x27::T = 0
    x28::T = 0
    x29::T = 0
    x30::T = 0
    x31::T = 0
    x32::T = 0
    x33::T = 0
    x34::T = 0
    x35::T = 0
    x36::T = 0
    x37::T = 0
[...]

PR #2114, opened 2024-12-23
Define cpu<->gpu conversions

charleskawczynski:
I ran into this again today, and I think it's about time we fix support for this.


charleskawczynski:
I spoke with @juliasloan25 earlier today, and she and I both ran into a few issues, that I'll summarize here:

#### Partially implemented mirrored gpu objects

We have mirror objects (e.g., `Grids.DeviceExtrudedFiniteDifferenceGrid`) for the gpu because (AFAICT) their counterparts (i.e., `Grids.ExtrudedFiniteDifferenceGrid`) are _mutable_, which cannot be put on the gpu. Furthermore, whether it was a design choice or accidental, `DeviceSpectralElementGrid2D` only has some of the objects that `SpectralElementGrid2D` has, so we cannot adapt to `CuArray` and adapt back to `Array`, since we throw out some information when adapting to `DeviceSpectralElementGrid2D`. We could use a different mechanism (i.e., define "`device_adapt`" or something), however, cuda kernels may not work in that case. We could try to synchronize that data structures more, but that could lead to meta data issues (if that was a choice to remove some things). Another option would be to add small bits into our data structures that would allow us to recreate them. For example, `IntervalMesh` discards `stretch` once `faces` is established. Not discarding `stretch` could allow us to lazily represent the interval mesh (discussed in #1790), and reduce the amount of data needed to export / import for restarts (plus we could recreate everything we want here on the fly).

I'm generally in favor of preserving small constructor meta because of these nice properties, and I think it could offer us some alternative solutions. But I'm not sure it's worth the effort for this particular task, at this moment.

I think the value of having convenient cpu<->gpu conversions is worth exploring the route of synchronizing the structs so that we can convert everything back and forth.

#### ClimaComms device ambiguity

If we start with a `ClimaComms.CUDADevice`, and we want to adapt to the cpu, then it's unclear whether that conversion should be to a `CPUSingleThreaded` or `CPUMultiThreaded`, because there's simply no history to determine which the user wants. We could write wrappers around our adapt functions to do this, but that will require threading the device information all the way through to wherever it's needed. I'm inclined to say that, for this particular situation, using `CPUSingleThreaded` as a default may be worth the pain in lack of flexibility.

charleskawczynski:
I also spoke with @Sbozzolo recently about fixing this, and I think the conclusion is that we'll keep the `on-device` objects (or perhaps better named `in-kernel` objects) as slimmed down versions of the cpu versions, but we can allow adapting the cpu versions to on-device (gpu), with all of the same information.

This way we can still keep some distance from the meta data issues, while still being able to switch between cpu and gpu.

The way this works is:
 - `Adapt.adapt(Array, x)` returns a new object of `x` on the cpu
 - `Adapt.adapt(CUDA.CuArray, x)` returns a new object of `x` on the gpu
 - `Adapt.adapt(CUDA.KernelAdaptor, x)` returns a new object of `x` that is in-kernel compatible

Closes #2091.

charleskawczynski:
Closes #1296.

Sbozzolo:
I think it would still be useful to provide a `todevice` function, at least defined on `Fields` and `FieldVectors`, that internally calls the correct Adapt.

Maybe something like this?

```julia
"""
     todevice(device, field)

Move `Field` field to the given `device`.

This is particularly useful to move `Field`s from CPUs to GPUs and viceversa.

If the `field` is already defined on the target device, return a copy.
"""
function todevice(device::ClimaComms.AbstractDevice, field)
    return Adapt.adapt(ClimaComms.array_type(device), field)
end

todevice(::ClimaComms.CPUMultiThreaded, _) = error("Not supported")
```

(I don't know if adapting returns a copy)

This is so that downstream packages don't have to explicitely depend on Adapt and use it. I think Adapt comes with a learning curve and is not the most intuitive of the packages. The [documentation](https://github.com/JuliaGPU/Adapt.jl) does not make it clear how to use it and is not very beginner friendly. In addition to this, `ClimaComms` defines a notion of devices, and it would be more coherent to stick to them, instead of requiring one to understand Array vs CuArray.

In any case, all of this (both the user facing aspects and the implementation details) should be documented in the documentation.

charleskawczynski:
One downside of defining a new method, `todevice` ourselves, is that we need to define where it lives. Should it live in `ClimaCore.Fields`? in `ClimaCore`?

charleskawczynski:
It feels like `todevice` can be added after the fact, since it's just a wrapper method. Let me see if there's a way to document the adapt methods.

charleskawczynski:
Ah, Aqua is (rightfully so) complaining about piracy. Let's move those definitions to ClimaComms.

charleskawczynski:
I need to strip out the ClimaComms pieces. This PR will depend on https://github.com/CliMA/ClimaComms.jl/pull/103.

charleskawczynski:
The main functionality is fully implemented without the wrapper, so let's follow up with a wrapper in a subsequent PR.

Sbozzolo:
Can you make sure to add documentation in the PR? You added a new feature and you clarified the role of certain structs, but this GitHub discussion is the only place where one can find information about them

charleskawczynski:
> Can you make sure to add documentation in the next PR? You added a new feature and you clarified the role of certain structs, but this GitHub discussion is the only place where one can find information about them

Yes, there are a handful of doc pieces I’d like to update, and I thought that it’d be easier to group them together

PR #2115, opened 2025-01-03
Introduce post-operation callback

charleskawczynski:
It seems to me that, to date, one of the most challenging tasks that users have is: debug a large simulation that is breaking, e.g., yielding NaNs somewhere. This is especially complex for large models with many terms and implicit time-stepping with all the bells and whistles that we offer.

This PR attempts to provide tools to tackle this giant task, by introducing two functions:

 - `call_post_op_callback`
 - `post_op_callback`

One way that users may leverage this is, for example:

```julia
import ClimaCore
ClimaCore.DataLayouts.call_post_op_callback() = true
function ClimaCore.DataLayouts.post_op_callback(data)
    if any(isnan, data)
        error("Found NaNs!")
    end
end

# run simulation
```
Now, this assums that `any(isnan, data)` will work on all of the results-- it may not, but users can catch whatever data `post_op_callback` is called with and handle them accordingly. As such, this sort of debugging tool is capable of revealing that a given simulation errors in a wide range of locations, like [here](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/95e6109c49e232c57e9e2c9f8d11c06ee08b835e/src/nl_solvers/newtons_method.jl#L162), [here](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/95e6109c49e232c57e9e2c9f8d11c06ee08b835e/src/solvers/imex_ark.jl#L133), or [here](https://github.com/CliMA/ClimaCore.jl/blob/92f3b2beec8e292c5b32453d0d96fb121a017999/src/MatrixFields/field_matrix_solver.jl#L174), which may provide hints to users as to what is wrong.

I'm not sure exactly how useful this will be, but I'm curious myself to try in order to debug the flaky ClimaTimestepper issues (where I'm having this experiencing this problem of hunting down NaNs).

Since this would be a purely debugging tool, I would argue that it should not be a feature used in production (i.e., removing it should not be a breaking change).

Thoughts?


charleskawczynski:
Here's an [example](https://buildkite.com/clima/climatimesteppers-ci/builds/1437#01943905-37d7-4f70-94fb-ec65f2c3f991) of this feature "in action" where I'm using it to debug the flaky job(s) in ClimaTimesteppers.

Indeed, NaNs is first introduced (spatially) in `solve_newton!` -> `ldiv!` -> `run_field_matrix_solver!` -> `single_field_solve!`.

It looks like, since this was caught early enough, there is spatial information in where NaNs is showing up, which I hope turns out to be useful (I after-all suspected that the issue with this problem related to the boundary conditions).

PR #2117, opened 2025-01-06
Upgrade to Julia 1.11

charleskawczynski:
This PR upgrades us to Julia 1.11, and lifts some of the conditionals.

Closes #2030


charleskawczynski:
We were able to sidestep https://github.com/JuliaLang/julia/issues/56295 by dispatching from `Base.copyto!` directly, so we can still overload `getindex` on linear indices. The commit where we originally saw this issue, https://github.com/CliMA/ClimaCore.jl/pull/2018/commits/979558162e142dcf8ed7424d02444c3bd02c3a2e, had a different dispatch pattern in `Base.copyto!` for datalayouts. The main branch is dispatching more carefully.

charleskawczynski:
The only remaining issue is that there is still a [compile-time regression](https://github.com/julialang/julia/issues/56093), but perhaps that's okay with us since we're using the depot.

ext/cuda/data_layouts_copyto.jl
```diff
@@ -16,18 +16,26 @@ function knl_copyto_linear!(dest, src, us)
     return nothing
 end
 
-if VERSION ≥ v"1.11.0-beta"
```
Sbozzolo, line 19: We should keep this check for versions prior to 1.11.0-beta

ext/cuda/data_layouts_fill.jl
```diff
@@ -19,7 +19,7 @@ function Base.fill!(dest::AbstractData, bc, ::ToCUDA)
     us = DataLayouts.UniversalSize(dest)
     args = (dest, bc, us)
     if Nv > 0 && Nh > 0
-        if !(VERSION ≥ v"1.11.0-beta") && dest isa DataLayouts.EndsWithField
```
Sbozzolo, line 22: We should keep this check for versions prior to 1.11.0-beta

ext/cuda/data_layouts_fused_copyto.jl
```diff
@@ -120,8 +120,7 @@ function launch_fused_copyto!(fmb::FusedMultiBroadcast)
     destinations = map(p -> p.first, fmb.pairs)
     bcs = map(p -> p.second, fmb.pairs)
     if all(bc -> DataLayouts.has_uniform_datalayouts(bc), bcs) &&
-       all(d -> d isa DataLayouts.EndsWithField, destinations) &&
-       !(VERSION ≥ v"1.11.0-beta")
```
Sbozzolo, line 124: Same here

src/DataLayouts/DataLayouts.jl
```diff
@@ -2051,88 +2051,23 @@ end
 const EndsWithField{S} =
     Union{IJHF{S}, IHF{S}, IJF{S}, IF{S}, VF{S}, VIJHF{S}, VIHF{S}}
 
-if VERSION ≥ v"1.11.0-beta"
```
Sbozzolo, line 2054: And here

ext/cuda/data_layouts_copyto.jl
charleskawczynski: The main branch doesn't need this check, if you check out an older version of ClimaCore, it has this check, so removing it is still backwards compatible.

charleskawczynski: (Same for the ones below)

.buildkite/scaling/pipeline.yml
```diff
@@ -1,5 +1,5 @@
 env:
-  JULIA_VERSION: "1.10.0"
+  JULIA_VERSION: "1.11.2"
```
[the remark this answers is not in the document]
charleskawczynski: I think @sriharshakandala is working on reviving this pipeline for the dycore paper, so I'd rather not removing. I don't think there's harm in updating the version, whether it's running or not.

ext/cuda/data_layouts_copyto.jl
Sbozzolo: Just to confirm I understand, does ClimaCore main run on 1.10 and 1.11 independently of these checks? (So, removing this check is unrelated to upgrading to 1.11?)

charleskawczynski: Yes

Sbozzolo: Perfect, thank you!

.github/workflows/ClimaCoreMakie.yml
```diff
@@ -15,12 +15,13 @@ jobs:
   lib-climacore-makie:
     runs-on: ubuntu-20.04
     timeout-minutes: 45
+    strategy:
+      matrix:
+        version: ['1.10', '1.11']
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
```
Sbozzolo, line 24: You need this to:
```
        with:
          version: ${{ matrix.version }}
```

Same for the other github workflows below where you changed from 1.10 to 1.11

charleskawczynski:
Ah, this requires changing the github status, that's why the required checks aren't starting.

charleskawczynski:
I think JET, which is used in the test suite, is not yet compatible with Julia 1.11. So we may need to wait for that afterall. (at least in ClimaCore CI).

cc @Sbozzolo, @sriharshakandala

Sbozzolo:
> I think JET, which is used in the test suite, is not yet compatible with Julia 1.11. So we may need to wait for that afterall. (at least in ClimaCore CI).
> 
> cc @Sbozzolo, @sriharshakandala

Could you paste a link to the error? I don't see clear signs in the repo for JET saying that it is not compatible with 1.11 (but I can totally believe that it doesn't work)

charleskawczynski:
> Could you paste a link to the error? I don't see clear signs in the repo for JET saying that it is not compatible with 1.11 (but I can totally believe that it doesn't work)

https://buildkite.com/clima/climacore-ci/builds/4899#0194487a-5e29-48c7-a04e-524846f2a78f/198-1147

Sbozzolo:
> > Could you paste a link to the error? I don't see clear signs in the repo for JET saying that it is not compatible with 1.11 (but I can totally believe that it doesn't work)
> 
> https://buildkite.com/clima/climacore-ci/builds/4899#0194487a-5e29-48c7-a04e-524846f2a78f/198-1147

Interesting. I can install JET locally in my base env with 1.11.2, but not in the climacore enviornment. It probably conflicts with something else. I can look into this

Sbozzolo:
I think I found it! SnoopCompile was pinned to 2 (3 is available)

Sbozzolo:
Yes, it was SnoopCompile, which force JET to use version 0.8 instead of 0.9. I pushed a new commit with the updated Manifest.

charleskawczynski:
I'll let someone else revive this.

PR #2118, opened 2025-01-06
Fix Adapt calls, and remove device-side structs

charleskawczynski:
In efforts to support conversion between cpu and gpu, we need the device-side objects to retain as much as possible so that we can convert back and forth without losing information.

This PR does a few things:
 - Abstracts away getproperty calls to mpi context for `mpicomm`
 - Fixes some adapt calls-- we only want `adapt_structure` to return device structs when called with a `CUDA.KernelAdaptor`.
 - Remove the device side structs, and instead use the ClimaComms objects.

This should hopefully work for all cases except for CUDA-MPI jobs, since adapting from CPU to GPU with `mpicomm` (a mutable object) cannot reside on the device. This limitation should hopefully be fixed by https://github.com/CliMA/ClimaComms.jl/pull/102.


sriharshakandala:
Would we be able to 
1. Move all Adapt code out of `src/` to `ext/`.
2. Avoid adding `Adapt` explicitly to the source project and extensions.
For example, something similar to https://github.com/CliMA/ClimaInterpolations.jl/pull/3/files#diff-3702a35dcf2bad10ba7fc72fc0c50366627cd0679c4e3afda9d24c602ad82a8eR117 &

Sbozzolo (approved):

charleskawczynski:
Actually, we can overload `getproperty` on the ClimaComms context, and make it backwards compatible. So we don't even need to change the `.mpicomm` calls here. Going to revert that.

charleskawczynski:
Closing in favor of #2120 and #2114.

PR #2120, opened 2025-01-08
Move in-kernel specific adapt functions to cuda ext

charleskawczynski:
This PR is probably the smallest and least regrettable step towards fixing #2091. This is a subset of #2118.

I realized that we actually probably want to leave the generic adapt methods in `src/`, since they could be applied for both cuda and metal, for example. So, I've only moved the ones that are explicitly device-specific.

### DIFF
--- a/ext/ClimaCoreCUDAExt.jl
+++ b/ext/ClimaCoreCUDAExt.jl
@@ -20,6 +20,7 @@ import ClimaCore.RecursiveApply:
 import ClimaCore.DataLayouts: get_N, get_Nv, get_Nij, get_Nij, get_Nh
 import ClimaCore.DataLayouts: UniversalSize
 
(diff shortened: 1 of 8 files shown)


Sbozzolo (approved):
We could add another extension ClimaCoreAdapt for all these methods if we want to support more in the future. Adapt.jl is loaded by CUDA.jl and Metal.jl, so the extension would be automatically loaded when one of these backends is loaded

charleskawczynski:
> We could add another extension ClimaCoreAdapt for all these methods if we want to support more in the future. Adapt.jl is loaded by CUDA.jl and Metal.jl, so the extension would be automatically loaded when one of these backends is loaded

Agreed, but let's cross that bridge when we get there. I kind of like having only two orthogonal extensions at the moment.

Issue #2108, opened 2024-12-20
Distributed remapping interpolation bug


charleskawczynski:
Our unit tests seem to be flaky: https://buildkite.com/clima/climacore-ci/builds/4846#0193e0c9-7253-46f0-b029-699ab6c5bf95 (the corresponding PR does not touch anything related to distributed remapping / interpolation).

Sbozzolo:
I don't think I've ever seen this error, and we haven't touched that code in a long time.

I am wondering if something else has changed that led to this

Sbozzolo:
I still have no clue what this could be. What I see is that the entire array is not exactly wrong: it seems that the values are out of order.

Sbozzolo:
Here's what I am finding:
- On `clima`, I ran the test 20 times and it always passed
- On the centeral cluster, the following script fails most of the time
```julia
using Logging
using Test
using IntervalSets

import ClimaCore:
    Domains,
    Fields,
    Geometry,
    Meshes,
    Operators,
    Spaces,
    Quadratures,
    Topologies,
    Remapping,
    Hypsography
using ClimaComms
ClimaComms.@import_required_backends
const context = ClimaComms.context()
const pid, nprocs = ClimaComms.init(context)
const device = ClimaComms.device()
ArrayType = ClimaComms.array_type(device)

# log output only from root process
logger_stream = ClimaComms.iamroot(context) ? stderr : devnull
prev_logger = global_logger(ConsoleLogger(logger_stream, Logging.Info))
atexit() do
    global_logger(prev_logger)
end

@testset "3D sphere" begin
    vertdomain = Domains.IntervalDomain(
        Geometry.ZPoint(0.0),
        Geometry.ZPoint(1000.0);
        boundary_names = (:bottom, :top),
    )

    vertmesh = Meshes.IntervalMesh(vertdomain, nelems = 30)
    verttopo = Topologies.IntervalTopology(
        ClimaComms.SingletonCommsContext(ClimaComms.device()),
        vertmesh,
    )
    vert_center_space = Spaces.CenterFiniteDifferenceSpace(verttopo)

    horzdomain = Domains.SphereDomain(1e6)

    quad = Quadratures.GLL{4}()
    horzmesh = Meshes.EquiangularCubedSphere(horzdomain, 6)
    horztopology = Topologies.Topology2D(context, horzmesh)
    horzspace = Spaces.SpectralElementSpace2D(horztopology, quad)

    hv_center_space =
[...]

PR #2121, opened 2025-01-10
Update dependencies

charleskawczynski:
This PR updates the dependencies.


charleskawczynski:
The only thing that is failing is the multi-gpu distributed remapping , but since #2108 is pre-existing, I think we can merge this PR.

I'm going to merge, but we should probably fix the issue or split out that test and make it a soft-fail until we have time to address it.

PR #2124, opened 2025-01-13
Save the stretch type in the IntervalMesh

charleskawczynski:
This PR is a first step towards #1790. I think I made this backwards compatible in `InputOutput` by checking which properties are available.

Also, this approach is probably nicer in that we save more information and less data in the HDF5 file, which should result in smaller output files.

The next step will be to define and use a `LazyFace` object in place of `faces` in the `IntervalMesh`. This `LazyFace` should be able to determine any given face, without heap reads/writes.

### DIFF
--- a/src/InputOutput/readers.jl
+++ b/src/InputOutput/readers.jl
@@ -30,6 +30,17 @@ import ..Geometry:
     Covariant3Vector
 using ..DataLayouts
 
+function get_key(file, key)
+    if haskey(file, key)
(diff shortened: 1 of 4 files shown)


charleskawczynski:
We justifiably use the faces from the extruded FD space, which requires computing faces from `Geometry.product_geometry` with the horizontal space. I'm not sure that this will be fruitful afterall, unless we want to speed up column or 3D box simulations.

PR #2126, opened 2025-01-13
Fix typo in debugging docs

Sbozzolo:
`result` is not defined.

I also updated the indentation so that the block can be copied and pasted without upsetting the formatter

### DIFF
--- a/docs/src/debugging.md
+++ b/docs/src/debugging.md
@@ -14,20 +14,32 @@ this data after any operation that ClimaCore performs.
 
 ## Example
 
-```@example
-import ClimaCore
-using ClimaCore: DataLayouts
-ClimaCore.DebugOnly.call_post_op_callback() = true
-function ClimaCore.DebugOnly.post_op_callback(result, args...; kwargs...)
-  if any(isnan, parent(data))
-    println("NaNs found!")
-  end
-end
+### Print `NaNs` when they are found
 
-FT = Float64;
-data = DataLayouts.VIJFH{FT}(Array{FT}, zeros; Nv=5, Nij=2, Nh=2)
-@. data = NaN
+In this example, we add a callback that simply prints `NaNs found` every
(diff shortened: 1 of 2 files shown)


docs/src/debugging.md
```diff
@@ -18,10 +18,10 @@ this data after any operation that ClimaCore performs.
 import ClimaCore
 using ClimaCore: DataLayouts
 ClimaCore.DebugOnly.call_post_op_callback() = true
-function ClimaCore.DebugOnly.post_op_callback(result, args...; kwargs...)
-  if any(isnan, parent(data))
-    println("NaNs found!")
-  end
+function ClimaCore.DebugOnly.post_op_callback(data, args...; kwargs...)
+    if any(isnan, parent(data))
```
charleskawczynski, line 22:
```suggestion
    if any(isnan, parent(result))
```

docs/src/debugging.md
```diff
@@ -14,27 +14,61 @@ this data after any operation that ClimaCore performs.
...
+ClimaCore.DebugOnly.call_post_op_callback() = true
+function ClimaCore.DebugOnly.post_op_callback(result, args...; kwargs...)
+    if any(isnan, parent(result))
+        println("NaNs found!")
+    end
+end
+
+FT = Float64
+data = ClimaCore.DataLayouts.VIJFH{FT}(Array{FT}, zeros; Nv=5, Nij=2, Nh=2)
+@. data = NaN
```
charleskawczynski, line 63:
```suggestion
@. data = NaN
ClimaCore.DebugOnly.call_post_op_callback() = false # hide
```
If the docs preserves this session for later jobs (in `@example`), then `call_post_op_callback` will continue returning `true`, and could fail for other examples. (it looks like the docs run with a float, and not an array, which will fail due to no method matching `parent`).

PR #2129, opened 2025-01-14
Make 2-proc cuda distributed remap soft-fail

charleskawczynski:
This PR makes the `distributed remapping with CUDA (2 processes)` job soft-fail for now, since it's failing pretty consistently.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -366,6 +366,7 @@ steps:
       - label: "Unit: distributed remapping with CUDA (2 processes)"
         key: distributed_remapping_gpu_2procs
         command: "srun julia --color=yes --check-bounds=yes --project=.buildkite test/Remapping/distributed_remapping.jl"
+        soft_fail: true
         env:
           CLIMACOMMS_CONTEXT: "MPI"
           CLIMACOMMS_DEVICE: "CUDA"


Sbozzolo:
I worry that this will just hide the problem and issue will not be addressed. 

In the past, this job showed no signs of flakiness, so we must have introduced something new relatively recently. 

Maybe before making this a soft fail, we can try spending two hours to see if we can reproduce the problem reliably. If in two hours of work, we cannot find come up with a reproducer, we can turn this into a soft fail.

charleskawczynski:
> I worry that this will just hide the problem and issue will not be addressed.
>
> In the past, this job showed no signs of flakiness, so we must have introduced something new relatively recently.

I understand, but this has been an issue as far back as December 20, 2024 (#2108). This is why I opened the issue, to make sure that we don't lose track.

> Maybe before making this a soft fail, we can try spending two hours to see if we can reproduce the problem reliably. If in two hours of work, we cannot find come up with a reproducer, we can turn this into a soft fail.

In this build: https://buildkite.com/clima/climacore-ci/builds/4912, I retried the same test multiple times, and it's clearly a race condition since the number of tests that passed were different on different attempts. So, it's not clear to me that we can make a reproducer aside from running the tests multiple times. I suppose we could try putting it inside a loop and call that a reproducer.

I'm happy to take a look over it, my thought was that we could merge this in the mean time, since it kind of clouds the overall CI status.

PR #2131, opened 2025-01-14
add Field sum reproducer

juliasloan25:
In https://github.com/CliMA/ClimaLand.jl/pull/970, I encountered that `sum(f, field) != sum(f, parent(field))`. I'm not sure if this is the expected behavior but it was surprising to me, so I wanted to open a reproducer for it.

@charleskawczynski do you have any insight?

### DIFF
--- a/test/Fields/sum_repro.jl
+++ b/test/Fields/sum_repro.jl
@@ -1,0 +1,24 @@
+import ClimaComms
+ClimaComms.@import_required_backends
+import ClimaCore
+import ClimaCore: Fields
+using Test
+
+include(
+    joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"),
+)
+import .TestUtilities as TU
+
+@testset "Sum function over field" begin
+    FT = Float32
+    context = ClimaComms.context()
+
+    space = TU.SpectralElementSpace2D(FT; context)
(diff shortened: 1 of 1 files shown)


juliasloan25:
The behavior of `sum` is described somewhat in [this thread](https://github.com/CliMA/ClimaCore.jl/issues/943), but it would be good to have more complete documentation of the expected behavior

charleskawczynski:
Yes, I hope I can shine some light on this. Our `sum` by default uses a weighting by the jacobian. From ClimaCore:

```julia
function Base.sum(
    field::Union{Field, Base.Broadcast.Broadcasted{<:FieldStyle}},
    dev::ClimaComms.CUDADevice,
)
    context = ClimaComms.context(axes(field))
    localsum = mapreduce_cuda(identity, +, field, weighting = true)
    ClimaComms.allreduce!(context, parent(localsum), +)
    call_post_op_callback() && post_op_callback(localsum[], field, dev)
    return localsum[]
end
```

Basically, it's an integral

charleskawczynski:
That said, there is likely still inconsistencies with how we handle sum when slicing data, as mentioned in #943.

I think we need to assemble a list of requirements / expectations, and perhaps define `integral` in ClimaCore, and document expected behavior in different situations.

PR #2140, opened 2025-01-17
Define zero on some MatrixField types

charleskawczynski:
This should allow us to use `zero` in ClimaTimesteppers, resulting in safer initialization. [xref](https://github.com/CliMA/ClimaTimeSteppers.jl/pull/346).

Does this need exercised anywhere else? I think this should work since it's recursive. I'm able to run CTS unit tests with this.

### DIFF
--- a/src/MatrixFields/field_matrix_solver.jl
+++ b/src/MatrixFields/field_matrix_solver.jl
@@ -168,6 +168,9 @@ end
 LazySchurComplement(A₁₁, A₁₂, A₂₁, A₂₂) =
     LazySchurComplement(A₁₁, A₁₂, A₂₁, A₂₂, nothing, nothing, nothing, nothing)
 
+Base.zero(lsc::LazySchurComplement) =
+    LazySchurComplement(map(fn -> zero(getfield(lsc, fn)), fieldnames(lsc))...)
+
 NVTX.@annotate function lazy_mul(A₂₂′::LazySchurComplement, x₂)
     (; A₁₁, A₁₂, A₂₁, A₂₂, alg₁, cache₁, A₁₂_x₂, invA₁₁_A₁₂_x₂) = A₂₂′
     zero_rows = setdiff(keys(A₁₂_x₂), matrix_row_keys(keys(A₁₂)))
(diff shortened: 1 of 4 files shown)


src/MatrixFields/field_matrix_solver.jl
```diff
@@ -66,6 +66,9 @@ function FieldMatrixSolver(
     return FieldMatrixSolver(alg, cache)
 end
 
+Base.zero(solver::FieldMatrixSolver) =
+    FieldMatrixSolver(zero(solver.alg), zero(solver.cache))
```
dennisYatunin, line 70: This method is unnecessary. The algorithm is purely symbolic, and the cache does not need to be initialized.

src/MatrixFields/field_matrix_with_solver.jl
```diff
@@ -41,6 +41,9 @@ Base.:(==)(A1::FieldMatrixWithSolver, A2::FieldMatrixWithSolver) =
 Base.similar(A::FieldMatrixWithSolver) =
     FieldMatrixWithSolver(similar(A.matrix), A.solver)
 
+Base.zero(A::FieldMatrixWithSolver) =
+    FieldMatrixWithSolver(zero(A.matrix), zero(A.solver))
```
dennisYatunin, line 45:
```suggestion
    FieldMatrixWithSolver(zero(A.matrix), A.solver)
```

src/MatrixFields/field_name_dict.jl
```diff
@@ -182,6 +182,16 @@ function Base.similar(dict::FieldNameDict)
...
 
+function Base.zero(dict::FieldMatrix)
+    entries = unrolled_map(values(dict)) do entry
+        entry isa UniformScaling ? entry : zero(entry)
+    end
+    return FieldNameDict(keys(dict), entries)
+end
+
+Base.zero(dict::FieldNameDict) =
+    FieldNameDict(keys(dict), map(zero, values(dict)))
```
dennisYatunin, line 193:
```suggestion
function Base.zero(dict::FieldNameDict)
    entries = unrolled_map(values(dict)) do entry
        entry isa UniformScaling ? entry : zero(entry)
    end # TODO: Replace UniformScaling with mutable Ref to make this consistent
    return FieldNameDict(keys(dict), entries)
end
```

charleskawczynski:
Thanks for the suggestions, @dennisYatunin! It looks like these changes worked in ClimaTimesteppers CI 🙂

src/MatrixFields/field_name_dict.jl
charleskawczynski: It looks like the matrix-field method was needed after all: https://buildkite.com/clima/climaatmos-ci/builds/22412#01949b97-4f3c-477b-8ea5-b0a0e49491d7

charleskawczynski: Oh, wait, that build shows that it's not using the latest ClimaCore.

PR #2144, opened 2025-01-20
Update to latest LazyBroadcast

charleskawczynski:


charleskawczynski:
I think the one failure is unrelated to the PR. Well, it could be due to the dependencies, but I'm happy to tackle that in a followup PR. Tracking down the KrylovKit.jl compat issue was more involved than I anticipated.

PR #2147, opened 2025-01-21
Add do-syntax support for HDF5Reader/Writer

### LINKED ISSUE #1480: Allow do-syntax with HDF5Writer
It would be nice if we could use `HDF5Writer` with

```julia
HDF5Writer(...) do writer
...
end
```
and ensure that the file is closed.

charleskawczynski:
Closes #1480.

### DIFF
--- a/src/InputOutput/writers.jl
+++ b/src/InputOutput/writers.jl
@@ -1,7 +1,12 @@
 abstract type AbstractWriter end
 
 """
-    HDF5Writer(filename::AbstractString[, context::ClimaComms.AbstractCommsContext];
+    HDF5Writer(filename::AbstractString[,
+               context::ClimaComms.AbstractCommsContext];
+               overwrite::Bool = true)
+    HDF5Writer(::Function,
+               filename::AbstractString[,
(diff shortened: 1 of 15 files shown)


src/InputOutput/writers.jl
```diff
@@ -42,6 +47,18 @@ struct HDF5Writer{C <: ClimaComms.AbstractCommsContext} <: AbstractWriter
...
+function HDF5Writer(
+    f::Function,
+    filename::AbstractString,
+    context::ClimaComms.AbstractCommsContext;
+    overwrite::Bool = true,
+)
+
+    writer = HDF5Writer(filename, context; overwrite)
+    f(writer)
+    Base.close(writer)
```
Sbozzolo, line 59: I think this should also be enclosed in a `try-finally`, otherwise there is no guarentee that the file will be closed (e.g., if `f` fails)

Sbozzolo (approved):
Thank you!

src/InputOutput/writers.jl
charleskawczynski: Ah, whoops, yes, thank you!

PR #2154, opened 2025-01-23
Improve robustness of distributed remapping

Sbozzolo:
Starting December, the distributed remapping with GPU test started failing often. I could not reproduce this failure on the `clima` node, but I could reproduce a form of it on the Caltech cluster.

Early on, I noticed that it seemed that output array contained correct values and also incorrect values. This, coupled with the erratic nature of the failure, pointed to a problem with MPI. However, identifying the problem required several hours of debugging.

I tried slimming down the test to come up with a minimum broken example. This proved difficult because removing adding code that was seemingly unrelated led to different outcomes and interactive work was impossible. I also went into the rabbit hole of exploring what has changed in the dependencies, but that led to no fruit. Eventually, I landed on a relatively small example that I could run in 90 seconds or so and trigger the problem.

In this, I realized that the first value of the output array was often 0. This was a hint that that value was not being set and the value at initialization was being used. I investigated this deeper and started adding info statements around the function where the reduction is performed.

At some point, I noticed that the order of print statements was off: process 2 would print its output before it was supposed to. With this I mean that if we have a blocking operation like MPI.Reduce, and have

```
@info mypid, "BEFORE"
MPI.Reduce
@info mypid, "AFTER"
```

It has to be that all the `BEFORE` are printed before all the `AFTER`. I noticed that this was not happening and I would see
```
2, BEFORE
2, AFTER
1, BEFORE
1, AFTER
```
Process 2 always gets there first because it has less work to do.

This is not what I was expecting because my mental model for MPI.Reduce is that it implies a barrier. However, if this is what was happening, then it would explain the bug: the processes are using stale data, so what is returned to root is not the correctly processed values.

I tried adding a MPI.Barrier and I could no longer reproduce the problem!

Now, I don't fully understand this. I think that `reduce` should be blocking, but that's not what I observed. I don't know if it is a buggy MPI implementation, or I am missing something. In any case, the correct behavior is to ensure synchronization between the processes before the reduction, so I added a barrier. If this is implied by reduce, adding an extra barrier will only add a very small overhead.

Aside, why is this only happening for CUDA? I think that this might be because of the additional overhead brought it by CUDA that allow two processes to be more out of sync compared to the CPU case.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,330 +40,330 @@ steps:
 
   - wait
 
-  - group: "Unit: Package-wide"
-    steps:
+  # - group: "Unit: Package-wide"
+  #   steps:
 
-      - label: "Unit: aqua"
-        key: unit_aqua
-        command: "julia --color=yes --check-bounds=yes --project=.buildkite test/aqua.jl"
+  #     - label: "Unit: aqua"
+  #       key: unit_aqua
+  #       command: "julia --color=yes --check-bounds=yes --project=.buildkite test/aqua.jl"
 
-      - label: "Unit: cuda"
-        key: unit_cuda_functional
-        command:
-          - "julia --project=.buildkite -e 'using CUDA; CUDA.versioninfo()'"
-          - "julia --color=yes --check-bounds=yes --project=.buildkite test/gpu/cuda.jl CUDA"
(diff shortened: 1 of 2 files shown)


Sbozzolo:
And this didn't work...

PR #2156, opened 2025-01-24
Simplify CUDA remapping

Sbozzolo:
The CUDA kernels for remapping worked in spite of a bug: the kernels used a full 3D launch grid to distribute the work, but only a 1D grid was being launched.

This commit changes three things:
- it removes the additional two dimensions from the kernels. This is the same behavior as in `main`, except for all the dead code being removed
- it reorders some for loops in the kernel so that the outermost loop is the field (in practice, this should not have consequence downstream because we are not using this feature yet)
- it removes a duplicated `env` in a buildkite step

### DIFF
--- a/ext/cuda/remapping_distributed.jl
+++ b/ext/cuda/remapping_distributed.jl
@@ -4,7 +4,6 @@ import CUDA
 using CUDA: @cuda
 import ClimaCore.Remapping: _set_interpolated_values_device!
 
-
 function _set_interpolated_values_device!(
     out::AbstractArray,
     fields::AbstractArray{<:Fields.Field},
@@ -21,7 +20,12 @@ function _set_interpolated_values_device!(
     purely_vertical_space = isnothing(interpolation_matrix)
     num_horizontal_points =
         purely_vertical_space ? 1 : prod(size(local_horiz_indices))
-    num_points = num_horizontal_points * length(vert_interpolation_weights)
+    num_vertical_points = length(vert_interpolation_weights)
+
+    # We parallelize over horizontal points, unless the space is purely vertical
+    # (in which case we parallelize over vertical points)
+    num_points =
+        purely_vertical_space ? num_vertical_points : num_horizontal_points
     max_threads = 256
     nthreads = min(num_points, max_threads)
     nblocks = cld(num_points, nthreads)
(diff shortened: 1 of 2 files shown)


ext/cuda/remapping_distributed.jl
```diff
@@ -67,32 +71,26 @@ function set_interpolated_values_kernel!(
...
-                            )
-                    end
+    for k in 1:num_fields
+        for i in hindex:totalThreadsX:num_horiz
+            h = local_horiz_indices[i]
+            for j in 1:num_vert
+                v_lo, v_hi = vert_bounding_indices[j]
+                A, B = vert_interpolation_weights[j]
+                out[i, j, k] = 0
+                for t in 1:Nq, s in 1:Nq
```
charleskawczynski, line 86: Changing the loop order, here, could have a notable impact on performance. I'm fine with first simplifying the ranges, and then maybe just update the threading pattern / parallelism.

Sbozzolo: I changed the order precisely because I expect this to be faster because retrieving values from different fields is not the most efficient memory access pattern, so I moved the field index to the outermost loop.

In my tests, this does not degrade performance (it improves them very slightly).

But I can remove this change from this PR if you prefer

charleskawczynski: Ok, that makes sense. I still think that the best path forward is to improve the parallelism.

charleskawczynski: I opened #2159, to give an example of how I think this should be refactored, and how we should improve the parallelism.

Sbozzolo: I didn't unify the indices because doing what I did in this PR was an easy fix, whereas thinking about how to do indices properly would have required more work.

Also, it's not immediately clear to me that #2159 will be better because every thread has to read the various arrays that are passed in, whereas this PR reduces some of the reads because some values are reused (e.g., `local_horiz_indices[i]`). The parallelization is over horizontal points, and for a typical output, this is order of 10^4-10^5 points, which should be plenty of threads to expose enough work to the GPU.

But if you want to go ahead and finish #2159, I'd be happy with it.

PR #2159, opened 2025-01-24
Improve threading in `_set_interpolated_values_device`

charleskawczynski:
This is a more parallelized alternative to #2156. cc @Sbozzolo 

Closes #2156

### DIFF
--- a/ext/cuda/remapping_distributed.jl
+++ b/ext/cuda/remapping_distributed.jl
@@ -18,13 +18,11 @@ function _set_interpolated_values_device!(
     # FIXME: Avoid allocation of tuple
     field_values = tuple(map(f -> Fields.field_values(f), fields)...)
 
-    purely_vertical_space = isnothing(interpolation_matrix)
-    num_horizontal_points =
-        purely_vertical_space ? 1 : prod(size(local_horiz_indices))
-    num_points = num_horizontal_points * length(vert_interpolation_weights)
-    max_threads = 256
-    nthreads = min(num_points, max_threads)
-    nblocks = cld(num_points, nthreads)
+    purely_vert_space = isnothing(interpolation_matrix)
+    nitems = length(out)
+
+    # For purely vertical spaces, `Nq` is not used, so we pass in -1 here.
+    _, Nq = purely_vert_space ? (-1, -1) : size(interpolation_matrix[1])
     args = (
         out,
         interpolation_matrix,
@@ -32,12 +30,15 @@ function _set_interpolated_values_device!(
         vert_interpolation_weights,
         vert_bounding_indices,
         field_values,
+        Val(Nq),
     )
+    threads = threads_via_occupancy(set_interpolated_values_kernel!, args)
+    p = linear_partition(nitems, threads)
     auto_launch!(
         set_interpolated_values_kernel!,
         args;
-        threads_s = (nthreads),
-        blocks_s = (nblocks),
+        threads_s = (p.threads),
+        blocks_s = (p.blocks),
     )
     call_post_op_callback() && post_op_callback(
         out,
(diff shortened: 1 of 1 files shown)


charleskawczynski:
It's surprising to me that this is erroring in `_apply_mpi_bitmask!`, I think there are just bugs in `interpolate` logic. Maybe related to #2108?

ext/cuda/remapping_distributed.jl
```diff
@@ -18,26 +18,28 @@ function _set_interpolated_values_device!(
...
-    num_horizontal_points =
-        purely_vertical_space ? 1 : prod(size(local_horiz_indices))
-    num_points = num_horizontal_points * length(vert_interpolation_weights)
-    max_threads = 256
-    nthreads = min(num_points, max_threads)
-    nblocks = cld(num_points, nthreads)
+    num_horiz = length(local_horiz_indices)
+    num_vert = length(vert_bounding_indices)
+    num_fields = length(field_values)
+    nitems = num_horiz * num_vert * num_fields
```
Sbozzolo, line 24: This will fail for purely vertical or purely horizontal spaces

charleskawczynski: Then isn’t the main branch is wrong, because these are extracted from within the kernel in the same way?

charleskawczynski: Rather, how would it fail?

Sbozzolo: The `main` branch has a switch to check if you have a purely vertical space:
```julia
    purely_vertical_space = isnothing(interpolation_matrix)
    num_horizontal_points =
        purely_vertical_space ? 1 : prod(size(local_horiz_indices))
    num_points = num_horizontal_points * length(vert_interpolation_weights)
    max_threads = 256
    nthreads = min(num_points, max_threads)
    nblocks = cld(num_points, nthreads)
```

`local_horiz_indices` is probably not defined on a purely verticla space

charleskawczynski: Yes, but `local_horiz_indices` is still passed to `set_interpolated_values_kernel!`, and `num_horiz = length(local_horiz_indices)` is inside that kernel. I'm very confused.

charleskawczynski: The only way that I could see this not breaking on the main branch is if CUDA adapts `nothing` to the gpu via `Ref`:

```julia
julia> Ref(1)
Base.RefValue{Int64}(1)

julia> length(Ref(1))
1
```

Sbozzolo: There are different kernels for different spaces, the kernel for vertical spaces doesn't use `num_horiz` and the value is `nothing`:

```julia
# GPU, vertical case
function set_interpolated_values_kernel!(
    out::AbstractArray,
    ::Nothing,
    ::Nothing,
    vert_interpolation_weights,
    vert_bounding_indices,
    field_values,
)
    # TODO: Check the memory access pattern. This was not optimized and likely inefficient!
    num_fields = length(field_values)
    num_vert = length(vert_bounding_indices)

    vindex = (blockIdx().y - Int32(1)) * blockDim().y + threadIdx().y
    findex = (blockIdx().z - Int32(1)) * blockDim().z + threadIdx().z

    totalThreadsY = gridDim().y * blockDim().y
    totalThreadsZ = gridDim().z * blockDim().z

    CI = CartesianIndex
    for j in vindex:totalThreadsY:num_vert
        v_lo, v_hi = vert_bounding_indices[j]
        A, B = vert_interpolation_weights[j]
        for k in findex:totalThreadsZ:num_fields
            if j ≤ num_vert && k ≤ num_fields
                out[j, k] = (
                    A * field_values[k][CI(1, 1, 1, v_lo, 1)] +
                    B * field_values[k][CI(1, 1, 1, v_hi, 1)]
                )
            end
        end
    end
    return nothing
end
```

charleskawczynski: But then we'd get many wrong values

charleskawczynski: I see, the same `_set_interpolated_values_device!` is calling different `set_interpolated_values_kernel!`s

ext/cuda/remapping_distributed.jl
```diff
@@ -60,41 +61,36 @@ function set_interpolated_values_kernel!(
...
+            (CUDA.blockIdx().x - Int32(1)) * CUDA.blockDim().x +
+            CUDA.threadIdx().x
+        inds = (num_vert, num_horiz, num_fields)
+
+        1 ≤ i_thread ≤ prod(inds) || return nothing
 
-    totalThreadsX = gridDim().x * blockDim().x
-    totalThreadsY = gridDim().y * blockDim().y
-    totalThreadsZ = gridDim().z * blockDim().z
+        (j, i, k) = CartesianIndices(map(x -> Base.OneTo(x), inds))[i_thread].I
```
Sbozzolo, line 78: Shouldn't the order be `(k, j, i)`?

Sbozzolo: Similar with the other ones

charleskawczynski: I think this is correct. The indices should match up with the ranges given to CartesianIndices

Sbozzolo: I think I am still confused by the order: Is the first or the last the fastest index?

charleskawczynski: The first index, returned by `CartesianIndices`, changes the fastest.

permuting the order will change the memory access pattern (which can impact vectorization on the cpu and coalescence on the gpu), but it will not change the correctness.

charleskawczynski: We could follow up with a benchmark if you’d like. For now, I matched the order with fast memory access to `out`

charleskawczynski: Oh, wait, I didn’t. Not sure why I did that order, we can try different orders, but it should probably be performance-driven

charleskawczynski: I probably did that order with the VIJFH datalayouts in mind

Sbozzolo: Okay, maybe we can leave the comment about double checking memory access patterns because this might not be the most effective permutation of indices.

PR #2162, opened 2025-01-27
Add support for HDF5 read and write with PointSpaces

### LINKED ISSUE #2171: InputOutput does not support PointSpaces
## Describe the bug

The `write(::HDF5Writer, ::Fields.Field, ::AbstractString)` method  does not work for Fields with a PointSpace because
the method expects the input field's space to have a grid and topology. Similarly, `read_field` cannot recreate the space of a Field with a PointSpace

## To Reproduce

```Julia
using Test
import ClimaCore

using ClimaComms
const comms_ctx = ClimaComms.context(ClimaComms.CPUSingleThreaded())

@testset "HDF5 write/read test for 0d PointSpace" begin
    FT = Float32

    # instead of directly constructing a PointSpace, we construct Field with a ColumnSpace,
    # and call the level function to get a Field with a PointSpace
    column_space = ClimaCore.CommonSpaces.ColumnSpace(;
        z_min = FT(0),
        z_max = FT(100),
        z_elem = 10,
        staggering = ClimaCore.Grids.CellCenter(),
    )

    field_1d = ClimaCore.Fields.local_geometry_field(column_space)
    field_0d = ClimaCore.Fields.level(field_1d, 5)

    filename = tempname()

    writer = ClimaCore.InputOutput.HDF5Writer(filename, comms_ctx)
    ClimaCore.InputOutput.write!(writer, field_0d, "field_0d")
    close(writer)

    reader = ClimaCore.InputOutput.HDF5Reader(filename, comms_ctx)
    restart_field_0d = ClimaCore.InputOutput.read_field(reader, "field_0d")
    close(reader)
    @test restart_field_0d == field_0d
end
```

## Related issues / PRs

Needed to close : [PR](https://github.com/CliMA/ClimaAtmos.jl/pull/3398)

imreddyTeja:
`write!` on Fields is now dispatched off the input
field's space. When it is a point space, the local
geometry data is written. When read_field is called,
and no grid is present, then the field is assumed
to be on a point space.

When the reader recreates a field from a HDF5, it called
eval on a parsed input string. This replaces that with a function
that ensures there are no function calls inside the parsed string
before eval'ing it. This is based on JLD.jl code.

Closes #2171 

- [] Code is exercised in an integration test OR N/A.


Sbozzolo:
If we can write, we should also be able to read, so maybe there's some other changes needed here

src/InputOutput/readers.jl
```diff
@@ -513,11 +515,24 @@ function read_field(reader::HDF5Reader, name::AbstractString)
...
+        elseif haskey(attrs(obj), "space")
             space = read_space(reader, attrs(obj)["space"])
+        else
+            # if the there is no grid, then the field is on a PointSpace
+            lg_obj = reader.file["local_geometry_data/$name"]
+            # TODO: Is this array type correct? should we assume SingletonCommsContext?
+            ArrayType = ClimaComms.array_type(ClimaComms.device(reader.context))
+            lg_data = ArrayType(read(lg_obj))
+            # because it is a point space, the data layout of local_geometry_data is always DataF
+            lg_type = eval(Meta.parse(attrs(lg_obj)["value_type"]))
```
Sbozzolo, line 527: I think that this allows for arbitrary code execution just by adding the desired code into an attribute of the HDF5 file, making it a large security vulnerability. A malicious actor could modify a restart file in ClimaAtmos to change the `value_type` of one of its fields to silently execute arbitrary code when a user restarts the simulation with that file. While this is unlikely to be a concern right now, I think we should avoid using this pattern here and address the problem in a future PR.

When we parse something that could come from users, we should always sanitize it and check it against the allowed inputs.

src/InputOutput/writers.jl
```diff
@@ -451,16 +451,88 @@ function write_new!(
...
+
+"""
+    write!(
+        writer::HDF5Writer,
+        field::Fields.Field,
+        name::AbstractString,
+        space::Spaces.AbstractPointSpace,
+    )
+
+Writes a `Field`, with `axes` of type `PointSpace`,  to the HDF5 file. The field
```
Sbozzolo, line 472:
```suggestion
Write a `Field`, with `axes` of type `PointSpace`,  to the HDF5 file. The field
```

src/InputOutput/writers.jl
```diff
@@ -451,16 +451,88 @@ function write_new!(
...
+
+"""
+    write!(
+        writer::HDF5Writer,
+        field::Fields.Field,
+        name::AbstractString,
+        space::Spaces.AbstractSpace,
+    )
+
+Writes an object of type 'Field' and name 'name' to the HDF5 file.
```
Sbozzolo, line 519:
```suggestion
Write an object of type 'Field' and name 'name' to the HDF5 file.
```

test/InputOutput/unit_point.jl
```diff
@@ -0,0 +1,43 @@
+using Test
+import ClimaCore
+
+using ClimaComms
+const comms_ctx = ClimaComms.context(ClimaComms.CPUSingleThreaded())
+
+
+@testset "HDF5 write/read test for 0d PointSpace" begin
+    # need to define equality because with no Grid to cache, the read space will not be "egal"
```
Sbozzolo, line 9:
```suggestion
    # need to define equality because with no Grid to cache, the read space will not be "equal"
```

src/InputOutput/writers.jl
```diff
@@ -451,16 +451,88 @@ function write_new!(
 end
 
 # write fields
+"""
+    write!(writer::HDF5Writer, field::Fields.Field, name::AbstractString)
+
+Calls `write!` with the `axes` of `field` as an extra positional argument, which
+writes an object of type 'Field' and name 'name' to the HDF5 file.
```
Sbozzolo, line 458: This docstring describes the implementation detail of the function (calling another method), but I don't think that they are imortant here.

```julia
Write the `field` to the HDF5 in `writer` and assign it the given `name`.
```

test/InputOutput/unit_point.jl
```diff
@@ -0,0 +1,43 @@
...
+    # to the written space.
+    function Base.:(==)(
+        lg1::ClimaCore.Geometry.LocalGeometry{I, C, FT, <:Any},
+        lg2::ClimaCore.Geometry.LocalGeometry{I, C, FT, <:Any},
+    ) where {I, C, FT}
+        for f in [:coordinates, :J, :WJ, :invJ, :∂x∂ξ, :∂ξ∂x, :gⁱʲ, :gᵢⱼ]
```
Sbozzolo, line 15:
```suggestion
        for f in propertynames(lg1)
```

Sbozzolo: This function could also be rewritten as
```julia
return all(getproperty(lg1, p) == getproperty(lg2, p) for p in propertynames(lg1))
```

Sbozzolo (commented):
Could you also add an item to the NEWS?

src/InputOutput/readers.jl
```diff
@@ -41,6 +41,31 @@ function get_key(file, key)
     end
 end
 
+"""
+    parse_type(ts::AbstractString)
+
+Parse a string 'ts' into an expression, and then evaluate it into a type if it is a valid type expression.
+"""
+function parse_type(ts::AbstractString)
+    # based on the old HDF5.jl `julia_type` function
```
charleskawczynski, line 50: Can you please add a comment pointing to where you saw this?

src/InputOutput/readers.jl
```diff
@@ -513,11 +540,24 @@ function read_field(reader::HDF5Reader, name::AbstractString)
                 staggering = Grids.CellFace()
             end
             space = Spaces.space(grid, staggering)
-        else
+        elseif haskey(attrs(obj), "space")
             space = read_space(reader, attrs(obj)["space"])
+        else
+            # if the there is no grid, then the field is on a PointSpace
+            lg_obj = reader.file["local_geometry_data/$name"]
+            # TODO: Is this array type correct? should we assume SingletonCommsContext?
```
charleskawczynski, line 548: Yeah, I think it's fine to assume singleton for the point spaces.

src/InputOutput/writers.jl
```diff
@@ -451,16 +451,87 @@ function write_new!(
...
+Write an object of type 'Field' and name 'name' to the HDF5 file.
+"""
+function write!(
+    writer::HDF5Writer,
+    field::Fields.Field,
+    name::AbstractString,
+    space::Spaces.AbstractSpace,
+)
     values = Fields.field_values(field)
-    space = axes(field)
```
Sbozzolo, line 456: Why is the `space` needed as as argument? Is it no longer the case that `space = axes(field)`?

imreddyTeja: That is still the case. The `space` argument is used to dispatch. If the space is a PointSpace, it needs to be dealt with differently. `write(::HDF5Writer, ::Field, ::String)` still works, but it calls this method. I can remove `space` as an argument and have two branches within one method, if you think that is better.

Sbozzolo: No, that's good, thank you!

src/InputOutput/readers.jl
```diff
@@ -41,6 +41,42 @@ function get_key(file, key)
...
+expressions, with the expection of @NamedTuple.
+This function is based on the JLD.jl `is_valid_type_exp`.
+See https://github.com/JuliaIO/JLD.jl/blob/80ac89643e3ad87545e48f4d361a00a29cdf4e2f/src/JLD.jl#L922
+"""
+is_type_expr(s::Symbol) = true
+is_type_expr(q::QuoteNode) = is_type_expr(q.value)
+function is_type_expr(expr::Expr)
+    if expr.head == :macrocall && expr.args[1] == Symbol("@NamedTuple")
+        # skip the LineNumberNode, as eval does nothing for it
+        if length(expr.args) > 2
```
imreddyTeja, line 70: "All macros receive an extra argument __source__::LineNumberNode which describes the parser location in the source file for the @ of the macro call. It can be accessed as a normal argument variable in the body of the macro. This is implemented by inserting an extra leading argument into the Expr(:macrocall, :@name, LineNumberNode(...), args...) surface syntax." from [here](https://github.com/nextjournal/julia-lang/blob/d00deb7ce962ae76f99e93b4ce97f6a1c4e27f18/NEWS.md?plain=1#L99-L104)

and [here](https://github.com/JuliaLang/julia/blob/c8cd8fdb832b008221fc2cab31b002e7555e607a/src/julia-parser.scm#L2586-L2599)

For example:

```
julia> dump(Meta.parse("@NamedTuple{foo::bar}"))
Expr
  head: Symbol macrocall
  args: Array{Any}((3,))
    1: Symbol @NamedTuple
    2: LineNumberNode
      line: Int64 1
      file: Symbol none
    3: Expr
      head: Symbol braces
      args: Array{Any}((1,))
        1: Expr
          head: Symbol ::
          args: Array{Any}((2,))
            1: Symbol foo
            2: Symbol bar
```
I have not found documentation or proof that a `LineNumberNode` will always `eval` to `nothing`, but it is mentioned [here](https://github.com/JuliaLang/julia/issues/38439). To be extra safe, this could also filter and drop any `LineNumberNodes` in the expression.

Sbozzolo (approved):

src/InputOutput/readers.jl
```diff
@@ -535,14 +585,16 @@ function read_field(reader::HDF5Reader, name::AbstractString)
         # For when `Nh` is added back to the type space
         #     Nhd = Nh_dim(data_layout)
         #     Nht = Nhd == -1 ? () : (size(data, Nhd),)
-        ElType = eval(Meta.parse(attrs(obj)["value_type"]))
+        ElType = read_type(attrs(obj)["value_type"])
```
charleskawczynski, line 588:
```suggestion
        ElType = read_type(attrs(obj)["field_eltype_type"])
```
and can we move this to right after `elseif data_layout in ("DataF",)`?

test/InputOutput/unit_point.jl
```diff
@@ -0,0 +1,36 @@
+using Test
+import ClimaCore
+
+using ClimaComms
+const comms_ctx = ClimaComms.context(ClimaComms.CPUSingleThreaded())
```
charleskawczynski, line 5:
```suggestion
const comms_ctx = ClimaComms.context()
```
And let's add a cuda job for this, if it's not already on buildkite

test/InputOutput/unit_point.jl
```diff
@@ -0,0 +1,36 @@
+using Test
+import ClimaCore
+
```
charleskawczynski, line 3:
```suggestion
import ClimaCore: Spaces, Fields, InputOutput
```
Then use qualified modules below

test/InputOutput/unit_read_value_type.jl
```diff
@@ -0,0 +1,54 @@
+using Test
+import ClimaCore
+
```
charleskawczynski, line 3:
```suggestion
compare_read_type(x) = InputOutput.read_type(string(eltype(x))) == eltype(x)
```
Let's define a helper above and then use `@test compare_read_type(ClimaCore.Fields.local_geometry_field(space))`

charleskawczynski (approved):
I think this looks good. The biggest risk I think is with `is_type_expr`, we can expand a bit on the tests, but that function can see nearly anything. So, maybe we can run atmos CI with this first?

If it passes, then I think we should merge, as this is currently safer than what we have.

src/InputOutput/readers.jl
imreddyTeja: This change makes this PR a breaking change. Old hdf5 files containing `fields` will no longer be able to be read unless the  "value_type" attribute is changed to "field_eltype".

charleskawczynski: Ah, I see. I guess it's fine to leave it for now. We could rename it and preserve backward compatibility, but I don't want to hold up this work.

PR #2163, opened 2025-01-27
Define `to_device` and `to_cpu`

charleskawczynski:
This PR adds `to_device`, which allow users to move data between the CPU and GPU, and `to_cpu`, which allow users to move data to the CPU.

### DIFF
--- a/src/to_device.jl
+++ b/src/to_device.jl
@@ -1,0 +1,60 @@
+import Adapt
+import ClimaComms
+
+"""
+    out = to_device(device, x::Union{
+        DataLayouts.AbstractData,
+        Spaces.AbstractSpace,
+        Fields.Field,
+        Fields.FieldVector,
+    })
+
+Move `x` to the given `device`.
+
+This is particularly useful to move different types of `Space.AbstractSpace`s,
+`Fields.Field`s, and `Fields.FieldVector`s from CPUs to GPUs and viceversa.
+
+If the input is already defined on the target device, returns a copy.
+
+This means that `out === x` will not in general be satisfied.
+"""
+function to_device(
+    device::ClimaComms.AbstractDevice,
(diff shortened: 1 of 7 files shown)


src/to_device.jl
```diff
@@ -0,0 +1,60 @@
...
+
+"""
+    out = to_device(device, x::Union{
+        DataLayouts.DataLayout,
+        Spaces.AbstractSpace,
+        Fields.Field,
+        Fields.FieldVector,
+    })
+
+Move `x` backing data to the given `device`.
```
Sbozzolo, line 12:
```suggestion
Move `x` to the given `device`.
```
sounds simpler to me.

src/to_device.jl
```diff
@@ -0,0 +1,60 @@
+import Adapt
+import ClimaComms
+
+"""
+    out = to_device(device, x::Union{
```
Sbozzolo, line 5:
```suggestion
    out = to_device(device::ClimaComms.AbstractDevice, x::Union{
```

Sbozzolo: I don't think I've ever seen docstrings with a named return value

Sbozzolo (commented):
Could you also add a page to the documentation describing this feature?

In the future, this might be mentioned in the page for `Field`s and `Space`s. For now, we could just start a "Frequently Asked Questions" page and describe this feature

src/to_device.jl
charleskawczynski: I've occasionally seen them. I feel that it can sometimes help explain them in the remaining docstring.

Sbozzolo: Yes, I agree.

Sbozzolo (approved):
Awesome, thank you!

PR #2167, opened 2025-01-30
Fix vertical metric terms, add SEM option for horizontal metric terms

dennisYatunin:
This PR makes our vertical metric terms more consistent, and it adds a new option for our horizontal metric terms.

### Content
- Change the vertical center coordinate of element `n` from `(z(n, 0) + z(n, 1)) / 2` to `z(n, 1/2)`, where `ξ³ = 1/2` represents the center in generalized (unstretched) coordinates, and modify the vertical metric Jacobian to reflect this change. This makes the stretching of center coordinates consistent with the stretching of face coordinates.
- Add an `autodiff_metric` option for `SpectralElementGrid2D` in order to distinguish between metric terms computed with automatic differentiation and metric terms computed with the SEM. The default setting of `autodiff_metric = true` matches what we had previously, while `autodiff_metric = false` is the setting recommended by Paul Ullrich. Both of these exhibit similar performance in ClimaCore unit tests, so we will determine which setting is optimal through testing in ClimaAtmos.


charleskawczynski:
@dennisYatunin mentioned to me offline that this was already tested in Atmos CI + the longruns. Can you please post builds here?

PR #2168, opened 2025-01-30
Remove some unused dependencies

charleskawczynski:
Found with

```julia
function julia_files(root)
	all_files = String[]
  for (root, dirs, files) in walkdir(root)
      for file in files
          endswith(file, ".jl") && push!(all_files, joinpath(root, file))
      end
  end
  return all_files
end

function find_unecessary_deps(root, deps)
  found = Dict{String,Bool}()
  for dep in deps
    found[dep] = false
  end
  jl_files = julia_files(root)
  N = length(jl_files)
  @show N
  for (i, file) in enumerate(jl_files)
    percent = i / N
  	jl_contents = readlines(file)
  	for dep in deps
  		found[dep] = found[dep] || occursin(dep, join(jl_contents, "\n"))
  	end
  end
  unfound = String[]
  for k in keys(found)
  	if !found[k]
  		push!(unfound, k)
  	end
  end
  return unfound
end

contents = readlines(".buildkite/Project.toml");
_deps = deleteat!(map(x->first(split(x, " = ")), contents), 1);
compat_start = findfirst(x->x=="", _deps)
if !isnothing(compat_start)
  _deps = _deps[1:compat_start-1]
end
@show length(_deps)
ud = find_unecessary_deps(".", _deps)
isempty(ud) || (using Pkg; Pkg.rm(ud))
```


charleskawczynski:
Maybe this script would be a nice addition to PkgDevTools?

Sbozzolo:
> Maybe this script would be a nice addition to PkgDevTools?

Yes, I think it'd be useful! I'd suggest you also describe how the script works because it is not immediately clear how unused dependencies are defined. In this case, unused is a package that is imported but never used, but a package that is never imported or mentioned in source codes.

Sbozzolo (approved):

.buildkite/Project.toml
```diff
@@ -55,7 +53,6 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 TerminalLoggers = "5d786b92-1e48-4d6f-9151-6b4477ca9bed"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 ThreadsX = "ac1d9e8a-700a-412c-b207-f0111f4b6c0d"
-URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
 
 [compat]
 SnoopCompileCore = "2"
```
Sbozzolo, line 58:
```suggestion
SnoopCompileCore = "2, 3"
```

Sbozzolo: Wasn't this needed?

Sbozzolo:
Also, I think you are missing the update to the Manifest-v1.11.toml

charleskawczynski:
> Also, I think you are missing the update to the Manifest-v1.11.toml

We're not using 1.11

Sbozzolo:
> > Also, I think you are missing the update to the Manifest-v1.11.toml
> 
> We're not using 1.11

Ah, right!

.buildkite/Project.toml
charleskawczynski: Not sure, I don't remember changing that

Sbozzolo: It was needed for julia 1.11

Sbozzolo: I forgot we haven't merged that yet

PR #2169, opened 2025-01-30
Fix distributed remapping bug

### LINKED ISSUE #2108: Distributed remapping interpolation bug

Sbozzolo:
#2108 partially describes the odissey I went through to understand and fix this bug.

After discussing this with folks at MPI.jl (https://github.com/JuliaParallel/MPI.jl/issues/892), a potential problem was identified:
```
remapper._interpolated_values[remapper.colons..., begin]
```
in the `ClimaComms.reduce!` call allocates a new copy. This can mess up CUDA's synchronization, so that the correct data is not sent over. The simple solution is to use `view` instead of slices.

In addition to this, this PR changes `interpolate` to use `interpoalte!`, which leads much simpler code (the reason it was not done this way originally is because `interpolate!` is a more recent addition and behaved slightly differently).

Closes #2108 
Closes #2132

### DIFF
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -751,12 +751,10 @@ function _collect_interpolated_values!(
     index_field_end::Int;
     only_one_field,
 )
-    # NOTE: MPI barriers for #2108
-    ClimaComms.barrier(remapper.comms_ctx)
     if only_one_field
         ClimaComms.reduce!(
             remapper.comms_ctx,
-            remapper._interpolated_values[remapper.colons..., begin],
+            view(remapper._interpolated_values, remapper.colons..., 1),
             dest,
             +,
         )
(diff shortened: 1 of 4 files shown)


Sbozzolo:
It looks like that my earlier [tests](https://buildkite.com/clima/climacore-ci/builds/5086) passed just by luck :(

src/Remapping/distributed_remapping.jl
```diff
@@ -777,38 +754,26 @@ function _collect_interpolated_values!(
     if only_one_field
         ClimaComms.reduce!(
```
vchuravy, line 755: Where is the synchronization of the CUDA stream?

https://github.com/CliMA/ClimaComms.jl/blob/c968350313d5dc81d007721bc27a4bc9c97a79e3/ext/ClimaCommsMPIExt.jl#L69

Issue #1989, opened 2024-09-14
gradient on combination of horizontal and 3d fields does not work


haakon-e:
## Describe the bug
The following code produces an error:
```julia
ᶜgradᵥ = Operators.GradientF2C()

level_field = Fields.level(Fields.Field(Float64, center_space), 1)
ᶠscalar_field = Fields.Field(Float64, face_space)
# Does not work:
@. ᶜgradᵥ(level_field + ᶠscalar_field)
```
<details> <summary> produces the following error </summary>

```julia
ERROR: InvalidIRError: compiling MethodInstance for ClimaCoreCUDAExt.copyto_stencil_kernel!(::ClimaCore.Fields.Field{…}, ::ClimaCore.Operators.StencilBroadcasted{…}, ::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{…}, ::NTuple{…}, ::ClimaCore.DataLayouts.UniversalSize{…}) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to level)
Stacktrace:
 [1] copyto_stencil_kernel!
   @ ~/.julia/packages/ClimaCore/wHC4M/ext/cuda/operators_finite_difference.jl:0
Hint: catch this exception as `err` and call `code_typed(err; interactive = true)` to introspect the erronous code with Cthulhu.jl
Stacktrace:
  [1] check_ir(job::GPUCompiler.CompilerJob{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams}, args::LLVM.Module)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Y4hSX/src/validation.jl:147
  [2] macro expansion
    @ ~/.julia/packages/GPUCompiler/Y4hSX/src/driver.jl:458 [inlined]
  [3] macro expansion
    @ ~/.julia/packages/TimerOutputs/Lw5SP/src/TimerOutput.jl:253 [inlined]
  [4] macro expansion
    @ ~/.julia/packages/GPUCompiler/Y4hSX/src/driver.jl:457 [inlined]
  [5] emit_llvm(job::GPUCompiler.CompilerJob; libraries::Bool, toplevel::Bool, optimize::Bool, cleanup::Bool, only_entry::Bool, validate::Bool)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Y4hSX/src/utils.jl:103
  [6] emit_llvm
[...]

PR #2172, opened 2025-02-03
Add GPU support for extruded 1D spaces

dennisYatunin:
This PR adds GPU support for extruded 1D spaces, with the following modifications:
- Remove the assumption that `Ni == Nj` from `mapreduce_cuda_kernel`, `columnwise_partition`, `columnwise_universal_index`, and `multiple_field_solve_universal_index`
- Add methods for `operator_shmem` `operator_fill_shmem!`, and `operator_evaluate` corresponding to the 1D versions of all SEM operators
- Rename `DSSDataTypes` to `DSSTypes2D`, and define a new constant `DSSTypes1D`
- Add a method for `dss_1d!` corresponding to `CUDADevice`, and extend `dss!` to support both 1D and 2D data types
- Adapt the `local_geometry` and `dss_weights` of a `SpectralElementSpace1D` from `Array`s to `CuArray`s

These changes are also tested through CI:
- Add a CUDA job that runs `unit_2d.jl` and `convergence_2d.jl`
- Add tests for `IFH`, `IHF`, `VIFH`, and `VIHF`  to `unit_mapreduce.jl`, which already has a CUDA job

To fix several test failues, I had to make a few additional changes:
- Simplify the CUDA implementation of `Gradient` by directly differentiating vectors instead of breaking them up into components, as the original design could not distinguish the gradient of a scalar from the gradient of a single-component vector, and one of the tests in `unit_2d.jl` takes the gradient of a `UVector` field
- Extend `strip_space` to `FiniteDifferenceOperator` boundary conditions, as one of the tests in `unit_2d.jl` uses a 1D field for the `CurlC2F` boundary condition
- Reduce the number of methods for `level(space::ExtrudedFiniteDifferenceSpace, v)` from 4 to 1 to avoid a GPU inference failure in `reconstruct_placeholder_space` when evaluating boundary conditions (we hit some sort of compiler heuristic when we use 2 or 4 specialized methods, so we need to just have one method with union-splitting)
  - By making this change, I have inadvertently fixed the issue in #1989
- Adapt several `CuArray`s to `Array`s in `unit_2d.jl` and `unit_spaces.jl` to avoid scalar indexing errors
- Bump the number of allowed JET failures for `SpectralElementSpace1D` in `opt_spaces.jl` from 141 to 825 on GPUs and from 137 to 150 on CPUs

---


ext/cuda/remapping_distributed.jl
```diff
@@ -130,9 +130,7 @@ function set_interpolated_values_kernel!(
         out[i, j, k] = 0
         for t in 1:Nq
             out[i, j, k] +=
-                I[i, t] *
-                I[i, s] *
-                (
+                I[i, t] * (
```
charleskawczynski, line 133: Was this a bug? Why aren't our unit tests catching this?

ext/cuda/remapping_distributed.jl
```diff
@@ -275,9 +273,9 @@ function set_interpolated_values_kernel!(
 
         h = local_horiz_indices[i]
         out[i, k] = 0
-        for t in 1:Nq, s in 1:Nq
+        for t in 1:Nq
```
charleskawczynski, line 276: Again, is this a bug?

dennisYatunin: Yes, this is also an untested bug.

charleskawczynski (commented):
> Bump the number of allowed JET failures for SpectralElementSpace1D in opt_spaces.jl from 141 to 825 on GPUs and from 137 to 150 on CPUs

This is alarming, without manifest changes. Would you mind taking a look at what is causing the spike in inferences? (I think that script makes it pretty easy to modify & see)

ext/cuda/remapping_distributed.jl
charleskawczynski: Ok, good to know. Can we add a unit test in ClimaCore to exercise this?

ext/cuda/remapping_distributed.jl
charleskawczynski: Also, cc @Sbozzolo

src/Hypsography/Hypsography.jl
```diff
@@ -77,10 +77,12 @@ end
 Locate the levels by linear interpolation between the surface field and the top
 of the domain, using the method of [GalChen1975](@cite).
 """
-struct LinearAdaption{F <: Fields.Field} <: HypsographyAdaption
+struct LinearAdaption{F} <: HypsographyAdaption
     surface::F
 end
 
+strip_field(::LinearAdaption) = LinearAdaption(nothing)
```
charleskawczynski, line 84: Shouldn't this be `strip_space`? And it should keep the field values

ext/cuda/remapping_distributed.jl
dennisYatunin: Sure, that's easy to add. I only discovered this bug last night so I'll add some tests today.

src/Hypsography/Hypsography.jl
dennisYatunin: There is no need to keep the field values in this function because they are passed in separately through `ref_z_to_physical_z`. I think it is best to avoid passing duplicate field values through here. In general, the design of this section should probably be refactored to split the surface elevations from the adaption specification.

src/Topologies/dss.jl
```diff
@@ -684,7 +677,14 @@ end
 
 Computed unweighted/pure DSS of `data`.
 """
-function dss!(data::DSSDataTypes, topology::Topology2D)
+function dss!(data::DSSTypes1D, topology::IntervalTopology)
+    Nij = DataLayouts.get_Nij(data)
+    length(parent(data)) == 0 && return nothing
```
charleskawczynski, line 682: I think there might be more cases here that we can short circuit. What do you think?

test/Operators/unit_operators_examples.jl
```diff
@@ -57,21 +57,12 @@ grid = Grids.ExtrudedFiniteDifferenceGrid(
...
-        @test_broken begin
-            @. ᶜgradᵥ(level_field + ᶠscalar_field)
-        end
-    else
-        @. ᶜgradᵥ(level_field + ᶠscalar_field)
-    end
-end
+using Test
+@testset "Example broadcast expression on GPUs" begin
+    @test !isnothing(@. ᶜgradᵥ(level_field + ᶠscalar_field))
```
charleskawczynski, line 67: 🎉, thank you!

test/Spaces/opt_spaces.jl
```diff
@@ -35,7 +35,7 @@ end
 #! format: off
     if ClimaComms.device(context) isa ClimaComms.CUDADevice
         test_n_failures(91,   TU.PointSpace, context)
-        test_n_failures(141,  TU.SpectralElementSpace1D, context)
+        test_n_failures(825,  TU.SpectralElementSpace1D, context)
```
charleskawczynski, line 38: Can you please look and see what this spike is from?

charleskawczynski: We might be able to use a function barrier to substantially reduce it

dennisYatunin: This spike is caused by the conversion of `Array`s to `CuArray`s, and it is similar to the spike in JET failures from CPUs to GPUs for `SpectralElementSpace2D`.

dennisYatunin: If we need to reduce this for `SpectralElementSpace1D`, we should probably apply the same fix to `SpectralElementSpace2D`

src/Topologies/dss.jl
dennisYatunin: You mean like whether we are using GL or GLL quadrature? I think we're just assuming GLL quadrature everywhere outside of a few unit tests. Or did you have something else in mind?

charleskawczynski:
> This is alarming, without manifest changes. Would you mind taking a look at what is causing the spike in inferences? (I think that script makes it pretty easy to modify & see)

It turns out that these are all false warnings, induced by our `get!` caching. Commenting out `get!` in spectral element grids and interval meshes shows no inference failures.

test/Spaces/opt_spaces.jl
charleskawczynski: It turns out that these are all false warnings, induced by our `get!` caching. Commenting out `get!` in spectral element grids and interval meshes shows no inference failures.

src/Topologies/dss.jl
charleskawczynski: Just checking `Nh > 1` should be good (+GLL?)

ext/cuda/remapping_distributed.jl
Sbozzolo: There are unit tests for 1D extruded spaces, but they were skipped on GPU. There are https://github.com/CliMA/ClimaCore.jl/blob/a1c6470b8ea8f32143be190feea56e91749e9128/test/Remapping/distributed_remapping.jl#L46-L147.

ext/cuda/remapping_distributed.jl
Sbozzolo: This should been tested here: https://github.com/CliMA/ClimaCore.jl/blob/a1c6470b8ea8f32143be190feea56e91749e9128/test/Remapping/distributed_remapping.jl#L666-L760

charleskawczynski (approved):
I chatted with @dennisYatunin about this PR. These changes look fine, and I approve once we add some unit tests exercising the new pieces/features.

Sbozzolo:
In https://github.com/CliMA/ClimaCore.jl/pull/2176, I am adding tests for GPU for the CommonSpaces module. I excluded the `SliceXZ` case from those tests, so you should be able to turn it on (it if #2176 gets merged before this).

ext/cuda/remapping_distributed.jl
dennisYatunin: This test has been reenabled on GPUs and appears to be passing now.

ext/cuda/remapping_distributed.jl
dennisYatunin: I think you meant here:
https://github.com/CliMA/ClimaCore.jl/blob/a1c6470b8ea8f32143be190feea56e91749e9128/test/Remapping/distributed_remapping.jl#L866-L897

This test has been reenabled on GPUs and appears to be passing now.

src/Hypsography/Hypsography.jl
dennisYatunin: I have redesigned this to avoid passing the field values separately, instead using a lazy broadcast to pass them through the adaption. Keeping everything in the adaption seems more in line with the original intention of this data structure, and the new pattern is also more generalizable.

src/Topologies/dss.jl
dennisYatunin: We cannot optimize out DSS in a horizontally periodic space when `Nh == 1`, since the opposite sides of the single horizontal element are still collocated. We can only optimize it out when we are using GL instead of GLL, or when there is no data to DSS. To check the first condition, I have defined the function `Quadratures.requires_dss(quadrature_style(space))`. To check the second condition, I am now using `sizeof(eltype(data))`, which is somewhat clearer than the `length(parent(data))` that was originally here.

PR #2179, opened 2025-02-11
Make buildkite env reproducible

### LINKED ISSUE #2178: Environment is not fully reproducible
We have an issue where our current buildkite environment is:

```yml
- echo "--- Instantiate buildkite env"
- "julia --project=.buildkite --check-bounds=yes -e 'using Pkg; Pkg.develop(path=\".\"); Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true)'"
- "julia --project=.buildkite -e 'using CUDA; CUDA.precompile_runtime()'"
```

The issue here is that `Pkg.develop` can (may?/will?) trigger the resolver (at least we just observed this after the depot was cleared), resulting in potentially different versions being run despite our manifest file being checked in.

Over the years, I've seen `Pkg.develop` added and removed from pipelines, likely due to this issue. One reason why `Pkg.develop` is typically added is because, without it, users may update the manifest locally, and forget to `Pkg.develop` themselves, effectively pinning CI to that version.

I think a better pattern to use moving forwards is to:

 - Remove the use of `Pkg.develop` on buildkite, to guarantee reproducibility
 - Document how users should setup their local environment
 - Add a test (whether that's via julia code or in the buildkite pipeline), that tests that the repo package (ClimaCore in this case), is developed in that environment. In Julia terms, that is testing that `TOML.parse(join(readlines(".buildkite/Manifest.toml")[7:end], "\n"))["deps"]["ClimaCore"][1]["path"] == ".."` (this check unfortunately may be julia version-dependent).

We should probably spread this pattern, as this can easily bite users on all of our repos.

charleskawczynski:
This is an attempt to close #2178. I don't know that this is the best way to do this, but it should work.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,15 +21,18 @@ steps:
           rm -rf ${JULIA_DEPOT_PATH}
         fi
 
+      - echo "--- Check that ClimaCore is developed in Manifest"
+      - julia --check-bounds=yes -e 'using Pkg; Pkg.add("TOML"); using TOML; using Test; @test TOML.parse(join(readlines(".buildkite/Manifest.toml")[7:end], "\n"))["deps"]["ClimaCore"][1]["path"] == ".."'
+
       - echo "--- Instantiate buildkite env"
-      - "julia --project=.buildkite --check-bounds=yes -e 'using Pkg; Pkg.develop(path=\".\"); Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true)'"
+      - "julia --project=.buildkite --check-bounds=yes -e 'using Pkg; Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true); Pkg.status()'"
       - "julia --project=.buildkite -e 'using CUDA; CUDA.precompile_runtime()'"
 
       - echo "--- Instantiate lib/ClimaCoreTempestRemap"
-      - "julia --project=lib/ClimaCoreTempestRemap -e 'using Pkg; Pkg.develop(path=\".\"); Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true)'"
+      - "julia --project=lib/ClimaCoreTempestRemap -e 'using Pkg; Pkg.develop(path=\".\"); Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true); Pkg.status()'"
 
       - echo "--- Instantiate benchmarks"
-      - "julia --project=benchmarks/bickleyjet -e 'using Pkg; Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true)'"
+      - "julia --project=benchmarks/bickleyjet -e 'using Pkg; Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true); Pkg.status()'"
 
     agents:
       slurm_mem: 80G


.buildkite/pipeline.yml
```diff
@@ -21,15 +21,18 @@ steps:
           rm -rf ${JULIA_DEPOT_PATH}
         fi
 
+      - echo "--- Check that ClimaCore is developed in Manifest"
+      - julia --check-bounds=yes -e 'using Pkg; Pkg.add("TOML"); using TOML; using Test; @test TOML.parse(join(readlines(".buildkite/Manifest.toml")[7:end], "\n"))["deps"]["ClimaCore"][1]["path"] == ".."'
```
Sbozzolo, line 25: I wonder if we can avoid hardcoding the path to the Manifest and rely on julia itself. For example, `pkgdir(ClimaCore)` would tell us. I don't know if Pkg has some functions to inpect a project without instantiating it

Also, we should avoid continuing the bulid if this check fails.

charleskawczynski: Unfortunately, `pkgdir(ClimaCore)` requires loading (and precompileing `ClimaCore`), which is why I went with just using `TOML` like this. It's ugly, but much more lightweight.

Sbozzolo: Yes, I was wondering if `Pkg` had some way to probe the environment programmatically without instanting it. I mean, it has to do it to instantiate up the env in the first place, but I don't know if this is somehow exposed to the user.

Alternatively, a slightly cleaner julia one liner would be:
```julia
using TOML; @assert TOML.parsefile(".buildkite/Manifest.toml")["deps"]["ClimaCore"][1]["path"] == ".."
```
This is to for N manifests:
```julia
using TOML; for f in filter(x->startswith("Manifest",x), readdir(".buildkite")); @assert TOML.parsefile(joinpath(".buildkite",f))["deps"]["ClimaCore"][1]["path"] == ".."; end
```

charleskawczynski: > Yes, I was wondering if Pkg had some way to probe the environment programmatically without instanting it.

I didn't see anything from a quick look through `Pkg.<tab>`. Another option, just using Pkg would be to `Pkg.develop(".")` and then check if the manifest has changed (via git).

Sbozzolo: A quick look seems to indicate that `Pkg.dependencies()[UUID("d414da3d-4745-48bb-8d80-42e94e092884")].source` is ClimaCore's path in this environment.

We can probably clean this up to avoid hardcoding the uuid and use the name instead

charleskawczynski: I kind of like that `using TOML; @assert TOML.parsefile(".buildkite/Manifest.toml")["deps"]["ClimaCore"][1]["path"] == ".."` is explicitly looking at the manifest file. The `Pkg.dependencies` business requires us to look into the implementation to see where it's getting that from. Also:

```
  Pkg.dependencies()::Dict{UUID, PackageInfo}

  This feature is considered experimental.
```

Sbozzolo: I think the strength of using `Pkg` is that we are guaranteed to access the same information that the julia is using in instantiating the environment. 

`Pkg.dependencies().source` is part of the documented API of this function, so I don't think that this is looking into the implementation. 

Actually, looking at the documentation, I see that there's directly a field `isdeveloped` to check what we want to check
```
    Pkg.dependencies()::Dict{UUID, PackageInfo}
Query the dependecy graph.
The result is a `Dict` that maps a package UUID to a `PackageInfo` struct representing the dependency (a package).
PackageInfo fields:
| `field`        | `Description`                                              |
|:---------------|:-----------------------------------------------------------|
| `name`         | The name of the package                                    |
| `version`      | The version of the package (this is `Nothing` for stdlibs) |
| `isdeveloped`  | Whether a package is directly tracking a directory         |
| `ispinned`     | Whether a package is pinned                                |
| `source`       | The directory containing the source code for that package  |
| `dependencies` | The dependencies of that package as a vector of UUIDs      |
```

So the test could be
```julia
using Pkg; @assert Pkg.dependencies()[Base.UUID("d414da3d-4745-48bb-8d80-42e94e092884")].isdeveloped
```

>  This feature is considered experimental.

This feature was introduced in julia 1.4 and has not changed at all since then. I opened an issue to Pkg to ask why they are still considered experimental.

All of this said, I don't feel strongly about this. I think that the benefits compared to manually parsing the file are more theoretical than practical, so I am happy to go with the other solution too. I'll leave it up to you.

charleskawczynski: Odd, I went to try `isdeveloped`, but I don't see that at all:

```julia
help?> Pkg.dependencies
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Pkg.dependencies

  Pkg.dependencies()::Dict{UUID, PackageInfo}

  This feature is considered experimental.

  Query the dependency graph of the active project. The result is a Dict that maps a package UUID to a PackageInfo struct representing the dependency (a package).

  PackageInfo fields
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  Field                Description                                                                      
  –––––––––––––––––––– –––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
  name                 The name of the package                                                          
  version              The version of the package (this is Nothing for stdlibs)                         
  tree_hash            A file hash of the package directory tree                                        
  is_direct_dep        The package is a direct dependency                                               
  is_pinned            Whether a package is pinned                                                      
  is_tracking_path     Whether a package is tracking a path                                             
  is_tracking_repo     Whether a package is tracking a repository                                       
  is_tracking_registry Whether a package is being tracked by registry i.e. not by path nor by repository
  git_revision         The git revision when tracking by repository                                     
  git_source           The git source when tracking by repository                                       
  source               The directory containing the source code for that package                        
  dependencies         The dependencies of that package as a vector of UUIDs  
```

charleskawczynski: I'm just going to move forward with `using TOML; @assert TOML.parsefile(".buildkite/Manifest.toml")["deps"]["ClimaCore"][1]["path"] == ".."`, I think this is a nice upgrade compared to the main branch.

Issue #52, opened 2021-06-21
Better error messages for invalid boundary conditions for certain FD stencil operators


jakebolewski:
We should probably throw an error at the constructor level when creating the Operator for invalid boundary condition combinations.  Ex. calling `SetGradient()` on `GradientF2C` in this example:

```
 bcs_bottom = Operators.SetValue(FT(b))   
    bcs_top = Operators.SetValue(FT(a))   
    gradc2f = Operators.GradientC2F()
    gradf2c = Operators.GradientF2C(bottom = bcs_bottom, top = bcs_top)
    @. dθ = gradf2c(ν * gradc2f(θ)) 
```

PR #2180, opened 2025-02-11
Add FD operator checks for allowable boundary conditions

charleskawczynski:
This PR adds assertions when constructing operators, to ensure that boundary conditions passed to those operators are supported. ~We could make these inner constructors~ I made them inner constructors, so that users are strictly unable to use incorrect boundary conditions for each operator.

This addresses the contructor-side checks for #52. Next, we should check before executing these broadcast expressions that the keys in the BCs match the space's `boundary_names`.

Testing that any BCs are given must be done right before execution, since the need for BCs in any particular operator depends on if and how they are composed.

This PR also updates / adds some documentation.

### DIFF
--- a/src/InputOutput/readers.jl
+++ b/src/InputOutput/readers.jl
@@ -284,8 +284,8 @@ function read_mesh_new(reader::HDF5Reader, name::AbstractString)
                 nelems = nelements,
             )
         end
-        if stretch_type ≠ "UnknownStretch"
-            stretch_params = attrs(group)["stretch_params"]
+        stretch_params = get(attrs(group), "stretch_params", nothing)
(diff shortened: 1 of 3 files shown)


Sbozzolo (approved):
Thank you!

charleskawczynski:
This PR passed ClimaCore CI [here](https://buildkite.com/clima/climacore-ci/builds/5276) and ClimaAtmos CI runs to completion ([here](https://buildkite.com/clima/climaatmos-ci/builds/23092))-- the failures there are due to reproducibility changes, likely from the metric term changes in #2167. So, I'm going to go ahead and manually merge this.

PR #2181, opened 2025-02-12
Update examples to use new implicit solver interface

dennisYatunin:
This PR updates the implicit solver in our ClimaCore examples to use the `MatrixFields` interface, brining it in alignment with the implicit solver in ClimaAtmos.

## Updates to Examples and Tests
- Replace `compose`, `operator2stencil`, and `StencilCoefs` in the tendency Jacobian with their new versions: `⋅`, `operator_matrix`, and `BandMatrixRow` from `MatrixFields`
  - Since `operator_matrix` supports upwinding, add the missing upwinding derivative to the tendency Jacobian
- Replace scalar Jacobian entries with vector, covector, and tensor entries, so that they match the ones in ClimaAtmos
  - To avoid modifying the `Wfact!` function and making it set `dtγ * ∂Yₜ∂Y - I` instead of `∂Yₜ∂Y`,  this required adding support for scalar multiplication/division of matrices, as well as support for identity matrices with non-scalar elements to `MatrixFields`
- Remove the autodiff testing functionality from the implicit solver, as it needs to be reimplemented in a way that supports non-scalar entries
- Remove the `linsolve` test from CI, as it was testing an outdated duplicate of the Thomas algorithm from `MatrixFields`

## Updates to Internals
- Allow `FieldMatrix` broadcasts to include multiplication and division by single values (`Number`s, `AxisTensor`s, or broadcast expressions that return `Number`s or `AxisTensor`s)
- Allow multiples of identity matrix blocks in a `FieldMatrix` to be represented as either `UniformScaling`s for `Number`s or `DiagonalMatrixRow`s for `Axis2Tensor`s
- Fix the behavior of `one(::FieldMatrix)`
  - Ensure that the identity matrix block for a `ColumnwiseBandMatrixField` of `Axis2Tensor`s is also an `Axis2Tensor`
  - Enable better handling of sparse matrices, so that `one(matrix)` will contain an identity matrix block for each variable that appears as a row or column index, even if `matrix` does not contain the corresponding diagonal entry
- Add an alternative to `one(::FieldMatrix)`, `identity_field_matrix(::FieldVector)`, which captures missing identity matrix blocks for very sparse matrices
  - E.g., if `matrix` does not contain any blocks with `Y.c.uₕ` as a row or column value, `one(matrix)` will not contain an identity block for `Y.c.uₕ`, but `identity_field_matrix(Y)` will contain `(@name(c.uₕ), @name(c.uₕ)) => I`
  - This can be used to fill in missing diagonal blocks in a sparse `FieldMatrix`, so that it can be passed to a `FieldMatrixSolver`
- These changes will allow us to simplify the implicit solver in ClimaAtmos, since we can now subtract `I` and multiply by `dtγ` automatically, instead of relying on developers to do this manually for each matrix block
- Add broadcasting support to `FieldMatrixWithSolver`, as well as a definition of `one(::FieldMatrixWithSolver)`, so that we do not need to directly access the internal `matrix` when subtracting `I` and multiplying by `dtγ`

---


examples/hybrid/implicit_equation_jacobian.jl
```diff
@@ -0,0 +1,113 @@
...
+    zero(A.∂Yₜ∂Y),
+    zero(A.∂R∂Y),
+    A.transform,
+    A.flags,
+    A.temp1,
+    A.temp2,
+)
+
+# This method for ldiv! is called by Newton's method from ClimaTimeSteppers.jl.
+# It solves ∂R∂Y * x = b for x, where R is the implicit residual.
```
charleskawczynski, line 91: Can you expand on this a bit more? Perhaps add a proper doc string to `ImplicitEquationJacobian`?

charleskawczynski: I know it's just an example, but this logic in the code is pretty complex.

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -35,16 +41,16 @@ const ᶠinterp = Operators.InterpolateC2F(
     top = Operators.Extrapolate(),
 )
 const ᶜdivᵥ = Operators.DivergenceF2C(
-    top = Operators.SetValue(Geometry.Contravariant3Vector(FT(0))),
-    bottom = Operators.SetValue(Geometry.Contravariant3Vector(FT(0))),
+    top = Operators.SetValue(CT3(FT(0))),
```
charleskawczynski, line 44: Is `FT` just a global constant defined somewhere?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -273,46 +287,52 @@ end
...
-        ᶜinterp_stencil(one(ᶠw_data))
+    #     ACT12(ᶜuₕ) * ᶜuₕ / 2 + ACT3(ᶜinterp(ᶠw)) * ᶜinterp(ᶠw) / 2
+    # ∂(ᶜK)/∂(ᶠw) = ACT3(ᶜinterp(ᶠw)) * ᶜinterp_matrix()
+    @. ∂ᶜK∂ᶠw = DiagonalMatrixRow(adjoint(CT3(ᶜinterp(ᶠw)))) ⋅ ᶜinterp_matrix()
 
     # ᶜρₜ = -ᶜdivᵥ(ᶠinterp(ᶜρ) * ᶠw)
-    # ∂(ᶜρₜ)/∂(ᶠw_data) = -ᶜdivᵥ_stencil(ᶠinterp(ᶜρ) * ᶠw_unit)
-    @. ∂ᶜρₜ∂ᶠ𝕄 = -(ᶜdivᵥ_stencil(ᶠinterp(ᶜρ) * one(ᶠw)))
+    # ∂(ᶜρₜ)/∂(ᶠw) = -ᶜdivᵥ_matrix() * ᶠinterp(ᶜρ) * ᶠg³³
+    @. ∂ᶜρₜ∂ᶠ𝕄 = -(ᶜdivᵥ_matrix()) ⋅ DiagonalMatrixRow(ᶠinterp(ᶜρ) * g³³(ᶠgⁱʲ))
```
charleskawczynski, line 335: Just to make sure I understand this, this is a matrix-matrix multiplication? (yielding a matrix field?)

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -273,46 +287,52 @@ end
...
+        Geometry.components(gⁱʲ)[end],
+    )
+
+    # If ∂(ᶜχ)/∂(ᶠw) = 0, then
+    # ∂(ᶠupwind_product(ᶠw, ᶜχ))/∂(ᶠw) =
+    #     ∂(ᶠupwind_product(ᶠw, ᶜχ))/∂(CT3(ᶠw)) * ∂(CT3(ᶠw))/∂(ᶠw) =
+    #     vec_data(ᶠupwind_product(ᶠw + εw, ᶜχ)) / vec_data(CT3(ᶠw + εw)) * ᶠg³³
+    # The ε is only necessary when w = 0. Since ᶠupwind_product is undefined at
+    # the boundaries, we also need to wrap it in a call to ᶠno_flux.
+    vec_data(vector) = vector[1]
```
charleskawczynski, line 324: Can we document what `vec_data` is for?

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Yeah it's defined in the driver: https://github.com/CliMA/ClimaCore.jl/blob/930a892a89f3e826afd4e30a9c91967f7d7b250c/examples/hybrid/driver.jl#L55

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Yes each use of `⋅` in this function represents a matrix-matrix multiplication, corresponding to the composition of finite difference operators and scalar multiplications.

examples/hybrid/implicit_equation_jacobian.jl
dennisYatunin: I've added a detailed docstring to `ImplicitEquationJacobian`. If I have time to clean these examples up further, I will eventually turn them into a full tutorial section in the documentation.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: I've added some more details to this comment about `vec_data` and `εw`.

PR #2184, opened 2025-02-12
Use shared memory in `DivergenceF2C` stencil operators

charleskawczynski:
This PR implements shared memory for the finite difference stencil `DivergenceF2C`. This is a step towards #1746, where the [high-level design](https://github.com/CliMA/ClimaCore.jl/issues/1746#issuecomment-2682159755) is discussed (it's basically the same design that we use for spectral element kernels).

### Some notes on implementation and behavior details:

 - Like our SEM, shared memory (shmem) is supported through a single layer of operator composition. That is, composed operators like `div(grad(f))` will put the result of `grad(f)` into shared memory per thread (or, per point), but duplicate reads of `f` will still occur (thread `i` will read `f` at `i+1` and thread `i+1` will read `f` at  `i`). To improve on this, we can write combined operators (e.g., `divgrad`), where shmem is managed, and this will result in a minimal number of required reads/writes. Ideally, we could automate this somehow, but I think that's the next step. One nice aspect of automating this is that we wouldn't need to introduce new terminology to users.
 - If the vertical resolution is too large (e.g., 1000 vertical levels), then we dispatch to a shmem-free version to support higher vertical resolution.
 - This PR only adds shared memory for a F2C operator, we should probably add one C2F operator, so that we exercise the cell center-filling shared memory branch code (which is currently not exercised, and is therefore likely not correct).
 - I've also added some specializations for `JContravariant3` to convince cuda to emit fused instructions (found through nsight compute).

### Direct kernel performance impact

This [comment](https://github.com/CliMA/ClimaCore.jl/pull/2184#issuecomment-2672013425) discusses the direct impact on kernels.

### Overall performance impact

 - Here is a GPU target pipeline [build](https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/405) of ClimaAtmos, and the latest nearby [main](https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/406). For moist held suarez, 1 gpu:
 - main: SYPD 1.88
 - This branch: SYPD 2.056

### Adding new operators

To add new operators, developers need to update two files:

 - `ext/cuda/operators_fd_shmem.jl`, where you need to define:
   - `fd_operator_shmem` - allocates shared memory
   - `fd_operator_fill_shmem_interior!` - defines how interior stencil shmem is populate
   - `fd_operator_fill_shmem_left_boundary!` - defines how left stencil shmem is populate
   - `fd_operator_fill_shmem_right_boundary!` - defines how right stencil shmem is populate
   - `fd_operator_evaluate` - defines how to get the result of `getidx` from shmem
  - `ext/cuda/operators_fd_shmem_is_supported.jl`, where you need to define `fd_shmem_is_supported` for the newly supported operator (and boundary conditions)


charleskawczynski:
Note to self: need to look at threading pattern

Sbozzolo:
I think the expectation of a 40 % improvement relies on the assumption that you have complete L1 cache misses, but if different threads within a block share faces and not much more data is needed, the effective reads to global memory will be on average less than 5 because some values will be cached in the L1 cache of the streaming multiprocessor (which has the same latency as the shared memory).

charleskawczynski:
Even slightly complicated cases are showing a nice improvement (~2x), so it may just depend on additional factors (e.g., register pressure / if there are errors / traps emitted by LLVM).

charleskawczynski:
In this PR, shared memory (shmem) is supported through a single layer of operator composition. That is, composed operators like `div(grad(f))` will put the result of `grad(f)` into shared memory per thread (or, per point), but duplicate reads of `f` will still occur (thread `i` will read `f` at `i+1` and thread `i+1` will read `f` at  `i`). To improve on this, we can write combined operators (e.g., `divgrad`), where shmem is managed, and this will result in a minimal number of required reads/writes.

Unfortunately, there are probably a lot of combinations of combined operators, however, we may be able to automatically transform the composed operators into the combined operators on the back-end, so that we don't need to introduce a slew of new terminology to users.

charleskawczynski:
I think that this is in near merge-worthy shape, the only remaining issue(s) I see are:

 - This PR changes the default threading pattern for all stencil operations, which may not be best until shmem is supported. We could add a check before the kernel launch to look to see if any operators support shmem.
 - I think this may change the limit of vertical resolution that we support (1,000 levels in the test was lowered)-- should we dispatch to a shmem-free version to support higher vertical resolution? I don't think we'll ever need more than 256, which is supported by this implementation.

charleskawczynski:
I've fixed the main bug I was running into, I think that this is ready to go. Here are the preliminary results for some of the relevant benchmarks (in ):

```
       main           shmem (this branch)
ᶜout1: 245.218 μs,    200.758 μs
ᶜout2: 287.298 μs,    228.957 μs
ᶜout3: 578.095 μs,    281.637 μs
ᶜout4: 320.777 μs,    225.958 μs
ᶜout5: 335.117 μs,    249.007 μs
ᶜout6: 568.145 μs,    261.858 μs
ᶜout7: 439.927 μs,    249.397 μs
ᶜout8: 440.407 μs,    239.448 μs
```

src/Geometry/conversions.jl
```diff
@@ -541,6 +541,16 @@ Required for statically infering the result type of the divergence operation for
 ) where {FT, A1, A2 <: LocalAxis, S <: StaticMatrix{S1, S2}} where {S1, S2} =
     AxisVector{FT, A2, SVector{S2, FT}}
 
+# TODO: can we better generalize this?
```
charleskawczynski, line 544: @dennisYatunin, can we better generalize this?

Sbozzolo:
The build that you run in ClimaAtmos shows a significant regression in the "Benchmark: GPU prog edmf" test compared to `main`. So maybe that's something that you want to look into before we look at this?

I also leave my first impression here:
> I think this may change the limit of vertical resolution that we support (1,000 levels in the test was lowered)-- should we dispatch to a shmem-free version to support higher vertical resolution? I don't think we'll ever need more than 256, which is supported by this implementation.

I think that this should be addressed. We will not need more than 256 levels for a global simulation, but we might want it in other settings. For example, I used more than 256 levels to study self-convergence in ClimaTimeSteppers. I think that the restriction of 256 points on a finite difference method is very strong for most applicaitons that are not global simulations. Moreover, this would further differentiate our CPU and GPU capabilities. (I think it'd be perfectly acceptable to have a slower path for when the number of points is more than 256, but users should still be able to run with such a configuration)

charleskawczynski:
I'll of course make sure we address the performance before merging.

Allowing a larger number of vertical levels would be nice too. I'm going to think about this. If we could preserve both code-paths somehow, it might fix both issues.

Do you have any other comments?

charleskawczynski:
I just pushed a fix that should fix all of these issues. Now we check ahead of time and transform the broadcasted style to disable the shmem broadcasted object if no shmem is supported (which should fix the regression) or if the resolution is too high (to maintain super-high resolution support).

charleskawczynski:
> The build that you run in ClimaAtmos shows a significant regression in the "Benchmark: GPU prog edmf" test compared to `main`. So maybe that's something that you want to look into before we look at this?

Which build are you comparing against?

Sbozzolo:
xref: https://github.com/CliMA/ClimaAtmos.jl/issues/2632

Meeting 24 Feb notes:

High level goal:
- Shared memory is required to allow for larger kernels that would not otherwise fit for finite difference operators
- Performance benefits due to reduced reads

Tradeffs:
- Code complexity and duplication
- Developer time

Path forward:
- Start with one operator (this PR)
- Progressively add more and more operators.
- Experiment with this by supporting three code paths: CPU, GPU without shared memory, GPU with shared memory. 
- Later down the line, experiment with unifying shared memory backends 

High level design:
- 

Concerns:
- Dennis: we should not manually fuse operators, we should use lazy representations as in MatrixFields

charleskawczynski:
> Experiment with this by supporting three code paths: CPU, GPU without shared memory, GPU with shared memory.

Just to reiterate on this point: the amount of duplication for the "three" code paths is really very small, `copyto_stencil_kernel!` is the only method that we need to support the non-shmem version.

charleskawczynski:
I realized something recently this PR. We don't need to define shmem operators for combinations of operators because the read operation is collocated and any global memory read performed by a thread will be put into a register, and that can be reused for any part of the broadcasted object. So, I'm now thinking that it's worth moving forward with this for at least a handful of operators.

Given this, fusing operations with this pattern should basically minimize the number of reads, which is great, but it could be limited in terms of how many operators that we can fuse, since shared memory is a limited resource, and we would need shmem allocated per operator.

So, it's possible that a super fused operator (`ColumnStateShmem` or something) could still be better in the limit of fusing many operations. See #1763 for an initial sketch, where I was applying this to the existing `apply_stencil!`. I should have defined a new operator in that PR to explicitly request that behavior-- that will prevent performance regressions for our existing operators.

I think a super fused is possible in principle, but I think I'd first need to count the number of needed metric terms in the local geometry to find out at what point it will benefit us.

dennisYatunin (approved):
There's definitely some duplicated code patterns here, but overall this is fine to merge in as an experimental feature. In the future, we should try to figure out a way to avoid duplicating the stencil definition for each vertical operator (by somehow generalizing the stencil coefficients in `operator_matrices.jl`, akin to what #782 was doing), and a way to combine the vertical and horizontal shared memory interfaces (`allocate_shmem`, `resolve_shmem`, etc). As long as this duplication is confined to the CUDA extension, it shouldn't be too hard to simplify things later.

PR #2188, opened 2025-02-14
Pin DiskArrays to fix ClimaCoreTempestRemap CI

dennisYatunin:
Using DiskArrays `v0.4.9` instead of `v0.4.8` breaks our ClimaCoreTempestRemap CI---a deprecation warning is printed for `DiskArrays.canscalar` and then the test hangs.

This package is only used by NCDatasets, which is currently on `v0.14.6` and has a DiskArrays compatibility entry of `v0.4.5`, which corresponds to the range `[0.4.5, 0.5.0)`. However, NCDatasets does not call any of the functions that were changed in DiskArrays `v0.4.9`, so I have no idea why the deprecation warning is printed and what's causing the ClimaCoreTempestRemap test to hang. In the interest of getting ClimaCore CI working again, I'm adding a compatibility entry for DiskArrays of `<v0.4.9`. Hopefully we can remove this after an update to NCDatasets is released.


Sbozzolo (approved):

charleskawczynski (approved):
Need to update the manifest files

PR #2085, opened 2024-11-15
Add element mask to `SpectralElementGrid2D`

sriharshakandala:
Add element mask to `SpectralElementGrid2D`

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #2201, opened 2025-02-25
Add nodal mask

charleskawczynski:
This PR adds a nodal mask into the `SpectralElementGrid2D`, and allows users to update it with `set_mask!(::Function, space::AbstractSpace)`. This is an initial step towards this [SDI](https://github.com/CliMA/ClimaLand.jl/issues/488).

I imagine that just the pointwise and stencil operations will go a long way, but we may need to iterate to find out what else is needed.

Note to self: I think we also need to thread the mask through in some `materialize!` calls for implicit pieces, assuming that they call `copyto!`

Supersedes #2085

Here is a rough summary of how this works

### Summary

 - We add the mask, in the form of a DataLayout to the horizontal grid, or a `DataLayouts.NoMask` object
   - Using different types like this allows us to dispatch on behavior when the mask is present and absent, effectively becoming a no-op when the mask is absent
 - We then extract and pass that mask to our kernels (i.e., `Base.copyto!`)
 - Inside `Base.copyto!`, if the mask is a datalayout, we check if `mask == 0` for a particular `i, j, f, v, h`, before doing the broadcast assignment (`dest[I] = bc[I]`) then we skip, via `continue` (or `return nothing`, in the case of cuda kernels).

### DIFF
--- a/docs/src/masks.md
+++ b/docs/src/masks.md
@@ -1,0 +1,199 @@
+# Masks
+
+## Motivation
+
+ClimaCore spaces, `SpectralElementSpace2D`s in particular, support masks, where
+users can set horizontal nodal locations where operations are skipped.
+
+This is especially helpful for the land model, where they may have degrees of
+freedom over the ocean, but do not want to evaluate expressions in regions where
+data is missing.
+
+Masks in ClimaCore offer a solution to this by, ahead of time prescribing
+regions to skip. This helps both with the ergonomics, as well as performance.
+
+## User interface
+
+There are two user-facing parts for ClimaCore masks:
+
+ - set the `enable_mask = true` keyword in the space constructor (when available),
+   which is currently any constructor that returns/contains a `SpectralElementSpace2D`.
+ - use `set_mask!` to set where the mask is `true` (where compute should occur)
+   and `false` (where compute should be skipped)
+
+Here is an example
+
+```julia
+using ClimaComms
+ClimaComms.@import_required_backends
+import ClimaCore: Spaces, Fields
+using ClimaCore.CommonSpaces
+using Test
+
+FT = Float64
+ᶜspace = ExtrudedCubedSphereSpace(FT;
+    z_elem = 10,
+    z_min = 0,
+    z_max = 1,
+    radius = 10,
+    h_elem = 10,
+    n_quad_points = 4,
+    staggering = CellCenter(),
+    enable_mask = true,
+)
+
+# How to set the mask
+Spaces.set_mask!(ᶜspace) do coords
+    coords.lat > 0.5
+end
+# Or
+mask = Fields.Field(FT, ᶜspace)
+mask .= map(cf -> cf.lat > 0.5 ? 0.0 : 1.0, Fields.coordinate_field(mask))
+Spaces.set_mask!(ᶜspace, mask)
+```
+
+Finally, operations over fields will be skipped where `mask == 0`, and applied
+where `mask == 1`:
+
+```
+@. f = 1 # only applied where the mask is equal to 1
+```
+
+## Example script
+
+Here is a more complex script where the mask is used:
+
+```julia
+using ClimaComms
+ClimaComms.@import_required_backends
+import ClimaCore: Spaces, Fields, DataLayouts, Geometry, Operators
+using ClimaCore.CommonSpaces
+using Test
+
+FT = Float64
+ᶜspace = ExtrudedCubedSphereSpace(FT;
+    z_elem = 10,
+    z_min = 0,
+    z_max = 1,
+    radius = 10,
+    h_elem = 10,
+    n_quad_points = 4,
+    staggering = CellCenter(),
+    enable_mask = true,
+)
+ᶠspace = Spaces.face_space(ᶜspace)
+ᶠcoords = Fields.coordinate_field(ᶠspace)
+
+# How to set the mask
+Spaces.set_mask!(ᶜspace) do coords
+    coords.lat > 0.5
+end
+
+# We also support the syntax `Spaces.set_mask!(::AbstractSpace, ::Field)`
+
+# We can check the mask directly: (internals, only for demonstrative purposes)
+mask = Spaces.get_mask(ᶜspace)
+@test count(parent(mask.is_active)) == 4640
+@test length(parent(mask.is_active)) == 9600
+
+# Let's skip operations that use fill!
+ᶜf = zeros(ᶜspace) # ignores mask
+@. ᶜf = 1 # tests fill! # abides by mask
+
+# Let's show that 4640 columns were impacted:
+@test count(x->x==1, parent(ᶜf)) == 4640 * Spaces.nlevels(axes(ᶜf))
+@test length(parent(ᶜf)) == 9600 * Spaces.nlevels(axes(ᶜf))
+
+# Let's skip operations that use copyto!
+ᶜz = Fields.coordinate_field(ᶜspace).z
+ᶜf = zeros(ᶜspace)
+@. ᶜf = 1 + 0 * ᶜz # tests copyto!
+
+# Let's again show that 4640 columns were impacted:
+@test count(x->x==1, parent(ᶜf)) == 4640 * Spaces.nlevels(axes(ᶜf))
+@test length(parent(ᶜf)) == 9600 * Spaces.nlevels(axes(ᶜf))
+
+# Let's skip operations in FiniteDifference operators
+ᶠf = zeros(ᶠspace)
+c = Fields.Field(FT, ᶜspace)
+div = Operators.DivergenceF2C()
+foo(f, cf) = cf.lat > 0.5 ? zero(f) : sqrt(-1) # results in NaN in masked out regions
+@. c = div(Geometry.WVector(foo(ᶠf, ᶠcoords)))
+
+# Check that this field should never yield NaNs
+@test count(isnan, parent(c)) == 0
+
+# Doing the same thing with a space without a mask will yield NaNs:
+ᶜspace_no_mask = ExtrudedCubedSphereSpace(FT;
+    z_elem = 10,
+    z_min = 0,
+    z_max = 1,
+    radius = 10,
+    h_elem = 10,
+    n_quad_points = 4,
+    staggering = CellCenter(),
+)
+ᶠspace_no_mask = Spaces.face_space(ᶜspace_no_mask)
+ᶠcoords_no_mask = Fields.coordinate_field(ᶠspace_no_mask)
+c_no_mask = Fields.Field(FT, ᶜspace_no_mask)
+ᶠf_no_mask = Fields.Field(FT, ᶠspace_no_mask)
+@. c_no_mask = div(Geometry.WVector(foo(ᶠf_no_mask, ᶠcoords_no_mask)))
+@test count(isnan, parent(c_no_mask)) == 49600
+```
+
+## Supported operations and caveats
+
+Currently, masked _operations_ are only supported for `Fields` (and not
+`DataLayouts`) with `SpectralElementSpace2D`s. We do not yet have support for
+masked `SpectralElement1DSpace`s, and we will likely never offer masked
+operation support for `DataLayouts`, as they do not have the space, and can
+therefore not use the mask.
+
+In addition, some operations with masked fields skip masked regions
+(i.e., mask-aware), and other operations execute everywhere
+(i.e., mask-unaware), effectively ignoring the mask. Here is a list of
+operations of mask-aware and mask-unaware:
+
+ - `DataLayout` operations (`Fields.field_values(f) = 1`) mask-unaware (will likely never be mask-aware).
+ - `fill!` (`@. f = 1`) mask-aware
+ - point-wise `copyto!` (`@. f = 1 + z`) mask-aware
+ - stencil `copyto!` (`@. ᶜf = 1 + DivergenceF2C()(Geometry.WVector(ᶠf))`) mask-aware (vertical derivatives and interpolations interpolations)
+ - spectral element operations `copyto!` (`@. f = 1 + Operators.Divergence()(f)`), where `Operators.Divergence` carries out a divergence operation in horizontal directions. mask-unaware
+ - fieldvector operations `copyto!` (`@. Y += 1`) mask-unaware
+ - reductions:
+   - `sum` (mask-unaware, warning is thrown)
+   - `extrema` (mask-unaware, warning is thrown)
+   - `min` (mask-unaware, warning is thrown)
+   - `max` (mask-unaware, warning is thrown)
+ - field constructors (`copy`, `Fields.Field`, `ones`, `zeros`) are mask-unaware.
+   This was a design implementation detail, users should not generally depend on the results where `mask == 0`, in case this is changed in the future. 
+ - internal array operations (`fill!(parent(field), 0)`) mask-unaware.
+
+## Developer docs
+
+In order to support masks, we define their types in `DataLayouts`, since 
+we need access to them from within kernels in `DataLayouts`. We could have made
+an API and kept them completely orthogonal, but that would have been a bit more
+complicated, also, it was convenient to make the masks themselves data layouts,
+so it seemed most natural for them to live there.
+
+We have a couple types:
+
+ - abstract `AbstractMask` for subtyping masks and use for generic interface
+   methods
+ - `NoMask` (the default), which is a lazy object that should effectively result
+   in a no-op, without any loss of runtime performance
+ - `IJHMask` currently the only supported horizontal mask, which contains
+   `is_active` (defined in `set_mask!`), `N` (the number of active columns),
+   and maps containing indices to the `i, j, h` locations where `is_active` is
+   true. The maps are defined in `set_mask_maps!`, allows us to launch cuda
+   kernels to only target the active columns, and threads are not wasted on
+   non-existent columns. The logic to handle this is relatively thin, and
+   extends our current `ext/cuda/datalayouts_threadblock.jl` api
+   (via `masked_partition` and `masked_universal_index`).
+
+An important note is that when we set the mask maps for active columns, the
+order that they are assigned can be permuted without impacting correctness, but
+this could have a big impact on performance on the gpu. We should investigate
+this.
+
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -92,6 +92,41 @@ abstract type AbstractDispatchToDevice end
 struct ToCPU <: AbstractDispatchToDevice end
 struct ToCUDA <: AbstractDispatchToDevice end
 
+"""
+    AbstractMask
+
+An abstract mask type, for marking domains as present or absent.
+"""
+abstract type AbstractMask end
+
+"""
+    NoMask
+
+A lazy non-mask type (default), used to indicate that an entire domain is
+present.
+"""
+struct NoMask <: AbstractMask end
+
+"""
+    IJHMask
+
+A mask type, used to house and use information for which columns
+in a grid are active.
+ - `is_active` a bool mask, the size of the entire grid, that indicates if the
+   column is active or not
+ - `N` a Int array with 1 element, containing number of active columns
+ - `i_map` a Int array, containing i-indices of active columns
+ - `j_map` a Int array, containing j-indices of active columns
+ - `h_map` a Int array, containing h-indices of active columns
+"""
+struct IJHMask{B, NT, V} <: AbstractMask
+    is_active::B
+    N::NT
+    i_map::V
+    j_map::V
+    h_map::V
+end
+
 include("struct.jl")
 
 abstract type AbstractData{S} end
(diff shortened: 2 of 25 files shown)


src/Fields/Fields.jl
```diff
@@ -22,6 +22,7 @@ import ..Quadratures
 import ..Grids: ColumnIndex, local_geometry_type
```
juliasloan25, line 22: There's a method for `copyto!` in line 278 of this file - do we need to include the mask for that method too?

juliasloan25: Same question for `copyto!` defined in `src/Fields/fieldvector.jl` and `src/Operators/common.jl`

test/Spaces/unit_spaces.jl
```diff
@@ -25,6 +25,48 @@ import ClimaCore.DataLayouts: IJFH, VF, slab_index
 
 on_gpu = ClimaComms.device() isa ClimaComms.CUDADevice
 
+@testset "2D spaces with mask" begin
```
juliasloan25, line 28: It would be good to add small tests for `PointSpace` and an `Extruded` space too, to check the `get_mask` definitions in `src/Spaces/Spaces.jl`

test/Spaces/unit_spaces.jl
```diff
@@ -25,6 +25,48 @@ import ClimaCore.DataLayouts: IJFH, VF, slab_index
...
+    # Tests with mask
+    hspace = Spaces.SpectralElementSpace2D(htopology, quad; enable_mask = true)
+    @test hspace.grid.mask isa DataLayouts.IJFH
+    @test eltype(hspace.grid.mask) <: Bool
+    Spaces.set_mask!(hspace) do coords
+        coords.x > 0.5
+    end
+    @test count(parent(hspace.grid.mask)) == 2
+    @test length(parent(hspace.grid.mask)) == 4
+
```
juliasloan25, line 63:
```suggestion
@test !(hspace.grid.mask isa DataLayouts.NoMask)

```
Do you think this would be useful, or is it redundant with the other checks here

src/DataLayouts/copyto.jl
```diff
@@ -61,31 +65,39 @@ function Base.copyto!(
...
+    fill!(dest, bc0, mask)
+    call_post_op_callback() && post_op_callback(dest, dest, bc, to, mask)
 end
 
 #####
 ##### DataLayouts
 #####
 
+compute(::NoMask, idx) = true
+compute(mask, idx) = !iszero(mask[idx])
```
juliasloan25, line 77: This is minor, but to me the name `compute` implies that some computation is being done, rather than just checking if computation should be done. Maybe something like `mask_check_compute` would be more informative

src/Operators/finitedifference.jl
```diff
@@ -4032,7 +4032,9 @@ function _serial_copyto!(field_out::Field, bc, Ni::Int, Nj::Int, Nh::Int)
     space = axes(field_out)
     bounds = window_bounds(space, bc)
     bcs = bc # strip_space(bc, space)
+    mask = Spaces.get_mask(axes(field_out))
     @inbounds for h in 1:Nh, j in 1:Nj, i in 1:Ni
+        DataLayouts.compute(mask, CartesianIndex(i, j, 1, 1, h)) || continue
```
juliasloan25, line 4037: Just curious, why do you use `continue` here instead of `return nothing`?

juliasloan25 (commented):
This looks very helpful, thank you! I left a few comments throughout, and it would also be good to add documentation advertising this new feature and demonstrating how to use it

test/Spaces/unit_spaces.jl
charleskawczynski: I agree that we can / should write tests for extruded spaces, but I don't know that masks make sense for point spaces. If you've called `Fields.column` on a field, it's already too late to apply a mask operation. Same for FiniteDifferenceSpace-- if you mask a FD space, then no operations will ever happen.

test/Spaces/unit_spaces.jl
charleskawczynski: I think this is probably a good idea

src/Operators/finitedifference.jl
charleskawczynski: Because we're in a loop, we want to continue to the next iteration of the loop.

src/DataLayouts/copyto.jl
charleskawczynski: Yeah, I punted on the name because I couldn't think of a good one. A name that I think is descriptive would be `skip_compute`, but I prefer and try to avoid double negatives (e.g., `!skip_compute`)

src/Operators/finitedifference.jl
charleskawczynski: For cuda kernels, there is no more work to be done by the thread, so we can just `return nothing`.

src/Fields/Fields.jl
charleskawczynski: Ah, good catch for 278. It's basically never used (we typically use `@. x = y`, which calls `Base.copyto!(::Field, ::Base.Broadcasted)`), but perhaps there are edge cases. I'll add it there.

For the fieldvector, I intentionally skipped adding mask support for two reasons:
 - we would need to de-optimize those operations to support masks for them (our current implementations are very fast and embarrassingly parallel)
 - I thought that this might be okay because most if not all field vector operations are uniform and or are performed in the timestepper, and I figured that land tendencies over the ocean would just be zeros

I don't think that we'd likely see any performance gain on the cpu because we can currently do linear indexing, which we've found to be 2x faster than cartesian indexing. So, filtering out unused columns would involve de-optimizing that linear memory access.

On the gpu, it may actually be beneficial, so I'll thread it through, but I'm going to leave it unused on the cpu, and we can experiment on the gpu.

src/DataLayouts/copyto.jl
charleskawczynski: Or `is_active`?

src/DataLayouts/DataLayouts.jl
```diff
@@ -92,6 +92,21 @@ abstract type AbstractDispatchToDevice end
 struct ToCPU <: AbstractDispatchToDevice end
 struct ToCUDA <: AbstractDispatchToDevice end
 
+"""
+    AbstractMask
+
+An abstract mask type, for marking domains as present or absent.
+"""
+abstract type AbstractMask end
```
Sbozzolo, line 100: We should add documentation of properties/methods a type that subtypes `AbstractMask` should implement.

src/DataLayouts/copyto.jl
juliasloan25: I like `should_compute`!

test/Spaces/unit_spaces.jl
charleskawczynski: Actually, it is redundant with the test above (at 56)

test/Spaces/unit_spaces.jl
charleskawczynski: I've added extruded spaces tests

src/Fields/Fields.jl
charleskawczynski: I looked at the two in `common`, and we don't need them there since we already get the mask in the `copyto!` for the stencils (which is what those methods are calling).

src/DataLayouts/copyto.jl
charleskawczynski: Updated!

src/DataLayouts/DataLayouts.jl
charleskawczynski: Yes, I'm going to start adding more docs now, since some things have changed since the initial sketch (in order to avoid compute on columns in the cuda kernels).

charleskawczynski: I added some docs that refer to this in the developer docs, but this is really an internal type, since the only two things that users need are: `enable_mask = true` in the space constructor, and `set_mask!`

charleskawczynski:
I'm on the fence regarding the mask index maps:

```julia
main
julia> using Revise; include("test/standalone/Soil/mask_test.jl")
BenchmarkTools.Trial: 1559 samples with 1 evaluation per sample.
 Range (min … max):  1.176 ms …   6.669 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.243 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.206 ms ± 343.094 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                      ▅█▅▆▅▄▂  
  ▄▄▄▆▇▅▆▁▄▅▄▄▆▁▄▄▄▅▄▁▆▅▄▁▄▄▁▁▁▁▁▁▁▄▆▄▄▅▅█▅▇▆▁▁▄▁▁▁▄▆████████ █
  1.18 ms      Histogram: log(frequency) by time      3.46 ms <

 Memory estimate: 6.14 KiB, allocs estimate: 243.

with mask
julia> using Revise; include("test/standalone/Soil/mask_test.jl")
BenchmarkTools.Trial: 1606 samples with 1 evaluation per sample.
 Range (min … max):  1.006 ms …   3.441 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.154 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.112 ms ± 332.086 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                      ▅█▆▅▆▅▃  
  ▄▄▁▁▅▆▇▆▄▄▆▁▄▄▁▁▄▆▄▄▄▄▄▄▄▅▁▄▁▁▄▁▁▄▁▆▆▄▅▄█▆▇▁▁▄▄▁▄▄▄▆███████ █
  1.01 ms      Histogram: log(frequency) by time      3.35 ms <

 Memory estimate: 10.52 KiB, allocs estimate: 523.
```

The performance characteristics with the mask index maps could be pretty unpredictable without some work. Should we just do the Boolean mask? In theory I imagine that the mask index maps could be made faster, but until more profiling and work is put into it, performance could degrade in different ways.

I'm happy to leave the arrays in there and leave them as comments / todos, if people think that's a good idea. Or open an issue and revert it.

Sbozzolo (commented):
I just went thorugh the PR at an extremely high level and left a few comments (I will have more).

Two pieces I don't see here are the guardrails to prevent users from getting incorrect results (e.g., in the reductions or the Remapper) and support for checkpoints.

src/DataLayouts/DataLayouts.jl
```diff
@@ -92,6 +92,41 @@ abstract type AbstractDispatchToDevice end
...
+"""
+struct NoMask <: AbstractMask end
+
+"""
+    IJHMask
+
+A mask type, used to house and use information for which columns
+in a grid are active.
+ - `is_active` a bool mask, the size of the entire grid, that indicates if the
+   column is active or not
```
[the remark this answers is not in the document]
charleskawczynski: Ah, it is process-local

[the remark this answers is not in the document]
charleskawczynski: The reason I used "entire" here was to differentiate it from the maps, which are subsets of the process-local grid. I'll update the doc string.

docs/src/masks.md
```diff
@@ -0,0 +1,198 @@
...
+ - `DataLayout` operations (`Fields.field_values(f) = 1`) unsupported (these )
+ - `fill!` (`@. f = 1`) supported
+ - point-wise `copyto!` (`@. f = 1 + z`) supported
+ - stencil `copyto!` (`@. ᶜf = 1 + DivergenceF2C()(Geometry.WVector(ᶠf))`) supported (vertical derivatives and interpolations interpolations)
+ - spectral element operations `copyto!` (`@. ᶜf = 1 + Operators.Divergence()(ᶠf)`) (unsupported)
+ - reductions:
+   - `sum` (unsupported)
+   - `extrema` (unsupported)
+   - `min` (unsupported)
+   - `max` (unsupported)
```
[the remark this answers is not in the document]
charleskawczynski: @juliasloan25 suggested to throw warnings since `sum` and `min` / `max` may still be useful.

docs/src/masks.md
```diff
@@ -0,0 +1,198 @@
...
+ - field constructors (`copy`, `Fields.Field`, `ones`, `zeros`) unsupported.
+   Not supporting masks for newly allocated fields is an intentional design
+   choice, because it's difficult to reason about what `ones` should return
+   where `mask == 0`-- should it return `NaN`? `1`? `0`? So, we chose to 
+   use `NoMask` when allocating new fields.
```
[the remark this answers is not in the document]
charleskawczynski: That's fine, but I'm fairly certain that this is less ergonomic-- users will likely need to use `fill!(parent(field), 0)` after creating fields with masks, to ensure that unmasked data is not full of garbage. Operations on subnormals and NaNs can be more expensive than ordinary numbers, so this could even impact performance (in FieldVector operations, not yet supported). As I mentioned, not all operations with masks can be supported, nor do they make sense.

IMO, we should strike a balance between ease of use and solving the primary pain point: the land model currently needs to prescribe nonsense land parameters over the ocean, because many operations will otherwise error. IMO, Implementing masks consistently with `zeros` will be less user-friendly. Any place where this is an issue, and users simply use `zeros(masked_space)` will require users to change this to
`z = zeros(masked_space); fill!(parent(z), 0)`-- which is itself an issue in that it reaches into internals (via `parent`).

All arrays in the space are mutable, so anything could be mutated and become invalid. I don't think we need to worry about this.

src/Grids/spectralelement.jl
```diff
@@ -137,18 +138,26 @@ mutable struct SpectralElementGrid2D{
...
 
 """
-    SpectralElementSpace2D(topology, quadrature_style; enable_bubble, autodiff_metric, horizontal_layout_type = DataLayouts.IJFH)
+    SpectralElementSpace2D(
+        topology,
+        quadrature_style;
+        enable_bubble,
+        autodiff_metric,
+        horizontal_layout_type = DataLayouts.IJFH
+        enable_mask::Bool,
```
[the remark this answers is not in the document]
charleskawczynski: I really don't see why the mask values need to be computed ahead of time. It's unlike every other property in the space. IMO, this is a cleaner way to express the mask, and honestly I feel that topography could be similarly improved. Injecting closures into the space constructor is difficult to reason about from within ClimaCore, and it's even more difficult to document.

Also, doing things this way would also require a dozen checks that the input type is compatible with the horizontal data layout type, which we internally guarantee with the `enable_mask`-- and that type cannot be mutated once the space is created.

I'd like a second opinion. @dennisYatunin, @juliasloan25, what do you think?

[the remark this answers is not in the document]
Sbozzolo: In my opinion, it is extremely important that the system doesn't silently return incorrect value because it using masked point somewhere. This would lead to very difficult to debug problems.

For this reason, I think that once a space is created with a mask, the mask should not be allowed to change (of course, one can hack its way into changing the mask by changing its parent array, but this is something we explicitely don't support, so it's user's responsability here). 

The system would allow doing something like this:
```
space = create_space()
space.set_mask(some_mask)
create_fields()
do_operations_with_fields()
space.set_mask(NoMask)  # Or any other mask that is a subset of the original one
# from here on, the fields would contain incorrect values
```

Having an immutable mask at constructor would also remove the possible failure mode where one forgets to set the mask and I would argue that makes reasoning about the code easier because the developer knows that once the mask is set, it can never be changed, and they don't have to do code tracing (the same reason Julia embraces immutability).


An intermediate solution would be to use the [builder pattern](https://refactoring.guru/design-patterns/builder), where one starts from a "base space" and builds more complex spaces out of it. It would look something like:
```julia
masked_space = with_mask(space, mask)
```
And you can imagine extending it with other properties
```julia
my_space = with_topography(with_mask(space, mask), topography)
```
In OOP languages, this looks a little cleaner
```python
my_space = space.add_mask(mask).add_topography(topography)
```

(However, as I understand it, topography is not a mere property of the space, but it is much more in the construction of the space itself, so I don't think it could be easily unbundled)

kmdeck:
- From a user point of view, I dont necessarily think the mask needs to be made in the same step as the space itself.  ClimaLand has already abstracted a lot of things from the user with respect to making our "domain" (topology...etc) , this would just be one other piece in those functions.  I also dont quite see how it is possible without making ClimaCore depend on ClimaUtilities (when the mask comes from data, and CU handles the interpolation).
- I agree with @Sbozzolo that we need to be careful that masked points arent used in computations, but for me my worry is more along the lines of someone using an unsupported function and the masked points affecting the results, rather than the mask being immutable and being reset. I think the only example is reductions (where the columns are no longer independent): tracking conserved quantities, `min` being used in implicit solver convergence criteria. This might be well documented here, but someone down the line working with ClimaLand wont know to check here, necessarily, theyll just assume masked points are not part of the picture at all
- I think it would be nice to show an example of how to set the mask from data (I think this is the more common use case, since we dont have analytic functions of coordinates for continents). Even in pseudo code or using a low dimensional space and a matrix of 1s and zeros.


I saw that the performance was mixed when using the mask vs not - looking into that further would be good! We saw a 2x speedup when we reduced the number of points in the simulation by half, so I think there is still a potential 3x speedup from this. Getting to that point first, and then addressing issues about the design would be my preference.

docs/src/masks.md
```diff
@@ -0,0 +1,197 @@
...
+mask.
+
+In addition, not all operations support masks. Here is a list of operations of
+supported and unsupported mask operations:
+
+ - `DataLayout` operations (`Fields.field_values(f) = 1`) unsupported (these operations are unaware of the mask and will likely never support masked operations).
+ - `fill!` (`@. f = 1`) supported
+ - point-wise `copyto!` (`@. f = 1 + z`) supported
+ - stencil `copyto!` (`@. ᶜf = 1 + DivergenceF2C()(Geometry.WVector(ᶠf))`) supported (vertical derivatives and interpolations interpolations)
+ - spectral element operations `copyto!` (`@. ᶜf = 1 + Operators.Divergence()(ᶠf)`) (unsupported, will execute everywhere)
```
kmdeck, line 159: I didnt quite get this example

src/DataLayouts/DataLayouts.jl
```diff
@@ -92,6 +92,41 @@ abstract type AbstractDispatchToDevice end
...
+"""
+abstract type AbstractMask end
+
+"""
+    NoMask
+
+A lazy non-mask type (default), used to indicate that an entire domain is
+present.
+"""
+struct NoMask <: AbstractMask end
```
kmdeck, line 108: Could you explain why we have both enable_mask and NoMask? It seems like then we have three options:'
- no mask enabled
- mask enabled but NoMask used (is this one needed?)
- mask enabled and mask used

charleskawczynski: `enable_mask` is user-facing, and turns on a mask for the space. `NoMask` (an internal object) is the default (when `enable_mask = false`), and `IJHMask` (also an internal object) is used when `enable_mask = true` for the appropriate space. If we want to support `SpectralElementSpace1D`, then we'll need to define `IHMask`.

docs/src/masks.md
charleskawczynski: Yeah, I was trying to fit everything on one line. Basically any spectral element operation. I can go and fine one and expand these sections to be more syntactically correct.

docs/src/masks.md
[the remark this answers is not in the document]
charleskawczynski: I applied masks to a bunch of the allocating methods, but here is what looks like an issue:

```julia
struct Field{V <: AbstractData, S <: AbstractSpace}
    values::V
    space::S
    # add metadata/attributes?
    function Field{V, S}(values::V, space::S) where {V, S}
        #TODOneed to enforce that the data size matches the space
        return new{V, S}(values, space)
    end
end
Field(values::V, space::S) where {V <: AbstractData, S <: AbstractSpace} =
    Field{V, S}(values, space)

Field(::Type{T}, space::S) where {T, S <: AbstractSpace} =
    Field(similar(Spaces.coordinates_data(space), T), space)
```

What is `Field(data, space)` supposed to do? If `space` has a mask, then this would fundamentally be a different operation-- the current method passes a pointer, but, if I understand correctly, @Sbozzolo is suggesting that this should only copy data where `mask == 1`, which means that passing a pointer is impossible (we would need to call `copyto!` on the field values). Also, `copy` calls this method, and `zeros` and `ones` call `copy`. So, I'm not sure that supporting allocating methods with masks can be done consistently, afterall.

IMO, this is all very difficult to reason about, and I still believe that a much simpler model to understand (and document) is: masked regions skip all pointwise and stencil broadcasting, and every other operation uses `NoMask`. Ideally, we could add support for reductions and fieldvector operations, but the behavior I sketched out should fix 90% of our issues related to prescribing nonsense land values over the ocean.

The push to support these other operations feels to me like your requesting that we design this perfectly-- and there is no perfect solution. Can we please iterate on this with something that works for our use case? There are many other issues that we're trying to fix, and I see no problem with what I had in https://github.com/CliMA/ClimaCore.jl/pull/2201/commits/4e1a097ed88f809b214d48edd63fdeee95c54ea0.

charleskawczynski:
I'm going to effectively revert to https://github.com/CliMA/ClimaCore.jl/pull/2201/commits/4e1a097ed88f809b214d48edd63fdeee95c54ea0 (I'll keep doc improvements and apply suggestions from @kmdeck).

charleskawczynski:
@Sbozzolo and I chatted and hashed through a few things. I've updated the docs and made a few things more clear. I've reverted to https://github.com/CliMA/ClimaCore.jl/commit/4e1a097ed88f809b214d48edd63fdeee95c54ea0 and fixed some terminology with the documentation.

docs/src/masks.md
```diff
@@ -143,30 +143,31 @@ c_no_mask = Fields.Field(FT, ᶜspace_no_mask)
...
+In addition, some operations with masked fields skip masked regions
+(i.e., mask-aware), and other operations execute everywhere
+(i.e., mask-unaware), effectively ignoring the mask. Here is a list of
+operations of mask-aware and mask-unaware:
+
+ - `DataLayout` operations (`Fields.field_values(f) = 1`) mask-unaware (will likely never be mask-aware).
+ - `fill!` (`@. f = 1`) mask-aware
+ - point-wise `copyto!` (`@. f = 1 + z`) mask-aware
+ - stencil `copyto!` (`@. ᶜf = 1 + DivergenceF2C()(Geometry.WVector(ᶠf))`) mask-aware (vertical derivatives and interpolations interpolations)
+ - spectral element operations `copyto!` (`@. ᶜf = 1 + Operators.Divergence()(ᶠf)`) mask-unaware
```
kmdeck, line 161: spectral element operations `copyto!` (`@. f = 1 + Operators.Divergence()(f)`), where `Operators.Divergence` carries out a divergence operation in horizontal directions.  mask-unaware

charleskawczynski:
Generic IO for datalayouts is not supported, and I don't want to hack in another implementation just for masks (i.e., duplicate what we have in `read_field`), so I'll followup with restarts for masks in a followup PR. (cc @Sbozzolo)

Also, that work will collide with https://github.com/CliMA/ClimaCore.jl/pull/2162, so it makes sense to split this up. I'll open an issue to track.

PR #2209, opened 2025-03-05
Try avoid reconstructing space in getidx

charleskawczynski:
This is an experimental PR, which tries to avoid reconstructing the space in `getidx`.

The basic idea is: under the assumption that compiling methods with big types is expensive, can we reduce the number of methods that we compile with big types (i.e., spaces)?

To achieve this, I think we need two pieces of information in the placeholder space from the original space:
 - is the vertical topology periodic?
 - number of levels in the space

### DIFF
--- a/src/MatrixFields/field_name_dict.jl
+++ b/src/MatrixFields/field_name_dict.jl
@@ -107,7 +107,10 @@ end
 function Operators.strip_space(dict::FieldNameDict)
     vals = unrolled_map(values(dict)) do val
         if val isa Fields.Field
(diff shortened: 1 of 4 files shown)


charleskawczynski:
This turned out to be a bit complicated because the space reconstruction actually changes the space used between center and faces, so using `axes(bc)` is not an option. So, we would still need some sort of reconstruction inside `getidx`.

Since #2284 had notable impact on latency, I think the right approach is to re-profile to see where the current bottleneck is.

PR #2210, opened 2025-03-05
Define divergence_result_type for missing case

### LINKED ISSUE #2208: bug in divergence_result_type
## Describe the bug

When calling the following function
```
function boundary_tendency_momentum(ᶜρ, ᶜuₕ, ρ_flux_uₕ)
    FT = eltype(ᶜρ)
    ᶠgradᵥ = Operators.GradientC2F()
    ᶜdivᵥ_uₕ = Operators.DivergenceF2C(
        top = Operators.SetValue(C3(FT(0)) ⊗ C12(FT(0), FT(0))),
        bottom = Operators.SetValue(ρ_flux_uₕ),
    )
    return @.  ᶜdivᵥ_uₕ((0 * ᶠgradᵥ(ᶜuₕ))) / ᶜρ
end
```
it results in an error
```
ERROR: LoadError: MethodError: no method matching divergence_result_type(::Type{ClimaCore.Geometry.AxisTensor{Float32, 2, Tuple{…}, StaticArraysCore.SMatrix{…}}})
The function `divergence_result_type` exists, but no method is defined for this combination of argument types.
```

Changing `@.  ᶜdivᵥ_uₕ((0 * ᶠgradᵥ(ᶜuₕ))) / ᶜρ` to `@.  lazy(ᶜdivᵥ_uₕ((0 * ᶠgradᵥ(ᶜuₕ))) / ᶜρ)` fixes the issue.

## To Reproduce
I'll try to make a minimal working example later.

charleskawczynski:
Attempt to close #2208

### DIFF
--- a/src/Geometry/conversions.jl
+++ b/src/Geometry/conversions.jl
@@ -538,7 +538,7 @@ Required for statically infering the result type of the divergence operation for
 # this isn't quite right: it only is true when the Christoffel symbols are zero
 @inline divergence_result_type(
     ::Type{Axis2Tensor{FT, Tuple{A1, A2}, S}},
-) where {FT, A1, A2 <: LocalAxis, S <: StaticMatrix{S1, S2}} where {S1, S2} =
+) where {FT, A1, A2 <: AbstractAxis, S <: StaticMatrix{S1, S2}} where {S1, S2} =
     AxisVector{FT, A2, SVector{S2, FT}}
 
 """


src/Geometry/conversions.jl
```diff
@@ -540,6 +540,14 @@ Required for statically infering the result type of the divergence operation for
...
 ) where {FT, A1, A2 <: LocalAxis, S <: StaticMatrix{S1, S2}} where {S1, S2} =
     AxisVector{FT, A2, SVector{S2, FT}}
+@inline divergence_result_type(
+    ::Type{AxisTensor{FT, 2, Tuple{A1, A2}, S}},
+) where {
+    FT,
+    A1 <: CovariantAxis{(3,)},
+    A2 <: CovariantAxis{(1, 2)},
+    S <: SMatrix{1, 2, FT, 2},
+} = Covariant12Vector{FT}
```
dennisYatunin, line 550:
```suggestion
) where {FT, A1, A2 <: AbstractAxis, S <: StaticMatrix{S1, S2}} where {S1, S2} =
    AxisVector{FT, A2, SVector{S2, FT}}
```

dennisYatunin (changes requested):
I think we can simplify this by extending previous code instead of defining new methods.

src/Geometry/conversions.jl
charleskawczynski: Sure, that's fine with me

dennisYatunin (approved):

PR #2212, opened 2025-03-06
Add restart support for masked spaces

### LINKED ISSUE #2216: Add restart support for masked spaces
We need to add restart support for masked spaces.

charleskawczynski:
Closes #2216

### DIFF
--- a/src/InputOutput/writers.jl
+++ b/src/InputOutput/writers.jl
@@ -321,6 +321,7 @@ end
 
 # Grids
 #
+defaultname(::DataLayouts.IJHMask) = "IJHMask"
 defaultname(::Grids.SpectralElementGrid1D) = "horizontal_grid"
 defaultname(::Grids.SpectralElementGrid2D) = "horizontal_grid"
 defaultname(::Grids.ExtrudedFiniteDifferenceGrid) =
@@ -374,6 +375,13 @@ function write_new!(
     )
     write_attribute(group, "bubble", grid.enable_bubble ? "true" : "false")
     write_attribute(group, "topology", write!(writer, Spaces.topology(grid)))
+    if !(grid.mask isa DataLayouts.NoMask)
+        write_attribute(
+            group,
+            "grid_mask",
+            write!(writer, grid.mask, Spaces.topology(grid)),
+        )
+    end
     return name
 end
 
@@ -450,6 +458,17 @@ function write_new!(
     return name
 end
 
+function write!(
+    writer::HDF5Writer,
+    mask::DataLayouts.IJHMask,
+    topology::Topologies.AbstractTopology,
+    name::AbstractString = defaultname(mask),
+)
+    group = create_group(writer.file, "grid_mask/$name")
+    write!(writer, group, mask.is_active, "is_active", topology)
+    return name
+end
+
 # write fields
 """
     write!(writer::HDF5Writer, field::Fields.Field, name::AbstractString)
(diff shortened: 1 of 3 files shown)


src/InputOutput/writers.jl
```diff
@@ -511,6 +531,118 @@ function write!(
...
+This is an internal method, meant to be used for writing data layouts to the
+HDF5 file.
+
+This method should be used when
+
+```julia
+!(topology isa Topologies.Topology2D && !(writer.context isa ClimaComms.SingletonCommsContext))
+```
+"""
+function _write_mpi!(
```
Sbozzolo, line 592: I haven't gone through the PR in details yet, but this caught my eye. When InputOutput writes other objects (including Fields and their associated data) it can directly handle the distributed case. Why is there a need to distinguish the MPI and non-MPI case here?

Also,
```
!(topology isa Topologies.Topology2D && !(writer.context isa ClimaComms.SingletonCommsContext))
```

The double negation makes this hard to understand. It seems that this says that the function should be used when the topology is not 2D and it is singleton, the opposite of what the function name suggests. Is this a typo?

Sbozzolo: Okay, I looked at the existing code and think I have answered my own question about the need for two functions: they are not really handled directly and there are still if-else statements

charleskawczynski: Yeah, I wish there was a better way, but you really need to know topology information to handle datalayout IO correctly. I split the method into two pieces in case either of them could be reused. For now, I'm only using these methods for the mask but we could eventually reuse them in other places to reduce some of the duplicate logic.

I think  happy to remove the double negative from the docs, I tend to avoid them, but I think I just did that to somewhat match the source code.

juliasloan25:
Could you add some notes in the PR description about the main changes being made here to support restarts with masked spaces? It would probably be good to add something like that to the NEWS too

src/InputOutput/writers.jl
```diff
@@ -374,6 +375,14 @@ function write_new!(
     )
     write_attribute(group, "bubble", grid.enable_bubble ? "true" : "false")
     write_attribute(group, "topology", write!(writer, Spaces.topology(grid)))
+    if !(grid.mask isa DataLayouts.NoMask)
+        write_attribute(
+            group,
+            "grid_mask",
+            write!(writer, grid.mask, Spaces.topology(grid)),
+        )
+        # write!(writer, grid.mask, Spaces.topology(grid))
```
Sbozzolo, line 384:
```suggestion
```

src/InputOutput/writers.jl
```diff
@@ -450,6 +459,17 @@ function write_new!(
...
 
+function write!(
+    writer::HDF5Writer,
+    mask::DataLayouts.IJHMask,
+    topology::Topologies.AbstractTopology,
+    name::AbstractString = defaultname(mask),
+)
+    group = create_group(writer.file, "grid_mask/$name")
+    write!(writer, group, mask.is_active, "is_active", topology)
+    name
```
Sbozzolo, line 470:
```suggestion
    return name
```

src/InputOutput/writers.jl
```diff
@@ -511,6 +531,118 @@ function write!(
...
 
+"""
+    write!(
+        writer::HDF5Writer,
+        values::DataLayouts.AbstractData,
+        name::AbstractString,
+        topology::Topologies.AbstractTopology,
+    )
+
+Write an object of type 'AbstractData' and name 'name' to the HDF5 file.
```
Sbozzolo, line 542:
```suggestion
Write an object of type `AbstractData` and name `name` to the HDF5 file.
```

src/InputOutput/writers.jl
```diff
@@ -511,6 +531,118 @@ function write!(
...
+        local_elem_gidx
+    )
+
+This is an internal method, meant to be used for writing data layouts to the
+HDF5 file.
+
+This method should be used when
+
+```julia
+!(topology isa Topologies.Topology2D && !(writer.context isa ClimaComms.SingletonCommsContext))
```
Sbozzolo, line 589: Is this restriction on Topology 2D needed because we don't support MPI for Topology1D?

Sbozzolo: Also, I think this conditional is incorrect for this case and should just be 
```
topology isa Topologies.Topology2D && !(writer.context isa ClimaComms.SingletonCommsContext)
```

src/InputOutput/writers.jl
```diff
@@ -511,6 +531,118 @@ function write!(
...
+        name::AbstractString,
+    )
+
+This is an internal method, meant to be used for writing data layouts to the
+HDF5 file.
+
+This method should be used when
+
+```julia
+!(topology isa Topologies.Topology2D && !(writer.context isa ClimaComms.SingletonCommsContext))
```
Sbozzolo, line 631: I find this hard to read (because of the double/nested negation). I'd suggest

```suggestion
is_distributed = !(writer.context isa ClimaComms.SingletonCommsContext)
!is_distributed || !(topology isa Topologies.Topology2D) 
```

src/InputOutput/writers.jl
```diff
@@ -511,6 +531,118 @@ function write!(
...
+```
+"""
+function _write!(group, values::DataLayouts.AbstractData, name::AbstractString;)
+    h_dim = DataLayouts.h_dim(DataLayouts.singleton(values))
+    array = parent(values)
+    # dataset = write_plain_array!(writer, array, "data/$name")
```
Sbozzolo, line 637:
```suggestion
```

test/InputOutput/unit_spectralelement2d.jl
```diff
@@ -67,31 +63,86 @@ end
...
+        end
+    end
+end
+
+@testset "restart test for 2D spectral element simulations" begin
     device = ClimaComms.device()
     @info "Using device" device
-    context = ClimaComms.SingletonCommsContext(device)
-    grid_topology = Topologies.Topology2D(context, mesh)
+    context = ClimaComms.context(device)
```
Sbozzolo, line 87: I think this test only works for SingletonCommsContext.

If you use MPI, `mktempfile` will create two different files on in the `/tmp/` of different nodes.

test/InputOutput/unit_spectralelement2d.jl
```diff
@@ -67,31 +63,86 @@ end
...
+    space = init_space(context; enable_bubble = true, enable_mask = true)
+    y0 = init_state.(Fields.local_geometry_field(space), Ref(parameters))
+    Y = Fields.FieldVector(y0 = y0)
 
-        # write field vector to hdf5 file
-        filename = tempname(pwd())
+    Spaces.set_mask!(space) do coords
+        rand() > 0.5
+    end
+    is_active₀ = Spaces.get_mask(space).is_active
```
Sbozzolo, line 122: This doesn't seem to be used anywhere

Sbozzolo (commented):
Thank you! I left minor comments

src/InputOutput/writers.jl
charleskawczynski: > Is this restriction on Topology 2D needed because we don't support MPI for Topology1D?

I think that is true, but I used this condition mostly to preserve the existing behavior

test/InputOutput/unit_spectralelement2d.jl
charleskawczynski: Ah, right, I think this is sort of a rebase artifact

src/InputOutput/writers.jl
charleskawczynski: I simplified the doc string.

src/InputOutput/writers.jl
charleskawczynski: It's been a while, but I think these comments came after the comments above, where you said [Okay, I looked at the existing code and think I have answered my own question about the need for two functions](https://github.com/CliMA/ClimaCore.jl/pull/2212#discussion_r1989523629), so I'm going to close. Please reopen if this is not the case.

test/InputOutput/unit_spectralelement2d.jl
charleskawczynski: This might be a really nice helper function in ClimaComms

charleskawczynski: Can you suggest how we do this here, temporarily?

Sbozzolo: Using temporary directories is a little tricky. If you create a directory with one MPI process, you have no guarantee that the other process sees the same directory (because it can take a little bit of time for the distributed filesystem to be synchronized across nodes)

What you can do instead is to create a temporary file in the current directory. This is not a problem because the creation of the file is handled by the HDF5+MPI library, which knows how to properly talk to the filesystem.

So, I think this would work
```julia
function mktempfile(f, context)
filename = ClimaComms.iamroot(context) ? tempname(pwd(); cleanup = true) : ""
filename = ClimaComms.bcast(context, filename)
f(filename)
end
```

charleskawczynski: I applied the suggestion, seems to be working.

Sbozzolo (approved):

PR #2229, opened 2025-03-11
Use new MultiplyColumnwiseBandMatrixField alias

charleskawczynski:
The `\cdot` alias for `MultiplyColumnwiseBandMatrixField` is a pain-- it collides with LinearAlgebra's `\cdot`, and it's exported by `ClimaCore.MatrixFields`.

This PR doesn't fix this issue per-say, but I would like to at least isolate where we are doing this and fix collisions for the ClimaCore examples.

I think we unfortunately can't remove the export from `ClimaCore.MatrixFields`, as this will technically be a breaking change. Perhaps we can deprecate this change and phase it out.

I spoke with @dennisYatunin, and he believes that we can replace all of these `\cdot`s with `*`. This PR applies these changes to all of the appropriate places, but instead replacing it with `\star`. This will allow us to do a replace-all once we are ready to apply this fix.

### DIFF
--- a/examples/hybrid/implicit_equation_jacobian.jl
+++ b/examples/hybrid/implicit_equation_jacobian.jl
@@ -2,6 +2,7 @@ import LinearAlgebra: ldiv!
 using ClimaCore: Spaces, Fields, Operators
 using ClimaCore.Utilities: half
 using ClimaCore.MatrixFields
+import ClimaCore.MatrixFields: ⋆
 using ClimaCore.MatrixFields: @name
 
 """
--- a/examples/hybrid/staggered_nonhydrostatic_model.jl
+++ b/examples/hybrid/staggered_nonhydrostatic_model.jl
@@ -331,11 +331,11 @@ function implicit_equation_jacobian!(j, Y, p, δtγ, t)
     #     norm_sqr(C123(ᶜuₕ) + C123(ᶜinterp(ᶠw))) / 2 =
     #     ACT12(ᶜuₕ) * ᶜuₕ / 2 + ACT3(ᶜinterp(ᶠw)) * ᶜinterp(ᶠw) / 2
     # ∂(ᶜK)/∂(ᶠw) = ACT3(ᶜinterp(ᶠw)) * ᶜinterp_matrix()
-    @. ∂ᶜK∂ᶠw = DiagonalMatrixRow(adjoint(CT3(ᶜinterp(ᶠw)))) ⋅ ᶜinterp_matrix()
+    @. ∂ᶜK∂ᶠw = DiagonalMatrixRow(adjoint(CT3(ᶜinterp(ᶠw)))) ⋆ ᶜinterp_matrix()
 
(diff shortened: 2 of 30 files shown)


Sbozzolo:
This also caused me a lot of troubles and hard-to-debug problems (see https://github.com/CliMA/ClimaCore.jl/pull/1771).

However, I think that the difference between ⋆ and * is way too small for ⋆ to be an acceptable symbol to use: it's extremely hard to distinguish and will cause a lot of confusion

charleskawczynski:
> This also caused me a lot of troubles and hard-to-debug problems (see #1771).

💯

> However, I think that the difference between ⋆ and * is way too small for ⋆ to be an acceptable symbol to use: it's extremely hard to distinguish and will cause a lot of confusion

> EDIT: I misread the PR. Using ⋆ is not the intended interface, so the problem won't be as big. That said, if this is internal, I would still lean towards a different symbol.

That's fair. I really don't have a preference on what we use, especially, but I went through the effort of replacing all of the places where we need to use a new symbol, while skipping places where we need to use LinearAlgebra's `\cdot`. I only care that it's uniquely different from `\cdot` and `*`, so that we can easily do a replace-all when we're ready to settle on something more final.

How about `◌`?

```julia
help?> ◌
"◌" can be typed by \dottedcircle<tab>
```

Sbozzolo:
> > This also caused me a lot of troubles and hard-to-debug problems (see #1771).
> 
> 💯
> 
> > However, I think that the difference between ⋆ and * is way too small for ⋆ to be an acceptable symbol to use: it's extremely hard to distinguish and will cause a lot of confusion
> 
> > EDIT: I misread the PR. Using ⋆ is not the intended interface, so the problem won't be as big. That said, if this is internal, I would still lean towards a different symbol.
> 
> That's fair. I really don't have a preference on what we use, especially since this is temporary. I went through the effort of replacing all of the places where we need to use a new symbol, while skipping places where we need to use LinearAlgebra's `\cdot`. I only care that it's uniquely different from `\cdot` and `*`, so that we can easily do a replace-all when we're ready to settle on something more final.
> 
> How about `◌`?
> 
> ```julia
> help?> ◌
> "◌" can be typed by \dottedcircle<tab>
> ```

Sounds good with me!

dennisYatunin:
◌ is extremely similar to Base's function composition operator, and it isn't generally an appropriate symbol for multiplication. I think it would be much simpler and clearer to use the standard multiplication symbol * instead, so I'll open a PR for that.

charleskawczynski:
> ◌ is similar to Base's function composition operator

Which operator is that?

dennisYatunin:
I was referring to the following operator:
```
julia> ∘
∘ (generic function with 3 methods)
```

I really don't see a need to export any new symbol (either ⋆ or ◌), when we could just use the standard *. Adding more symbols that are hard to distinguish is not the direction we want to go in, even if they are temporary. I have updated #2232 to retain the value of ⋅ exported in this PR, while also allowing us to switch ClimaAtmos/ClimaLand/ClimaCoupler to *, so let's use that PR instead.

charleskawczynski:
Superseded by #2232.

PR #2232, opened 2025-03-11
Use * to denote matrix multiplication

### LINKED ISSUE #1774: Hard-to-debug error when LinearAlgebra.cdot and MatrixFields.cdot are both imported
```julia
    n_elem_z = 2

    domain = Domains.IntervalDomain(Geometry.ZPoint(0.0), Geometry.ZPoint(1.0), boundary_names = (:bottom, :top))
    mesh = Meshes.IntervalMesh(domain, nelems = 2)
    space = Spaces.FaceFiniteDifferenceSpace(mesh)

    diverg = Operators.DivergenceC2F(; bottom = Operators.SetDivergence(0.0), top = Operators.SetDivergence(0.0))
    grad = Operators.GradientF2C()

    diverg_matrix = MatrixFields.operator_matrix(diverg)
    grad_matrix = MatrixFields.operator_matrix(grad)

    name = @name(u)
    jacobian = MatrixFields.FieldMatrix(
        (@name(u), @name(u)) => similar(zeros(space), ClimaCore.MatrixFields.TridiagonalMatrixRow{Float64}),
    )

    @. jacobian[name, name] = diverg_matrix() ⋅ grad_matrix() - (I,)
```

The error here is "Broadcasted spaces are not the same". The real issue is that `cdot` is not the correct function.

Tracking this problem down took a couple of hours of debugging, but it is something we can easily catch in ClimaCore (there is no instance where users want to use LinearAlgebra.dot when working with MatrixFields).

dennisYatunin:
This PR removes the alias `const ⋅ = MultiplyColumnwiseBandMatrixField()` and allows us to use the standard `*` symbol to denote matrix multiplication. All internal uses of `⋅` have been replaced with `*`, and the exported `⋅` symbol has been replaced with `const ⋅ = *`. This export is necessary to avoid breaking ClimaAtmos for now, but it should be removed before the next major release of ClimaCore. (The `⋅` symbol cannot be marked as deprecated because `@deprecate` only works for function calls.)

This PR supersedes #2229 and closes #1774.

### DIFF
--- a/src/MatrixFields/MatrixFields.jl
+++ b/src/MatrixFields/MatrixFields.jl
@@ -12,8 +12,8 @@ matrices, one for each column of the `Field`. Such `Field`s are called
 for them:
 - Constructors, e.g., `matrix_field = @. BidiagonalMatrixRow(field1, field2)`
 - Linear combinations, e.g., `@. 3 * matrix_field1 + matrix_field2 / 3`
-- Matrix-vector multiplication, e.g., `@. matrix_field ⋅ field`
-- Matrix-matrix multiplication, e.g., `@. matrix_field1 ⋅ matrix_field2`
+- Matrix-vector multiplication, e.g., `@. matrix_field * field`
+- Matrix-matrix multiplication, e.g., `@. matrix_field1 * matrix_field2`
 - Compatibility with `LinearAlgebra.I`, e.g., `@. matrix_field = (4I,)` or
     `@. matrix_field - (4I,)`
 - Integration with `RecursiveApply`, e.g., the entries of `matrix_field` can be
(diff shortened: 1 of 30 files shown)


src/MatrixFields/MatrixFields.jl
```diff
@@ -78,7 +78,10 @@ export DiagonalMatrixRow,
     QuaddiagonalMatrixRow,
     PentadiagonalMatrixRow
 export FieldVectorKeys, FieldMatrixKeys, FieldVectorView, FieldMatrix
-export FieldMatrixWithSolver, ⋅
+export FieldMatrixWithSolver
+
+export ⋅
+const ⋅ = * # TODO: Remove this in the next major release of ClimaCore.
```
charleskawczynski, line 84: This seems very dangerous to me. I do not want ClimaCore to export this symbol, and make it an alias for `*`.

charleskawczynski (changes requested):
Changing the meaning of `\cdot`, via an exported symbol seems like it could lead to a _lot_ of confusion.

I prefer that we start with #2229-- you can pick the symbol if you don't like either of my two suggestions. Then, we can deprecate the use of `\cstar`, or whatever symbol we decide, update ClimaAtmos and all dependencies accordingly.

Frankly, I would suggest that we simply stick to a new symbol altogether, but I'm also fine with eventually overloading things without exporting any symbols.

dennisYatunin:
@charleskawczynski, are you not also [exporting the ⋅ symbol](https://github.com/CliMA/ClimaCore.jl/blob/d720850be985cf3df9c9ce737c489fe1b622b000/src/MatrixFields/matrix_multiplication.jl#L202) in #2229? I was merely trying to recreate your PR without the introduction of yet another symbol. Regardless, this symbol is only being exported in order to avoid breaking ClimaAtmos until we make a major release of ClimaCore. In the interest of getting this PR merged in, I will revert `const ⋅ = *` to `const ⋅ = MultiplyColumnwiseBandMatrixField()` if that seems safer to you.

I am opposed to merging in #2229 as that introduces a new unnecessary symbol. We should stick to `*` for denoting multiplication, as that is clearly understandable for users of ClimaCore.

charleskawczynski:
> @charleskawczynski, are you not also [exporting the ⋅ symbol](https://github.com/CliMA/ClimaCore.jl/blob/d720850be985cf3df9c9ce737c489fe1b622b000/src/MatrixFields/matrix_multiplication.jl#L202) in #2229?

#2229 does still export `⋅`-- it's not changing the user interface. The main purpose of that PR was to be able to use `LinearAlgebra.⋅` and `MatrixFields.MultiplyColumnwiseBandMatrixField` in the same scope.

> I was merely trying to recreate your PR without the introduction of yet another symbol. Regardless, this symbol is only being exported in order to avoid breaking ClimaAtmos until we make a major release of ClimaCore.

Yes, but it's changing the meaning of the symbol, which can lead to a lot of confusion.

> In the interest of getting this PR merged in, I will revert `const ⋅ = *` to `const ⋅ = MultiplyColumnwiseBandMatrixField()` if that seems safer to you.

It does. Can we deprecate `MatrixFields.⋅` in this PR?

> I am opposed to merging in #2229 as that introduces a new unnecessary symbol. We should stick to `*` for denoting multiplication, as that is clearly understandable for users of ClimaCore.

That's fair, so long we have a path towards deprecating the old pattern. Why did we use `⋅` to begin with?

I think the right path here is to deprecate the existing pattern, and then (perhaps slowly) transition to a new pattern. Can we temporarily leave `const ⋅ = MultiplyColumnwiseBandMatrixField()` to avoid a breaking release, add support for using `*`, and `@deprecate` the use of `MatrixFields.⋅`?

dennisYatunin:
As we've discussed offline, it is not possible to deprecate the ⋅ symbol because `@deprecate` only works for function calls. I am currently trying out the ⋅ → * substitution enabled by this branch in CliMA/ClimaAtmos.jl#3693, and I am rerunning this branch's CI with `export ⋅` commented out. If both of these work, this should be good to go (though I'll add the export back in until we make a new major release of ClimaCore).

dennisYatunin:
All of CI passes with `export ⋅` commented out (except for the downstream ClimaAtmos and ClimaCoupler tests). I'll add the export back in now.

PR #2235, opened 2025-03-12
Updates to masked spaces

### LINKED ISSUE #2226: Masks don't inherit from broadcast expressions
Fields inherit masks from explicit allocations, like `zeros(masked_space)`, `copy(masked_space)`, or `similar(masked_space)`, but they don't (at the moment) from broadcast expressions, like `f = @. a + 1` where `axes(a)` is a masked space.

One reason this is not quite as trivial as the other cases is if we change the example to `f = @. a + b` where `axes(a)` is a masked space and `axes(b)` is not. I think that having default behavior here might be surprising.

I think, for spaces to match, the horizontal space _must_ match. So, perhaps we can just use `axes(bc)`, which uses `Base.Broadcast.combine_axes`, but we need to try this out and test it.

Alternatively, we could error here. I'll take a look.

charleskawczynski:
Closes #2226. This is somewhat of a peel off from #2212.

This PR:
 - initializes masks to true everywhere by default
 - Adds masks to the show method for spaces
 - Extracts the mask from the axes of the broadcasted objects

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -2330,8 +2330,17 @@ function ColumnMask(
     @assert Nq isa Integer
     @assert Nh isa Integer
     T = horizontal_layout_type
-    is_active = replace_basetype(T{FT, Nq}(Array{FT}, Nh), Bool)
-    Nijh = Nq * Nq * Nh
+    is_active = replace_basetype(T{FT, Nq}(DA{FT}, Nh), Bool)
+    parent(is_active) .= true
+    return IJHMask(is_active)
+end
+
+union_all_type(::Type{T}) where {T} = T.name.wrapper
+
+function IJHMask(is_active::Union{IJFH, IJHF})
+    DA = union_all_type(typeof(parent(is_active)))
+    (Ni, Nj, _, _, Nh) = size(is_active)
+    Nijh = Ni * Nj * Nh
     i_map = zeros(Int, Nijh)
     j_map = zeros(Int, Nijh)
     h_map = zeros(Int, Nijh)
--- a/test/Spaces/unit_spaces.jl
+++ b/test/Spaces/unit_spaces.jl
@@ -23,11 +23,15 @@ import ClimaCore:
     DeviceSideDevice
 
 using ClimaCore.CommonSpaces
+using ClimaCore.Utilities.Cache
 import ClimaCore.DataLayouts: IJFH, VF, slab_index
 
 on_gpu = ClimaComms.device() isa ClimaComms.CUDADevice
 
 @testset "2D spaces with mask" begin
+    # We need to test a fresh instance of the spaces, since
+    # masked spaces include data set by users.
+    Cache.clean_cache!()
     FT = Float64
     context = ClimaComms.context()
     x_max = FT(1)
@@ -57,6 +61,7 @@ on_gpu = ClimaComms.device() isa ClimaComms.CUDADevice
     hspace = Spaces.SpectralElementSpace2D(htopology, quad; enable_mask = true)
     mask = Spaces.get_mask(hspace)
     @test mask isa DataLayouts.IJHMask
+    @test all(x -> x == true, parent(mask.is_active)) # test that default is true
     Spaces.set_mask!(hspace) do coords
         coords.x > 0.5
     end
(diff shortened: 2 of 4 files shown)


src/DataLayouts/DataLayouts.jl
```diff
@@ -2330,8 +2330,17 @@ function ColumnMask(
     @assert Nq isa Integer
     @assert Nh isa Integer
     T = horizontal_layout_type
-    is_active = replace_basetype(T{FT, Nq}(Array{FT}, Nh), Bool)
-    Nijh = Nq * Nq * Nh
+    is_active = replace_basetype(T{FT, Nq}(DA{FT}, Nh), Bool)
+    @. is_active = true
```
kmdeck, line 2334: Is this the part that sets the mask = 1 everywhere if enable_mask = true?

charleskawczynski: Yes

test/Spaces/unit_spaces.jl
```diff
@@ -71,6 +72,9 @@ on_gpu = ClimaComms.device() isa ClimaComms.CUDADevice
     @. f = 1 + ᶜx * 0 # tests copyto!
     @test count(iszero, parent(f)) == 2
 
+    fbc = @. 1 + ᶜx * 0 # tests copy
```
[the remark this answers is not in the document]
charleskawczynski: No, I don't think so because we'll get a space mismatch, as you pointed out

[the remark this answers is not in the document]
juliasloan25: Maybe it's still a good idea to add a test to ensure we get the space mismatch error we're expecting we'll get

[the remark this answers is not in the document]
charleskawczynski: Added!

juliasloan25 (approved):
Thanks Charlie! Could you also test that an error is thrown if we try to do an operation on two masked fields which have different masks?

src/DataLayouts/DataLayouts.jl
```diff
@@ -2330,8 +2330,17 @@ function ColumnMask(
...
-    is_active = replace_basetype(T{FT, Nq}(Array{FT}, Nh), Bool)
-    Nijh = Nq * Nq * Nh
+    is_active = replace_basetype(T{FT, Nq}(DA{FT}, Nh), Bool)
+    @. is_active = 1
+    return IJHMask(is_active)
+end
+
+union_all_type(::Type{T}) where {T} = T.name.wrapper
+
+function IJHMask(is_active::Union{IJFH, IJHF})
```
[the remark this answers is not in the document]
Sbozzolo: Ah no, now I see it

src/DataLayouts/DataLayouts.jl
```diff
@@ -2330,8 +2330,17 @@ function ColumnMask(
...
+    DA = union_all_type(typeof(parent(is_active)))
```
Sbozzolo, line 2341: What is this function doing?

Sbozzolo (commented):
Could you update che commit/PR messages to be more informative?

charleskawczynski:
> Thanks Charlie! Could you also test that an error is thrown if we try to do an operation on two masked fields which have different masks?

This is redundant to the existing test, we compare spaces by `===`.

charleskawczynski:
> Could you update che commit/PR messages to be more informative?

Done

src/DataLayouts/DataLayouts.jl
```diff
@@ -2330,8 +2330,17 @@ function ColumnMask(
...
-    Nijh = Nq * Nq * Nh
+    is_active = replace_basetype(T{FT, Nq}(DA{FT}, Nh), Bool)
+    parent(is_active) .= true
+    return IJHMask(is_active)
+end
+
+union_all_type(::Type{T}) where {T} = T.name.wrapper
+
+function IJHMask(is_active::Union{IJFH, IJHF})
+    DA = union_all_type(typeof(parent(is_active)))
```
Sbozzolo, line 2341: The comment was about this function `union_all_type`. I don't understand what it does and why it's needed

charleskawczynski: Ah, sorry. `union_all_type` returns the `UnionAll` version of the given type. For example:

```julia
julia> union_all_type(::Type{T}) where {T} = T.name.wrapper
union_all_type (generic function with 1 method)

julia> struct Foo{A, B}
               a::A
               b::B
       end

julia> union_all_type(typeof(Foo(1,2)))
Foo
```
So, in this context, `typeof(parent(is_active))` might be `Array{Float64, 2}`, and `union_all_type(Array{Float64, 2})` will yield `Array` (this needs to work for CuArrays, too).

We've had to do this sort of hack, using internal julia types (`name.wrapper`), in a few places, as Julia doesn't (yet) have a user-way of handling this. In this particular circumstance, we might be able to write a helper to handle this, but I'd rather avoid coupling this to other parts of the code.

charleskawczynski: Also, I'll note that there are other circumstances where doing this is generally not possible without using internals (because we may not own the types that we want the UnionAll version of).

Sbozzolo: I see, thank you! I opened #2238 to add the examples you had here to documentation and to avoid having two identical functions with slightly different names in the codebase.

PR #2256, opened 2025-03-27
Sync CUDA streams before MPI calls

Sbozzolo:
I don't claim to fully understand why an explicit stream synchronization is required before a MPI call (maybe because MPI calls are executed in their own stream?), but starting from CUDA 5.7, `interpolate` fails if this is not done (under certain conditions, unsure which exactly).

Adding a synchronization was also advice I got when debugging a similar bug a few months ago from the MPI.jl folks.

Passing PR: https://buildkite.com/clima/climaatmos-gpulongruns/builds/526
(check the summary.pdf)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,14 @@ ClimaCore.jl Release Notes
 main
 -------
 
+
+v0.14.29
+-------
+
+ - Fix empty interpolation with certain versions of CUDA+MPI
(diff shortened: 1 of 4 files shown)


sriharshakandala:
Looks reasonable to me!

sriharshakandala (approved):

PR #2259, opened 2025-03-28
Update debugging docs

### LINKED ISSUE #2258: Improve DebugOnly docs for finding broadcast expressions with NaNs

charleskawczynski:
Closes #2258.

This PR updates the docs for the DebugOnly module.

I've made the example slightly more complicated, to demonstrate how to debug more complicated cases (e.g., a complex broadcast expression) by leveraging highlighted printing using [StructuredPrinting.jl](https://github.com/charleskawczynski/StructuredPrinting.jl).

### DIFF
--- a/docs/src/debugging.md
+++ b/docs/src/debugging.md
@@ -41,14 +41,15 @@ function ClimaCore.DebugOnly.post_op_callback(result, args...; kwargs...)
     end
 end
 ```
-If needed, `post_op_callback` can be specialized or behave differently in
-different cases, but here, it only checks if `NaN`s are in the given that.
 
-Note that, due to dispatch, `post_op_callback` will likely need a very general
-method signature, and using `post_op_callback
-(result::DataLayouts.VIJFH, args...; kwargs...)` above fails (on the CPU),
-because `post_op_callback` ends up getting called multiple times with different
-datalayouts.
+If needed, multiple methods of `post_op_callback` can be defined, but here, we
+define a general method that checks if `NaN`s are in the given object.
+
+Note that we need `post_op_callback` to be called for a wide variety of inputs
+because it is called by many many different functions with many different
+objects. Therefore, we recommend that you define `post_op_callback` with a very
+general method signature, like the one above and perhaps use `Infiltrator` to
+inspect the arguemtns.
 
 Now, let us put everything together and demonstrate a complete example:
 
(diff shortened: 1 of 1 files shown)


docs/src/debugging.md
```diff
@@ -109,11 +111,50 @@ ClimaCore.DebugOnly.print_depth_limited_stack_trace(st; maxtypedepth=1)
...
+
+```julia
+using StructuredPrinting
+import ClimaCore: DataLayouts
+highlight_nans(x::DataLayouts.AbstractData) = any(y->isnan(y), parent(x));
+highlight_nans(_) = false;
+bc = Infiltrator.safehouse.args[2]; # we know that argument 2 is the broadcasted object
+(; result) = Infiltrator.safehouse; # get the result
+@structured_print bc Options(; highlight = x->highlight_nans(x))
+```
```
imreddyTeja, line 134: Could this maybe be in an `@example` block? I think actually seeing the highlighting would be nice

imreddyTeja (approved):
I think this would be even better if it was a

docs/src/debugging.md
charleskawczynski: Sounds good, I think it's worth it. Let's try it out. I'll to rebase after #2257, which introduces a new dep.

docs/src/debugging.md
```diff
@@ -98,7 +98,9 @@ end
 
 FT = Float64
```
juliasloan25, line 99: I can't comment on that line, but there's a typo in line 42:
"If needed, `post_op_callback` can be specialized or behave differently in
different cases, but here, it only checks if `NaN`s are in the given that."
Maybe it should say "in the given object."?

juliasloan25: Also it might be helpful to give more details about what `post_op_callback` should look like, particularly what the arguments should be

juliasloan25 (approved):
This looks good! I had a couple comments about the more simple example, but the StructuredPrinting part looks nice. I agree it would be cool to have the colored highlighting render in the docs

docs/src/debugging.md
charleskawczynski: It's unfortunately difficult to generalize, because `post_op_callback` is called by many many different functions, so there is not an easy way to document what the arguments will be. I'll try to improve the docs on this.

charleskawczynski:
I'm going to merge this for now, and I'll update the code block to an example block once StructuredPrinting.jl has been moved to the clima org.

PR #2261, opened 2025-03-28
Move from using `GaussQuadrature` to `FastGaussQuadrature.jl`

sriharshakandala:
Move from using `GaussQuadrature` to `FastGaussQuadrature.jl`


sriharshakandala:
@szy21 , @akshaysridhar : Please advise if we are currently using the functions `orthonormal_poly` and `spectral_filter_matrix`!

Sbozzolo:
> @szy21 , @akshaysridhar : Please advise if we are currently using the functions `orthonormal_poly` and `spectral_filter_matrix` in `ClimaCore.jl`!

`cutoff_filter_matrix` calls `spectral_filter_matrix` and is currently used in the `bickleyjet_dg` example. As far as I can tell (with grep), that's the only occurence.

sriharshakandala:
> > @szy21 , @akshaysridhar : Please advise if we are currently using the functions `orthonormal_poly` and `spectral_filter_matrix` in `ClimaCore.jl`!
> 
> `cutoff_filter_matrix` calls `spectral_filter_matrix` and is currently used in the `bickleyjet_dg` example. As far as I can tell (with grep), that's the only occurence.

Yes. To be specific, do we need these functions in the future are can we remove them (the functions and the `bickleyjet_dg` example)!

sriharshakandala:
@Sbozzolo : Do we need to build the `benchmarks` and `lib/ClimaCoreTempestRemap` projects in the pipeline here: https://github.com/CliMA/ClimaCore.jl/blob/main/.buildkite/pipeline.yml#L34
We do not seem to be using them! cc: @nefrathenrici

Sbozzolo:
> @Sbozzolo : Do we need to build the `benchmarks` and `lib/ClimaCoreTempestRemap` projects in the pipeline here: https://github.com/CliMA/ClimaCore.jl/blob/main/.buildkite/pipeline.yml#L34 ? We do not seem to be using them! cc: @nefrathenrici

Good catch! We don't need them!

PR #2270, opened 2025-04-01
updating field solver

kmdeck:
Makes the linear solves mask aware

### DIFF
--- a/src/MatrixFields/single_field_solver.jl
+++ b/src/MatrixFields/single_field_solver.jl
@@ -1,3 +1,4 @@
+import ..DataLayouts
 dual_type(::Type{A}) where {A} = typeof(Geometry.dual(A.instance))
 
 inv_return_type(::Type{X}) where {X} = error(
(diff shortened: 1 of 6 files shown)


src/MatrixFields/single_field_solver.jl
```diff
@@ -54,10 +55,9 @@ function single_field_solve_diag_matrix_row!(
...
     b,
 )
-    Aⱼs = unzip_tuple_field_values(Fields.field_values(A.entries))
-    b_vals = Fields.field_values(b)
-    x_vals = Fields.field_values(x)
-    (A₀,) = Aⱼs
+    # Use fields, and not field values, here so that this
+    # operation is mask-aware.
+    A₀ = A.entries.:1
     @. x_vals = inv(A₀) ⊠ b_vals
```
kmdeck, line 61: @charleskawczynski I think this is a bug now, x_vals is not defined yet

kmdeck: nor b_vals?

charleskawczynski: Yep

PR #2272, opened 2025-04-01
Clean up convenience `interpolate`

Sbozzolo:
I don't know if I lost a commit or something in a previous PR, but
`Remapping.interpolate` was missing some methods needed to make it useful. In this commit, I go over the methods for all the spaces and add what is missing (e.g., `interpolate(horizontal_space, target_hcoords)` instead of `interpolate(horizontal_space, target_hcoords, nothing).

I also change the default behavior for `default_target_zcoords` to interpolate on levels instead of linear interpolation. I do this because this is a much more common type of workflow.

With this commit, plotting `Field`s becomes very easy. For example, plotting a surface values of a 3D field:
```julia
using CairoMakie
import ClimaCore.Remapping: interpolate
heatmap(interpolate(field)[:, :, 1])
```

This commit also adds the defaults to the constructors for the `Remapper` too, so that creating a `Remapper` becomes easier:

```julia
import ClimaCore.Remapping: Remapper
remapper = Remapper(space)
```

(If one is fine with the defaults otherwise they can still pass the coordinates)

In the future, it might be nice to incorporate this into a `Makie` extension.

### DIFF
--- a/src/Remapping/distributed_remapping.jl
+++ b/src/Remapping/distributed_remapping.jl
@@ -231,6 +231,9 @@ The `Remapper` is designed to not be tied to any particular `Field`. You can use
 
 If you want to quickly remap something, you can call directly `interpolate`.
 
+By default, [`default_target_zcoords`](@ref) [`default_target_hcoords`](@ref) are used to
+determine the coordinates.
+
 Keyword arguments
 =================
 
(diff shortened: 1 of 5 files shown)


src/Remapping/distributed_remapping.jl
```diff
@@ -922,13 +945,17 @@ end
...
+                   target_zcoords = default_target_zcoords(space; zresolution)
                    )
 
 Interpolate the given fields on the Cartesian product of `target_hcoords` with
-`target_zcoords` (if not empty).
+`target_zcoords`. For `Space`s without horizontal/vertical component, the relevant
+`target_*coords` are ignored.
+
+When `zresolution` is set to `nothing`, use the Space levels, otherwise perform linear
+interpolation.
```
charleskawczynski, line 958: In the `nothing` case, there is no vertical interpolation, correct? It might be more clear to explicitly say that.

src/Remapping/distributed_remapping.jl
```diff
@@ -990,115 +1017,45 @@ This can be used directly for plotting.
...
 end
 
-function default_target_hcoords(
-    space::Spaces.SpectralElementSpace1D;
-    hresolution = 180,
+function interpolate(
+    field::Fields.Field,
+    space::Spaces.FiniteDifferenceSpace;
+    target_zcoords,
+    kwargs..., # needed to capture the unused target_hcoords
```
charleskawczynski, line 1047: Should we explicitly capture `target_hcoords`? I think that, this way, mis-spelled kwargs are caught early

charleskawczynski: (same elsewhere)

Sbozzolo: Good idea

charleskawczynski (approved):
I left a comment on the docs and catching kwargs a bit more restrictively, but overall I think this looks fine.

Sbozzolo:
> I left a comment on the docs and catching kwargs a bit more restrictively, but overall I think this looks fine.

Thank you for the super quick review!

I updated the docstring to mention that with `nothing` the vertical interpolation is disabled and added explicit kwards.

PR #2279, opened 2025-04-04
Improve some doc strings

### LINKED ISSUE #2265: `coordinate_field` returns a reference to the coordinates, so modifying the resulting Field modifies the underlying space
```julia
using ClimaCore.CommonSpaces
using ClimaCore.Fields
space = ExtrudedCubedSphereSpace(;
           z_elem = 10,
           z_min = 0,
           z_max = 1,
           radius = 10,
           h_elem = 10,
           n_quad_points = 4,
           staggering = CellFace()
       )
field = Fields.coordinate_field(space)
fill!(field.z, NaN)

all(isnan, parent(space.grid.face_local_geometry.coordinates.z))
# true
```

I mean, I guess it makes sense considering how Fields are created, but I think this is dangerous. 

I personally often use `Fields.coordinate_field` as a starting point to create new Fields (in particular, when I want to evaulate an analytical function of the coordinates) and I wasn't expecting that modifying this Field would change the underlying space.

At minimum, this should be documented.

charleskawczynski:
Closes #2265.

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -314,7 +314,7 @@ end
 """
     zeros(space::AbstractSpace)
 
-Construct a field on `space` that is zero everywhere.
+Create a new field on `space` that is zero everywhere.
 """
 function Base.zeros(::Type{FT}, space::AbstractSpace) where {FT}
     field = Field(FT, space)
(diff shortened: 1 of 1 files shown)


src/Fields/Fields.jl
```diff
@@ -521,12 +522,9 @@ Base.@propagate_inbounds Base.setindex!(field::PointField, val) =
...
 
-Apply function `f` to populate
-values in field `field`. `f` must
-have a function signature with signature
-`f(::LocalGeometry[, args...])`.
-Additional arguments may be passed to
-`f` with `args`.
+Apply function `f` to populate values in field `field`. `f` must have a function
+signature with signature `f(::LocalGeometry[, args...])`. Additional arguments
+may be passed to `f` with `args`.
```
Sbozzolo, line 527: Could we add an example for this?

Like, using `f` to `set!`a sine wave

Sbozzolo (approved):

charleskawczynski:
Going to merge now, we can keep at it.

PR #2280, opened 2025-04-05
Define getidx_return_type

charleskawczynski:
This PR adds a new method, `getidx_return_type`, which returns the type of a `getidx` call, independent of many other arguments, like the space, idx etc.

This new method is effectively branchless (as it doesn't depend on the `Interior`, `LeftBoundaryWindow` or `RightBoundaryWindow`), which means that inference should be far cheaper on this method than actually calling `getidex`. This may be helpful for conversions and or type assertions, which could help improve latency.

The reason this new method is being added, however, is because I'd like to see if and how easy it would be for our shared memory to instead store the result of `getidx` calls from within the stencil methods. Caching in this way would make the population of shared memory nearly independent of the stencil, resulting in perhaps far simpler code and less duplication.

One downside of this approach is that it does not apply shared memory to the local geometry, which involves several reads.

To ensure that the code is correct, I've added the use of this new function to assert the return type of `getidx` from within the `apply_stencil!` method, which is basically harmless.

Since this PR is mostly additive, I'm going to merge so long it passes CI, as this will allow me to more easily experiment with this new shared memory design that I've described above.

If it doesn't work out, we can remove it.

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -164,6 +164,8 @@ struct RightBoundaryWindow{name} <: BoundaryWindow end
 
 An abstract type for finite difference operators. Instances of this should define:
 
+- [`getidx_return_type`](@ref)
+- [`stencil_return_type`](@ref)
 - [`return_eltype`](@ref)
 - [`return_space`](@ref)
 - [`stencil_interior_width`](@ref)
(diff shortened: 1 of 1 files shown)


Sbozzolo:
The downstream test for ClimaCoupler failed here, but was passing on the previous commit on `main`

PR #2285, opened 2025-04-12
Fix fill with mask

charleskawczynski:
This PR fixes an observed issue with the masked `fill!` implementation, which was incorrectly mask-unaware. (found by @kmdeck)


Sbozzolo (approved):

kmdeck:
thanks, @charleskawczynski !

Issue #2277, opened 2025-04-03
Support boolean datalayout indexing


charleskawczynski:
For Julia arrays, users can do things like:

```julia
julia> A = rand(3,3)
3×3 Matrix{Float64}:
 0.628506  0.704686   0.205372
 0.390219  0.192054   0.118027
 0.27103   0.0604534  0.297599

julia> A[A .> 0.5] .= 1
2-element view(::Vector{Float64}, [1, 4]) with eltype Float64:
 1.0
 1.0

julia> A
3×3 Matrix{Float64}:
 1.0       1.0        0.205372
 0.390219  0.192054   0.118027
 0.27103   0.0604534  0.297599
```

It would be really convenient for users if we could somehow support this with DataLayouts, specifically for `AbstractData{Bool}` fields.

This may be a bit a bit of work to support `getindex` and `setindex!`, since `setindex!` will require some sort of lazy view, containing the indices, but I think it's possible in principle.

This might be a nice complement to the space masks.

charleskawczynski:
I started sketching something, but I think we might want to draw inspiration from what Base is doing, and make sure we don't duplicate too much code since there will be different requirements on the gpu

```julia
@inline function Base.getindex(
    data::VIJFH,
    I::VIJFH{Bool},
)
    s = singleton(data)
    # fdim = field_dim(s) # get a single field to compute logical indices
    # Ipre = ntuple(i -> Colon(), Val(fdim - 1))
    # Ipost = ntuple(i -> Colon(), Val(ndims(data) - fdim))
    # V = view(parent(data), Ipre..., 1:1, Ipost...)
    # (inds, ) = Base.to_indices(V, (parent(I),)) # Base.LogicalIndex

    N = count(parent(I))
    v = similar(parent(data), eltype(data), N)
    i = 1
    for CI in CartesianIndices(data)
        v[i] = get_struct(
            parent(data),
            eltype(data),
            Val(field_dim(s)),
            CartesianIndex(to_data_specific(s, CI)),
        )
        i+=1
    end
    return v
end
```

charleskawczynski:
With the ClimaCore syntax, we could write this as `extrema(data[mask::IJHMask])`, which will forward into a `extrema(Base.getindex(parent(data), !parent(mask.is_active)))`.

We can also then define `Base.:~` on Boolean datalayouts to support `extrema(.~data[mask::IJHMask])`.

PR #2286, opened 2025-04-14
Add support for temporary work-around for mask-aware reductions

charleskawczynski:
This PR implements a function, `full_bitmask`, which gives users a temporary work-around to perform mask-aware reductions. For now, I've only added this support for `VIJFH` datalayouts, since this function should ultimately be deprecated in favor of using natively supported mask-aware reductions.

### DIFF
--- a/docs/src/masks.md
+++ b/docs/src/masks.md
@@ -169,6 +169,60 @@ operations of mask-aware and mask-unaware:
    This was a design implementation detail, users should not generally depend on the results where `mask == 0`, in case this is changed in the future. 
  - internal array operations (`fill!(parent(field), 0)`) mask-unaware.
 
+## Temporary work-arounds
+
+We can perform mask-aware reductions with the following work-around
+
+```julia
+using ClimaComms
+ClimaComms.@import_required_backends
+import ClimaCore: Spaces, Fields, DataLayouts, Geometry, Operators
+using ClimaCore.CommonSpaces
+using Test
+
+FT = Float64
+ᶜspace = ExtrudedCubedSphereSpace(FT;
+    z_elem = 10,
+    z_min = 0,
+    z_max = 1,
+    radius = 10,
+    h_elem = 10,
+    n_quad_points = 4,
+    staggering = CellCenter(),
+    enable_mask = true,
+)
+ᶠspace = Spaces.face_space(ᶜspace)
+ᶠcoords = Fields.coordinate_field(ᶠspace)
+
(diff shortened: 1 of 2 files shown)


Sbozzolo (approved):
Thank you!

Maybe we also need the case for 2D surface fields?

docs/src/masks.md
```diff
@@ -169,6 +169,45 @@ operations of mask-aware and mask-unaware:
...
+Spaces.set_mask!(ᶜspace) do coords
+    coords.lat > 0.5
+end
+
+# get the mask
+mask = Spaces.get_mask(ᶜspace)
+# make a field of ones
+ᶜf = ones(ᶜspace) # ignores mask
+
+@show sum(parent(ᶜf)[DataLayouts.full_bitmask(mask, Fields.field_values(ᶜf))])
```
kmdeck, line 207: Maybe we should point out that sum(field) = integral over volume... while sum(array) is the typical sum (no weighting by volume element)

kmdeck:
> Thank you!
> 
> Maybe we also need the case for 2D surface fields?

I agree! In our test cases where we check that the masked values are zero, we operate a lot on surface (2d) fields.

src/DataLayouts/DataLayouts.jl
```diff
@@ -2352,4 +2352,30 @@ function IJHMask(is_active::Union{IJFH, IJHF})
...
+!!! warn
+
+    This function provides users a work-around to compute mask-aware reductions,
+    and should be deprecated in favor of providing native masked-reduction
+    support. Therefore, this function should be used sparingly.
+
+    This feature is extensible, but not performant in that it allocates
+    and, on the gpu, will launch many kernels.
+"""
+function full_bitmask(mask::IJHMask, data::VIJFH)
```
kmdeck, line 2370: would it be possible to add change this to be 
```
full_bitmask(mask::IJHMask, data::VIJFH; complement = false)
```

and on line 2372:
```
arr = complement ? .~ parent(is_active) : parent(is_active)
```

docs/src/masks.md
charleskawczynski: Yeah, good point. I'll emphasize that `sum` on fields include weighting, `sum` on datalayouts do not, and neither are mask-aware.

charleskawczynski:
This is basically a crude step towards #2277.

kmdeck (approved):
thank you!

PR #2289, opened 2025-04-14
Add scalar_fieldmatrix

imreddyTeja:
Add a function to convert a FieldMatrix where each matrix entry has an eltype of some struct into a FieldMatrix where each entry has an eltype of a scalar.


src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,74 @@ function Base.one(matrix::FieldMatrix)
     return FieldNameDict(inferred_diagonal_keys, entries)
 end
 
+"""
+    get_field_first_index_offset(name::FieldName, ::Type{T}, ::Type{S})
+
+Returns the offset of the the field with name `name` in an object of type `S`
+in multiples of `sizeof(T)`.
+"""
+function get_field_first_index_offset(
```
charleskawczynski, line 286: I think it'd be good to write some unit tests for this.

imreddyTeja: Tests are now added

src/MatrixFields/field_name.jl
```diff
@@ -59,6 +59,24 @@ get_field(x, ::FieldName{()}) = x
...
+    broadcasted_get_field_type(
+        fieldtype(X, extract_first(name)),
+        drop_first(name),
+    )
+if hasfield(Method, :recursion_relation)
+    dont_limit = (args...) -> true
+    for m in methods(broadcasted_get_field_type)
+        m.recursion_relation = dont_limit
+    end
+end
```
charleskawczynski, line 78: Can you confirm that this is actually needed?

imreddyTeja: Yes. I just tried running the tests with this removed, and it results in a 
```
failed to optimize due to recursion: ClimaCore.MatrixFields.broadcasted_get_field_type
```
with `@test_opt`

dennisYatunin: Yes, it makes sense that this is necessary for types with depth >= 2. My only suggestion would be to move this to the bottom of the file, where I've disabled the recursion limits of a bunch of other functions. It's a little cleaner if they're all in one place.

src/MatrixFields/field_name_dict.jl
```diff
@@ -181,11 +181,51 @@ function get_internal_entry(
         entry
     elseif name_pair[2] == @name() && broadcasted_has_field(T, name_pair[1])
         # multiplication case 2 or 4, second argument
-        Base.broadcasted(entry) do matrix_row
-            map(matrix_row) do matrix_row_entry
-                broadcasted_get_field(matrix_row_entry, name_pair[1])
-            end
-        end # Note: This assumes that the entry is in a FieldMatrixBroadcasted.
+        target_field_eltype = broadcasted_get_field_type(T, name_pair[1])
+        if target_field_eltype <: Number
```
dennisYatunin, line 185:
```suggestion
        if sizeof(target_field_eltype) == sizeof(eltype(parent(entry))
```
This could potentially be incorrect if we have a `Number` with multiple fields (e.g., `Rational` or `Dual`). What we really want to check here is whether the `target_field_eltype` contains exactly one value from the parent array of `entry`.

src/MatrixFields/field_name_dict.jl
```diff
@@ -181,11 +181,51 @@ function get_internal_entry(
...
+            T_band = eltype(entry)
+            singleton_datalayout =
+                DataLayouts.singleton(Fields.field_values(entry))
+            # BandMatrixRow with same lowest diagonal and bandwidth as `entry`, with a scalar eltype
+            scalar_band_type = BandMatrixRow{
+                T_band.parameters[1],
+                T_band.parameters[2],
+                eltype(parent(entry)),
+            }
```
dennisYatunin, line 194:
```suggestion
            scalar_band_type =
                band_matrix_row_type(outer_diagonals(T_band)..., eltype(parent(entry)))
```
I think we should generally avoid accessing internals like this, since public API methods are easier to read/debug than `<type>.parameters`. If you think the suggested version is too clunky, though, feel free to add a new method to `MatrixFields` that computes this in a single function call.

src/MatrixFields/field_name_dict.jl
```diff
@@ -181,11 +181,51 @@ function get_internal_entry(
...
+            field_dim_size = DataLayouts.ncomponents(Fields.field_values(entry))
+            scalar_field_offset = get_field_first_index_offset(
+                name_pair[1],
+                target_field_eltype,
+                T,
+            )
+            band_element_size = Int(div(sizeof(T), sizeof(target_field_eltype)))
```
dennisYatunin, line 201:
```suggestion
            band_element_size = div(sizeof(T), sizeof(target_field_eltype))
```
`div` already returns an `Int`, so you shouldn't need to convert it

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,96 @@ function Base.one(matrix::FieldMatrix)
...
+    ::Type{S},
+) where {T, S}
+    if name == @name()
+        return 0
+    end
+    child_name = extract_first(name)
+    child_type = fieldtype(S, child_name)
+    remaining_field_chain = drop_first(name)
+    field_index =
+        unrolled_filter(i -> fieldname(S, i) == child_name, 1:fieldcount(S))[1]
```
dennisYatunin, line 298:
```suggestion
    field_index = unrolled_findfirst(==(child_name), fieldnames(S))
```
This can be simplified using `unrolled_findfirst`

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,96 @@ function Base.one(matrix::FieldMatrix)
...
+
+"""
+    get_scalar_keys(dict::FieldMatrix)
+
+Returns a `FieldMatrixKeys` object that contains the keys of all the scalar
+entries in the `FieldMatrix` `dict`.
+"""
+function get_scalar_keys(dict::FieldMatrix)
+    keys_tuple = unrolled_flatmap(keys(dict).values) do key
+        _, entry = unrolled_filter(pair -> key == pair[1], pairs(dict))[1]
```
dennisYatunin, line 317:
```suggestion
        entry = values(dict)[unrolled_findfirst(==(key), keys(dict))]
```
This can also be written in terms of `unrolled_findfirst`. Feel free to use whichever version you think is simpler

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,96 @@ function Base.one(matrix::FieldMatrix)
...
+        entry =
+            entry isa ColumnwiseBandMatrixField ? entry.entries.:(1) : entry
+        unrolled_map(
+            filtered_child_names(
+                field -> eltype(field) <: Number,
```
dennisYatunin, line 322:
```suggestion
                field -> sizeof(eltype(field)) == sizeof(eltype(parent(field))),
```
As in one of my other suggestions, this should check for equality in `sizeof` rather than for a specific subtype

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,96 @@ function Base.one(matrix::FieldMatrix)
...
+                entry,
+                @name()
```
dennisYatunin, line 324:
```suggestion
```
You don't need to specify this last argument, since it's set to `@name()` by default

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +277,96 @@ function Base.one(matrix::FieldMatrix)
...
+end
+
+"""
+    scalar_fieldmatrix(field_matrix::FieldMatrix)
+
+Constructs a `FieldNameDict` where the keys and entries are views
+of the entries of `field_matrix`, which corresponding to the
+scalar components of entries of `field_matrix`.
+
+# Example usage
```
dennisYatunin, line 340: I love the names in this example, but maybe it would be a bit clearer if it had names that are recognizable to users of ClimaAtmos, like `ρ` and `uₕ`? And maybe `foo` could be called something like `atmos_state`? This version is quite funny though so use whichever names you prefer.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,273 @@
...
+import ClimaCore.Utilities: half
+import ClimaCore.RecursiveApply: ⊠
+import ClimaComms
+import LinearAlgebra: I, norm, ldiv!, mul!
+import ClimaCore.MatrixFields: @name
+ClimaComms.@import_required_backends
+include("matrix_field_test_utils.jl")
+
+@testset "get_field_first_index_offset" begin
+    struct singleton{T}
```
dennisYatunin, line 14:
```suggestion
    struct Singleton{T}
```
Types should generally be capitalized. Same for `TwoFields` below.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,273 @@
...
+    ᶜᶜmat3_scalar_uₕ = ᶜᶜmat3 .* (e¹²,)
+    ᶜᶜmat1_scalar_uₕ = ᶜᶜmat1 .* (e¹²,)
+    ᶜᶠmat2_uₕ_u₃ = ᶜᶠmat2 .* (e¹² * e₃',)
+    ᶠᶠmat3_u₃_u₃ = ᶠᶠmat3 .* (e³ * e₃',)
+    ᶜᶜmat3_ρχ_scalar = map(Base.Fix1(map, Base.Fix2(⊠, ρχ_unit)), ᶜᶜmat3)
+    ᶜᶜmat3_ρaχ_scalar = map(Base.Fix1(map, Base.Fix2(⊠, ρaχ_unit)), ᶜᶜmat3)
+    ᶜᶠmat2_ρχ_u₃ = map(Base.Fix1(map, Base.Fix2(⊠, ρχ_unit ⊠ e₃')), ᶜᶠmat2)
+    ᶜᶠmat2_ρaχ_u₃ = map(Base.Fix1(map, Base.Fix2(⊠, ρaχ_unit ⊠ e₃')), ᶜᶠmat2)
+
+    A = MatrixFields.FieldMatrix(
```
dennisYatunin, line 166: You might want to move the definition of this `FieldMatrix` into `matrix_fields_test_utils.jl` so that we don't have so much duplicate code

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,218 @@
...
+
+    ρχ_unit = (; ρq_tot = 1, ρq_liq = 1, ρq_ice = 1, ρq_rai = 1, ρq_sno = 1)
+    ρaχ_unit =
+        (; ρaq_tot = 1, ρaq_liq = 1, ρaq_ice = 1, ρaq_rai = 1, ρaq_sno = 1)
+
+
+    ᶠᶜmat2_u₃_scalar = ᶠᶜmat2 .* (e³,)
+    ᶜᶠmat2_scalar_u₃ = ᶜᶠmat2 .* (e₃',)
+    ᶠᶠmat3_u₃_u₃ = ᶠᶠmat3 .* (e³ * e₃',)
+    ᶜᶠmat2_ρχ_u₃ = map(Base.Fix1(map, Base.Fix2(⊠, ρχ_unit ⊠ e₃')), ᶜᶠmat2)
```
dennisYatunin, line 149: All of this can probably also be moved into the definition of `dycore_prognostic_EDMF_FieldMatrix` to reduce code duplication.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,218 @@
...
+    @test all(
+        f -> f isa MatrixFields.UniformScaling || eltype(eltype(f)) <: FT,
+        getindex.(Ref(A), scalar_keys),
+    )
+    @test (@allocated MatrixFields.get_scalar_keys(A)) == 0
+    @test_opt MatrixFields.get_scalar_keys(A)
+    foreach(scalar_keys) do key
+        @test_opt A[key]
+        @test (@allocated A[key]) == 0
+    end
```
dennisYatunin, line 178:
```suggestion
    @test_all length(MatrixFields.get_scalar_keys(A)) > length(keys(A))
    for key in MatrixFields.get_scalar_keys(A)
        @test_all A[key] isa MatrixFields.UniformScaling ||
            eltype(eltype(A[key])) == eltype(parent(A[key]))
    end
```
This could be a little less wordy if you use `@test_all` instead of writing separate tests for allocations and instabilities.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,218 @@
...
+        key′, entry′ =
+            filter(pair -> MatrixFields.is_child_value(key, pair[1]), pairs(A))[1]
+        if key != key′
+            parent(entry) isa SubArray ||
+                parent(entry) === parent(entry′) ||
+                return false
+        end
+        parent(parent(entry)) === parent(entry′) || return false
+        return true
+    end
```
dennisYatunin, line 197:
```suggestion
    for (key, entry) in MatrixFields.scalar_fieldmatrix(A)
        @test_all A[key] === entry
    end
```
This looks like it's testing a bunch of internal implementation details, but all we really need to check is that the parent arrays are identical (which we can do by checking for physical equality of the entries themselves). Also, I'd recommend writing loops over separate tests instead of combining the loops into `@test all(...)`, since the short-circuiting logic is a bit hard to read.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,218 @@
...
+        entry_copy = A_copy[key]
+        entry_original = A[key]
+        entry_flattened = A_scalar[key]
+        entry_flattened isa MatrixFields.UniformScaling && return true
+        entry_original == entry_copy || return false
+        entry_original === entry_flattened || return false
+        entry_flattened .*= 2
+        entry_original == 2 .* entry_copy || return false
+        return true
+    end
```
dennisYatunin, line 217:
```suggestion
```
I think this follows trivially from the previous test, so we don't need a separate test for it.

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,218 @@
...
+import ClimaComms
+import ClimaCore.MatrixFields: @name
+ClimaComms.@import_required_backends
+include("matrix_field_test_utils.jl")
+
+@testset "get_field_first_index_offset" begin
+    struct Singleton{T}
+        x::T
+    end
+    struct Two_fields{T1, T2}
```
dennisYatunin, line 15:
```suggestion
    struct TwoFields{T1, T2}
```
We typically use camel case for type names

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +276,103 @@ function Base.one(matrix::FieldMatrix)
...
+entries in the `FieldMatrix` `dict`.
+"""
+function get_scalar_keys(dict::FieldMatrix)
+    keys_tuple = unrolled_flatmap(keys(dict).values) do key
+        entry = values(dict)[UnrolledUtilities.unrolled_findfirst(
+            isequal(key),
+            keys(dict).values,
+        )]
+        entry =
+            entry isa ColumnwiseBandMatrixField ? entry.entries.:(1) : entry
```
dennisYatunin, line 321: This is correct for a `ColumnwiseBandMatrixField` or a `UniformScaling`, but not for a `DiagonalMatrixRow` (the other type of `ScalingFieldMatrixEntry`). When we have a single tensor in a `FieldMatrix`, we need to extract a number from that tensor for `scalar_fieldmatrix`. You'll also need to modify the relevant method for `get_internal_entry` to do this, and you'll probably want to add/modify a block in `dycore_prognostic_EDMF_FieldMatrix` to test this functionality.

test/MatrixFields/scalar_fieldmatrix.jl
imreddyTeja: This matrix is also used in `test/MatrixFields/field_matrix_solvers.jl`, which also creates other FieldMatrices and FieldVectors using these variables. I left these outside of the function because I thought it could be confusing if the FieldMatrix created by calling `dycore_prognostic_EDMF_FieldMatrix` in `field_matrix_solvers.jl` is made from different matrices as the rest of them. Should I still move these into the function?

dennisYatunin: Yeah I think it's fine if it allocates a new set of Fields there. We'll probably want to use this in several other unit tests so it's simpler if we avoid duplicating all this setup code.

src/MatrixFields/field_name_dict.jl
```diff
@@ -152,18 +152,29 @@ get_internal_key(child_name_pair::FieldNamePair, name_pair::FieldNamePair) = (
 get_internal_entry(entry, name::FieldName, key_error) = get_field(entry, name)
 get_internal_entry(entry, name_pair::FieldNamePair, key_error) =
     name_pair == (@name(), @name()) ? entry : throw(key_error)
-get_internal_entry(
-    entry::ScalingFieldMatrixEntry,
+get_internal_entry(entry::UniformScaling, name_pair::FieldNamePair, key_error) =
+    if name_pair[1] == name_pair[2]
+        entry
+    elseif is_overlapping_name(name_pair[1], name_pair[2])
+        throw(key_error)
```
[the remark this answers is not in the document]
charleskawczynski: (in case this code is exercised on-device).

src/MatrixFields/field_name_dict.jl
```diff
@@ -152,18 +152,29 @@ get_internal_key(child_name_pair::FieldNamePair, name_pair::FieldNamePair) = (
...
+)
     if name_pair[1] == name_pair[2]
         entry
+    elseif name_pair[2] == @name() && has_field(entry, name_pair[1])
+        DiagonalMatrixRow(get_field(entry, name_pair[1]))
     elseif is_overlapping_name(name_pair[1], name_pair[2])
         throw(key_error)
     else
         zero(entry)
     end
```
imreddyTeja, line 176: @dennisYatunin I'm not sure if these conditionals are correct. If they are this can probably be re-combined into a single `get_internal_entry(::UniformScaling, ...)`

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +287,103 @@ function Base.one(matrix::FieldMatrix)
...
+        unrolled_map(
+            filtered_names(entry) do field
+                if field isa UniformScaling
+                    return true
+                elseif field isa Fields.Field
+                    return eltype(field) == eltype(eltype(field))
+                else
+                    return eltype(field) == typeof(field)
+                end
+            end,
```
charleskawczynski, line 339:
```suggestion
            filtered_names(entry) do field
                if field isa UniformScaling
                    true
                elseif field isa Fields.Field
                    eltype(field) == eltype(eltype(field))
                else
                    eltype(field) == typeof(field)
                end
            end,
```
I think using `return` inside `filter` / `map` is confusing here.

test/MatrixFields/matrix_field_test_utils.jl
```diff
@@ -134,6 +138,112 @@ function test_field_broadcast(;
...
+    b = Fields.FieldVector(;
+        sfc = sfc_vec .* ((; T = 1),),
+        c = ᶜvec .* ((; center_gs_unit..., sgsʲs = (center_sgsʲ_unit,)),),
+        f = ᶠvec .* ((; u₃ = e³, sgsʲs = ((; u₃ = e³),)),),
+    )
+    A = MatrixFields.FieldMatrix(
+        # GS-GS blocks:
+        (@name(sfc), @name(sfc)) => I,
+        (@name(c.ρ), @name(c.ρ)) => I,
+        (@name(c.ρe_tot), @name(c.ρe_tot)) => deepcopy(ᶜᶜmat3),
```
charleskawczynski, line 193: Why are we using deepcopy here?

test/MatrixFields/scalar_fieldmatrix.jl
```diff
@@ -0,0 +1,152 @@
...
+        ::Type{T},
+        ::Type{S},
+        expected_offset,
+    ) where {T, S}
+        @test_all MatrixFields.get_field_first_index_offset(name, T, S) ==
+                  expected_offset
+    end
+    test_get_field_first_index_offset(
+        @name(x),
+        Float64,
```
charleskawczynski, line 30: Can we set `FT` above from within `@testset "get_field_first_index_offset" begin` and use `FT` here?

src/MatrixFields/field_name_dict.jl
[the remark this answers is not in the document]
imreddyTeja: Ok. I'm going to change `throw` to `error` for all the other `throw(key_error)` occurrences in this file.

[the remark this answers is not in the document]
dennisYatunin: I didn't think this code will ever be called from the GPU, so that probably isn't necessary

test/MatrixFields/matrix_field_test_utils.jl
imreddyTeja: One of the previous tests for `scalar_fieldmatrix` needed each entry in the matrix to have different parents, but that was removed in the last commit. I'll change this to get rid of the deep copy

src/MatrixFields/field_name_dict.jl
dennisYatunin: I think this is slightly wrong; you want to check the entry's value `entry[0]` instead of `entry` itself, so your condition should be `has_field(entry[0], name_pair[1])` and your result value should be `get_field(entry[0], name_pair[1])`. It would be good to add a `DiagonalMatrixRow(::Axis2Tensor)` block to `dycore_prognostic_EDMF_FieldMatrix` to test this functionality.

You can also do a similar optimization for scalars here like you did in the method for `ColumnwiseBandMatrixField`, wrapping `get_field(entry[0], name_pair[1])` in a `UniformScaling` instead of a `DiagonalMatrixRow` when `get_field(entry[0], name_pair[1]) isa Number`.

And yes, it would be nice to combine this with the method for `UniformScaling` so that there is just one method for `ScalingFieldMatrixEntry`. I think the method for `DiagonalMatrixRow` will work for both types of scaling entries, so you can just remove the other method.

src/MatrixFields/field_name_dict.jl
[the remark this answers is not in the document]
charleskawczynski: It looks like `get_internal_entry` is called in `getindex` on a `FieldNameDict`, this is never called in a gpu kernel?
```julia
function Base.getindex(dict::FieldNameDict, key)
    key in keys(dict) || throw(KeyError(key))
    key′, entry′ =
        unrolled_filter(pair -> is_child_value(key, pair[1]), pairs(dict))[1]
    internal_key = get_internal_key(key, key′)
    return get_internal_entry(entry′, internal_key, KeyError(key))
end
```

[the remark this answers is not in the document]
dennisYatunin: We are currently only parallelizing over Fields, rather than FieldNameDicts. Though I suppose this could change in the future if we want to achieve greater levels of kernel fusion.

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +399,251 @@ function Base.one(matrix::FieldMatrix)
...
+    end
+end
+
+function bypass_adjoint(d::T) where {T}
+    if T <: Adjoint
+        return typeof(d.parent)
+    else
+        return typeof(d)
+    end
+end
```
charleskawczynski, line 445:
```suggestion
bypass_adjoint(x::Fields.Field) = bypass_adjoint(Fields.field_values(x))
bypass_adjoint(x::DataLayouts.AbstractData{<:Adjoint}) = eltype(x.parent)
bypass_adjoint(x::DataLayouts.AbstractData) = eltype(x)
bypass_adjoint(x::Adjoint) = typeof(x.parent)
bypass_adjoint(x) = typeof(x)
```
This might be easier to extend?

src/MatrixFields/field_name_dict.jl
```diff
@@ -428,23 +391,29 @@ if hasfield(Method, :recursion_relation)
...
+If x is a Field, return the type of the elements inside it, ignoring Adjoint wrappers.
+Otherwise return the type of x, also ignoring Adjoint wrappers.
+"""
+inner_type_ignore_adjoint(x::Fields.Field) =
+    inner_type_ignore_adjoint(Fields.field_values(x))
+inner_type_ignore_adjoint(x::DataLayouts.AbstractData{<:Adjoint}) =
+    eltype(x.parent)
+inner_type_ignore_adjoint(x::DataLayouts.AbstractData) = eltype(x)
+inner_type_ignore_adjoint(x::Adjoint) = typeof(x.parent)
+inner_type_ignore_adjoint(x) = typeof(x)
```
charleskawczynski, line 400: Should this also support when `x` is a broadcasted object that contains an Adjoint?

imreddyTeja: It currently is never used where `x` could be broadcasted, but maybe it should be added so this could be used elsewhere? Although  this should be used with caution, because it is a bit of a hack (it is only used with axis vectors)

dennisYatunin (approved):
Thank you Teja! We should eventually try to simplify the scalar indexing logic, as it's ended up very entangled with the AxisTensor internals, but I think you've captured every case we could possibly support. The new section in the docs is also helpful. This will be really useful for implementing the sparse autodiff Jacobian in ClimaAtmos.

charleskawczynski (approved):
It's nice to see that this works. My only feedback is that a lot of those functions are pretty big with seemingly lots of edge cases. So, I question how maintainable it is. That said, I'm fine with it if you are.

PR #2290, opened 2025-04-15
Add shmem support for InterpolateC2F

charleskawczynski:


charleskawczynski:
The performance tuning is in agreement with what me and @kmdeck found today, while looking at some nsight reports for the land. That should be somewhat easy. I'm going to merge this for now, since it seems like a further generalization for the shmem implementation.

PR #2291, opened 2025-04-15
Add performance tuning tests

charleskawczynski:
This PR:

 - adds some unit tests and some comments on some thread-block configurations that could be improved.
 - Slightly refactors the `masked_partition` so that we can more easily unit-test it
 - Renames `fd_stencil_partition` to `fd_shmem_stencil_partition`

It's unfortunate that https://github.com/maleadt/StaticCartesian.jl doesn't work. If it did, we could basically eliminate all of these functions and just use `linear_partition` + `CartesianIndices` with static ranges, and we would get fairly good performance for all of our kernels.

Instead, we occasionally get 2x slower with `linear_partition`, and need to carefully tune each of our `partition` methods.

Once we merge this, I can fix some of the configurations that result in the expensive surface kernels.

Also, it looks like we could even improve performance for the masked partition. So, that could give us a significant boost.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -137,6 +137,16 @@ steps:
         agents:
           slurm_gpus: 1
 
+      - label: "Unit: data cuda threadblocks"
+        key: unit_data_threadblock
+        command:
+          - "julia --project=.buildkite -e 'using CUDA; CUDA.versioninfo()'"
+          - "julia --color=yes --check-bounds=yes --project=.buildkite test/DataLayouts/unit_cuda_threadblocks.jl"
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+        agents:
+          slurm_gpus: 1
+
       - label: "Unit: data fill"
         key: gpu_unit_data_fill
         command:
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Note to self: I need to add this to buildkite.

kmdeck (approved):
As far as I can tell, this PR establishes tests which demonstrate the current behavior but doesnt really make any real changes to src code.

A future PR can then use these tests to see if the thread counts increase and prevent regressions

charleskawczynski:
> As far as I can tell, this PR establishes tests which demonstrate the current behavior but doesnt really make any real changes to src code.
> 
> A future PR can then use these tests to see if the thread counts increase and prevent regressions

Exactly. Now that the thread-block configuration is unit tested, we can fix the obvious issues and prevent future regressions.

This is one of the benefits of the refactoring that I did a while back (which ironically resulted in some of these performance regressions to start with). Previously, there were no `partition` functions to unit test.

Next, we can fix the regressions, which I've identified by inspection in the unit tests, and they should (hopefully) not crop up again.

On the technical side, we could hack in simple fixes, and that may actually be fine. But it'd be really nice if https://github.com/maleadt/StaticCartesian.jl worked as intended (see the first issue), as we could use that to fix them all in one go.

PR #2299, opened 2025-04-18
Refactor FD shmem index management

charleskawczynski:
This PR refactors the FD shmem so that, for example, changing the number of dimensions for shmem can be done in a more isolated section.


charleskawczynski:
I think that this PR is fine to merge. Going to tag some people for review.

imreddyTeja (commented):
Could you add a docstring for `ShmemIndex`? I'm a bit confused how `col_id` is intended to be used. Would this be to index into a 2d array of shape (n,m) where n*m=ncols per block?

dennisYatunin:
This shmem implementation was dropped in #2526, superseded by the changes in #2544.

PR #2304, opened 2025-04-21
Use ClimaCartesianIndices

charleskawczynski:
This PR adds the use of [ClimaCartesianIndices.jl](https://github.com/CliMA/ClimaCartesianIndices.jl), which can yield better [gpu performance](https://clima.github.io/ClimaCartesianIndices.jl/dev/#Results-(NVIDIA-A100)) for simple `copyto!` kernels.

This PR should move us from `perf_cart_index!(X, Y, CartesianIndices(...))`-like performance to `perf_cart_index!(X, Y, fast_ci(...))`-like performance for most datalayouts (all but those that end with fields, which exhibit `perf_linear_index!`-like performance).


dennisYatunin:
Would it be possible to run ClimaAtmos CI with this change and measure the effect on SYPD for GPU runs? The simple array example you've linked looks promising but it would be good to measure a real-world example.

charleskawczynski:
> My only concern is that `length(CI) > typemax(Int32)` might be type-unstable and trigger allocations. If you think that's fine then I'm okay with these changes.

Agreed. I changed it to a type assert and always use `FastCartesianIndices`. I don't think that we'll need more than 2 billion points per process (I'm not sure that we can even make a space that big)

charleskawczynski:
Let's just make this a direct dependency.

charleskawczynski:
I think these two builds are fair to compare (here I'm comparing the dry baro wave):

 - https://buildkite.com/clima/climaatmos-ci/builds/24041#01967d72-2013-4b68-916f-0b19dccbbf0a (main branch, sypd=68.483, P100)
 - https://buildkite.com/clima/climaatmos-ci/builds/24043#01967db1-4b79-4c59-aad5-794c3183d862 (this branch, sypd=78.787, P100)

We should probably be doing a comparison on the A100 (though I posted microbenchmarks on an A100 in the ClimaCartesianIndices docs), but it's at least in agreement with microbenchmarks and shows that there is some benefit for real-world cases.

cc @dennisYatunin, @Sbozzolo

Sbozzolo:
Here's a build on A100 with ClimaCore 0.14.30
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/418

1 GPU dry baro: 3.583
4 GPU dry baro: 10.728
1 GPU Moist HS: 1.856
4 GPU Moist HS: 5.616
1 GPU Aquaplanet: 0.694
2 GPU Aquaplanet: 1.358

Here's one with 0.14.31
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/419

1 GPU dry baro: 3.461
4 GPU dry baro: 10.564
1 GPU Moist HS: 1.801
4 GPU Moist HS: 5.745
1 GPU Aquaplanet: 0.503
2 GPU Aquaplanet: 1.005

Here's one with ClimaCartesianIndices:
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/420

1 GPU dry baro: 3.434
4 GPU dry baro: 10.646
1 GPU Moist HS: 1.755
4 GPU Moist HS: 5.519
1 GPU Aquaplanet: 0.504
2 GPU Aquaplanet: 1.005

akshaysridhar:
> I think these two builds are fair to compare (here I'm comparing the dry baro wave):
> 
>     * https://buildkite.com/clima/climaatmos-ci/builds/24041#01967d72-2013-4b68-916f-0b19dccbbf0a (main branch, sypd=68.483, P100)
> 
>     * https://buildkite.com/clima/climaatmos-ci/builds/24043#01967db1-4b79-4c59-aad5-794c3183d862 (this branch, sypd=78.787, P100)
> 
> 
> We should probably be doing a comparison on the A100 (though I posted microbenchmarks on an A100 in the ClimaCartesianIndices docs), but it's at least in agreement with microbenchmarks and shows that there is some benefit for real-world cases.
> 
> cc @dennisYatunin, @Sbozzolo

A100 estimates are much more reliable, I agree e.g. 68.48 SYPD in the linked branch above for the dry barowave problem, vs 83.905 SYPD  in the latest RRTMGP interface update PR 3786.

akshaysridhar:
https://gist.github.com/akshaysridhar/fca4d50242a8bcc65faa8289b03cdc8f shows the target-gpu timings across 4 older commits between 0.14.30 and 0.14.31.

charleskawczynski:
> Here's a build on A100 with ClimaCore 0.14.30 https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/418

The 0.14.31 also included [upgraded dependencies](https://github.com/CliMA/ClimaAtmos.jl/commit/4fdf7a46f5abbe459f27752a2f9baf55410dc9b5) for a bunch of other packages, including SciMLBase, GPUCompiler and a Tracy extension. Are you sure that this is a fair comparison? The tracy extension, which @kmdeck and I observed showed up in the nsight report, very well may have added overhead to those jobs. Did anyone look into this?

Reporting sypd across versions can help identify if regressions occur, but it doesn't explain why. Has anyone looked at the nsight reports? The build @Sbozzolo posted seems to have a bunch of canceled jobs, so it's not clear to me where those numbers are coming from. I'd suggest we report links to specific jobs with nsight reports, so that we can identify what exactly has slowed down.

akshaysridhar:
> > Here's a build on A100 with ClimaCore 0.14.30 https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/418
> 
> The 0.14.31 also included [upgraded dependencies](https://github.com/CliMA/ClimaAtmos.jl/commit/4fdf7a46f5abbe459f27752a2f9baf55410dc9b5) for a bunch of other packages, including SciMLBase, GPUCompiler and a Tracy extension. Are you sure that this is a fair comparison? The tracy extension, which @kmdeck and I observed showed up in the nsight report, very well may have added overhead to those jobs. Did anyone look into this?
> 
> Reporting sypd across versions can help identify if regressions occur, but it doesn't explain why. Has anyone looked at the nsight reports? The build @Sbozzolo posted seems to have a bunch of canceled jobs, so it's not clear to me where those numbers are coming from. I'd suggest we report links to specific jobs with nsight reports, so that we can identify what exactly has slowed down.

I think it is a reasonable comparison (w.r.t additional noise from unrelated commits I mean) the buildkite IDs I've reported are a sampling of commits between the two releases between which the degradation was noticed. I'm not aware of the Tracy extension but we can take a look at this shortly. It's possible that identifying issues from these updated dependencies does indeed help regain some performance improvements - but this [list](https://gist.github.com/akshaysridhar/fca4d50242a8bcc65faa8289b03cdc8f) attempts to simply modify the ClimaCore versions against `[b2c96348] ClimaAtmos v0.30.0` It seems that, with these updated dependencies listed above, disabling fd shmem helped recover a significant portion of the SYPD degradation.

charleskawczynski:
> It seems that, with these updated dependencies listed above, disabling fd shmem helped recover a significant portion of the SYPD degradation. (It's worth re-running an updated version of this branch against ClimaAtmos main in any case for reference - I'll start that job shortly).

Yeah, disabling shmem will improve performance for low vertical resolution jobs, but it will degrade performance for high-resolution jobs. The performance regression came from switching from a multi-dimensional launch configuration to a linear one + CartesianIndices. The shmem difference is orthogonal to versions 0.14.30 and 0.14.31.

akshaysridhar:
> > It seems that, with these updated dependencies listed above, disabling fd shmem helped recover a significant portion of the SYPD degradation. (It's worth re-running an updated version of this branch against ClimaAtmos main in any case for reference - I'll start that job shortly).
> 
> Yeah, disabling shmem will improve performance for low vertical resolution jobs, but it will degrade performance for high-resolution jobs. The performance regression came from switching from a multi-dimensional launch configuration to a linear one + CartesianIndices. The shmem difference is orthogonal to versions 0.14.30 and 0.14.31.

Noted, thanks. The strong-scaling jobs in the linked Gist above all use 63 layers (64 levels) so comparisons across those jobs are reasonable. What's considered low-res vertical resolution here - something like less than ~20 vertical elements ?

charleskawczynski:
It's really difficult to reason about the results of the gist.

charleskawczynski:
And yeah, 20 levels is low resolution

Sbozzolo:
> > It seems that, with these updated dependencies listed above, disabling fd shmem helped recover a significant portion of the SYPD degradation. (It's worth re-running an updated version of this branch against ClimaAtmos main in any case for reference - I'll start that job shortly).
> 
> Yeah, disabling shmem will improve performance for low vertical resolution jobs, but it will degrade performance for high-resolution jobs. The performance regression came from switching from a multi-dimensional launch configuration to a linear one + CartesianIndices. The shmem difference is orthogonal to versions 0.14.30 and 0.14.31.

Could you share builds where you see performance increase due to shared memory in an AMIP/Aquaplanet setup? 

What we observed is that disabling shared memory makes everything that is not baro wave/held suarzed faster. We see this acorss multiple pipelines, the nightly AMIP (39 vertical elements, 5 % faster without shared memory), the benchmark AMIP in ClimaCoupler (63 elements, 30 % faster without shared memory), the gpu_aquaplanet case in the atmos target gpu pipeline (63 elements, 20-30 % faster without shared memory).

charleskawczynski:
> Could you share builds where you see performance increase due to shared memory in an AMIP/Aquaplanet setup?

> What we observed is that disabling shared memory makes everything that is not baro wave/held suarez faster. We see this acorss multiple pipelines, the nightly AMIP (39 vertical elements, 5 % faster without shared memory), the benchmark AMIP in ClimaCoupler (63 elements, 30 % faster without shared memory), the gpu_aquaplanet case in the atmos target gpu pipeline (63 elements, 20-30 % faster without shared memory).

I'm more interested to know of any kernels that are slower with shmem.

dennisYatunin:
The performance issues that motivated these changes have not been seen since #2522, so using `ClimaCartesianIndices` is no longer necessary.

PR #2308, opened 2025-04-21
Fix allocations with pow

### LINKED ISSUE #1126: Broadcasting ^2 causes allocations
## Describe the bug
While working on https://github.com/CliMA/ClimaAtmos.jl/pull/1333, it became clear that using constant exponents in broadcasting (e.g. `x.^2`) caused allocations, where `abs2.(x)` did not.

charleskawczynski:
Closes #1126.

Using LazyBroadcast here was pretty convenient, because we could quickly get the broadcasted object to experiment.

### DIFF
--- a/src/Fields/broadcast.jl
+++ b/src/Fields/broadcast.jl
@@ -313,7 +313,7 @@ end
 Base.Broadcast.broadcasted(
     ::typeof(Base.literal_pow),
     ::typeof(^),
-    f::Field,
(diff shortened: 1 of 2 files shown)


charleskawczynski:
> Looks like the method you've added is identical to the one above it. Could you reduce code duplication by instead defining a single method for `Union{Field, Base.Broadcast.Broadcasted{<:AbstractFieldStyle}}`?

Agreed, updated.

PR #2315, opened 2025-04-28
Remove some soft-fails

charleskawczynski:
I noticed in a recent build that a handful of gpu jobs that were previously failing with `soft_fail: true` are now passing. I looked back, and I think they were fixed by #2284.

There's only 4 left, and I think that 2 of them may just need fixes in the test. The other two are probably compiler heuristics.


Sbozzolo (approved):

charleskawczynski:
Ah, bummer, it looks like at least some of these are flakey. Perhaps it would be good to generate the PTX (as recommended in the log) and file an issue?

PR #2324, opened 2025-05-13
Add `DebugOnly.allow_mismatched_spaces_unsafe`

Sbozzolo:
This commit adds an option to skip the consistency checks when working with `ClimaCore` spaces. As discussed in the documentation page, this is useful to enable an interactive "checkpointing" workflow with `deepcopy`.

This commit also adds a description of another debugging workflow to the documentation.

### DIFF
--- a/docs/src/debugging.md
+++ b/docs/src/debugging.md
@@ -1,20 +1,23 @@
 # Debugging
 
+`ClimaCore` comes with a set of debugging tools for a variety of situations.
+These tools are collected in the [`ClimaCore.DebugOnly`](@ref) modules and are
+presented in this page.
+
+## Finding where `NaN` arise
+
 One of the most challenging tasks that users have is: debug a large simulation
 that is breaking, e.g., yielding `NaN`s somewhere. This is especially complex
 for large models with many terms and implicit time-stepping with all the bells
 and whistles that the CliMA ecosystem offers.
 
-ClimaCore has a module, [`ClimaCore.DebugOnly`](@ref), which contains tools for
-debugging simulations for these complicated situations.
-
 Because so much data (for example, the solution state, and many cached fields)
-is typically contained in ClimaCore data structures, we offer a hook to inspect
-this data after any operation that ClimaCore performs.
+is typically contained in `ClimaCore` data structures, we offer a hook to
+inspect this data after any operation that `ClimaCore` performs.
 
-## Example
+### Example
 
-### Print `NaNs` when they are found
+#### Print `NaNs` when they are found
 
 In this example, we add a callback that simply prints `NaNs found` every
 instance when they are detected in a `ClimaCore` operation.
@@ -46,10 +49,10 @@ If needed, multiple methods of `post_op_callback` can be defined, but here, we
 define a general method that checks if `NaN`s are in the given object.
 
 Note that we need `post_op_callback` to be called for a wide variety of inputs
-because it is called by many many different functions with many different
-objects. Therefore, we recommend that you define `post_op_callback` with a very
-general method signature, like the one above and perhaps use `Infiltrator` to
-inspect the arguemtns.
+because it is called by many different functions with many different objects.
+Therefore, we recommend that you define `post_op_callback` with a very general
+method signature, like the one above and perhaps use `Infiltrator` to inspect
+the arguments.
 
 Now, let us put everything together and demonstrate a complete example:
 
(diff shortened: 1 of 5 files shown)


src/DebugOnly/DebugOnly.jl
```diff
@@ -98,4 +98,77 @@ function print_depth_limited_stack_trace(
...
+   @ ClimaCore.Fields ~/repos/ClimaCore.jl/src/Fields/broadcast.jl:227
+
+# Turning `allow_mismatched_spaces` on
+julia> ClimaCore.DebugOnly.allow_mismatched_spaces() = true;
+
+julia> one .+ other_one
+Float64-valued Field:
+  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
+"""
+function allow_mismatched_spaces()
```
charleskawczynski, line 170: If we're going to make this method user-facing, and being that this method opens the door for incorrect use, I'd suggest that we use the word `unsafe`, or similar, somewhere in the name

Sbozzolo: Okay

docs/src/debugging.md
```diff
@@ -72,14 +75,127 @@ ClimaCore.DebugOnly.call_post_op_callback() = false # hide
...
+import Infiltrator # must be in your default environment
+ClimaCore.DebugOnly.call_post_op_callback() = true
+function ClimaCore.DebugOnly.post_op_callback(result, args...; kwargs...)
+    has_nans = result isa Number ? isnan(result) : any(isnan, parent(result))
+    has_inf = result isa Number ? isinf(result) : any(isinf, parent(result))
+    @infiltrate has_nans || has_inf
+end
+```
+
+"Infiltrating" means being dropped into a new REPL where in the scope of the
```
charleskawczynski, line 135: Is this a precise way to describe "infiltrating"?

charleskawczynski: I just looked at the docs and they does seem to reflect this. The implementation seems a bit nebulous to me, though. I'm not exactly sure how this differs from `@exfiltrate`

docs/src/debugging.md
```diff
@@ -72,14 +75,127 @@ ClimaCore.DebugOnly.call_post_op_callback() = false # hide
...
+infil>
+```
+Here, we are dropped into a new REPL with full access to the variables in the scope where the `NaN` occurred. However, because of how `post_op_callback`, this is at a low level within `ClimaCore`, which is typically not useful. Hence, the next step is to type `@trace`, which prints out
+```julia
+[1] post_op_callback(::ClimaCore.DataLayouts.IJFH{…}, ::ClimaCore.DataLayouts.IJFH{…}, ::Vararg{…}; kwargs::@Kwargs{})
+    at REPL[40]:4
+[2] post_op_callback
+    at REPL[40]:1
+[3] copyto!
+    at /home/sbozzolo/repos/ClimaCore.jl/src/DataLayouts/copyto.jl:18
```
charleskawczynski, line 153: I tend to try to avoid leaving my personal footprint in code / documentation. Can / should we curate this directory/file?

charleskawczynski (approved):
Overall, I think this is fine. I think the docs example could be simpler and perhaps less physics based, but I don't think it's a big deal. The only other minor issue with the docs example is that it doesn't seem "performant". I know it's not important, but I tend to try to leave good patterns for users, because patterns can very easily spread in unintended ways.

I left a comment about fixing up directories in the docs, but I approve if this is addressed.

Sbozzolo:
> Overall, I think this is fine. I think the docs example could be simpler and perhaps less physics based, but I don't think it's a big deal. The only other minor issue with the docs example is that it doesn't seem "performant". I know it's not important, but I tend to try to leave good patterns for users, because patterns can very easily spread in unintended ways.

What do you mean it's not "performant"?

docs/src/debugging.md
Sbozzolo: I find infiltrating simpler than exfiltrating because there's no safehouse and no scope pollution. It also stops the evaluation of the program at the infiltration point, but you can continue/toggle the infiltration point, and add as many infiltration points as you want.

Sbozzolo:
> > What do you mean it's not "performant"?
> 
> These functions:
> 
> ```julia
> kinetic_energy(field) = field .* field ./ 2
> other_energy(field) = field ./ sum(field)
> offset_by_one(field) = field .- 1
> 
> function specific_energy(rho, P, u)
>     density_without_restmass = offset_by_one(rho)
>     return (kinetic_energy(u) .+ other_energy(P)) ./ density_without_restmass
> end
> 
> function renormalized_energy(rho, P, u)
>     energy = specific_energy(rho, P, u)
>     return energy ./ sum(energy)
> end
> ```
> 
> yield intermediate allocations.

Ah, okay. Yes. I can remove the first, but I will leave the second to avoid unnecessary complexity.

Sbozzolo:
> > > What do you mean it's not "performant"?
> > 
> > 
> > These functions:
> > ```julia
> > kinetic_energy(field) = field .* field ./ 2
> > other_energy(field) = field ./ sum(field)
> > offset_by_one(field) = field .- 1
> > 
> > function specific_energy(rho, P, u)
> >     density_without_restmass = offset_by_one(rho)
> >     return (kinetic_energy(u) .+ other_energy(P)) ./ density_without_restmass
> > end
> > 
> > function renormalized_energy(rho, P, u)
> >     energy = specific_energy(rho, P, u)
> >     return energy ./ sum(energy)
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > yield intermediate allocations.
> 
> Ah, okay. Yes. I can remove the first, but I will leave the second to avoid unnecessary complexity.

Actually no, the reason I had put `density_without_restmass` in a different field is to mimic allow one to inspect its value interactively with infiltrator. This way, it is closer to a real-world usage.

I'll add a note about the allocations

PR #2326, opened 2025-05-16
Refactor `promote_parent_array_type`

charleskawczynski:
The essence of the PR is that `promote_parent_array_type` returns a type, which includes a type, which includes the backing array, which is never used! We might actually reduce compile times by returning a type that is independent of the field size (thermo state has 5 field variables).

I found that these changes were needed when experimenting with KernelAbstractions in https://github.com/CliMA/ClimaAtmos.jl/pull/3655.


Sbozzolo (approved):

charleskawczynski:
I think that this was superseded by the `columnwise!` PR.

PR #2328, opened 2025-05-16
Implement a columnwise shmem operator

charleskawczynski:
This PR abstracts all but the physics from https://github.com/CliMA/ClimaAtmos.jl/pull/3655, and implements `columnwise!`.

This operator can be used to assign a large number of broadcast operations, all within the same kernel.

The idea is that we can use this function

 - once in `implicit_tendency!` (which will set all of the tendencies in a single kernel)
 - 2 or 3 times in `remaining_tendency!` (before / after dss, before / after horizontal terms)

Some preliminary results from the dry baro wave:

```
unfused & multiple broadcasts (set-precomputed + implicit): 4.31997 ms
single fused broadcast without shmem: 7.18 ms
single fused broadcast with shmem for state: 3.47 ms
single fused broadcast with shmem for state + LG: 4.73 ms
single fused broadcast with shmem for state + needed parts of LG: 3.93 ms (** though shmem is allocated for all of LG)
single fused broadcast with shmem for state + PartialLocalGeometry from #2332: 1.37 ms 🚀
```

For `ldiv!` and `wfact!`, I think we may need to rethink their implementations.

The only optimizations I've done in this operator is shared memory. There is probably a lot of room in tuning and improving memory access. A big benefit of this approach is that this will drastically reduce the number of kernel launches, and sequential memory reads.

This approach also helps push us towards https://github.com/CliMA/ClimaAtmos.jl/issues/3594.


charleskawczynski:
I think that one of the most encouraging aspects of this approach is that we likely won't have sharp performance characteristics due to compute, since all of it happens in shared memory.

charleskawczynski:
Not bad:

image

All estimated speedups are inside 10%. Un-coalesced memory reads is the highest estimated speedup at `9.98%`. While it'd be nice to improve on that, it might be difficult to attain if it's connected to the strides of the underlying data, and rearranging that could be difficult. There might be other ways we could tune it, but I think optimizing for the last 10% of peak performance is premature.

My recommendation is:
 - Merge #2332
 - I can then push the local changes with the additional needed fixes
 - Merge this new feature
 - We incrementally use this in atmos

charleskawczynski:
During today's meeting, there was some discussion on limitations and barriers, and how that impacts things. Specifically:

 - horizontal operations?
 - broadcasted level?
 - coriolis?

My short-answer response is: it'd be best if we can categorize the widest range of operators, to maximize the span of tendency computations, without barriers.

charleskawczynski:
These are the components that I found were needed in the local geometry:

```julia
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.coordinates.z[ᶜui] = ᶜlg.coordinates.z[ᶜui]) # needed
    is_valid_index(ᶠus, ᶠui) && (ᶠlg_col.coordinates.z[ᶠui] = ᶠlg.coordinates.z[ᶠui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.J[ᶜui] = ᶜlg.J[ᶜui]) # needed
    is_valid_index(ᶠus, ᶠui) && (ᶠlg_col.J[ᶠui] = ᶠlg.J[ᶠui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.invJ[ᶜui] = ᶜlg.invJ[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:1[ᶜui] = ᶜlg.gⁱʲ.components.data.:1[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:2[ᶜui] = ᶜlg.gⁱʲ.components.data.:2[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:3[ᶜui] = ᶜlg.gⁱʲ.components.data.:3[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:4[ᶜui] = ᶜlg.gⁱʲ.components.data.:4[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:5[ᶜui] = ᶜlg.gⁱʲ.components.data.:5[ᶜui]) # needed
    is_valid_index(ᶜus, ᶜui) && (ᶜlg_col.gⁱʲ.components.data.:6[ᶜui] = ᶜlg.gⁱʲ.components.data.:6[ᶜui]) # needed
    is_valid_index(ᶠus, ᶠui) && (ᶠlg_col.gⁱʲ.components.data.:9[ᶠui] = ᶠlg.gⁱʲ.components.data.:9[ᶠui]) # needed
```

charleskawczynski:
I spent about 10 minutes and improved the cpu performance by ~2x (via inlining), but it is still almost 10x slower than the cached version. There is still a lot of room on the cpu optimization front:

image

I think the right strategy here is to eliminate the large platforms that are basically not doing any compute. Once the flame graph is full of compute, we should probably look at vectorization (SIMD.jl ?), multithreading (and trying local memory with MArrays?).

That should make the performance not terrible compared to what it currently is on the cpu. Multithreading should be especially effective here, given the amount of work per thread that we'll get with these large kernels.

PR #2330, opened 2025-05-19
Add missing methods for column spaces and FieldVectors

dennisYatunin:
This PR adds several missing method definitions, which were discovered while working on CliMA/ClimaAtmos.jl#2730:
- `adapt_structure` was not defined for `ColumnGrid`, making vertical slices of 3D spaces incompatible with GPUs
- `ncolumns` was not defined for `FiniteDifferenceSpace`, making it hard to count columns of arbitrary spaces
- `ClimaComms.device` and `ClimaComms.context` were not defined for `FieldVector`, so that we had to extract specific `Field`s to get the device

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -459,6 +459,17 @@ import ClimaComms
 ClimaComms.array_type(x::FieldVector) =
     promote_type(unrolled_map(ClimaComms.array_type, _values(x))...)
 
+ClimaComms.device(x::FieldVector) = ClimaComms.device(ClimaComms.context(x))
+function ClimaComms.context(x::FieldVector)
+    isempty(_values(x)) && error("Empty FieldVector has no device or context")
+    # We don't have promotion for devices or contexts, so we use the first value
+    # that isn't a PointField (a PointField's data can be stored on a different
+    # device from other Fields to avoid scalar indexing on GPUs). If there is no
+    # such value, fall back to using the first PointField.
+    index = unrolled_findfirst(Base.Fix1(!isa, PointField), _values(x))
+    return ClimaComms.context(_values(x)[isnothing(index) ? 1 : index])
+end
+
 function __rprint_diff(
     io::IO,
     x::T,
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -30,7 +30,7 @@ import ..Utilities: PlusHalf, half
 using ..RecursiveApply
 using ClimaComms
 import Adapt
-import UnrolledUtilities: unrolled_map
+import UnrolledUtilities: unrolled_map, unrolled_findfirst
 
 import StaticArrays, LinearAlgebra, Statistics, InteractiveUtils
 
--- a/src/Grids/column.jl
+++ b/src/Grids/column.jl
@@ -36,6 +36,8 @@ struct ColumnGrid{
     colidx::C
 end
 
+Adapt.@adapt_structure ColumnGrid
+
 local_geometry_type(::Type{ColumnGrid{G, C}}) where {G, C} =
     local_geometry_type(G)
 
--- a/src/Spaces/finitedifference.jl
+++ b/src/Spaces/finitedifference.jl
@@ -103,6 +103,7 @@ function center_space(space::FiniteDifferenceSpace)
     return FiniteDifferenceSpace(grid(space), CellCenter())
 end
 
+ncolumns(::FiniteDifferenceSpace) = 1
 nlevels(space::FiniteDifferenceSpace) = length(space)
 # TODO: deprecate?
 Base.length(space::FiniteDifferenceSpace) = length(coordinates_data(space))


charleskawczynski (approved):
I'd suggest skipping the changes in [src/Fields/fieldvector.jl](https://github.com/CliMA/ClimaCore.jl/pull/2330/files#diff-a6811a052c442b6024962a45c82e3eae6e441211caef3c2fddef1cd8405a2d84), but the other changes are fine.

src/Fields/fieldvector.jl
```diff
@@ -443,6 +443,10 @@ import ClimaComms
 ClimaComms.array_type(x::FieldVector) =
     promote_type(unrolled_map(ClimaComms.array_type, _values(x))...)
 
+# We don't have promotion for devices or contexts, so just use the first value.
+ClimaComms.device(x::FieldVector) = ClimaComms.device(first(_values(x)))
+ClimaComms.context(x::FieldVector) = ClimaComms.context(first(_values(x)))
```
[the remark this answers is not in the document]
charleskawczynski: Yeah, so I'd recommend skipping this part.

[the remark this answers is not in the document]
dennisYatunin: Good catch! I assume that a `PointField` can be defined on CPUs to avoid scalar indexing on GPUs? I've modified this code to avoid using `PointField`s whenever possible.

dennisYatunin:
I've addressed your comment by modifying `ClimaComms.device` and `ClimaComms.context` to avoid using `PointField`s whenever a `FieldVector` contains anything other than a `PointField`. This should ensure that the functions work correctly across all possible uses of `FieldVector`.

charleskawczynski:
That’s a pretty complicated solution..

src/Fields/fieldvector.jl
```diff
@@ -459,6 +459,17 @@ import ClimaComms
...
+ClimaComms.device(x::FieldVector) = ClimaComms.device(ClimaComms.context(x))
+function ClimaComms.context(x::FieldVector)
+    isempty(_values(x)) && error("Empty FieldVector has no device or context")
+    # We don't have promotion for devices or contexts, so we use the first value
+    # that isn't a PointField (a PointField's data can be stored on a different
+    # device from other Fields to avoid scalar indexing on GPUs). If there is no
+    # such value, fall back to using the first PointField.
+    index = unrolled_findfirst(Base.Fix1(!isa, PointField), _values(x))
+    return ClimaComms.context(_values(x)[isnothing(index) ? 1 : index])
+end
```
charleskawczynski, line 471: Can we just not include this?

charleskawczynski: Not including this won't result in potential edge cases where this method simply can't be used (e.g., where the prognostic state is only a PointField). I really don't see what benefit that this method brings to the table.

dennisYatunin: If the prognostic state is only a PointField, then its device is that of the PointField, no? I don't see why you would want to use another device in that situation. If you're trying to access the default device of the machine, you would call `device()` instead of `device(::FieldVector)`.

The benefit is that we can access the device of any generic FieldVector, without needing to assume a specific structure. So, for example, we could access the device of any FieldVector from low-level packages like ClimaCore or ClimaTimeSteppers.

charleskawczynski: > If the prognostic state is only a PointField, then its device is that of the PointField, no?

I don't think that we can generally infer a device based only on a given field (specifically because of the edge cases for fields, like PointFields, that doe not contain device).

I think that this is an important place to enforce the Dependency Inversion Principle:

 - High-level modules should depend on abstractions, not concrete implementations
 - Invert the dependency flow to make code more flexible and testable
 - Utilize dependency injection and interfaces to decouple modules

Assuming that we can infer a device from a field is not part of the field abstraction, it's a concrete implementation (of only certain fields, which contain the device).

dennisYatunin: I was under the impression that the device is actually part of the Field abstraction. Every Field is stored on a unique device, which is determined by its underlying data and space. I was also under the impression that the device is part of the FieldVector abstraction. FieldVectors are types of arrays, which are inherently associated with specific devices. A small part of an array may be stored on a separate low-latency device (like in the case of a PointField), but the rest of it will always be stored on a single device.

If the only time a device cannot be inferred from a Field is when it's a PointField, we should clarify that as part of our interface. Also, if we are going to generalize my new autodiff work to arbitrary FieldVectors in ClimaCore (so that we can utilize the same implicit solver across ClimaAtmos, ClimaLand, etc), we need to define `device(::FieldVector)`. I mean, we could also just use the default `device()`, but that is much less efficient than inferring the device from a valid Field.

charleskawczynski: I forgot that we added the context back to the PointSpace:
```julia
struct PointSpace{
    C <: ClimaComms.AbstractCommsContext,
    LG <: DataLayouts.Data0D,
} <: AbstractPointSpace
    context::C
    local_geometry::LG
end
```
So, I think that this is fine.

charleskawczynski (approved):

PR #2331, opened 2025-05-20
Fix slicing of lazy Field broadcasts

dennisYatunin:
This PR fixes a bug that was discovered while working on CliMA/ClimaAtmos.jl#2730:
- Materializing `Fields.column(lazy.(<extruded_field_broadcast>), indices...)` allocates an extruded field instead of a column field
- Materializing `Fields.column(lazy.(<horizontal_field_broadcast>), indices...)` allocates a horizontal field instead of a point field

This is fixed by extending `DataColumnStyle` to horizontal layouts and defining a new `FieldColumnStyle`, which is used in place of the original `FieldStyle`. The same bug also exists for `Fields.slab`, so I've similarly added a `DataSlabStyle` and `FieldSlabStyle`. There was another bug for `Fields.slab` where lazy broadcasts that were sliced with a single index instead of two indices threw an error, so I've generalized `Fields.column` and `Fields.slab` for lazy broadcasts to handle arbitrary numbers of indices.

The unit tests for `Fields.column` have been extended to check lazy broadcasts and extruded spaces, and similar unit tests have been added for `Fields.slab`. While all of these tests pass on CPUs, some of the tests for `Fields.slab` fail on GPUs. It's not clear why this would be the case, but, since `Fields.slab` is not required for ClimaAtmos development, I'll just mark those tests as broken for now.

### DIFF
--- a/src/DataLayouts/broadcast.jl
+++ b/src/DataLayouts/broadcast.jl
@@ -21,37 +21,45 @@ struct VFStyle{Nv, A} <: DataColumnStyle end
 DataStyle(::Type{VF{S, Nv, A}}) where {S, Nv, A} =
     VFStyle{Nv, parent_array_type(A)}()
 DataColumnStyle(::Type{VFStyle{Nv, A}}) where {Nv, A} = VFStyle{Nv, A}
+Data0DStyle(::Type{VFStyle{Nv, A}}) where {Nv, A} = DataFStyle{A}
 
 abstract type Data1DStyle{Ni} <: DataStyle end
 struct IFHStyle{Ni, A} <: Data1DStyle{Ni} end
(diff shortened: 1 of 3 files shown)


charleskawczynski:
What was the error that you saw?

dennisYatunin:
@charleskawczynski, I've added unit tests to `test/Fields/unit_field.jl` that illustrate the bugs I mentioned in the PR description. These tests fail for the current main branch, but pass on this branch.

charleskawczynski (approved):
It looks like the broken test is unexpectedly passing, so you can update that.

I think the changes look fine to me. My only nit pick is that it might be clearer how this changes behavior by only adding to the test suite (assuming no changes) or adding to the test suite and updating tests that break (if there are behavioral changes).

PR #2332, opened 2025-05-20
Make an AbstractLocalGeometry type

charleskawczynski:
This PR:
 - makes an `AbstractLocalGeometry` type, and makes `LocalGeometry` a subtype of it
 - Updates method signatures to work with `AbstractLocalGeometry`
 - Defines `PartialLocalGeometry`, which we can use to reduce memory bandwidth and allocated shared memory in `columnwise!` kernels (https://github.com/CliMA/ClimaCore.jl/pull/2328)

This could impact `single fused broadcast with shmem for state + needed parts of LG: 3.93 ms (** though shmem is allocated for all of LG)` in #2328, as this may help increase the number of active blocks per SM.


charleskawczynski:
I just need to rebase and fix the docs git conflict. But this is otherwise ready (cc @dennisYatunin)

dennisYatunin (changes requested):
This seems to me like it introduces an unnecessary amount of complexity for what is currently an experimental feature. Instead of having two types of `LocalGeometry` with duplicate fields, I would recommend keeping a single type with fewer restrictions on the `∂ξ∂x` and `∂x∂ξ` fields.

From what I can see, this PR is effectively creating a duplicate `LocalGeometry` type with `∂ξ∂x` set to `nothing` and `∂x∂ξ` set to `(nothing, nothing, ..., nothing, Δz)`. So, you could simply remove the restrictions in the original `LocalGeometry` type and define a `simplified_local_geometry(::LocalGeometry)` function that replaces the `StaticArray`s in `∂ξ∂x` and `∂x∂ξ` with the simplified values `nothing` and `(nothing, nothing, ..., nothing, ∂x∂ξ[end])`. This would remove the need for any other changes in `ClimaCore` and `ClimaAtmos`.

charleskawczynski:
> This seems to me like it introduces an unnecessary amount of complexity for what is currently an experimental feature. Instead of having two types of `LocalGeometry` with duplicate fields, I would recommend keeping a single type with fewer restrictions on the `∂ξ∂x` and `∂x∂ξ` fields.

> From what I can see, this PR is effectively creating a duplicate `LocalGeometry` type with `∂ξ∂x` set to `nothing` and `∂x∂ξ` set to `(nothing, nothing, ..., nothing, Δz)`. So, you could simply remove the restrictions in the original `LocalGeometry` type and define a `simplified_local_geometry(::LocalGeometry)` function that replaces the `StaticArray`s in `∂ξ∂x` and `∂x∂ξ` with the simplified values `nothing` and `(nothing, nothing, ..., nothing, ∂x∂ξ[end])`. This would remove the need for any other changes in `ClimaCore` and `ClimaAtmos`.

I thought that this was actually a pretty simple refactor-- `LocalGeometry` doesn't change at all. The addition of the new type can be tested and doesn't need to be user-facing. I'd rather not entangle these changes with  #2328, but this PR (or something similar) is needed to better understand the performance that `columnwise!` might deliver.

Would

```julia
ᶜTS_lg = DataLayouts.typesize(FT, eltype(ᶜlg))
ᶜlg_arr = local_mem(device, FT, Val((ᶜNv, ᶜTS_lg)))
```
still work with what you suggest?

Rather than going through the whole process of changing this PR, I'll open a separate one with your suggestion, and I'll leave it up to you.

dennisYatunin:
Yeah I think `DataLayouts.typesize` should still work! As long as everything is concretely typed, `nothing` has size 0 and `(nothing, nothing, ..., nothing, ∂x∂ξ[end])` has the same size as `∂x∂ξ[end]`. I guess this strategy requires a few additional bytes of parameter memory, so if that's an issue we can totally go with the route of defining a new type. If you're just trying to replace some fields with low-storage altenatives without optimizing parameter memory, though, it seems much simpler to just expand the current struct definition instead of defining separate methods for different structs.

charleskawczynski:
>  I guess this strategy requires a few additional bytes of parameter memory, so if that's an issue we can totally go with the route of defining a new type.

I don't think that this is an issue with parameter memory, since it seems we can construct huge broadcasted objects on the device. But, I can't guarantee that #2335 won't change compile times due to increased type information in general. This PR will certainly not, as it leaves the type information unchanged.

charleskawczynski:
Superseded by #2335.

PR #2333, opened 2025-05-20
Fix 2329

charleskawczynski:
It seems that whatever https://github.com/JuliaLang/julia/pull/35948 is changing results in `Base.Broadcast.BroadcastStyle` getting called with a different order of arguments. I thought that Julia was supposed to swap and try both, but I don't recall where in the docs.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -155,16 +155,32 @@ Base.Broadcast.BroadcastStyle(
     as::Base.Broadcast.DefaultArrayStyle{0},
 ) = fs
 Base.Broadcast.BroadcastStyle(
+    as::Base.Broadcast.DefaultArrayStyle{0},
+    fs::FieldVectorStyle,
+) = fs
+Base.Broadcast.BroadcastStyle(
     fs::FieldVectorStyle,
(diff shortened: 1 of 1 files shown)


Sbozzolo (approved):

marius311:
Would it be possible for you to create a MWE showing where my PR makes it necessary to define both orders? That shouldn't be happening. The place in the docs I think you're thinking of is here: https://docs.julialang.org/en/v1/manual/interfaces/#writing-binary-broadcasting-rules

It could be that what you're seeing is just the need to define the Val constructor correctly masquerading as needing both orders.

charleskawczynski:
> Would it be possible for you to create a MWE showing where my PR makes it necessary to define both orders? That shouldn't be happening. The place in the docs I think you're thinking of is here: https://docs.julialang.org/en/v1/manual/interfaces/#writing-binary-broadcasting-rules
> 
> It could be that what you're seeing is just the need to define the Val constructor correctly masquerading as needing both orders.

I'm not actually sure where the error came from, as the reported stack trace was quite truncated. But if someone can report a stack trace, I can look into it if I have the bandwidth.

PR #2335, opened 2025-05-21
Allow for more flexibly constructed LocalGeometry

charleskawczynski:
Alternative to #2332.


src/Geometry/localgeometry.jl
```diff
@@ -18,28 +18,52 @@ struct LocalGeometry{I, C <: AbstractPoint, FT, S}
...
+
+full_lg_type(I, C, FT, S) = LocalGeometry{
+    I,
+    C,
+    FT,
+    Axis2Tensor{FT, Tuple{LocalAxis{I}, CovariantAxis{I}}, S},
+    Axis2Tensor{FT, Tuple{ContravariantAxis{I}, LocalAxis{I}}, S},
+    Axis2Tensor{FT, Tuple{ContravariantAxis{I}, ContravariantAxis{I}}, S},
+    Axis2Tensor{FT, Tuple{CovariantAxis{I}, CovariantAxis{I}}, S},
+}
```
dennisYatunin, line 38:
```suggestion
const FullLocalGeometry{I, C, FT, S} = LocalGeometry{
    I,
    C,
    FT,
    Axis2Tensor{FT, Tuple{LocalAxis{I}, CovariantAxis{I}}, S},
    Axis2Tensor{FT, Tuple{ContravariantAxis{I}, LocalAxis{I}}, S},
    Axis2Tensor{FT, Tuple{ContravariantAxis{I}, ContravariantAxis{I}}, S},
    Axis2Tensor{FT, Tuple{CovariantAxis{I}, CovariantAxis{I}}, S},
}
```
This is a little more consistent with how we typically write types.

src/Geometry/localgeometry.jl
```diff
@@ -18,28 +18,52 @@ struct LocalGeometry{I, C <: AbstractPoint, FT, S}
...
+    Jinv = inv(J)
+    gⁱʲ = ∂ξ∂x * ∂ξ∂x'
+    gᵢⱼ = ∂x∂ξ' * ∂x∂ξ
+    isapproxsymmetric(components(gⁱʲ)) || error("gⁱʲ is not symmetric.")
+    isapproxsymmetric(components(gᵢⱼ)) || error("gᵢⱼ is not symmetric.")
+    ∂x∂ξT = typeof(∂x∂ξ)
+    ∂ξ∂xT = typeof(∂ξ∂x)
+    gⁱʲT = typeof(gⁱʲ)
+    gᵢⱼT = typeof(gᵢⱼ)
+    return full_lg_type(I, C, FT, S)(
```
dennisYatunin, line 57:
```suggestion
    return LocalGeometry(
```
All of the type parameters should be inferable now.

src/Geometry/localgeometry.jl
```diff
@@ -55,7 +79,9 @@ struct SurfaceGeometry{FT, N}
     normal::N
 end
 
-undertype(::Type{LocalGeometry{I, C, FT, S}}) where {I, C, FT, S} = FT
+undertype(
+    ::Type{LocalGeometry{I, C, FT, ∂x∂ξT, ∂ξ∂xT, gⁱʲT, gᵢⱼT}},
+) where {I, C, FT, ∂x∂ξT, ∂ξ∂xT, gⁱʲT, gᵢⱼT} = FT
```
dennisYatunin, line 84:
```suggestion
undertype(::Type{LocalGeometry{I, C, FT}}) where {I, C, FT} = FT
```
The new type parameters aren't used here.

dennisYatunin (approved):
A few small syntax nitpicks, but overall this looks great! I think this is a very good alternative to #2332, as it allows us to set any part of the tensors in `LocalGeometry` to `nothing` based on the requirements of our kernels, without needing to define a new data structure for each variation.

src/Geometry/localgeometry.jl
charleskawczynski: I think all type parameters are needed, unless we use `<:`

charleskawczynski: I'll just use the `FullLocalGeometry` alias

src/Geometry/localgeometry.jl
```diff
@@ -18,28 +18,48 @@ struct LocalGeometry{I, C <: AbstractPoint, FT, S}
...
+    ∂x∂ξ::Axis2Tensor{FT, Tuple{LocalAxis{I}, CovariantAxis{I}}, S},
+) where {FT, I, S}
+    ∂ξ∂x = inv(∂x∂ξ)
+    C = typeof(coordinates)
+    Jinv = inv(J)
+    gⁱʲ = ∂ξ∂x * ∂ξ∂x'
+    gᵢⱼ = ∂x∂ξ' * ∂x∂ξ
+    isapproxsymmetric(components(gⁱʲ)) || error("gⁱʲ is not symmetric.")
+    isapproxsymmetric(components(gᵢⱼ)) || error("gᵢⱼ is not symmetric.")
+    return FullLocalGeometry{I, C, FT, S}(
```
charleskawczynski, line 53: I just went with `FullLocalGeometry` since it's more easily mapped.

PR #2338, opened 2025-05-26
Add localgeometry interface

charleskawczynski:
Thoughts on doing this, @dennisYatunin?


charleskawczynski:
I'll leave this open. We'll probably want to re-think the LocalGeometry design, if/when `columnwise!` starts getting more use. At that point, we'll want to get better interface control over LocalGeometry.

PR #2341, opened 2025-05-30
Try improving low vertical resolution cases

charleskawczynski:
An attempt to improve `columnwise!` performance for low vertical resolution.


charleskawczynski:
I recommend that someone take this over.

PR #2343, opened 2025-06-02
Remove DeviceIntervalTopology

charleskawczynski:
This PR removes the `DeviceIntervalTopology`, and instead adapts `IntervalTopology` to the GPU. This allows us to compute things like `z_min` / `z_max` on the device (which needs the vertical mesh, which the `DeviceIntervalTopology` does not have).

Consequently, I've updated the columnwise test to include `rayleigh_sponge_tendency_uₕ`, which includes that computation.

### DIFF
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -33,10 +33,8 @@ A tuple or vector of unique boundary names of a spatial domain.
 """
 function unique_boundary_names end
 
-struct IntervalDomain{CT, B} <: AbstractDomain where {
-    CT <: Geometry.Abstract1DPoint{FT},
-    B,
-} where {FT}
+struct IntervalDomain{CT, B} <:
+       AbstractDomain where {CT <: Geometry.Abstract1DPoint{FT}, B} where {FT}
     coord_min::CT
     coord_max::CT
 end
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -35,7 +35,7 @@ function unique_boundary_names end
 
 struct IntervalDomain{CT, B} <: AbstractDomain where {
     CT <: Geometry.Abstract1DPoint{FT},
-    B <: BCTagType,
+    B,
 } where {FT}
     coord_min::CT
     coord_max::CT
@@ -43,7 +43,7 @@ end
 
 isperiodic(::IntervalDomain{CT, B}) where {CT, B} = B == nothing
 unique_boundary_names(domain::IntervalDomain{CT, B}) where {CT, B} =
-    isperiodic(domain) ? () : unique(B)
+    isperiodic(domain) ? Symbol[] : unique(B)
 boundary_names(::IntervalDomain{CT, B}) where {CT, B} = B
 
 """
@@ -184,7 +184,7 @@ Base.show(io::IO, domain::SphereDomain) =
     print(io, nameof(typeof(domain)), ": radius = ", domain.radius)
 
 boundary_names(::SphereDomain) = ()
-unique_boundary_names(::SphereDomain) = ()
+unique_boundary_names(::SphereDomain) = Symbol[]
 coordinate_type(::SphereDomain{FT}) where {FT} = Geometry.Cartesian123Point{FT}
 
 end # module
--- a/src/Domains/Domains.jl
+++ b/src/Domains/Domains.jl
@@ -32,12 +32,12 @@ struct IntervalDomain{CT, B} <: AbstractDomain where {
 } where {FT}
     coord_min::CT
     coord_max::CT
-    boundary_names::B
 end
 
-isperiodic(domain::IntervalDomain) = isnothing(domain.boundary_names)
-boundary_names(domain::IntervalDomain) =
-    isperiodic(domain) ? () : unique(domain.boundary_names)
+isperiodic(::IntervalDomain{CT, B}) where {CT, B} = B == nothing
+boundary_names(domain::IntervalDomain{CT, B}) where {CT, B} =
+    isperiodic(domain) ? () : unique(B)
+boundary_names_type(::IntervalDomain{CT, B}) where {CT, B} = B
 
 """
     IntervalDomain(coord⁻, coord⁺; periodic=true)
@@ -60,7 +60,13 @@ function IntervalDomain(
             ),
         )
     end
-    IntervalDomain(promote(coord_min, coord_max)..., boundary_names)
+    c = promote(coord_min, coord_max)
+    boundary_names = if isnothing(boundary_names)
+        boundary_names
+    else
+        Tuple(boundary_names)
+    end
+    IntervalDomain{eltype(c), boundary_names}(c...)
 end
 IntervalDomain(coords::IntervalSets.ClosedInterval; kwargs...) =
     IntervalDomain(coords.left, coords.right; kwargs...)
@@ -95,7 +101,7 @@ function print_interval(io::IO, domain::IntervalDomain{CT}) where {CT}
     if isperiodic(domain)
         print(io, "(periodic)")
     else
-        print(io, domain.boundary_names)
+        print(io, boundary_names_type(domain))
     end
 end
 function Base.show(io::IO, domain::IntervalDomain)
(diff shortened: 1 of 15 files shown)


src/InputOutput/writers.jl
```diff
@@ -166,10 +166,10 @@ function write_new!(
     )
     write_attribute(group, "coord_min", Geometry.component(domain.coord_min, 1))
     write_attribute(group, "coord_max", Geometry.component(domain.coord_max, 1))
-    !isnothing(domain.boundary_names) && write_attribute(
+    !isnothing(Domains.boundary_names_type(domain)) && write_attribute(
```
imreddyTeja: Doesn't `boundary_names_type(domain)` return a type? Should this be `boundary_names_type(domain) <: Nothing`?

src/InputOutput/writers.jl
```diff
@@ -166,10 +166,10 @@ function write_new!(
...
         group,
         "boundary_names",
-        [String(bname) for bname in domain.boundary_names],
+        [String(bname) for bname in Domains.boundary_names_type(domain)],
```
imreddyTeja: Should this be `Domains.boundary_names(domain)`?

src/Domains/Domains.jl
```diff
@@ -32,12 +32,12 @@ struct IntervalDomain{CT, B} <: AbstractDomain where {
...
-    boundary_names::B
 end
 
-isperiodic(domain::IntervalDomain) = isnothing(domain.boundary_names)
-boundary_names(domain::IntervalDomain) =
-    isperiodic(domain) ? () : unique(domain.boundary_names)
+isperiodic(::IntervalDomain{CT, B}) where {CT, B} = B == nothing
+boundary_names(domain::IntervalDomain{CT, B}) where {CT, B} =
+    isperiodic(domain) ? () : unique(B)
+boundary_names_type(::IntervalDomain{CT, B}) where {CT, B} = B
```
imreddyTeja: It seems like these treat `B` in `IntervalDomain{CT, B})` as a tuple of symbols or `nothing`, but `B` is constrained to `B <: BCTagType`, so won't `B` always be a type?

src/InputOutput/writers.jl
charleskawczynski: It returns the value in the type parameter, so maybe `boundary_names_type_values` would be a better name?

src/Domains/Domains.jl
charleskawczynski: Indeed, I'm not actually sure how this worked. Or perhaps it's not working. Also, the atmos run is breaking so I need to see what's wrong there.

src/InputOutput/writers.jl
charleskawczynski: They're basically the same thing, assuming the boundary names are not nothing. `boundary_names_type` returns a tuple and `boundary_names` will return a vector because it calls `unique`, and (annoyingly) `unique(::Tuple) -> Vector`. We could use TupleTools.jl, but I just thought maybe better to avoid adding a new dependency just for this.

src/Domains/Domains.jl
imreddyTeja: I can't think of a good name for this function. Maybe `boundary_names_value_type` or `boundary_names_type_values` or  `boundary_names_param`.

I think it is only a little confusing because `boundary_names` also returns values, so I would expect `boundary_names_type` to return the typeof whatever `boundary_names` returns.

src/Domains/Domains.jl
```diff
@@ -32,12 +32,12 @@ struct IntervalDomain{CT, B} <: AbstractDomain where {
...
     coord_max::CT
-    boundary_names::B
 end
 
-isperiodic(domain::IntervalDomain) = isnothing(domain.boundary_names)
-boundary_names(domain::IntervalDomain) =
-    isperiodic(domain) ? () : unique(domain.boundary_names)
+isperiodic(::IntervalDomain{CT, B}) where {CT, B} = B == nothing
+boundary_names(domain::IntervalDomain{CT, B}) where {CT, B} =
+    isperiodic(domain) ? () : unique(B)
```
imreddyTeja: should this be 
`isperiodic(domain) ? Symbol[] : unique(B)`
for consistency of return type?

charleskawczynski: Yeah, that's probably better-- it's still technically type stable in that the result of `isperiodic` is compile-time known, but I still don't like that the types differ. I'll try switching it (hopefully no logic depends on that).

src/Domains/Domains.jl
charleskawczynski: I've removed the type restriction from the struct, as it doesn't really make sense with `B` now in the type domain.

src/Domains/Domains.jl
charleskawczynski: Applied, and it seems to work

charleskawczynski:
Ok, this is now working in ClimaCore CI, and ClimaAtmos.

As this PR developed, I came to better understand the trade-offs. On the one hand, this will allow users to write things like 

```julia
function rayleigh_sponge_tendency_uₕ(ᶜuₕ, s)
    s isa Nothing && return NullBroadcasted()
    ᶜz = Fields.coordinate_field(axes(ᶜuₕ)).z
    ᶠz = Fields.coordinate_field(Spaces.face_space(axes(ᶜuₕ))).z
    zmax = Spaces.z_max(axes(ᶠz))
    return @. lazy(-β_rayleigh_uₕ(s, ᶜz, zmax) * ᶜuₕ)
end
```

On the other hand, more operations on vertical meshes need to be gpu compatible if we make the `FiniteDifferenceSpace` constructor adapt to the given device, which is a kind of annoying set of requirements to satisfy (it's much easier to write code that operates on CPU data structures compared to GPU ones). In addition, the way I handled this at the moment is to adapt intermediate things to the CPU, which is not ideal, and normally I wouldn't do this, but the context was relevant: the data is already sent to the host with the explicit `Array` call (look for `ArrayType(vertical_interpolation_weights(cpu_space, target_zcoords))`). For InputsOutputs, there might be a way to do this somewhat efficiently, but that too showed that there is an explicit use of `Array`, so an explicit sync doesn't seem too bad there.

The alternative is that we tell users: you must put `z_min` / `z_max` in the cache, and instead write this function as:

```julia
function rayleigh_sponge_tendency_uₕ(ᶜuₕ, s, p)
    s isa Nothing && return NullBroadcasted()
    (; zmax) = p
    ᶜz = Fields.coordinate_field(axes(ᶜuₕ)).z
    return @. lazy(-β_rayleigh_uₕ(s, ᶜz, zmax) * ᶜuₕ)
end
```

(the same will go for any variables in the vertical mesh). I'm torn. I almost think that the latter is just easier, and handle this on a case-by-case basis. The most I've seen so far is needing `z_min` / `z_max`. @imreddyTeja, what do you think?

src/Remapping/distributed_remapping.jl
```diff
@@ -463,10 +469,21 @@ function _Remapper(
     FT = Spaces.undertype(space)
     ArrayType = ClimaComms.array_type(space)
 
+    cpu_space = if ClimaComms.device(space) isa ClimaComms.AbstractCPUDevice
+        space
+    else
+        device = ClimaComms.device(space)
+        cpu_space = if device isa ClimaComms.AbstractCPUDevice
+            space
```
imreddyTeja, line 477: I don't think the `if device isa ClimaComms.AbstractCPUDevice` case will ever be hit because of line 472

imreddyTeja:
It seems like from a user POV, making `FiniteDifferenceSpace` adapt to the device would be simpler (I'm assuming any operations on vertical meshes are defined in ClimaCore). Moving `z_min` and `z_max` to the cache means the user must know that the vertical mesh if available on host but not device. 

From a developer POV, moving what's needed on device to the cache seems much easier.

I'm not sure how many operations need to be made gpu compatible, but maybe this would be easier with Denis's `ClimaComms.@threaded` macro PR

src/Remapping/distributed_remapping.jl
charleskawczynski: Yeah, I think that was a copy-paste mistake.

charleskawczynski: I'm going to leave it, as I think this approach turned out to be not ideal.

dennisYatunin:
Superseded by #2631

PR #2344, opened 2025-06-03
Define FiniteDifferenceSpace for extruded spaces

charleskawczynski:
The failure [here](https://buildkite.com/clima/climaatmos-ci/builds/24411#01973684-acab-4b2e-89ff-e5469efe038c) revealed that ClimaDiagnostics.jl is reconstructing the vertical grid, via `_FiniteDifferenceGrid`, which is not efficient and will become gpu-incompatible in #2343 (as the mesh, once made, will be moved to the gpu).

This PR defines `FiniteDifferenceSpace(space::ExtrudedFiniteDifferenceSpace)` which avoids that code path and reconstructs a new grid with the same backing data.

I suppose I should add this to ClimaDiagnostics, so that I can test it in ClimaAtmos, as this is the only place where this functionality seems to be exercised. I'll do that now, but this feature should probably exist in ClimaCore.


charleskawczynski:
I don't this is controversial, so I'm going to merge.

PR #2346, opened 2025-06-11
Tr/refactor fm internal index

imreddyTeja:
TODO:

### DIFF
--- a/docs/src/matrix_fields.md
+++ b/docs/src/matrix_fields.md
@@ -107,9 +107,9 @@ scalar_field_matrix
 
 A FieldMatrix entry can be:
 
-- An `UniformScaling`, which contains a `Number`
-- A `DiagonalMatrixRow`, which can contain aything
-- A `ColumnwiseBandMatrixField`, where each row is a [`BandMatrixRow`](@ref) where the band element type is representable with the space's base number type.
+- A `UniformScaling`, which contains a `Number`
+- A `DiagonalMatrixRow`, which can contain either a `Number` or a tensor (represented as a `Geometry.Axis2Tensor`)
+- A `ColumnwiseBandMatrixField`, where each value is a [`BandMatrixRow`](@ref) with entries of any type that can be represented using the field's base number type.
 
 If an entry contains a composite type, the fields of that type can be extracted.
 This is also true for nested composite types.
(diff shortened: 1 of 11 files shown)


src/Geometry/axistensors.jl
```diff
@@ -61,7 +61,7 @@ coordinate_axis(::Type{<:LatLongPoint}) = (1, 2)
 
 coordinate_axis(coord::AbstractPoint) = coordinate_axis(typeof(coord))
 
-@inline idxin(I::Tuple{Int}, i::Int) = 1
+@inline idxin(I::Tuple{Int}, i::Int) = I[1] == i ? 1 : nothing
```
imreddyTeja: I'm not sure if the original function is correct here because it behaves differently than the other `idxin` functions. 
I found this when I was experimenting with using  axis symbols to index with vs numbers.

dennisYatunin: Yeah the original function looks like it was incorrect and this change is fine. Also, all of these `idxin` methods could be simplified to
```
@inline idxin(I, i) = unrolled_findfirst(==(i), I)
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
 field2arrays_view
+scalar_field_matrix
 ```
+
+## Indexing a FieldMatrix
+
+A FieldMatrix entry can be:
+
+- An `UniformScaling`, which contains a `Number`
+- A `DiagonalMatrixRow`, which can contain aything
```
dennisYatunin:
```suggestion
- A `DiagonalMatrixRow`, which can contain either a `Number` or a tensor (represented as a `Geometry.Axis2Tensor`)
```
The constructor does allow anything, but in terms of how this is used we only support numbers and tensors.

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
+- A `ColumnwiseBandMatrixField`, where each row is a [`BandMatrixRow`](@ref) where the band element type is representable with the space's base number type.
```
dennisYatunin:
```suggestion
- A `ColumnwiseBandMatrixField`, where each value is a [`BandMatrixRow`](@ref) with entries of any type that can be represented using the field's base number type.
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
+```julia
+ A = MatrixFields.FieldMatrix((@name(name1), @name(name2)) => sample_entry)
+```
+
+Now consider what happens indexing `A` with the key `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))`.
+
+First, a function searches the keys of `A` for a key that `(@name(foo.bar.buz), @name(biz.bop.fud))`
+is a child of. In this example, `(@name(foo.bar.buz), @name(biz.bop.fud))` is a child of
+the key `(@name(name1), @name(name2))`, and
+`(@name(foo.bar.buz), @name(biz.bop.fud))` is referred to as the internal key.
```
dennisYatunin:
```suggestion
First, `getindex` finds a key in `A` that contains the key being indexed. In this example, `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))` is contained within `(@name(name1), @name(name2))`, so `(@name(name1), @name(name2))` is called the "parent key" and `(@name(foo.bar.buz), @name(biz.bop.fud))` is referred to as the "internal key".
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
+works as follows:
+
+1. If the  `internal_name_pair` is blank, return `entry`
+2. If the element type of each band of `entry` is an `Axis2Tensor`, and `internal_name_pair` is of the form
+`(@name(components.data.1...), @name(components.data.2...))` (potentially with different numbers),
+then extract the specified component, and recurse on it with the remaining `internal_name_pair`.
+3. If the element type of each band of `entry` is a `Geometry.AdjointAxisVector`, then recurse on the parent of the adjoint.
+4. If `internal_name_pair[1]` is not empty, and the first name in it is a field of the element type of each band of `entry`,
+extract that field from `entry`, and recurse on the it with the remaining names of `internal_name_pair[1]` and all of `internal_name_pair[2]`
+5. If `internal_name_pair[2]` is not empty, and the first name in it is a field of the element type of each row of `entry`,
```
dennisYatunin:
```suggestion
5. If `internal_name_pair[2]` is not empty, and the first name in it is a field of the element type of each band of `entry`,
```
You used "band" elsewhere in this file

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
+The recursive indexing of an internal entry given some entry `entry` and internal_key `internal_name_pair`
+works as follows:
+
+1. If the  `internal_name_pair` is blank, return `entry`
+2. If the element type of each band of `entry` is an `Axis2Tensor`, and `internal_name_pair` is of the form
+`(@name(components.data.1...), @name(components.data.2...))` (potentially with different numbers),
+then extract the specified component, and recurse on it with the remaining `internal_name_pair`.
+3. If the element type of each band of `entry` is a `Geometry.AdjointAxisVector`, then recurse on the parent of the adjoint.
+4. If `internal_name_pair[1]` is not empty, and the first name in it is a field of the element type of each band of `entry`,
+extract that field from `entry`, and recurse on the it with the remaining names of `internal_name_pair[1]` and all of `internal_name_pair[2]`
```
dennisYatunin:
```suggestion
extract that field from `entry`, and recurse into it with the remaining names of `internal_name_pair[1]` and all of `internal_name_pair[2]`
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,97 @@ column_field2array
...
+5. If `internal_name_pair[2]` is not empty, and the first name in it is a field of the element type of each row of `entry`,
+extract that field from `entry`, and recurse on the it with all of `internal_name_pair[1]` and the remaining names of `internal_name_pair[2]`
```
dennisYatunin:
```suggestion
extract that field from `entry`, and recurse into it with all of `internal_name_pair[1]` and the remaining names of `internal_name_pair[2]`
```

src/MatrixFields/field_name_dict.jl
```diff
@@ -149,45 +149,110 @@ get_internal_key(child_name_pair::FieldNamePair, name_pair::FieldNamePair) = (
...
-        entry
-    elseif is_overlapping_name(name_pair[1], name_pair[2])
-        throw(key_error)
+    full_key::FieldNamePair,
+) where {T}
+    if name_pair == (@name(), @name())
+        return entry
+    elseif T <: Geometry.Axis2Tensor &&
+           all(n -> is_child_name(n, @name(components.data)), name_pair)
+        # two indices needed to index into a 2d tensor (one can be Colon())
```
dennisYatunin, line 183: Is the `Colon()` comment correct? I'm not sure if we can handle vector/covector slices of tensors.

src/MatrixFields/field_name_dict.jl
```diff
@@ -149,45 +149,110 @@ get_internal_key(child_name_pair::FieldNamePair, name_pair::FieldNamePair) = (
     extract_internal_name(child_name_pair[2], name_pair[2]),
 )
 
-get_internal_entry(entry, name::FieldName, key_error) = get_field(entry, name)
-get_internal_entry(entry, name_pair::FieldNamePair, key_error) =
-    name_pair == (@name(), @name()) ? entry : throw(key_error)
-get_internal_entry(
-    entry::ScalingFieldMatrixEntry,
+get_internal_entry(entry, name::FieldName) = get_field(entry, name)
```
dennisYatunin, line 156: Could you add a docstring for `get_internal_entry` like you have for `field_offset_and_type`?

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
 field2arrays
 field2arrays_view
+scalar_field_matrix
+```
+
+## Indexing a FieldMatrix
+
+A FieldMatrix entry can be:
+
+- An `UniformScaling`, which contains a `Number`
```
dennisYatunin:
```suggestion
- A `UniformScaling`, which contains a `Number`
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+```
+
+Now consider what happens indexing `A` with the key `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))`.
+
+First, `getindex` finds a key in `A` that contains the key being indexed. In this example, `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))` is contained within `(@name(name1), @name(name2))`, so `(@name(name1), @name(name2))` is called the "parent key" and `(@name(foo.bar.buz), @name(biz.bop.fud))` is referred to as the "internal key".
+
+Next, the entry that `(@name(name1), @name(name2))` is paired with is recursively indexed
+by the internal key.
+
+The recursive indexing of an internal entry given some entry `entry` and internal_key `internal_name_pair`
```
dennisYatunin:
```suggestion
The recursive indexing of an internal entry given some entry `entry` and internal key `internal_name_pair`
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+### Further Indexing Details
+
+Let key `(@name(name1), @name(name2))` correspond to entry `sample_entry` in `FieldMatrix` `A`.
+An example of this is:
+
+```julia
+ A = MatrixFields.FieldMatrix((@name(name1), @name(name2)) => sample_entry)
+```
+
+Now consider what happens indexing `A` with the key `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))`.
```
dennisYatunin:
```suggestion
Now consider what happens when indexing `A` with the key `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))`.
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+First, `getindex` finds a key in `A` that contains the key being indexed. In this example, `(@name(name1.foo.bar.buz), @name(name2.biz.bop.fud))` is contained within `(@name(name1), @name(name2))`, so `(@name(name1), @name(name2))` is called the "parent key" and `(@name(foo.bar.buz), @name(biz.bop.fud))` is referred to as the "internal key".
+
+Next, the entry that `(@name(name1), @name(name2))` is paired with is recursively indexed
+by the internal key.
+
+The recursive indexing of an internal entry given some entry `entry` and internal_key `internal_name_pair`
+works as follows:
+
+1. If the  `internal_name_pair` is blank, return `entry`
+2. If the element type of each band of `entry` is an `Axis2Tensor`, and `internal_name_pair` is of the form
```
dennisYatunin:
```suggestion
2. If the element type of each band of `entry` is an `Axis2Tensor`, and `internal_name_pair` is of the form\
```
You need to add a backslash to the end of each line in a multiline list item, or else the next line will be rendered as a new paragraph. Alternatively, you could just avoid using newlines in list items, which is what I usually do. This goes for all lists in this file.

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+## Optimizations
+
+Each entry of a `FieldMatrix` can be a `ColumnwiseBandMatrixField`, a `DiagonalMatrixRow`, or an
+`UniformScaling`.
+
+A `ColumnwiseBandMatrixField` is a `Field` with a `BandMatrixRow` at each point. It is intended
+to represent a collection of banded matrices, where there is one band matrix for each column
+of the space the `Field` is on. Beyond only storing the diagonals of the band matrix, an `entry`
+can be optimized to use less memory. Each optimized representation can be indexed equivalently to
+non optimized representation, and used in addition, subtraction, Matrix-vector multiplication,
```
dennisYatunin:
```suggestion
non optimized representations, and used in addition, subtraction, matrix-vector multiplication,
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+`T_f` and `T_g`. $f_i$ and $g_i$ refers to the values of $f$ and $g$ at the $ 0 < i \leq N_v$ level.
+
+Let $M$ be a $N_v \times N_v$ banded matrix with lower and upper bandwidth of $b_1$ and $b_2$.
+$M$ represents $\frac{\partial f}{\partial g}$, so $M_{i,j} = \frac{\partial f_i}{\partial g_j}$
+
+### `ScalingFieldMatrixEntry` Optimization
+
+Consider the case where $b_1 = 0$ and $b_2 = 0$, i.e $M$ is a diagonal matrix, and
+where $M = k * I$, and $k$ is of type `T_k`. This would happen if
+$\frac{\partial f_i}{\partial g_j} = [i=j] * k$. Instead of storing
```
dennisYatunin:
```suggestion
$\frac{\partial f_i}{\partial g_j} = \delta_{ij} * k$. Instead of storing
```
The Kronecker delta is typically used for this.

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+each element on the diagonal, less memory can be used by storing a single value that represents a scaling of the identity. This reduces memory usage by a factor of $N_v$.
```
dennisYatunin:
```suggestion
each element on the diagonal, the `FieldMatrix` can store a single value that represents a scaling of the identity matrix, reducing memory usage by a factor of $N_v$:
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+
+### Implicit Tensor Structure Optimization
+
+The functions that index an entry with an internal key assume the implicit tensor structure optimization is being used
+when all of the following are true for `entry` where `T_k` is the element type of each band, and
+`(internal_key_1, internal_key_2)` is the internal key indexing `entry`.
+
+- the first key in the `internal_key_1` name chain is not a fieldname of `T_k`
+- the first key in the `internal_key_2` name chain is not a fieldname of `T_k`
+- `internal_key_1` and `internal_key_2` are both not empty
```
dennisYatunin:
```suggestion
- the `internal_key_1` name chain is not empty and its first name is not a field of `T_k`
- the `internal_key_2` name chain is not empty and its first name is not a field of `T_k`
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+
+For most use cases, `T_k` is a scalar.
+
+If the above conditions are met, the optimization assumes that the user intends the
+entry to have an implicit tensor structure, with the values of type `T_k` representing a scaling of the
+tensor identity. If both the first and second name in the name pair are equivalent, then they index onto the diagonal,
+and the value is returned. Otherwise, they index off the diagonal, and a zero value
```
dennisYatunin:
```suggestion
identity tensor. If both the first and second names in the name pair are equivalent, then they index onto the diagonal,
and the scalar value of `k` is returned. Otherwise, they index off the diagonal, and a zero value
```

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+is returned.
+
+This optimization is intended to be used when `T_f = T_g`, and they are both `AxisVectors`
+The notation $f_{n}[i]$ where $0 < n \leq N_v$  refers to the $i$ component of the element
+at the $n$ vertical level of $f$. In the following example, `T_f=T_g=Covariant12Vector`, and
```
dennisYatunin:
```suggestion
This optimization is intended to be used when `T_f = T_g`.
The notation $f_{n}[i]$ where $0 < n \leq N_v$  refers to the $i$-th component of the element
at the $n$-th vertical level of $f$. In the following example, `T_f` and `T_g` are both `Covariant12Vector`s, and
```
I don't think this optimization requires `T_f` and `T_g` to be `AxisVector`s, it should apply the same way to something like $\partial$`NamedTuple` $/\partial$`NamedTuple`.

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+
+```@example 2
+J[(@name(f.components.data.:(2)), @name(g.components.data.:(1)))]
+```
+
+This can be more optimally stored with the implicit tensor structure optimization:
+
+```@example 2
+∂f_∂g = fill(MatrixFields.TridiagonalMatrixRow(-0.5, 1.0, -0.5), space) # hide
+J = MatrixFields.FieldMatrix((@name(f), @name(g))=> ∂f_∂g) # hide
```
dennisYatunin:
```suggestion
J = MatrixFields.FieldMatrix((@name(f), @name(g))=> ∂f_∂g) # hide
nothing # hide
```
This is currently rendering the value of `J` instead of hiding it. You need to either return `nothing` from the `@example` block or replace `@example` with `@setup`.

docs/src/matrix_fields.md
```diff
@@ -98,4 +100,252 @@ column_field2array
...
+
+```julia
+∂f_∂g = fill(MatrixFields.DiagonalMatrixRow(k), space)
+```
+
+can equivalently be represented with
+
+```julia
+∂f_∂g = MatrixFields.DiagonalMatrixRow(k)
 ```
```
dennisYatunin:
```suggestion
where $k$ is a constant scalar, the implicit tensor structure optimization and
`ScalingFieldMatrixEntry` optimization can both be applied.
```
I don't think we need a copy of the previous section's example here.

src/MatrixFields/field_name_dict.jl
```diff
@@ -237,6 +334,251 @@ function Base.one(matrix::FieldMatrix)
...
+"""
+    field_offset_and_type(name_pair::FieldNamePair, ::Type{T}, ::Type{S}, full_key::FieldNamePair)
+
+Returns the offset of the field with name `name_pair` in an object of type `S` in
+multiples of `sizeof(T)`, the type of the field with name `name_pair`, and a `Val` indicating
+what method can index a ClimaCore `Field` of `S` with `name_pair`.
+
+The third return value is one of the following:
+- `Val(:view)`: indexing with a view is possible
+-  `Val(:view_of_blocks)`: indexing with a view of blocks is possible (this is not implemented)
```
dennisYatunin: Could you add a line to clarify this? I'm interpreting this to mean that a `view_of_blocks` would involve a strided view with nonuniform stride length, but I'm not sure if I got that right.

dennisYatunin (approved):
Thank you for all the new documentation! And the recursive implementations of `get_internal_entry` and `field_offset_and_type` are much clearer now.

Couple of minor suggestions for the docs, but overall this is good to merge in. This will be the last change to ClimaCore required for sparse autodiff, so we can make a new release after it's merged.

PR #2347, opened 2025-06-13
Tr/minor fixes

### LINKED ISSUE #2320: `field2array` error statement is inaccurate
The `field2array`  error when the field is not a field of scalars is inaccurate. 

```julia
julia> ClimaCore.Fields.field2array(ClimaCore.Fields.fill((1.0, 2.0), ClimaCore.Spaces.PointSpace(ClimaCore.Geometry.YPoint(0.5))))
ERROR: unable to use field2array because each Field element is represented by 0 array elements (must be 1)
Stacktrace:
```

I think this is caused by 
```
f_axis_size = sizeof(eltype(parent(field))) ÷ sizeof(eltype(field))
        error("unable to use field2array because each Field element is \
               represented by $f_axis_size array elements (must be 1)")
```

which should be 

```
f_axis_size = sizeof(eltype(field)) ÷ sizeof(eltype(parent(field))) 
```

### LINKED ISSUE #2337: `Spaces.nlevels` doesn't error for invalid level input arg
```
Build an arbitrary extruded geometry (in this case, spherical) simulation and query some level information : 

julia> Spaces.axes(Y.c.ρ)
CenterExtrudedFiniteDifferenceSpace:
  context: SingletonCommsContext using CPUSingleThreadedSingletonCommsContext using CPUSingleThreaded
  horizontal:
    mesh: 6×6×6-element EquiangularCubedSphere of SphereDomain: radius = 6.371e6
    node_horizontal_length_scale: 512229.34
    element_horizontal_length_scale: 1.536688e6
    quadrature: 4-point Gauss-Legendre-Lobatto quadrature
  vertical:
    mesh: 10-element IntervalMesh of IntervalDomain: z ∈ [0.0,30000.0] (:bottom, :top)

julia> test_face_field
Float32-valued Field:
  Float32[-0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]

julia> size(parent(test_face_field))
(11, 4, 4, 1, 216)

julia> Spaces.nlevels(test_face_field)
11

julia> Fields.level(test_face_field, 11+half)
Float32-valued Field:
  Float32[-0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0  …  -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, -0.0, 0.0]

julia> Fields.level(test_face_field, 10+half)
Float32-valued Field:
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]

julia> Fields.level(test_face_field, 200+half)
Float32-valued Field:
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  1.11116f-36, 163965.0, 9332.46, 9.26442f-23, 9.26442f-23, 9.49672f-18, 3.06812f-10, 2.37873f6, 9.67983f-12, 1.08037f-40]
```

I'd expect the last line to error since the level ind exceeds `Spaces.nlevels(field_of_interest)`.

### LINKED ISSUE #2348: Inconsistent AxisVector indexing
Indexing a 1d axis vector behaves differently than indexing a 2d axis vector:

```
julia> v1 = ClimaCore.Geometry.Covariant1Vector(1.0)
1-element ClimaCore.Geometry.Covariant1Vector{Float64} with indices ClimaCore.Geometry.CovariantAxis{(1,)}():
 1.0

julia> v2 = ClimaCore.Geometry.Covariant12Vector(1.0, 2.0)
2-element ClimaCore.Geometry.Covariant12Vector{Float64} with indices ClimaCore.Geometry.CovariantAxis{(1, 2)}():
 1.0
 2.0

julia> v2.u
u₁
u₂
u₃
julia> v2.u₃
0.0

julia> v1.u₃
1.0
```

Ran on ClimaCore v0.14.34

```
Julia Version 1.10.9
Commit 5595d20a287 (2025-03-10 12:51 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 8 × Apple M3
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 4 virtual cores)
```

imreddyTeja:
This PR addresses three issues:
1. An incorrect error message is shown when `field2array` is called with incompatible elements. 
2. For face fields, the `level` function does not throw `BoundsError`s
3. The indexing behavior for AxisTensors is inconsistent. With any 1d axis tensor vs 2d and 3d.

For the first issue, the message itself is corrected.
For the second issue, tests are added for `level(::Field)`. 
The fix for the second issue requires removing an `@inbounds`, so there may be a performance penalty.

This PR also bumps the UnrolledUtilities compat to v0.1.8. v0.1.7 adds a lot of useful unrolled_functions, including `unrolled_findfirst`, but the downgrade will throw JET errors unless using v0.1.8

Closes #2337 
Closes #2320 
Closes #2348

### DIFF
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -517,7 +517,7 @@ Base.@propagate_inbounds function level(
     v::PlusHalf,
 )
     hspace = level(axes(field), v)
-    @inbounds data = level(field_values(field), v.i + 1)
+    data = level(field_values(field), v.i + 1)
     Field(data, hspace)
 end
 
@@ -664,7 +664,7 @@ element type of the array is the same as the struct type of `field`.
 """
 function field2array(field::Field)
     if sizeof(eltype(field)) != sizeof(eltype(parent(field)))
-        f_axis_size = sizeof(eltype(parent(field))) ÷ sizeof(eltype(field))
+        f_axis_size = sizeof(eltype(field)) ÷ sizeof(eltype(parent(field)))
         error("unable to use field2array because each Field element is \
                represented by $f_axis_size array elements (must be 1)")
     end
--- a/Project.toml
+++ b/Project.toml
@@ -78,7 +78,7 @@ Statistics = "1"
 StatsBase = "0.34"
 TerminalLoggers = "0.1"
 Test = "1"
-UnrolledUtilities = "0.1.6"
+UnrolledUtilities = "0.1.9"
 julia = "1.10"
 
 [extras]
--- a/src/Geometry/Geometry.jl
+++ b/src/Geometry/Geometry.jl
@@ -2,6 +2,7 @@ module Geometry
 
 using ..RecursiveApply
 import LinearAlgebra
+import UnrolledUtilities: unrolled_findfirst
 
 using StaticArrays
 
--- a/src/Geometry/axistensors.jl
+++ b/src/Geometry/axistensors.jl
@@ -61,33 +61,8 @@ coordinate_axis(::Type{<:LatLongPoint}) = (1, 2)
 
 coordinate_axis(coord::AbstractPoint) = coordinate_axis(typeof(coord))
 
-@inline idxin(I::Tuple{Int}, i::Int) = 1
-
-@inline function idxin(I::Tuple{Int, Int}, i::Int)
-    @inbounds begin
-        if I[1] == i
-            return 1
-        elseif I[2] == i
-            return 2
-        else
-            return nothing
-        end
-    end
-end
-
-@inline function idxin(I::Tuple{Int, Int, Int}, i::Int)
-    @inbounds begin
-        if I[1] == i
-            return 1
-        elseif I[2] == i
-            return 2
-        elseif I[3] == i
-            return 3
-        else
-            return nothing
-        end
-    end
-end
+@inline idxin(I::NTuple{N, Int}, i::Int) where {N} =
+    unrolled_findfirst(isequal(i), I)
 
 struct PropertyError <: Exception
     ax::Any
--- a/test/Fields/unit_field.jl
+++ b/test/Fields/unit_field.jl
@@ -1196,6 +1196,43 @@ end
     bf_new = @. bf # test copy()
 end
 
+@testset "Level function boundschecking" begin
+    FT = Float32
+    extruded_center_space = ExtrudedCubedSphereSpace(
+        FT;
+        z_elem = 10,
+        z_min = 0,
+        z_max = 1,
+        radius = 10,
+        h_elem = 10,
+        n_quad_points = 4,
+        staggering = Grids.CellCenter(),
+    )
+    fd_center_space = ColumnSpace(
+        FT;
+        z_elem = 10,
+        z_min = 0,
+        z_max = 10,
+        staggering = CellCenter(),
+    )
+    for center_space in (extruded_center_space, fd_center_space)
+        face_space = Spaces.face_space(center_space)
+        center_field = fill(FT(1), center_space)
+        face_field = fill(FT(2), face_space)
+        if Base.JLOptions().check_bounds == 1
+            for level in (-1, 11, 0)
+                @test_throws BoundsError Fields.level(center_field, level)
+                level != 0 && @test_throws BoundsError Fields.level(
+                    face_field,
+                    PlusHalf(level),
+                )
+            end
+        else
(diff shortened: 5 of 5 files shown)


src/Spaces/finitedifference.jl
```diff
@@ -145,3 +145,10 @@ Base.@propagate_inbounds function level(
     local_geometry = level(local_geometry_data(space), v)
     PointSpace(ClimaComms.context(space), local_geometry)
 end
+function issubspace(subspace::PointSpace, space::FiniteDifferenceSpace)
+    all_local_geometry = local_geometry_data(space)
+    unrolled_any(
+        v -> subspace.local_geometry === level(all_local_geometry, v),
+        1:nlevels(space),
+    )
```
charleskawczynski: I'm not sure that we want to unroll over all `nlevels` for this. Isn't it sufficient to check the first level? When would it not be?

imreddyTeja: I thought the local geometry can vary spatially. For example:

```
julia> DataLayouts.level(Spaces.local_geometry_data(space), 1)[].coordinates
ZPoint(0.5)

julia> DataLayouts.level(Spaces.local_geometry_data(space), 3)[].coordinates
ZPoint(2.5)

julia> Fields.local_geometry_field(Fields.level(space, 1)).coordinates
ClimaCore.Geometry.ZPoint{Float64}-valued Field:
  z: [0.5]

julia> Fields.local_geometry_field(Fields.level(space, 3)).coordinates
ClimaCore.Geometry.ZPoint{Float64}-valued Field:
  z: [2.5]
```

charleskawczynski: Yeah, I suppose it's safe to do the comparison over every point (even checking the range is still subject errors). But, we still probably don't want to unroll this call, since the result is not compile time known and `nlevels` can be quite large.

imreddyTeja: I guess this method is not really needed. Its only use in src is to check broadcasting shapes, which already has  unique methods for `PointSpaces`.

charleskawczynski: Yeah, I don't think it's likely that we'll fall into such an error, but it is an extra layer of protection.

On the one hand, I don't think that this can be easily determined, with our current design, at compile-time. Ideally, we could add some sort of compile-time switches into our code that would allow / enforce these checks at runtime, and we could use it as a debugging feature. But it might be a hefty amount of work to make that feature available.

imreddyTeja: I'm not quite sure why I used unrolling here... do you think I should just remove this functionality, or leave it without unrolling? The only argument I have for keeping it is that if it exists for the extruded spectral element spaces, I'd expect it to also exist for finite difference spaces.

charleskawczynski: I think the performance is likely fine without the unrolling, since the `===` comparison is likely pretty fast.

The only issue that I could see is that this operation is not compile-time known, and the logic of where this is called includes a check that the spaces are compatible (we error if they're not). If the check is not compile-time known, then it won't work inside a cuda kernel (this check is currently performed before the kernel is launched, but that won't be the case and it will need to work inside a cuda kernel with my `columnwise!` design, so there may be clashes with that).

src/Fields/Fields.jl
```diff
@@ -517,7 +517,7 @@ Base.@propagate_inbounds function level(
     v::PlusHalf,
 )
     hspace = level(axes(field), v)
-    @inbounds data = level(field_values(field), v.i + 1)
+    data = level(field_values(field), v.i + 1)
```
akshaysridhar, line 520: Do we know what the performance penalty , if any, here might be ?

akshaysridhar: @dennisYatunin pointed out we call the outer function with propagate_inbounds; this is not an issue. Thanks @imreddyTeja

PR #2351, opened 2025-07-02
Bump patch version 0.14.35

costachris:
Bump patch version to 0.14.35 for release.


imreddyTeja (approved):
`NEWS.md` has not been updated in a while, but that can be done in a different PR. 
I think it would be good to at least add `NEWS.md` entries for v0.14.34 and v0.14.35, even if they are blank or incomplete, but I'm also fine leaving that for later.

PR #2356, opened 2025-07-17
Extend level and field_values to work with lazy Fields

dennisYatunin:
This PR extends `level` and `field_values` to work with `Broadcasted` expressions that represent lazy pointwise field operations. This builds on the work from #2331, which extended `column` and `slab` to work with lazy fields. In order to test `field_values` with lazy fields, it was also necessary to define `==` for DataLayouts.

### DIFF
--- a/test/Fields/unit_field.jl
+++ b/test/Fields/unit_field.jl
@@ -662,6 +662,33 @@ end
     end
 end
 
+@testset "Lazy Field broadcasts" begin
+    FT = Float64
+    for space in TU.all_spaces(FT)
+        field = fill((; x = FT(1)), space)
+        @test field == Base.materialize(lazy.(identity.(field)))
+        @test field .+ 1 == Base.materialize(lazy.(field .+ 1))
+        @test Fields.field_values(field .+ 1) ==
+              Base.materialize(Fields.field_values(lazy.(field .+ 1)))
+    end
+end
+
+@testset "Levels of Fields and Field broadcasts" begin
+    FT = Float64
+    for space in TU.all_spaces(FT)
+        TU.levelable(space) || continue
+        field = fill((; x = FT(1)), space)
+        level_of_field = Fields.Field(
+            Spaces.level(Fields.field_values(field), 1),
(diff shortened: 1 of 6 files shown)


imreddyTeja (approved):
At some point we may want to add documentation for the different broadcast style types, even though most of the names are self explanatory. I also couldn't find much about `NonExtrudedBroadcasted`, and I'm not exactly sure what it does, or where it is used.

test/Fields/unit_field.jl
```diff
@@ -662,6 +662,22 @@ end
     end
 end
 
+@testset "Levels of Fields and Field broadcasts" begin
```
akshaysridhar, line 676: This test seems reasonable. Thanks @dennisYatunin.

akshaysridhar (approved):

akshaysridhar:
> At some point we may want to add documentation for the different broadcast style types, even though most of the names are self explanatory. I also couldn't find much about `NonExtrudedBroadcasted`, and I'm not exactly sure what it does, or where it is used.

The type-name suggests that `NonExtrudedBroadcasted` is used only with the `ClimaCore.DataLayouts.IJFH` layouts (in ClimaCore there is a 2D example problem which uses `RectangleDomain`s in the `bickley_jet` directory.

dennisYatunin:
@akshaysridhar I think the word "extruded" in this context refers to broadcast axis extrusion (i.e., the ability to do something like `vector .+ matrix`, where the second axis of `vector` is extruded to match the second axis of `matrix), rather than the extrusion of our horizontal ClimaCore spaces into 3D spaces.

akshaysridhar:
> @akshaysridhar I think the word "extruded" in this context refers to broadcast axis extrusion (e.g., the ability to do something like `vector .+ matrix`, where the second axis of `vector` is extruded to match the second axis of `matrix`), rather than the extrusion of our horizontal ClimaCore spaces into 3D spaces.

Ah; noted. thanks.

costachris (approved):

PR #2361, opened 2025-08-01
Added remapping to ClimaCorePlots

ray-chew:
This feature might be useful for debugging in an infiltrator session.


lib/ClimaCorePlots/src/ClimaCorePlots.jl
```diff
@@ -492,26 +494,63 @@ end
...
+            Geometry.LatLongPoint(lat, long) for long in longpts, lat in latpts
+        ]
 
-    # set the plot attributes
-    seriestype := :heatmap
-    title --> "$nelem $quad_from_name{$dof_in} element space"
-    xguide --> "panel x1"
-    yguide --> "panel x2"
-    seriescolor --> :balance
+        # @Main.infiltrate
```
akshaysridhar, line 516:
```suggestion
```

lib/ClimaCorePlots/src/ClimaCorePlots.jl
```diff
@@ -492,26 +494,63 @@ end
...
 
-    (unfolded_panels)
+        # hcoords = Remapping.default_target_hcoords(axes(field))
```
akshaysridhar, line 518:
```suggestion
```

lib/ClimaCorePlots/src/ClimaCorePlots.jl
```diff
@@ -492,26 +494,63 @@ end
...
+        remapped = Array(Remapping.interpolate(remapper, field))
+
+        # Set plot attributes
+        seriestype := :heatmap
+        xguide --> "Longitude"
+        yguide --> "Latitude"
+        seriescolor --> :balance
+        title --> "Lat-Lon Remapped Field ($nlat × $nlon)"
+
+        return (longpts, latpts, transpose(remapped))
```
akshaysridhar, line 530: Could you please include an example of the output from your suggested changes and differentiate it from the default "unfolded panel" output in the `docs` for ClimaCorePlots (I realise the current doc section for `Libraries > ClimaCorePlots.jl` is empty..) . Thanks @ray-chew, this way we can visually inspect the different options available here. It seems the current tests https://github.com/CliMA/ClimaCore.jl/blob/main/lib/ClimaCorePlots/test/runtests.jl rely heavily on `png` generation with `isfile`.

PR #2364, opened 2025-08-04
update README: WeakValueDicts unmaintained

juliasloan25:
WeakValueDicts.jl is no longer maintained and is not threadsafe. ClimaCore.jl doesn't currently depend on that package, but previous versions do. This PR updates the README to explain that those older versions are unsupported and shouldn't be used.

See discussion in https://github.com/twavv/WeakValueDicts.jl/issues/2#issuecomment-3144544623 for more details.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@ ClimaCore.jl Release Notes
 main
 -------
 
+v0.14.37
+-------
+- Extend level and field_values to work with lazy Fields [2356](https://github.com/CliMA/ClimaCore.jl/pull/2356)
+
 v0.14.36
 -------
 - Refactor FieldMatrix indexing, add scalar_field_matrix [2346](https://github.com/CliMA/ClimaCore.jl/pull/2346)
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.14.36"
+version = "0.14.37"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


imreddyTeja:
> Tests here are failing after only changing the README, likely because of the new RootSolvers v0.4.5 release. Tagging @szy21 @tapios for visibility

I think the latest release dropped the default max iterations by a factor of 10.

szy21:
It fails at initialization though. Maybe we need to clear the depot?

szy21:
Oh I see that the unit tests are failing - yes that is probably from the RootSolvers

tapios:
> > Tests here are failing after only changing the README, likely because of the new RootSolvers v0.4.5 release. Tagging @szy21 @tapios for visibility
> 
> I think the latest release dropped the default max iterations by a factor of 10.

I think the issue is an early convergence check, not the iteration count. Patch on the way.

szy21:
Should we add downstream tests in RootSolvers?

tapios:
> Should we add downstream tests in RootSolvers?

I prefer not to. RootSolvers is a very basic package that is not geared toward any particular CliMA (or other) package. We should make sure that edge cases are properly handled, and we should add tests covering them to RootSolvers when we find them. But it does not make sense in my view to bloat the dependencies of RootSolver testing. Here, for example, it's noa RootSolvers problem from what I can see, but a problem of inappropriate tests.

akshaysridhar:
> > > > Tests here are failing after only changing the README, likely because of the new RootSolvers v0.4.5 release. Tagging @szy21 @tapios for visibility
> > > 
> > > 
> > > I think the latest release dropped the default max iterations by a factor of 10.
> > 
> > 
> > I think the issue is an early convergence check, not the iteration count. Patch on the way.
> 
> Actually, no, the issue is an unreasonable requirement to get a floating point precision match between the computed and expected mesh spacing. Yes, maybe with 10,000 iterations we can get there, but it really does not make sense to test that. Also if higher accuracy is desired, this should not be handled by default choices in the RootSolvers but by custom choices made locally.
> 
> @akshaysridhar Can you please replace the tests here (in `tests/interval.jl`) by tests with `isapprox` and `rtol = 1e-3` (or whatever makes sense, 1e-3 iseems to be used in `src/interval.jl`)?

I'll review these tests and make the changes today.

PR #2367, opened 2025-08-05
update README: WeakValueDicts unmaintained

akshaysridhar:
Supersedes #2364, Retains PR Description + author. 

WeakValueDicts.jl is no longer maintained and is not threadsafe. ClimaCore.jl doesn't currently depend on that package, but previous versions do. This PR updates the README to explain that those older versions are unsupported and shouldn't be used.

See discussion in https://github.com/twavv/WeakValueDicts.jl/issues/2#issuecomment-3144544623 for more details.

Relaxes test tolerances in grid-stretching checks.
- modified:   test/Meshes/interval.jl

Author : @juliasloan25, @akshaysridhar


juliasloan25 (approved):
thank you!

PR #2373, opened 2025-08-25
Fix Jacobian-free Newton Krylov method on GPUs

dennisYatunin:
This PR modifies `Krylov.ktypeof(::FieldVector)` so that the Jacobian-free Newton Krylov method used in CliMA/ClimaAtmos.jl#3976 works on GPUs. It also fixes a typo in the definition `ClimaComms.context(::FieldVector)`, where `Fix1` was used in place of `Fix2`.

### DIFF
--- a/ext/KrylovExt.jl
+++ b/ext/KrylovExt.jl
@@ -1,10 +1,14 @@
 module KrylovExt
 
-import ClimaComms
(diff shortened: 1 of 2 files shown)


tapios:
This is stale, right? JFNK now works after the recent changes in CTS. Can we close?

PR #2375, opened 2025-10-01
Add support for FiniteDifferenceSpaces

nefrathenrici:
This PR enables horizontal spectral element operators (gradient, divergence, curl) to work correctly on column spaces extracted from hybrid spaces, returning appropriate zero values since columns have no horizontal dimensions.

### DIFF
--- a/src/CommonGrids/CommonGrids.jl
+++ b/src/CommonGrids/CommonGrids.jl
@@ -40,7 +40,7 @@ z_domain = Domains.IntervalDomain(
 )
 z_mesh = Meshes.IntervalMesh(z_domain; nelems = z_elem)
 h_grid = Grids.SpectralElementGrid2D(h_topology, quad)
-z_topology = Topologies.IntervalTopology(context, z_mesh)
+z_topology = Topologies.IntervalTopology(ClimaComms.SingletonCommsContext(device), z_mesh)
 z_grid = Grids.FiniteDifferenceGrid(z_topology)
 grid = Grids.ExtrudedFiniteDifferenceGrid(
     h_grid,
@@ -440,6 +440,19 @@ function Box3DGrid(
             y_elem,
             periodic_x,
             periodic_y,
+        ),
+        Topologies.spacefillingcurve(
+            DefaultRectangleXYMesh(
+                FT;
+                x_min,
+                x_max,
+                y_min,
+                y_max,
+                x_elem,
+                y_elem,
+                periodic_x,
+                periodic_y,
+            ),
         ),
     ),
     z_mesh::Meshes.IntervalMesh = DefaultZMesh(
--- a/src/Geometry/axistensors.jl
+++ b/src/Geometry/axistensors.jl
@@ -219,6 +219,8 @@ const AxisVector{T, A1, S} = AxisTensor{T, 1, Tuple{A1}, S}
 AxisVector(ax::A1, v::SVector{N, T}) where {A1 <: AbstractAxis, N, T} =
     AxisVector{T, A1, SVector{N, T}}((ax,), v)
 
+(AxisVector{T, A, SVector{0, T}} where {T})() where {A} =
+    AxisVector(A.instance, SVector{0, T}())
 (AxisVector{T, A, SVector{1, T}} where {T})(arg1::Real) where {A} =
     AxisVector(A.instance, SVector(arg1))
 (AxisVector{T, A, SVector{2, T}} where {T})(arg1::Real, arg2::Real) where {A} =
@@ -306,11 +308,12 @@ const CovariantTensor = Union{CovariantVector, Covariant2Tensor}
 const ContravariantTensor = Union{ContravariantVector, Contravariant2Tensor}
 const CartesianTensor = Union{CartesianVector, Cartesian2Tensor}
 const LocalTensor = Union{LocalVector, Local2Tensor}
+for I in [(), (1,), (2,), (3,), (1, 2), (1, 3), (2, 3), (1, 2, 3)]
 
-for I in [(1,), (2,), (3,), (1, 2), (1, 3), (2, 3), (1, 2, 3)]
-    strI = join(I)
+    strI = isempty(I) ? "Null" : join(I)
     N = length(I)
-    strUVW = join(map(i -> [:U, :V, :W][i], I))
+
+    strUVW = isempty(I) ? "Null" : join(map(i -> [:U, :V, :W][i], I))
     @eval begin
         const $(Symbol(:Covariant, strI, :Axis)) = CovariantAxis{$I}
         const $(Symbol(:Covariant, strI, :Vector)){T} =
@@ -467,7 +470,7 @@ end
         end
         push!(vals, val)
     end
-    return :(@inbounds AxisVector(ato, SVector($(vals...))))
+    return :(@inbounds AxisVector(ato, SVector{$(length(Ito)), $T}($(vals...))))
 end
 
 function _transform(
--- a/src/Geometry/conversions.jl
+++ b/src/Geometry/conversions.jl
@@ -501,13 +501,16 @@ Curl is only defined for `CovariantVector`` field input types.
 |  Covariant12Vector | (1,2) | Contravariant3Vector |
 |  Covariant3Vector | (1,2) | Contravariant12Vector |
 |  Covariant123Vector | (1,2) | Contravariant123Vector |
-|  Covariant1Vector | (1,) | Contravariant1Vector |
+|  Covariant1Vector | (1,) | ContravariantNullVector |
 |  Covariant2Vector | (1,) | Contravariant3Vector |
 |  Covariant3Vector | (1,) | Contravariant2Vector |
+|  Covariant13Vector | (1,) | Contravariant2Vector |
+|  Covariant123Vector | (1,) | Contravariant23Vector |
 |  Covariant12Vector | (3,) | Contravariant12Vector |
 |  Covariant1Vector | (3,) | Contravariant2Vector |
 |  Covariant2Vector | (3,) | Contravariant1Vector |
 |  Covariant3Vector | (3,) | Contravariant3Vector |
+|  Any CovariantVector | () | ContravariantNullVector |
 """
 @inline curl_result_type(::Val{(1, 2)}, ::Type{Covariant3Vector{FT}}) where {FT} =
     Contravariant12Vector{FT}
@@ -516,9 +519,8 @@ Curl is only defined for `CovariantVector`` field input types.
 @inline curl_result_type(::Val{(1, 2)}, ::Type{Covariant123Vector{FT}}) where {FT} =
     Contravariant123Vector{FT}
 
-
 @inline curl_result_type(::Val{(1,)}, ::Type{Covariant1Vector{FT}}) where {FT} =
-    Contravariant1Vector{FT} # not strictly correct: should be a zero Vector
+    ContravariantNullVector{FT}
 @inline curl_result_type(::Val{(1,)}, ::Type{Covariant2Vector{FT}}) where {FT} =
     Contravariant3Vector{FT}
 @inline curl_result_type(::Val{(1,)}, ::Type{Covariant3Vector{FT}}) where {FT} =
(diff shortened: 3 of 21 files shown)


src/CommonGrids/CommonGrids.jl
```diff
@@ -441,6 +441,19 @@ function Box3DGrid(
             periodic_x,
             periodic_y,
         ),
+        Topologies.spacefillingcurve(
```
dennisYatunin, line 444: This can just be the default for Topology2D, unless there's some benefit to using linear ordering.

src/Geometry/axistensors.jl
```diff
@@ -244,6 +244,9 @@ AxisVector(ax::A1, v::SVector{N, T}) where {A1 <: AbstractAxis, N, T} =
     arg3::Real,
 ) where {A} = AxisVector(A.instance, SVector(arg1, arg2, arg3))
 
+(AxisVector{T, A, SVector{0, T}})() where {A, T} =
```
dennisYatunin:
```suggestion
(AxisVector{T, A, SVector{0, T}} where {T})() where {A} =
```
The previous 3 methods all have this type signature. Not sure if it matters but probably best to be consistent.

src/Geometry/axistensors.jl
```diff
@@ -355,6 +355,12 @@ for I in [(1,), (2,), (3,), (1, 2), (1, 3), (2, 3), (1, 2, 3)]
     end
 end
 
+const CovariantTensor = Union{CovariantVector, Covariant2Tensor}
+const ContravariantTensor =
+    Union{ContravariantVector, Contravariant2Tensor}
+const CartesianTensor = Union{CartesianVector, Cartesian2Tensor}
+const LocalTensor = Union{LocalVector, Local2Tensor}
```
dennisYatunin: This doesn't look like it needs to be moved.

src/Geometry/conversions.jl
```diff
@@ -550,6 +550,10 @@ The return type when taking the divergence of a field of `V`.
 Required for statically infering the result type of the divergence operation for `AxisVector` subtypes.
 """
 @inline divergence_result_type(::Type{V}) where {V <: AxisVector} = eltype(V)
+@inline divergence_result_type(::Type{V}) where {V <: CovariantNullVector} =
+    eltype(V)
+@inline divergence_result_type(::Type{V}) where {V <: ContravariantNullVector} =
+    eltype(V)
```
dennisYatunin:
```suggestion
```
These cases are already handled by the generic `AxisVector` method above.

src/Geometry/conversions.jl
```diff
@@ -571,6 +575,9 @@ Required for statically infering the result type of the gradient operation for `
     N = length(I)
     AxisVector{V, CovariantAxis{I}, SVector{N, V}}
 end
+@inline function gradient_result_type(::Val{()}, ::Type{V}) where {V <: Number}
+    CovariantNullVector{V}
+end
```
dennisYatunin:
```suggestion
```
This is already handled by the generic `Val{I}` method above.

src/Geometry/conversions.jl
```diff
@@ -612,6 +619,12 @@ Curl is only defined for `CovariantVector`` field input types.
     ::Val{(1, 2)},
     ::Type{Covariant123Vector{FT}},
 ) where {FT} = Contravariant123Vector{FT}
+@inline curl_result_type(
+    ::Val{()},
+    ::Type{CovariantNullVector{FT}},
+) where {FT} = ContravariantNullVector{FT}
+@inline curl_result_type(::Val{()}, ::Type{<:CovariantVector{FT}}) where {FT} =
+    ContravariantNullVector{FT}
```
dennisYatunin: You've covered null curls of null and non-null vectors, but you're missing a method for non-null curls of null vectors:
```
@inline curl_result_type(_, ::Type{CovariantNullVector{FT}}) where {FT} =
    ContravariantNullVector{FT}
```
You'll probably need to keep the method for null curls of null vectors to avoid introducing an ambiguity.

src/Operators/spectralelement.jl
```diff
@@ -195,6 +197,19 @@ Base.@propagate_inbounds function copyto_slab!(out, bc, slabidx)
...
+    bc,
+    slabidx,
+) where {T}
+    space = axes(out)
+    rbc = resolve_operator(bc, slabidx)
+    @inbounds for ij in node_indices(space)
+        set_node!(space, out, ij, slabidx, get_node(space, rbc, ij, slabidx))
+    end
+    return nothing
+end
```
```suggestion
```
This is already handled by the generic method above. You'll need to remove `QS` and `Nq` from that method (since they don't exist in a column), but those values aren't used so I'm not sure why they were extracted in the first place.

src/Operators/spectralelement.jl
```diff
@@ -340,11 +355,13 @@ Base.@propagate_inbounds function get_node(
     slabidx,
 )
     space = reconstruct_placeholder_space(axes(field), parent_space)
-    i, = Tuple(ij)
-    if space isa Spaces.FaceExtrudedFiniteDifferenceSpace
+    i, = space isa Spaces.FiniteDifferenceSpace ? (1,) : Tuple(ij)
```
dennisYatunin: Won't `Tuple(ij)` already return `(1,)` for a `FiniteDifferenceSpace`?

src/Operators/spectralelement.jl
```diff
@@ -1650,3 +1668,25 @@ function rmatmul2(W, S, i, j)
...
+        WeakCurl{()},
+    },
+    space,
+    slabidx,
+    arg,
+)
+    FT = Spaces.undertype(space)
+    RT = operator_return_eltype(op, eltype(arg))
+    zero_value = zero(RT)
+    out = DataLayouts.IF{RT, 1}(MArray, FT)
```
dennisYatunin: Is `IF` always the right layout here? We might want to use `IJF` when `space` has 2 horizontal dimensions and `DataF` when `space` is a column. You can define some sort of `slab_type` function in `DataLayouts.jl` to abstract this out, if we don't already have something like that.

src/Topologies/Topologies.jl
```diff
@@ -62,21 +62,12 @@ function print_context(io::IO, context::ClimaComms.SingletonCommsContext)
...
 function print_context(io::IO, context::ClimaComms.MPICommsContext)
-    if ClimaComms.MPI.Initialized()
-        print(
-            io,
-            "MPICommsContext with ",
-            ClimaComms.nprocs(context),
-            " processes",
-        )
-    else
-        print(io, "MPICommsContext (uninitialized)")
```
dennisYatunin, line 73: Do we not need to check if MPI is initialized?

dennisYatunin (approved):
Looks good! I left a few suggestions for reducing code duplication, and it would be good to test all of this by adding column fields to loops over extruded fields in the test suite.

src/Topologies/Topologies.jl
nefrathenrici: I don't think ClimaComms has an MPI module or an `Initialized` function any more, so I removed the check and it hasn't caused any issues.

src/Operators/spectralelement.jl
```diff
@@ -1650,3 +1655,36 @@ function rmatmul2(W, S, i, j)
...
+
+
+function apply_operator(
+    op::Union{
+        Divergence{()},
+        WeakDivergence{()},
+        Gradient{()},
+        WeakGradient{()},
+        Curl{()},
+        WeakCurl{()},
```
nefrathenrici: put SpectralOperator type or find subtype

src/Geometry/conversions.jl
```diff
@@ -515,7 +515,12 @@ Curl is only defined for `CovariantVector`` field input types.
     Contravariant3Vector{FT}
 @inline curl_result_type(::Val{(1, 2)}, ::Type{Covariant123Vector{FT}}) where {FT} =
     Contravariant123Vector{FT}
-
+@inline curl_result_type(::Val{()}, ::Type{CovariantNullVector{FT}}) where {FT} =
+    ContravariantNullVector{FT}
+@inline curl_result_type(::Val{()}, ::Type{<:CovariantVector{FT}}) where {FT} =
+    ContravariantNullVector{FT}
+@inline curl_result_type(_, ::Type{CovariantNullVector{FT}}) where {FT} =
+    ContravariantNullVector{FT}
```
dennisYatunin:
```suggestion
@inline curl_result_type(_, ::Type{<:CovariantVector{FT}}) where {FT} =
    ContravariantNullVector{FT}
```
Can this just be a fallback for all combinations of spaces and input vectors in the table above?

src/Geometry/conversions.jl
```diff
@@ -515,7 +515,12 @@ Curl is only defined for `CovariantVector`` field input types.
...
 
 @inline curl_result_type(::Val{(1,)}, ::Type{Covariant1Vector{FT}}) where {FT} =
     Contravariant1Vector{FT} # not strictly correct: should be a zero Vector
```
dennisYatunin: The comment here is correct; this needs to be a `CovariantNullVector{FT}`. A generic fallback like the one I suggested above should be able to catch this.

src/Operators/spectralelement.jl
```diff
@@ -128,6 +130,11 @@ function Base.Broadcast.instantiate(sbc::SpectralBroadcasted)
             Base.Broadcast.check_broadcast_axes(axes, args...)
         end
     end
+    # For FiniteDifferenceSpace, return zeros 
+    if axes isa Spaces.FiniteDifferenceSpace
+        RT = operator_return_eltype(op, map(eltype, args)...)
+        return Fields.zeros(RT, axes)
```
```suggestion
        return Broadcast.broadcasted(Returns(zero(RT)), Fields.coordinate_field(axes))
```
`Broadcast.instantiate` should return an `AbstractBroadcasted`, not a newly allocated `Field`. Returning a `Field` here will trigger allocations, even for in-place broadcast expressions, so it would be good to replace this with a lazy representation of a `Field` of `zero(RT)`s.

src/Operators/spectralelement.jl
```diff
@@ -147,6 +154,11 @@ function Base.Broadcast.instantiate(
         axes = bc.axes
         Base.Broadcast.check_broadcast_axes(axes, args...)
     end
+    # For FiniteDifferenceSpace, return zeros 
+    if axes isa Spaces.FiniteDifferenceSpace
+        RT = operator_return_eltype(bc.f, map(eltype, args)...)
+        return Fields.zeros(RT, axes)
```
dennisYatunin: This should also be lazy.

src/Operators/spectralelement.jl
```diff
@@ -1650,3 +1665,36 @@ function rmatmul2(W, S, i, j)
...
+        out = DataLayouts.IJF{RT, Nq}(MArray, FT)
+        fill!(parent(out), zero(FT))
+        out[slab_index(1, 1)] = zero_value
+    else
+        out = DataLayouts.IF{RT, 1}(MArray, FT)
+        out[slab_index(1)] = zero_value
+    end
+
+    return Field(SArray(out), space)
+end
```
dennisYatunin, line 1685:
```suggestion
function apply_operator(op::SpectralElementOperator{()}, space, _, arg)
    RT = operator_return_eltype(op, eltype(arg))
    return map(Returns(zero(RT)), space)
end
```
This can be simplified, especially if you make the dimensions a parameter of the abstract `SpectralElementOperator` type. If you don't want to change the abstract type you can also keep the current `Union{...}` type.

src/Spaces/finitedifference.jl
```diff
@@ -36,6 +36,8 @@ staggering(space::FiniteDifferenceSpace) = getfield(space, :staggering)
 space(grid::Grids.AbstractFiniteDifferenceGrid, staggering::Staggering) =
     FiniteDifferenceSpace(grid, staggering)
 
+horizontal_space(space::FiniteDifferenceSpace) = nothing
```
dennisYatunin: I feel like it would be more consistent to make this a `PointSpace`. Then you wouldn't need to change the `Spaces` API to accept `nothing` as an input.

dennisYatunin (approved):

PR #2376, opened 2025-10-13
Name CUDA kernels based on stack trace in `auto_launch`

petebachant:
This PR adds stack trace-based CUDA kernel naming to make performance benchmarks more informative. Kernel names are computed once per session and cached in a module-level Dict.

## TODO:

- [x] Confirm we get useful results, i.e., that we don't spend too much time computing the kernel name and wash out the timeline.

## Example results from ClimaAtmos Buildkite

From [here](https://buildkite.com/clima/climaatmos-ci/builds/26622/steps/canvas?sid=019a49af-90d4-4479-909b-8c512e3aee30)

image

### DIFF
--- a/ext/cuda/cuda_utils.jl
+++ b/ext/cuda/cuda_utils.jl
@@ -4,9 +4,69 @@ import ClimaCore.DataLayouts
 import ClimaCore.DataLayouts: empty_kernel_stats
 
 const reported_stats = Dict()
+const kernel_names = IdDict()
 # Call via ClimaCore.DataLayouts.empty_kernel_stats()
 empty_kernel_stats(::ClimaComms.CUDADevice) = empty!(reported_stats)
 collect_kernel_stats() = false
+
+# Robustly parse boolean-like environment variables
+function _getenv_bool(var::AbstractString; default::Bool = false)
+    raw = get(ENV, var, nothing)
+    raw === nothing && return default
+    s = lowercase(strip(String(raw)))
+    if s in ("1", "true", "t", "yes", "y", "on")
+        return true
+    elseif s in ("0", "false", "f", "no", "n", "off")
+        return false
+    else
+        # fall back to parse as integer (non-zero -> true)
+        try
+            return parse(Int, s) != 0
+        catch
+            @warn "Unrecognized boolean env var value; using default" var = var val = raw default =
+                default
+            return default
+        end
+    end
+end
+
+# Create a ref to hold the setting determining whether to name kernels from
+# stack trace
+const NAME_KERNELS_FROM_STACK_TRACE = Ref{Bool}(false)
+
+# Always reload when module is imported so precompilation doesn't make it "stick"
+function __init__()
+    NAME_KERNELS_FROM_STACK_TRACE[] = _getenv_bool(
+        "CLIMA_NAME_CUDA_KERNELS_FROM_STACK_TRACE"; default = false,
+    )
+end
+
+name_kernels_from_stack_trace() = NAME_KERNELS_FROM_STACK_TRACE[]
+
+# Modules to ignore when constructing kernel names from stack traces
+const IGNORE_MODULES = (
+    :Base,
+    :Core,
+    :GPUCompiler,
+    :CUDA,
+    :NVTX,
+    :ClimaCoreCUDAExt,
+    :ClimaCore,
+)
+
+# Helper function to check if a stack frame is relevant
+@inline function is_relevant_frame(frame::Base.StackTraces.StackFrame)
+    linfo = frame.linfo
+    linfo isa Core.MethodInstance || return false
+    mod = linfo.def.module::Module
+    mod_name = fullname(mod)[1]
+    return mod_name ∉ IGNORE_MODULES
+end
+
+# Extract file path from a MethodInstance as a string
+@inline function fpath_from_method_instance(mi::Core.MethodInstance)
+    return string(mi.def.file::Symbol)::String
+end
 
 """
     auto_launch!(f!::F!, args,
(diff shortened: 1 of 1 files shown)


imreddyTeja:
> ook suspiciously different. Reminds me of Simon's comment about the CUDA API not doing much in the previous implementation that computed the name every time a kernel was created. Do we need to somehow hydrate the kernel name cache initially to get a useful result here?

Are these profile results from the first step, or a later step? The kernel renaming example takes significantly longer, which makes me think `stacktrace` is getting called.

petebachant:
I think I figured this out. The compile step was running before kernel renaming was enabled.

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +41,60 @@ function auto_launch!(
...
+            kernel_names[key] = kernel_name
+        end
+        kernel_name = kernel_names[key]
+    end
+
     if auto
         @assert !isnothing(nitems)
         if nitems ≥ 0
-            kernel = CUDA.@cuda always_inline = true launch = false f!(args...)
+            kernel = CUDA.@cuda name = kernel_name always_inline = true launch =
```
imreddyTeja, line 157: It might be good to add a note somewhere that passing in `name = nothing` here will replicate default behavior. That behavior is not documented in the CUDA.jl docs.

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +41,60 @@ function auto_launch!(
...
+                frame.linfo isa Core.MethodInstance && (
+                    fullname(frame.linfo.def.module)[1] ∉ uninteresting_modules
+                )
+            end
+            if !isnothing(first_relevant_index)
+                frame = stack[first_relevant_index]
+                name_str =
+                    string(frame.func) *
+                    "_" *
+                    string(frame.linfo.def.file) *
```
imreddyTeja: It might make the names more readable if either the file path is not included in the name, or only the part after the package dir is included. For example, just `ClimaCore.jl/src/Fields/Fields.jl` or `Fields.jl` instead of `/home/treddy/ClimaCore.jl/src/Fields/Fields.jl`

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +41,60 @@ function auto_launch!(
...
+                    "_" *
+                    string(frame.line)
```
imreddyTeja: It might be a good idea to either handle the case where `frame.linfo.def.file` is inside NVTX.jl specially, or to leave a note here that filenames and line numbers of kernels will be incorrect if inside an NVTX annotation.

petebachant: What do you think about the format below? If it's an NVTX annotation, it's easy to see that. Likewise, it's easy to tell if the name is a function defined in a file:

image

imreddyTeja: I like this a lot. If a lot of kernels are getting caught in the NVTX annotations, it might make sense to reduce the number of NVTX annotations in downstream packages

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +42,82 @@ function auto_launch!(
...
     caller = :unknown,
 ) where {F!}
+    # If desired, compute a kernel name from the stack trace and store in
+    # a global Dict, which serves as an in memory cache
+    kernel_name = nothing
+    if name_kernels_from_stack_trace()
+        # Create a key from the method instance and types of the args
+        key = objectid(methodinstance(typeof(f!), typeof(args)))
+        kernel_name_exists = key in keys(kernel_names)
+        if !kernel_name_exists
```
imreddyTeja, line 109: These syntax changes may be less clear, so feel free to ignore

```
        kernel_name_exists =  haskey(kernel_names, key)
        if !kernel_name_exists
```

or 

```
kernel_name = get!(kernel_names, key) do
#calculate_name_here
end
```

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +42,82 @@ function auto_launch!(
...
+            end
+            if !isnothing(first_relevant_index)
+                # Don't include file if this is inside an NVTX annotation
+                frame = stack[first_relevant_index]
+                func_name = string(frame.func)
+                if contains(func_name, "#")
+                    func_name = split(func_name, "#")[1]
+                end
+                file_path = frame.linfo.def.file
+                fp_split = split(string(file_path), "/")
```
imreddyTeja:
```suggestion
                fp_split = splitpath(string(file_path))
```

also I think `file_path` may already by a string

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +42,82 @@ function auto_launch!(
...
+                if "NVTX" in fp_split
+                    fp_string = "_NVTX"
+                    line_string = ""
+                else
+                    # Trim base directory off of file path to shorten
+                    package_index = findfirst(fp_split) do part
+                        startswith(part, "Clima")
+                    end
+                    if isnothing(package_index)
+                        package_index = findfirst(p -> p == ".julia", fp_split)
```
imreddyTeja, line 131: Should this be searching for `.jl` instead of `.julia`?

ext/cuda/cuda_utils.jl
petebachant: I believe it's a `Symbol` at this point, but I should double check. If only the Julia language server could analyze types in VS Code...

ext/cuda/cuda_utils.jl
petebachant: I did it this way because I believe the `.jl` is omitted from the package name when it's installed under the `.julia/packages` directory. If I search for `.jl`, in some cases I'll only get the file name, not the path in the package.

petebachant:
Breaking on `src` if we can't find any other tokens:

image

ext/cuda/cuda_utils.jl
```diff
@@ -39,19 +42,85 @@ function auto_launch!(
     always_inline = true,
     caller = :unknown,
 ) where {F!}
+    # If desired, compute a kernel name from the stack trace and store in
+    # a global Dict, which serves as an in memory cache
+    kernel_name = nothing
+    if name_kernels_from_stack_trace()
+        # Create a key from the method instance and types of the args
+        key = objectid(methodinstance(typeof(f!), typeof(args)))
```
imreddyTeja:
```suggestion
        key = objectid(methodinstance(F!, typeof(args)))
```

The function signature could also be changed to 

```julia
function auto_launch!(
    f!::F!,
    args::ARGS,
    nitems::Union{Integer, Nothing} = nothing;
    auto = false,
    threads_s = nothing,
    blocks_s = nothing,
    always_inline = true,
    caller = :unknown,
) where {F!, ARGS}
.
.
.

 key = objectid(methodinstance(F!, ARGS))

```

but I think that forced specialization of the method (I'm not sure if that is desirable or not here)

ext/cuda/cuda_utils.jl
```diff
@@ -2,8 +2,11 @@ import CUDA
 import ClimaCore.Fields
 import ClimaCore.DataLayouts
 import ClimaCore.DataLayouts: empty_kernel_stats
+import ClimaCore.DebugOnly: name_kernels_from_stack_trace
+import CUDA.GPUCompiler: methodinstance
```
imreddyTeja: Not sure if importing a package through another package is recommended or not, and I can't find anything when I look online.

petebachant: It does feel wrong. Would it be smarter to make GPUCompiler an explicit dependency of ClimaCore? Is there a nice way to make it optional for just the CUDA extension?

petebachant: Reading a little more about this, I guess it would go in `weakdeps`?

imreddyTeja: I think [ClimaLand](https://github.com/CliMA/ClimaLand.jl/blob/be4b3eb348640bf0883d84de639ba9c33a8c6680/Project.toml#L26-L42) is a good example of using extensions/weak reps. I'm not sure if there is a way to manage extensions with the package manager though (I never found a way for Julia 1.10), so you might need to edit the Project.toml by hand

petebachant: I was able to add the weak dep to ClimaCore with `add --weak` in Julia v1.11, but I don't seem to be able to add it to the extension.

ext/cuda/cuda_utils.jl
```diff
@@ -2,12 +2,40 @@ import CUDA
 import ClimaCore.Fields
 import ClimaCore.DataLayouts
 import ClimaCore.DataLayouts: empty_kernel_stats
+# Import from CUDA.GPUCompiler, since we can't depend on GPUCompiler directly
+# in an extension package
+import CUDA.GPUCompiler: methodinstance
```
petebachant: This would probably be possible with the correct definitions in `Project.toml`, but is it worth adding all over the place?

dennisYatunin: It looks like CUDA.jl imports `methodinstance` internally, so you can just write `CUDA.methodinstance` to be consistent with the other `CUDA._` function calls in this file.

I suppose this isn't a great practice in general, as it involves making use of another package's internal implementation details, but in this case I think it's fine. GPUCompiler.jl is a fundamental part of CUDA.jl, so that internal import is unlikely to go away any time soon. And yes, Julia doesn't allow us to add dependencies for extensions, at least until [this issue](https://github.com/JuliaLang/Pkg.jl/issues/3641) is fixed.

imreddyTeja:
> I was finally able to get the JET tests to pass while retaining the ability to set this via environmental variable. It's not pretty, but it saves a bit of code downstream...

Does this allow a user to turn renaming on/off while a repl is active, and without restarting it?

Issue #1650, opened 2024-03-20
Inference failure when broadcasting `mapreduce` on gpus


charleskawczynski:
The following kernel fails on gpus in ClimaAtmos:

```julia
    @. u³⁰_halflevel = divide_by_ρa(
        ρ_level * u³_halflevel - mapreduce(*, +, ρaʲs_level, u³ʲs_halflevel),
        ρ_level,
        ρ_level * u³_halflevel,
        ρ_level,
        turbconv_model,
    )
```

Seen in this build: https://buildkite.com/clima/climaatmos-ci/builds/17608#018e5841-01f3-4410-a32d-a1276535c614

We could technically get this running by either
 - Removing `always_inline` from the finite difference kernels in ClimaCore, or
 - Figuring out a way to get mapreduce to work on the gpu with `always_inline`

Since inlining has proven to be a performance benefit (and it may be necessary for broadcast fusion to benefit), we should opt for solution 2.

charleskawczynski:
xref: https://github.com/CliMA/ClimaAtmos.jl/issues/2365

charleskawczynski:
For now, we can write custom unrolled workarounds, but it might be interesting to see if we can make a reproducer for this, since it seems to indicate a compiler limitation.

charleskawczynski:
I've removed this issue from the milestone since we were able to write a workaround.

PR #2390, opened 2025-11-13
Release Patch 0.14.42

imreddyTeja:
Add NEWS.md entry for kernel renaming.
Update .buildkite manifest.

Issue #1650 has a broken test that now passes. Should this issue be closed?

### DIFF
--- a/.buildkite/Manifest.toml
+++ b/.buildkite/Manifest.toml
@@ -1,13 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.9"
+julia_version = "1.10.8"
 manifest_format = "2.0"
 project_hash = "f124967e8a7148176db47cb208ec29c14f382b0f"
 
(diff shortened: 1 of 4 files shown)


imreddyTeja:
> Looks like updating the manifest must have broken the JET tests 😠 StackFrame is a hard type to deal with.

I think we might want to fix this. It is a bit worrying that JET is detecting problems even when the renaming option is not enabled. If we determine that this is OK, then we should add `auto_launch` to a list of functions for JET to ignore

petebachant:
> I think we might want to fix this. It is a bit worrying that JET is detecting problems even when the renaming option is not enabled. If we determine that this is OK, then we should add `auto_launch` to a list of functions for JET to ignore

I think the risk of a performance hit is very low, since the detected runtime dispatch is in the non-default path, but I can take a look and see if I can get the tests to pass anyway with the right allocations/annotations.

dennisYatunin (approved):

PR #2391, opened 2025-11-14
Extract kernel name creation into its own function

petebachant:
Trying to deal with type instability issues causing JET tests to fail. Can't reproduce them locally or on `clima`, so just giving it a shot here.


petebachant:
This isn't passing the JET tests, so I'm going to close.

PR #2396, opened 2025-11-24
Tr/mem access patterns

imreddyTeja:
When a low resolution simulation is ran, the solver, which runs one thread per column, will not saturate the gpu. With the current launch config, a portion of the multiprocessors remain unused. This adds a function to catch that case, and use smaller block sizes.

Threads for stencils and pointwise operations were being launched by converting a single block and single grid index into a linear index, and then converting that into a cartesian index of (I, J, V, H), but the data is stored as (VIJFH). This adds special cases for v=63 and v=64, when i=j=4, that uses multiple block and grid dimensions, and indexes so adjacent threads are likely in the same column. Profiling shows this massively increases cache access patterns.
Before: 2.0 of the 32 bytes transmitted per sector are utilized (L2)
After: 26.2 of the 32 bytes transmitted per sector are utilized (L2)

These changes to not get fully tested within ClimaCore because most of the tests running on gpus are at a lower vertical and horizontal resolution. It has been tested with ClimaAtmos and ClimaCoupler, as well as tested in ClimaCore by temporarily running the tests at a higher resolution.

### DIFF
--- a/ext/cuda/data_layouts_copyto.jl
+++ b/ext/cuda/data_layouts_copyto.jl
@@ -21,6 +21,20 @@ function knl_copyto_linear!(dest, src, us)
     return nothing
 end
 
+"""
+    knl_copyto_VIJFH_64!(dest, src, ::Val{P})
+
+Kernel for pointwise broadcasts on VIJFHStyle{63,4} and VIJFHStyle{64,4} datalayouts. P is a boolean
+indicating if the column is padded (true for 63, false for 64).
+"""
+function knl_copyto_VIJFH_64!(dest, src, ::Val{P}) where {P}
+    # P is a boolean, indicating if the column is padded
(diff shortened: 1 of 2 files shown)


ext/cuda/data_layouts_copyto.jl
```diff
@@ -104,6 +118,44 @@ else
     end
 end
 
+# Specialized kernel launch for VIJFHStyle{63,4} and VIJFHStyle{64,4} arrays. This uses block and grid indices
+# instead of computing cartesian indices from a linear index. The threads are launched so that
+# a set 64 threads covers a column.
```
petebachant, line 123: How do you decide when to use a simple comment versus a docstring for a function?

petebachant (approved):

PR #2400, opened 2025-12-04
add ConservativeRegridding extension

### LINKED ISSUE #2398: Write a function to convert 2D ClimaCore space to vector of coordinates
To interface with ConservativeRegridding.jl
Part of #2397 

requirements:
- each element of the space should be represented as one polygon
- the list of vertices of the polygon should start in one corner and continue clockwise, and repeat the first vertex at the end
- the order of vertices/elements should be recoverable later on
- at the end, the space should be represented as a vector of a vector of coordinate tuples

juliasloan25:
Prototypes regridding using ConservativeRegridding with FV approximation

Closes #2398 

## Content
- [x] Add extension to ClimaCore; update Project deps
- [x] Copy functions over from ConservativeRegridding.jl [test](https://github.com/JuliaGeo/ConservativeRegridding.jl/blob/main/test/climacore.jl): `get_element_vertices`, `integrate_each_element`, `get_value_per_element!`, `set_value_per_element!`
- [x] Copy test from ConservativeRegridding.jl test, update to use new functions
- [x] Add a Regridder constructor for ClimaCore spaces (and fields?)
- [x] Extend `regrid!` for ClimaCore fields - think about intermediate fields

## New method ambiguities
This PR increases the number of method ambiguities because ConservativeRegridding.jl pulls in GeometryOps, leading to more ambiguities of `broadcasted`. For example, in GeometryOps, they define `broadcasted(f, a::AbstractArray{T}, b::GeometryOps.UnitSpherical.UnitSphericalPoint)` and in ClimaCore, we define  `broadcasted(op::ClimaCore.Operators.FiniteDifferenceOperator, args...)`. Thank you to @ph-kev for finding this.

## A note on zero-area elements when using an odd number of elements
When using a space with odd `h_elem` (but not even) we encounter zero-area polygons. This is because when we have odd h_elem, there are elements centered around the north and south poles. Their vertices each lie along a line of constant latitude. When we then compute planar area of this element, we get 0 area, but if we were to compute the correct spherical area it would be nonzero.

ConservativeRegridding.jl currently computes simple planar areas, so it currently fails when using a space with even `h_elem`. This should be resolved once we switch to computing spherical areas (coming soon).

Thank you to @imreddyTeja and @ph-kev for tracking this down!

#### Source field
field1

#### Source field with one value per element
field1_one_value_per_element

#### Destination field (remapped)
field2

### DIFF
--- a/test/aqua.jl
+++ b/test/aqua.jl
@@ -20,7 +20,7 @@ using Aqua
     # then please lower the limit based on the new number of ambiguities.
     # We're trying to drive this number down to zero to reduce latency.
     # Uncomment for debugging:
-    n_existing_ambiguities = 25
+    n_existing_ambiguities = 28
     if !(length(ambs) ≤ n_existing_ambiguities)
         for method_ambiguity in ambs
             @show method_ambiguity


test/ClimaCoreConservativeRegriddingExt/test_conservative_regridding.jl
```diff
@@ -0,0 +1,128 @@
...
+using ClimaCore
+using ClimaCore:
+    CommonSpaces, Remapping, Fields, Spaces, RecursiveApply, Meshes, Quadratures
+using ConservativeRegridding
+
+# Extensions load lazily in Julia - they need to be explicitly triggered
+# This ensures the extension is loaded and its methods are available
+const ClimaConservativeRegrid =
+    Base.get_extension(ClimaCore, :ClimaCoreConservativeRegriddingExt)
+@assert !isnothing(ClimaConservativeRegrid) "ClimaCoreConservativeRegriddingExt extension not loaded"
```
juliasloan25: this is not right

Project.toml
```diff
@@ -8,9 +8,11 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
+ClimaCoreMakie = "908f55d8-4145-4867-9c14-5dad1a479e4d"
```
juliasloan25: we shouldn't need to add plotting packages

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+
+    # Put each polygon into a vector, with the first coordinate pair repeated at the end
+    vertices = collect(Iterators.partition(vertex_coords, 5))
+
+    # Check for zero area polygons (all latitude or longitude values are the same)
+    for polygon in vertices
+        if allequal(first.(polygon)) || allequal(last.(polygon))
+            @error "Zero area polygon found in vertices" polygon
+        end
+    end
```
ph-kev, line 51: I think the error message can be better here. They are only zero area polygons, because the polygons are on a longitude-latitude grid. If you view the polygons on a sphere, then they are not zero area. It might also be helpful to point out that the current implementation of ConservativeRegridding does not work with these zero area polygons in the docstring or error message.

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+"""
+function Remapping.get_element_vertices(space)
+    # Get the indices of the vertices of the elements, in clockwise order for each element
+    Nh = Meshes.nelements(space.grid.topology.mesh)
+    Nq = Quadratures.degrees_of_freedom(Spaces.quadrature_style(space))
+    vertex_inds = [
+        CartesianIndex(i, j, 1, 1, e) # f and v are 1 for SpectralElementSpace2D
+        for e in 1:Nh
+        for (i, j) in [(1, 1), (1, Nq), (Nq, Nq), (Nq, 1), (1, 1)]
+    ] # repeat the first coordinate pair at the end
```
ph-kev, line 34: Does this only work with a `SpectralElementSpace2D`? If it is, can you add a type annotation for that?

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+to the number of elements in the space.
+
+Here `ones_field` is a field on the same space as `field` with all
+values set to 1.
+"""
+function Remapping.get_value_per_element!(
+    value_per_element,
+    field,
+    ones_field,
+)
```
ph-kev, line 100: Is this function meant to be used by ClimaCore users or is it internal? If it is the latter, can you rename this to `_get_value_per_element!`? If it the former, can you add a check for `ones_field` that the field contains all ones or refactor this to remove that argument?

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+        end
+    end
+    return field
+end
+
+"""
+    Regridder(dst_space, src_space; kwargs...)
+
+Create a regridder between two ClimaCore Spaces.
+This works by finding the vertices of the elements of the source and
```
ph-kev, line 139:
```suggestion
Create a regridder between two ClimaCore spaces.
This works by finding the vertices of the elements of the source and
```

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+### rather than one value per node. Note that this will not be our long-term solution.
+"""
+    integrate_each_element(field)
+
+Integrate the field over each element of the space.
+Returns a vector with length equal to the number of elements in the space,
+containing the integrated value over the nodes of each element.
+"""
+function Remapping.integrate_each_element(field)
+    space = axes(field)
```
ph-kev, line 65: Can you add a type annotation for `field`?

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+    regrid!(dst_field, regridder_tuple, src_field)
+
+Perform conservative regridding from `src_field` to `dst_field` using a
+Regridder object and pre-allocated buffers.
+
+The `regridder_tuple` should be a NamedTuple containing:
+- `regridder`: The ConservativeRegridding.Regridder object
+- `value_per_element_src`: Pre-allocated buffer for source values
+- `value_per_element_dst`: Pre-allocated buffer for destination values
+- `ones_src`: Pre-allocated field of ones on the source space
```
ph-kev, line 173: Is it possible to encapsulate this in the `regridder` object? It seems error prone to expect an user to make this `NamedTuple`. Alternatively, is it possible to make a helper function for constructing this?

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+This is currently only defined for 2D spaces, but could be extended to
+3D spaces.
+
+This function is intended to be used with the finite volume approximation
+of the ClimaCore spectral element space.
+"""
+function ConservativeRegridding.Regridder(
+    dst_space::Spaces.SpectralElementSpace2D,
+    src_space::Spaces.SpectralElementSpace2D;
+    kwargs...,
```
ph-kev, line 151: What are the keyword arguments? I couldn't find them.

ext/ClimaCoreConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,232 @@
...
+
+This is a convenience function that allocates the buffers for you.
+Note that this is not efficient for repeated regridding with the same regridder,
+but it may be helpful for one-off regriddings or testing/debugging.
+"""
+function ConservativeRegridding.regrid!(
+    dst_field::Fields.Field,
+    regridder::ConservativeRegridding.Regridder,
+    src_field::Fields.Field,
+)
```
ph-kev, line 213: I am not sure if it is worth doing, but if you think the backend (what interpolator is being used) will change,  you can put this regridder in another struct and have the user be exposed to that as the public interface. Then, developers can change what is happening behind the scene without breaking anything.

PR #2403, opened 2025-12-11
add dispatch back in to_device

juliasloan25:
Revert the dispatch removal from #2375 to avoid errors from calling `Adapt.adapt` on types it isn't defined for - e.g. https://buildkite.com/clima/climacoupler-longruns/builds/991#019b0f73-6073-4684-a0f7-b1ba9eb91e0f

Also tags a patch release of 0.14.44

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,11 @@ main
 -------
 
 
+v0.14.44
+-------
+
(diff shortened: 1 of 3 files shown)


nefrathenrici (approved):
Thanks!

PR #2409, opened 2025-12-12
Add split divergence operator

nefrathenrici:

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -655,7 +655,18 @@ The split form is defined as the arithmetic mean of the conservative form and th
 ```
 
 ## Discrete (SBP)
-The implementation uses the Summation-By-Parts (SBP) property of the spectral element derivative matrix ``D``. The divergence is computed locally on the reference element as:
+The implementation uses the generalized summation-by-parts (SBP) property, where the divergence is the inverse of the mass matrix acting on the stiffness matrix. Let \\delta V_i = w_i J_i be the physical volume element (weights w, Jacobian J) and Q be the skew-symmetric stiffness matrix. The divergence is:
+```math
+(Div_{split})_i = \\frac{1}{\\delta V_i} \\sum_j Q_{ij} F_{ij}
+```
+where F_{ij} is the symmetric two-point flux.
+
+**Reduction to differentiation matrix form:** For Legendre-Gauss-Lobatto (LGL) elements, the stiffness matrix satisfies Q_{ij} = w_i D_{ij}. Substituting this into the equation above, the quadrature weights w_i cancel out:
+```math 
+\\frac{1}{w_i J_i} \\sum_j (w_i D_{ij}) F_{ij} = \\frac{1}{J_i} \\sum_j D_{ij} F_{ij}
+```
+
+This is the form implemented in the code:
 ```math
 (Div_{split})_i = \\frac{1}{J_i} \\sum_j D_{ij} F_{ij}
 ```
@@ -663,7 +674,7 @@ where ``F_{ij}`` is the symmetric **two-point flux** between nodes ``i`` and ``j
 ```math
 F_{ij} = \\frac{1}{2} \\left( U_i + U_j \\right) (\\psi_i + \\psi_j)
 ```
-Here, ``U`` represents the volume-weighted contravariant flux component (``U^1 = J u^1``). In 2D/3D tensor-product grids, this 1D split operator is 
+Here, ``U`` represents the Jacobian-weighted contravariant flux component (``U^1 = J u^1``). In 2D/3D tensor-product grids, this 1D split operator is 
 applied sequentially along each dimension.
 
 # Properties
@@ -754,8 +765,8 @@ function apply_operator(op::SplitDivergence{(1, 2)}, space, slabidx, arg1, arg2)
             val = zero(FT)
             for k in 1:Nq
                 flux_ik =
-                    0.5 * (Ju1_slab[i, j] + Ju1_slab[k, j]) *
-                    (psi_slab[i, j] + psi_slab[k, j])
+                    (Ju1_slab[i, j] + Ju1_slab[k, j]) *
+                    (psi_slab[i, j] + psi_slab[k, j]) / 2
                 val += D[i, k] * flux_ik
             end
             out[slab_index(i, j)] += val
@@ -767,8 +778,8 @@ function apply_operator(op::SplitDivergence{(1, 2)}, space, slabidx, arg1, arg2)
             val = zero(FT)
             for k in 1:Nq
                 flux_jk =
-                    0.5 * (Ju2_slab[i, j] + Ju2_slab[i, k]) *
-                    (psi_slab[i, j] + psi_slab[i, k])
+                    (Ju2_slab[i, j] + Ju2_slab[i, k]) *
+                    (psi_slab[i, j] + psi_slab[i, k]) / 2
                 val += D[j, k] * flux_jk
             end
             out[slab_index(i, j)] += val
@@ -799,7 +810,7 @@ for all ``\\phi\\in \\mathcal{V}_0``
 ```
 where ``\\mathcal{V}_0`` is the space of ``u``.
 
-This arises as the contribution of the volume integral after by applying
+This arises as the contribution of the volume integral after applying
 integration by parts to the weak form expression of the divergence
 ```math
 \\int_\\Omega \\phi (\\nabla \\cdot u) \\, d \\Omega
--- a/test/Operators/spectralelement/split_divergence.jl
+++ b/test/Operators/spectralelement/split_divergence.jl
@@ -20,26 +20,51 @@ include(
 import .TestUtilities as TU
 
 split_div = Operators.SplitDivergence()
-div = Operators.Divergence()
+wdiv = Operators.WeakDivergence()
 
 function create_test_fields(space, FT)
     coords = Fields.coordinate_field(space)
-    u = @. Geometry.UVVector(cos(coords.x) * sin(coords.y), sin(coords.x) * cos(coords.y))
-    psi = @. FT(2) + sin(coords.x) * cos(coords.y)
+    # Generic test fields
+    if space isa Spaces.SpectralElementSpace2D
+        # Trigonometric functions in Cartesian and spherical domains
+        if eltype(coords) <: Geometry.LatLongPoint
+            lon = coords.long
+            lat = coords.lat
+            u = @. Geometry.UVVector(cosd(lon), sind(lat))
+            psi = @. FT(2) + sind(lon) * cosd(lat)
+        else
+            x = coords.x
+            y = coords.y
+            u = @. Geometry.UVVector(cos(x) * sin(y), sin(x) * cos(y))
+            psi = @. FT(2) + sin(x) * cos(y)
+        end
+    else
+        error("Unsupported space")
+    end
     return u, psi
 end
 
-function SpectralElem2D(FT, context, x2periodic)
+function SpectralElem2D(FT, context, Ne, Nq)
+    # Doubly periodic Cartesian domain
     domain = Domains.RectangleDomain(
         Geometry.XPoint{FT}(-pi) .. Geometry.XPoint{FT}(pi),
         Geometry.YPoint{FT}(-pi) .. Geometry.YPoint{FT}(pi);
         x1periodic = true,
-        x2periodic,
+        x2periodic = true,
         x1boundary = nothing,
-        x2boundary = x2periodic ? nothing : (:south, :north),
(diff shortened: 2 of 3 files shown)


examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -112,6 +113,8 @@ function default_cache(ᶜlocal_geometry, ᶠlocal_geometry, Y, upwinding_mode)
         ᶜK = similar(ᶜlocal_geometry, FT),
         ᶜΦ = grav .* ᶜcoord.z,
         ᶜp = similar(ᶜlocal_geometry, FT),
+        ᶜΠ = similar(ᶜlocal_geometry, FT),
+        ᶜθ = similar(ᶜlocal_geometry, FT),
```
szy21: remove this

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -219,38 +222,38 @@ function default_remaining_tendency!(Yₜ, Y, p, t)
     ᶜρ = Y.c.ρ
     ᶜuₕ = Y.c.uₕ
     ᶠw = Y.f.w
-    (; ᶜuvw, ᶜK, ᶜΦ, ᶜp, ᶜω³, ᶠω¹², ᶠu¹², ᶠu³, ᶜf) = p
+    (; ᶜuvw, ᶜK, ᶜΦ, ᶜp, ᶜΠ, ᶜθ, ᶜω³, ᶠω¹², ᶠu¹², ᶠu³, ᶜf) = p
```
szy21: revert this

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -226,31 +227,31 @@ function default_remaining_tendency!(Yₜ, Y, p, t)
     @. ᶜK = norm_sqr(ᶜuvw) / 2
 
     # Mass conservation
-
-    @. Yₜ.c.ρ -= divₕ(ᶜρ * ᶜuvw)
+    FT = eltype(Yₜ.c.ρ)
+    @. Yₜ.c.ρ -= split_divₕ(ᶜρ * ᶜuvw, FT(1))
```
dennisYatunin:
```suggestion
    @. Yₜ.c.ρ -= split_divₕ(ᶜρ * ᶜuvw, 1)
```
This operator should be able to automatically promote integers to floating-point values.

Based on past experiments, specifying `FT` conversions in broadcast expressions incurs a higher latency penalty than performing automatic conversions when evaluating each point. While automatic conversions are easy for the compiler to optimize out, explicit conversions are much harder because they need to be constant-propagated through the entire broadcasting API.

test/Operators/spectralelement/split_divergence.jl
```diff
@@ -0,0 +1,98 @@
...
+
+            split_result = split_div.(u, psi)
+            Spaces.weighted_dss!(split_result)
+
+            div_result = div.(u .* psi)
+            Spaces.weighted_dss!(div_result)
+
+            # Test comparison - not sure what tolerance to use here
+            max_val = max(norm(split_result), norm(div_result), one(FT))
+            @test norm(split_result .- div_result) < FT(0.1) * max_val
```
nefrathenrici: This fails with 0.01:
```
Test Failed at REPL[107]:1
  Expression: norm(split_result .- div_result) < FT(0.01) * max_val
   Evaluated: 0.057408232f0 < 0.015798254f0
```

src/Operators/spectralelement.jl
```diff
@@ -631,6 +631,136 @@ Base.@propagate_inbounds function apply_operator(
...
+
+The discrete implementation uses a symmetric two-point flux formulation:
+```math
+(Div_{split})_i = \\frac{1}{J_i w_i} \\sum_j D_{ij} F_{ij}
+```
+where the flux ``F_{ij} = F_{ji}`` between quadrature points ``i`` and ``j`` is given by:
+```math
+F_{ij} = \\frac{1}{2} (J_i u^1_i + J_j u^1_j) (\\psi_i + \\psi_j)
+```
+Here, ``u^1_i`` denotes the first contravariant component of ``u`` at node ``i``. In 2D, the flux includes contributions from both contravariant components.
```
````suggestion
The discrete implementation uses a symmetric two-point flux formulation. For example, the one-dimensional divergence at quadrature point ``i`` is given by
```math
(\\nabla \\cdot \\psi u)_i|_{split} = \\frac{1}{WJ_i} \\sum_j D_{ij} WJ_j F^1_{ij},
```
where ``F^1_{ij}`` is the first contravariant component of the flux ``F_{ij} = F_{ji}`` between quadrature points ``i`` and ``j``:
```math
F^1_{ij} = \\frac{1}{2} (u^1_i + u^1_j) (\\psi_i + \\psi_j).
```
In two dimensions, the divergence includes contributions from both contravariant components.
````
This docstring can be made a little clearer, but more importantly the first equation is missing a factor of `WJ_j` and the second equation has unnecessary factors of `J_i` and `J_j`.

File: src/Operators/spectralelement.jl [hunk 6]
```diff
@@ -631,6 +631,136 @@ Base.@propagate_inbounds function apply_operator(
...
+
+"""
+struct SplitDivergence{I} <: SpectralElementOperator{I} end
+SplitDivergence() = SplitDivergence{()}()
+SplitDivergence{()}(space) = SplitDivergence{operator_axes(space)}()
+
+operator_return_eltype(::SplitDivergence{I}, ::Type{S1}, ::Type{S2}) where {I, S1, S2} =
+    RecursiveApply.rmaptype(Geometry.divergence_result_type, S1)
```
```suggestion
    Geometry.rmul_return_type(
        RecursiveApply.rmaptype(Geometry.divergence_result_type, S1),
        S2,
    )
```
This probably needs to account for the element type of the second argument, since it might not always be an `AbstractFloat` (e.g., it could be a `Dual` or a `NamedTuple`).

At some point, we should replace `operator_return_eltype` with the built-in `Base._return_type`, so that we can avoid such messy code duplication (in this case, expressing the fact that we are doing a divergence and a multiplication in two different ways, once for types and once for values).

test/Operators/spectralelement/split_divergence.jl
tapios: This looks fine at the coarse (4x4) resolution you have. The difference should decrease with resolution.

tapios: I checked that the error indeed converges with resolution.

src/Operators/spectralelement.jl
```diff
@@ -631,6 +631,156 @@ Base.@propagate_inbounds function apply_operator(
...
+        psi = get_node(space, arg2, ij, slabidx)
+        Ju1 = local_geometry.J * Geometry.contravariant1(u, local_geometry)
+        Ju1_slab[i] = Ju1
+        psi_slab[i] = psi
+    end
+
+    @inbounds for i in 1:Nq
+        val = zero(FT)
+        for j in 1:Nq
+            flux_ij = 0.5 * (Ju1_slab[i] + Ju1_slab[j]) * (psi_slab[i] + psi_slab[j])
```
```suggestion
            flux_ij = (Ju1_slab[i] + Ju1_slab[j]) * (psi_slab[i] + psi_slab[j]) / 2
```
0.5 is a `Float64`, so this is a type instability when `FT = Float32`

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin:
```suggestion
    @. Yₜ.c.ρ -= split_wdivₕ(ᶜρ * ᶜuvw, FT(1))
```
This is very outdated (it was last changed in #606), and it has not been updated to use the weak divergence operator, which we've been using to compute horizontal flux divergences since CliMA/ClimaAtmos.jl#2091. Using the strong divergence operator in this tendency is incorrect, as it violates conservation of mass. The same goes for all other horizontal flux divergences in this file.

It's fine to define the strong `split_divₕ` operator you're using here, but make sure to also define the analogous weak `split_wdivₕ` operator required by ClimaAtmos.

src/Operators/spectralelement.jl
dennisYatunin: To get conservation of mass/energy/tracers for the weak split divergence operator, the expression for `flux_ij` needs to look more like
```julia
    flux_ij = WJ_slab[j] * (u1_slab[i] + u1_slab[j]) * (psi_slab[i] + psi_slab[j]) / 2
```
I don't know what properties the strong split divergence operator needs to satisfy (@tapios?), but I'm guessing that it should also use the value of `J` from node `j`, not the values from both nodes `i` and `j`, i.e.,
```julia
    flux_ij = J_slab[j] * (u1_slab[i] + u1_slab[j]) * (psi_slab[i] + psi_slab[j]) / 2
```

src/Operators/spectralelement.jl
```diff
@@ -631,6 +631,156 @@ Base.@propagate_inbounds function apply_operator(
...
+        psi_slab[i, j] = psi
+    end
+
+    @inbounds for j in 1:Nq
+        for i in 1:Nq
+            val = zero(FT)
+            for k in 1:Nq
+                flux_ik =
+                    0.5 * (Ju1_slab[i, j] + Ju1_slab[k, j]) *
+                    (psi_slab[i, j] + psi_slab[k, j])
```
dennisYatunin: This line and the corresponding one for `flux_jk` have the same `Float32`-`Float64` type instability, and they should also probably only use the value of `J` from node `j`.

test/Operators/spectralelement/split_divergence.jl
```diff
@@ -0,0 +1,104 @@
...
+            div_result = div.(u .* psi)
+            Spaces.weighted_dss!(div_result)
+
+            # Test comparison
+            max_val = max(norm(split_result), norm(div_result), one(FT))
+            @test norm(split_result .- div_result) < FT(0.1) * max_val
+
+            # Test conservation - integral of divergence over periodic domain should be zero
+            integral = sum(split_result)
+            @test abs(integral) < 100 * eps(FT)
```
dennisYatunin: I think you'll be able to get a lower bound if you use the weak form of the split divergence operator here.

dennisYatunin (commented):
The writeup that @tapios shared on Slack, which I used to obtain a proof of conservation of mass/energy/tracers, uses the weak form of the split divergence operator. This PR only implements a strong form of the split divergence operator, and it also uses values of `J` at different nodal points than the writeup. Please let me know if there was an issue with the equations in the writeup, but I'm pretty sure the writeup was correct and this PR should be updated to reflect it.

src/Operators/spectralelement.jl
dennisYatunin: Note: I wrote this comment before the split divergence operator was apparently updated from its weak form to its strong form. The equations I provided here are what I used to derive conservation of mass/energy/tracers.

tapios: Yes, we should use the weak form, as in the write-up. I can look at it later tonight. However, in the meantime, feel free to push changes directly to the branch.

src/Operators/spectralelement.jl
tapios: I do think the Jacobians need to locally multiply the flux terms, as we have it. Let's discuss offline.

src/Operators/spectralelement.jl
tapios: Fixed type instability

src/Operators/spectralelement.jl
tapios: Upon further thought, I think what we have here is correct. And I also think the strong and weak forms we have are discretely equivalent because of the SBP property D^T = - W D W^{-1}, with diagonal quadrature weight matrix W (on GLL points) and differentiation matrix D. Let's discuss offline.

tapios: The split form we have is correct. There is some ongoing discussion with @dennisYatunin about inclusion of quadrature weights, but I am relatively confident they drop out.

src/Operators/spectralelement.jl
tapios: We have converged (offline) that this form of including the Jacobian terms is correct.

examples/hybrid/staggered_nonhydrostatic_model.jl
tapios: Let's only have one split divergence operator, and we have converged that the one we have is the correct one.

szy21 (approved):

src/Operators/spectralelement.jl
szy21: Note: I think this change makes FT in the following equation not working:
`@. Yₜ.c.ρ -= split_divₕ(ᶜρ * ᶜuvw, FT(1))`
error here: https://buildkite.com/clima/climaatmos-ci/builds/27460#019b6c61-87f2-467a-8148-6fdb3bd71289
Removing FT works.

PR #2413, opened 2025-12-16
Group/unrolling

nefrathenrici:
This PR refactors recursive functions to use UnrolledUtilites instead. From the hackathon on 12.12.25

### DIFF
--- a/src/DataLayouts/has_uniform_datalayouts.jl
+++ b/src/DataLayouts/has_uniform_datalayouts.jl
@@ -1,5 +1,5 @@
 @inline function first_datalayout_in_bc(args::Tuple, rargs...)
-    idx = unrolled_findfirst(x -> x isa AbstractData, args)
+    idx = unrolled_findfirst(Base.Fix2(isa, AbstractData), args)
     return isnothing(idx) ? nothing : args[idx]
 end
 
--- a/src/MatrixFields/field_name_dict.jl
+++ b/src/MatrixFields/field_name_dict.jl
@@ -51,8 +51,8 @@ struct FieldNameDict{
     end
 end
 function FieldNameDict{T}(key_entry_pairs::Pair{<:T}...) where {T}
-    keys = unrolled_map(pair -> pair[1], key_entry_pairs)
-    entries = unrolled_map(pair -> pair[2], key_entry_pairs)
+    keys = unrolled_map(first, key_entry_pairs)
+    entries = unrolled_map(last, key_entry_pairs)
     return FieldNameDict(FieldNameSet{T}(keys), entries)
 end
 
@@ -466,7 +466,7 @@ parent of the first entry in `dict` that is a `Fields.Field`. If no such entry
 is found, `target_type` defaults to `Number`.
 """
 function get_scalar_keys(dict::FieldMatrix)
-    first_field_idx = unrolled_findfirst(x -> x isa Fields.Field, dict.entries)
+    first_field_idx = unrolled_findfirst(Base.Fix2(isa, Fields.Field), dict.entries)
     target_type = Val(
         isnothing(first_field_idx) ? Number :
         eltype(parent(dict.entries[first_field_idx])),
@@ -597,7 +597,7 @@ function check_diagonal_matrix(matrix, error_message_start = "The matrix")
         !is_diagonal_matrix_entry(pair[2])
     end
     non_diagonal_entry_keys =
-        FieldMatrixKeys(unrolled_map(pair -> pair[1], non_diagonal_entry_pairs))
+        FieldMatrixKeys(unrolled_map(first, non_diagonal_entry_pairs))
     isempty(non_diagonal_entry_keys) || error(
         "$error_message_start has non-diagonal entries at the following keys: \
          $(set_string(non_diagonal_entry_keys))",
@@ -611,7 +611,7 @@ Checks whether the `FieldNameDict` `dict` contains any un-materialized
 `AbstractBroadcasted` entries.
 """
 is_lazy(dict) =
-    unrolled_any(entry -> entry isa Base.AbstractBroadcasted, values(dict))
+    unrolled_any(Base.Fix2(isa, Base.AbstractBroadcasted), values(dict))
 
 """
     lazy_main_diagonal(matrix)
--- a/src/MatrixFields/operator_matrices.jl
+++ b/src/MatrixFields/operator_matrices.jl
@@ -73,7 +73,7 @@ has_affine_bc(op) = any(
     op.bcs,
 )
 
-uses_extrapolate(op) = unrolled_any(bc -> bc isa Operators.Extrapolate, op.bcs)
+uses_extrapolate(op) = unrolled_any(Base.Fix2(isa, Operators.Extrapolate), op.bcs)
 
 ################################################################################
 


src/DataLayouts/fused_copyto.jl
```diff
@@ -18,11 +18,10 @@ Base.@propagate_inbounds function rcopyto_at_linear!(
     return nothing
 end
 Base.@propagate_inbounds function rcopyto_at_linear!(pairs::Tuple, I)
-    rcopyto_at_linear!(first(pairs), I)
-    rcopyto_at_linear!(Base.tail(pairs), I)
+    unrolled_foreach(pairs) do pair
+        rcopyto_at_linear!(pair, I)
+    end
```
dennisYatunin:
```suggestion
    unrolled_foreach(Base.Fix2(rcopyto_at_linear!, I), pairs)
```
I've seen several posts on Github and the Julia forums (like [this one](https://discourse.julialang.org/t/state-of-closures-fix1-fix2/118997/3?u=dyatunin)) which claim that `Base.Fix1` and `Base.Fix2` lead to lower latency than anonymous function closures. Extensive benchmarking in UnrolledUtilities has confirmed this, and I have a vague recollection that they also improved type stability in a few cases, so probably best to use them instead of closures throughout this PR.

src/DataLayouts/has_uniform_datalayouts.jl
```diff
@@ -1,32 +1,15 @@
...
-@inline _has_uniform_datalayouts_args(
-    truesofar,
-    start,
-    args::Tuple{Any},
-    rargs...,
-) = truesofar && _has_uniform_datalayouts(truesofar, start, args[1], rargs...)
-@inline _has_uniform_datalayouts_args(truesofar, _, args::Tuple{}, rargs...) =
-    truesofar
-
+    truesofar && unrolled_all(args) do arg
```
dennisYatunin:
```suggestion
    unrolled_all(args) do arg
```
Not sure whether I'm reading this correctly, but it looks like `truesofar` is always initialized to `true` and becomes `false` whenever `unrolled_all` returns false. If that's the case, you should be able to remove it from all these functions without changing their behavior.

src/DataLayouts/fused_copyto.jl
```diff
@@ -18,11 +18,10 @@ Base.@propagate_inbounds function rcopyto_at_linear!(
...
 end
-Base.@propagate_inbounds rcopyto_at_linear!(pairs::Tuple{<:Any}, I) =
-    rcopyto_at_linear!(first(pairs), I)
 @inline rcopyto_at_linear!(pairs::Tuple{}, I) = nothing
```
dennisYatunin:
```suggestion
```
This method is unnecessary, since `unrolled_foreach` already does nothing for empty iterators.

src/DataLayouts/non_extruded_broadcasted.jl
```diff
@@ -54,10 +55,11 @@ end
...
-    to_non_extruded_broadcasted_args(Base.tail(args))...,
-)
+@inline function to_non_extruded_broadcasted_args(args::Tuple)
+    UnrolledUtilities.unrolled_map(args) do arg
+        to_non_extruded_broadcasted(arg)
+    end
+end
 @inline to_non_extruded_broadcasted_args(args::Tuple{Any}) =
     (to_non_extruded_broadcasted(args[1]),)
 @inline to_non_extruded_broadcasted_args(args::Tuple{}) = ()
```
dennisYatunin:
```suggestion
```
These methods are unnecessary, since `unrolled_map` already has this behavior.

src/DataLayouts/non_extruded_broadcasted.jl
```diff
@@ -140,8 +142,11 @@ Base.@propagate_inbounds function _broadcast_getindex(
...
-Base.@propagate_inbounds _getindex(args::Tuple, I) =
-    (_broadcast_getindex(args[1], I), _getindex(Base.tail(args), I)...)
+Base.@propagate_inbounds function _getindex(args::Tuple, I)
+    UnrolledUtilities.unrolled_map(args) do arg
+        _broadcast_getindex(arg, I)
+    end
+end
 Base.@propagate_inbounds _getindex(args::Tuple{Any}, I) =
     (_broadcast_getindex(args[1], I),)
 Base.@propagate_inbounds _getindex(args::Tuple{}, I) = ()
```
dennisYatunin:
```suggestion
```
These methods are also unnecessary.

src/DataLayouts/non_extruded_broadcasted.jl
```diff
@@ -1,6 +1,7 @@
 #! format: off
 # ============================================================ Adapted from Base.Broadcast (julia version 1.10.4)
 import Base.Broadcast: BroadcastStyle
+import UnrolledUtilities
```
dennisYatunin:
```suggestion
import UnrolledUtilities: unrolled_map
```
The rest of this PR uses unqualified function names, so probably best to be consistent here. Switching to only using qualified names would be fine too.

src/DataLayouts/struct.jl
```diff
@@ -64,10 +64,10 @@ replace_basetype(::Type{T}, ::Type{T′}, ::Type{S}) where {T, T′, S} =
     S.name.wrapper{replace_basetypes(T, T′, S.parameters...)...}
 replace_basetype(::Type{T}, ::Type{T′}, ::Type{<:T}) where {T, T′} = T′
 replace_basetype(::Type{T}, ::Type{T′}, value) where {T, T′} = value
-replace_basetypes(::Type{T}, ::Type{T′}, value) where {T, T′} =
-    (replace_basetype(T, T′, value),)
-replace_basetypes(::Type{T}, ::Type{T′}, value, values...) where {T, T′} =
-    (replace_basetype(T, T′, value), replace_basetypes(T, T′, values...)...)
+replace_basetypes(::Type{T}, ::Type{T′}, values...) where {T, T′} =
```
dennisYatunin:
```suggestion
replace_basetypes(::Type{T}, ::Type{T′}, values) where {T, T′} =
```
Splatting iterator values like this can lead to type instabilities, since the compiler occasionally gives up for functions with more than 32 inputs. It would be better to directly pass the iterator to this function without splatting it.

src/Fields/broadcast.jl
```diff
@@ -180,7 +181,11 @@ end
...
 # Return underlying DataLayout object, DataStyle of broadcasted
 # for `Base.similar` of a Field
-_todata_args(args::Tuple) = (todata(args[1]), _todata_args(Base.tail(args))...)
+# _todata_args(args::Tuple) = (todata(args[1]), _todata_args(Base.tail(args))...)
+_todata_args(args::Tuple) =
+    unrolled_map(args) do arg
+        todata(arg)
+    end
 _todata_args(args::Tuple{Any}) = (todata(args[1]),)
 _todata_args(::Tuple{}) = ()
```
dennisYatunin:
```suggestion
```
These methods are also unnecessary.

src/Fields/fieldvector.jl
```diff
@@ -231,44 +231,26 @@ function Spaces.weighted_dss!(
...
-    rargs...,
-) where {TStart} =
-    truesofar && _is_diagonal_bc(truesofar, TStart, args[1], rargs...)
-@inline _is_diagonal_bc_args(
-    truesofar,
-    ::Type{TStart},
-    args::Tuple{},
-    rargs...,
-) where {TStart} = truesofar
+    truesofar && unrolled_all(args) do arg
```
dennisYatunin:
```suggestion
    unrolled_all(args) do arg
```
Looks like `truesofar` could also be removed from these functions.

src/Operators/finitedifference.jl
```diff
@@ -3674,10 +3674,10 @@ assymetric).
...
-    left_interior_window_idx(args[1], space, loc),
-    _left_interior_window_idx_args(Base.tail(args), space, loc)...,
-)
+@inline _left_interior_window_idx_args(args::Tuple, space, loc) =
+    unrolled_map(args) do arg
+        left_interior_window_idx(arg, space, loc)
+    end
 @inline _left_interior_window_idx_args(args::Tuple{Any}, space, loc) =
     (left_interior_window_idx(args[1], space, loc),)
 @inline _left_interior_window_idx_args(args::Tuple{}, space, loc) = ()
```
dennisYatunin:
```suggestion
```

src/Operators/finitedifference.jl
```diff
@@ -3725,10 +3725,10 @@ end
...
-    right_interior_window_idx(args[1], space, loc),
-    _right_interior_window_idx_args(Base.tail(args), space, loc)...,
-)
+@inline _right_interior_window_idx_args(args::Tuple, space, loc) =
+    unrolled_map(args) do arg
+        right_interior_window_idx(arg, space, loc)
+    end
 @inline _right_interior_window_idx_args(args::Tuple{Any}, space, loc) =
     (right_interior_window_idx(args[1], space, loc),)
 @inline _right_interior_window_idx_args(args::Tuple{}, space, loc) = ()
```
dennisYatunin:
```suggestion
```

src/Operators/spectralelement.jl
```diff
@@ -231,28 +233,25 @@ Base.@propagate_inbounds function resolve_operator(
...
+    _resolve_operator_args(slabidx, args)
 
 Calls `resolve_operator(arg, slabidx)` for each `arg` in `args`
 """
-@inline _resolve_operator_args(slabidx) = ()
-Base.@propagate_inbounds _resolve_operator_args(slabidx, arg, xargs...) = (
-    resolve_operator(arg, slabidx),
-    _resolve_operator_args(slabidx, xargs...)...,
-)
+Base.@propagate_inbounds _resolve_operator_args(slabidx, args::Tuple) =
```
```suggestion
Base.@propagate_inbounds _resolve_operator_args(slabidx, args) =
```
It doesn't look like the `Tuple` type constraint is doing anything here, so it can probably be removed. Same goes for most of the other functions with `Tuple` constraints in this PR.

src/Topologies/dss_transform.jl
```diff
@@ -35,14 +36,11 @@ Base.@propagate_inbounds dss_transform(
...
     weight,
 )
-    (
-        dss_transform(first(arg), local_geometry, weight),
-        dss_transform(Base.tail(arg), local_geometry, weight)...,
-    )
+    unrolled_map(arg -> dss_transform(arg, local_geometry, weight), args)
 end
 @inline dss_transform(
     arg::Tuple{Any},
```
dennisYatunin:
```suggestion
```
This method is also unnecessary.

src/interface.jl
```diff
@@ -14,10 +15,11 @@ Base.@propagate_inbounds slab(tup::Tuple, inds...) = slab_args(tup, inds...)
...
-Base.@propagate_inbounds slab_args(args::Tuple, inds...) =
-    (slab(args[1], inds...), slab_args(Base.tail(args), inds...)...)
-Base.@propagate_inbounds slab_args(args::Tuple{Any}, inds...) =
-    (slab(args[1], inds...),)
+Base.@propagate_inbounds function slab_args(args::Tuple, inds...)
+    unrolled_map(args) do arg
+        slab(arg, inds...)
+    end
+end
 Base.@propagate_inbounds slab_args(args::Tuple{}, inds...) = ()
```
dennisYatunin:
```suggestion
```

src/interface.jl
```diff
@@ -34,19 +36,21 @@ Base.@propagate_inbounds column(tup::Tuple, inds...) = column_args(tup, inds...)
...
-Base.@propagate_inbounds column_args(args::Tuple{Any}, inds...) =
-    (column(args[1], inds...),)
+Base.@propagate_inbounds function column_args(args::Tuple, inds...)
+    unrolled_map(args) do arg
+        column(arg, inds...)
+    end
+end
 Base.@propagate_inbounds column_args(args::Tuple{}, inds...) = ()
 Base.@propagate_inbounds column_args(args::NamedTuple, inds...) =
     NamedTuple{keys(args)}(column_args(values(args), inds...))
```
dennisYatunin:
```suggestion
```
`unrolled_map` already has this behavior for `NamedTuple`s and empty `Tuple`s.

src/interface.jl
```diff
@@ -34,19 +36,21 @@ Base.@propagate_inbounds column(tup::Tuple, inds...) = column_args(tup, inds...)
...
-Base.@propagate_inbounds level_args(args::Tuple, inds...) =
-    (level(args[1], inds...), level_args(Base.tail(args), inds...)...)
-Base.@propagate_inbounds level_args(args::Tuple{Any}, inds...) =
-    (level(args[1], inds...),)
+Base.@propagate_inbounds function level_args(args::Tuple, inds...)
+    unrolled_map(args) do arg
+        level(arg, inds...)
+    end
+end
 Base.@propagate_inbounds level_args(args::Tuple{}, inds...) = ()
```
dennisYatunin:
```suggestion
```

dennisYatunin (approved):
Thanks for adding this! There's definitely a lot of refactoring left to do in these modules, but this is a good start.

You'll probably want to test this with ClimaAtmos before merging it in, just to make sure everything is still type-stable.

PR #2417, opened 2026-01-10
Replace RecursiveApply functions with AutoBroadcaster wrappers

dennisYatunin:
### Purpose

This replaces the `RecursiveApply` module with a new interface that follows the opposite design pattern: instead of defining functions for `Tuple`s and `NamedTuple`s that need to be called in `Fields`, `Operators`, `MatrixFields`, and external libraries like `ClimaAtmos`, we can now use a wrapper for `Tuple`s and `NamedTuple`s that behaves the same way when passed to standard math functions. The number of places where this interface needs to be explicitly used is quite small, making it possible to decouple iteration over field variables from most of ClimaCore's internals.

### Content

- Introduce the `AutoBroadcaster` wrapper, which triggers `nested_broadcast` (an analogue of the old `RecursiveApply.rmap`) when passed to any common math function or constructor
  - Replace all `RecursiveApply` functions in ClimaCore's source code with standard math operations
  - Make it possible to add support for any iterator type `T` by defining the method `is_auto_broadcastable(::T) = true`, as long as `T` is compatible with UnrolledUtilities
- Preserve backward-compatibility with original `RecursiveApply` implementation
  - Add a `RecursiveApply` module that defines the same functions as the old version, but as simple wrappers around `add_auto_broadcasters` and `drop_auto_broadcasters`
  - Define an `auto_broadcasted` function that automatically introduces calls to either `add_auto_broadcasters` or `drop_auto_broadcasters` in `Broadcasted` expressions, allowing ClimaCore operators to be used in the same expressions as user-defined functions that are incompatible with `AutoBroadcaster`s (e.g., functions that dispatch on non-`AutoBroadcaster` types)
- Unify broadcasting across DataLayouts, Fields, and Operators
  - The `broadcastable` form of any `AbstractData` or `Field` has `add_auto_broadcasters` applied to its values
  - Calling `Base.broadcasted` triggers`auto_broadcasted` instead of the standard `Broadcasted` constructor
  - The result of `eltype` is the output of `unsafe_eltype`, which acts as an unrolled version of `Base.Broadcast.combine_eltypes`
  - The result of `similar` uses the output of `safe_eltype`, which can identify the first point during inference where a non-concrete type appears in a broadcast expression (essentially a simplified and corrected version of the old `call_with_first`)
- Update the implementation of `new` to avoid segfaults and other errors when dealing with `DataType` fields
- Refactor the `deformation_flow.jl` example by removing all loops over field variables, adding several missing cases, and clarifying precisely what needs to be tested

---


ext/cuda/operators_spectral_element.jl
```diff
@@ -198,11 +198,11 @@ Base.@propagate_inbounds function operator_evaluate(
...
     local_geometry = get_local_geometry(space, ij, slabidx)
 
-    DJv = D[i, 1] ⊠ Jv¹[1, vt]
+    DJv = D[i, 1] * Jv¹[1, vt]
     for k in 2:Nq
-        DJv = DJv ⊞ D[i, k] ⊠ Jv¹[k, vt]
+        DJv += D[i, k] * Jv¹[k, vt]
     end
-    return RecursiveApply.rmul(DJv, local_geometry.invJ)
+    return DJv / local_geometry.invJ
```
nefrathenrici:
```suggestion
    return DJv * local_geometry.invJ
```

nefrathenrici: I see this several instances of this change from rmul to division. I'm not familiar with this code but it looks incorrect. If this is an issue, it also appears on lines 231, 482, 489, 499–500, 529, 540, 556

src/MatrixFields/MatrixFields.jl
```diff
@@ -111,28 +101,30 @@ const FieldOrStencilStyleType = Union{
...
-    ::typeof(*),
-    single_value_or_broadcasted::SingleValueStyleType,
-    field_or_broadcasted::FieldOrStencilStyleType,
-    args...,
-) = Base.Broadcast.broadcasted(
-    ⊠,
-    single_value_or_broadcasted,
-    Base.Broadcast.broadcasted(*, field_or_broadcasted, args...),
-)
+# TODO: This is broken now!
```
nefrathenrici: Is this meant to be fixed before merging? Do we need this functionality?

nefrathenrici (commented):
This looks good to me. It's nice to get rid of RecursiveApply. I didn't have any significant comments.

ext/cuda/operators_spectral_element.jl
dennisYatunin: Excellent catch! Thank you

src/MatrixFields/MatrixFields.jl
dennisYatunin: Yeah this is needed to support matrix-matrix and matrix-vector multiplication. ~Still figuring out how to do it properly.~ Done!

src/Utilities/auto_broadcaster.jl
```diff
@@ -0,0 +1,355 @@
...
+"""
+    add_auto_broadcasters(itr)
+    add_auto_broadcasters(::Type)
+
+Recursively applies the [`AutoBroadcaster`](@ref) constructor to iterators for
+which [`is_auto_broadcastable`](@ref) is true, as well as their elements for
+which it is true, while leaving values for which it is false unmodified. Can
+also be passed an iterator's type to infer the result type for such an iterator.
+"""
+add_auto_broadcasters(itr) =
```
imreddyTeja, line 91: I don't think we'd ever see structs nested that much, but should this have the recursion limit removed?

src/Utilities/auto_broadcaster.jl
```diff
@@ -0,0 +1,355 @@
...
+arguments if doing so may prevent an error from being thrown.
+
+When the [`unsafe_eltype`](@ref) of `bc` indicates that an error will be thrown,
+an alternative `Broadcasted` wrapper is constructed with `add_auto_broadcasters`
+applied to every argument, and then another `Broadcasted` wrapper is constructed
+with `drop_auto_broadcasters` applied to every argument. If one of the new
+wrappers no longer corresponds to a guaranteed error, it is returned instead of
+the original `bc`. Otherwise, `bc` is returned without any modifications.
+"""
+function allow_auto_broadcasting(bc::Base.Broadcast.Broadcasted)
```
imreddyTeja: has this function been checked for allocations or dynamic function invokations?

src/Utilities/auto_broadcaster.jl
```diff
@@ -0,0 +1,355 @@
...
+    NamedTuple{names}(unwrap(x))
+Base.show(io::IO, x::AutoBroadcaster) = show(io, unwrap(x))
+Base.iterate(x::AutoBroadcaster, state...) = iterate(unwrap(x), state...)
+Base.getproperty(x::AutoBroadcaster, name::Symbol) =
+    getproperty(unwrap(x), name)
+Base.merge(args::AutoBroadcaster...) =
+    AutoBroadcaster(merge(unrolled_map(unwrap, args)...))
+
+for f in (:axes, :size, :firstindex, :lastindex)
+    @eval Base.$f(x::AutoBroadcaster) = $f(unwrap(x))
```
imreddyTeja: Could this be combined with the loop on line 145?

src/Utilities/auto_broadcaster.jl
```diff
@@ -0,0 +1,355 @@
...
+Base.mapreduce(
+    f::F,
+    op::O,
+    arg::AutoBroadcaster,
+    args::AutoBroadcaster...;
+    init...,
+) where {F, O} =
+    unrolled_mapreduce(f, op, unwrap(arg), unrolled_map(unwrap, args)...; init...)
+Base.map(f::F, arg::AutoBroadcaster, args::AutoBroadcaster...) where {F} =
+    AutoBroadcaster(unrolled_map(f, unwrap(arg), unrolled_map(unwrap, args)...))
```
imreddyTeja: How does this work if the arg is a AutoBroadcaster wrapping a tuple of AutoBroadcasters? Wouldn't it need to recurse with Base.map instead?

src/Utilities/Utilities.jl
```diff
@@ -118,18 +121,73 @@ to ensure that recursive functions over nested types have inferrable outputs.
...
+new(::Type{Type{T}}, ::Tuple{} = ()) where {T} = T
+new(::Type{T}, ::Tuple{} = ()) where {T <: Type} =
+    throw(ArgumentError("Cannot allocate ambiguous $T"))
+
+no_immutable_types(::Type{<:Type}) = false
+no_immutable_types(::Type{T}) where {T} =
+    ismutabletype(T) || unrolled_all(no_immutable_types, fieldtypes(T))
+
+drop_immutable_types(not_a_type) = not_a_type
+drop_immutable_types(::Type{<:Type}) = DataType
```
imreddyTeja: What is this doing?

src/Utilities/Utilities.jl
```diff
@@ -118,18 +121,73 @@ to ensure that recursive functions over nested types have inferrable outputs.
...
     ntuple(Val ∘ Base.Fix1(fieldtype, T), Val(fieldcount(T)))
 
+# Special handling of Type values, which can cause new and splatnew to segfault
+new(::Type{Type{T}}, ::Tuple{} = ()) where {T} = T
+new(::Type{T}, ::Tuple{} = ()) where {T <: Type} =
+    throw(ArgumentError("Cannot allocate ambiguous $T"))
+
+no_immutable_types(::Type{<:Type}) = false
+no_immutable_types(::Type{T}) where {T} =
+    ismutabletype(T) || unrolled_all(no_immutable_types, fieldtypes(T))
```
imreddyTeja: This function could use a comment or docstring. How does `unrolled_all(no_immutable_types, fieldtypes(T))` imply the type is immutible

imreddyTeja (commented):
I left some comments and questions, but this looks mostly good to me

imreddyTeja (approved):

src/Utilities/auto_broadcaster.jl
dennisYatunin: I've tested this with arbitrary levels of recursion and cannot find any case where this hits a recursion limit on Julia 1.10.8. For example, the following inference test passes:

```julia
julia> using ClimaCore.Utilities: add_auto_broadcasters
julia> tup = (1, 2, 3); for i in 1:100 tup = (1, 2, 3, tup) end
julia> using Test; @inferred add_auto_broadcasters(tup);
```

src/Utilities/auto_broadcaster.jl
dennisYatunin: Yeah, this is called every time we construct a `Broadcasted` object that uses some `DataStyle` or `AbstractFieldStyle`, so it's already covered by a bunch of allocation and inferrence tests

src/Utilities/auto_broadcaster.jl
dennisYatunin: This is treating `map`/`mapreduce` non-recursively, in contrast to the math operators that are passed to `nested_auto_broadcast`. I've rearranged things and added some comments to clarify this.

src/Utilities/Utilities.jl
dennisYatunin: This was a hacky way of accounting for the fact that a `Tuple{..., Type{T}, ...}` turns into `Tuple{..., DataType, ...}` when it is allocated (and likewise for `NamedTuple`s). I've refactored this to avoid direct manipulation type parameters; the updated version just maps over the `fieldtypes` of a `Tuple` (or `NamedTuple`) when at least one of the fields is a `DataType`.

src/Utilities/Utilities.jl
dennisYatunin: Yeah, "immutable" was not the correct word to use here. The purpose of this function was to check whether a type `T` can be allocated with uninitialized fields through `Expr(:new, :T)`, which appears to only be possible for structs that do not contain constant `DataType` fields. Allocating a `DataType` field in this manner causes segfaults and other bizarre errors, presumably because the compiler is not designed to handle different pointers to the same `DataType`. I've changed the name of this function to `can_alloc_uninitialized`, and added some code comments to clarify what's going on.

Issue #2421, opened 2026-01-14
Support for space with pressure as the vertical


ph-kev:
related to https://github.com/CliMA/ClimaDiagnostics.jl/issues/136

To facilitate writing diagnostics in pressure coordinates, a new point type, space with pressure in the vertical direction, and interpolation for fields between a space with z in the vertical direction to a space with pressure in the vertical direction should be implemented.

The new point could be called `PfullPoint` for full pressure level point. For the new space, this could be built off of the `ExtrudedFiniteDifferenceSpace`, since that is the use case that we are most interested in. Functionality for single column could be supported later. Furthermore, fields defined on this new space need to support basic arithmetic operations and the `Remapper` object should work with it.

Finally, for vertical interpolation, this relies on an existing pressure field and the field of interest. The algorithm is

1. Get a permutation matrix that sort the columns of the pressure field.
2. Apply the permutation matrix to both fields
3. Linearly interpolate each column of field of interest to the space with pressure as the vertical

In ClimaDiagnostics, this operation will be computed every compute step, but it will be infrequently called (e.g. 4 hours).

PR #2422, opened 2026-01-14
Add pressure point and space with pressure as the vertical direction

ph-kev:
As a first step towards https://github.com/CliMA/ClimaCore.jl/issues/2421, this PR introduces `PfullPoint` and interpolation from spaces whose vertical component consists of `ZPoint`s to spaces whose vertical component consists of `PfullPoint`s.


src/Geometry/coordinates.jl
```diff
@@ -89,6 +89,7 @@ end
 @pointtype XPoint x
 @pointtype YPoint y
 @pointtype ZPoint z
+@pointtype PfullPoint pfull
```
ph-kev: I wasn't sure if `PfullPoint` is a good name for this. `Pfull` stands for pressure at model full levels.

dennisYatunin: Either `P` or `Pfull` is fine, whichever makes more sense.

ext/ClimaCoreClimaInterpolationsExt.jl
```diff
@@ -0,0 +1,64 @@
...
+vertical direction is pressure and return the interpolated field.
+
+This mutates `field`.
+"""
+function Remapping.vertical_interpolate!(
+    field::Fields.Field,
+    pfull_intp::PressureInterpolator,
+)
+    (; pfull_field) = pfull_intp
+    dest = fill(eltype(pfull_field)(1.0), pfull_intp.pfull_space)
```
dennisYatunin:
```suggestion
    dest = fill(one(eltype(pfull_field)), pfull_intp.pfull_space)
```

src/CommonGrids/CommonGrids.jl
```diff
@@ -706,4 +706,124 @@ function RectangleXYGrid(
     )
 end
 
+"""
+    _ExtrudedPfullCubedSphereGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
```
ph-kev: I am planning to remove the code from CommonGrids and CommonSpaces, since I don't see an use for it for this PR.

src/Geometry/coordinates.jl
```diff
@@ -89,6 +89,7 @@ end
...
```
imreddyTeja: What do you mean by pressure at model "full" levels?

src/Geometry/coordinates.jl
```diff
@@ -120,6 +121,7 @@ Cartesian123Point(pt::Cartesian13Point{FT}) where {FT} =
 @pointtype LongPoint long
 @pointtype LatLongPoint lat long
 @pointtype LatLongZPoint lat long z
+@pointtype LatLongPfullPoint lat long pfull
```
imreddyTeja: I'm not sure where it should go, but I think it would be good to document that functionality is limited when using Pfull

src/Geometry/localgeometry.jl
```diff
@@ -63,8 +63,22 @@ struct SurfaceGeometry{FT, N}
...
 
+"""
+    CoordinateOnly
+
+The necessary coordinates information defined at each node.
+
+This is currently used for constructing a space with `PfullPoint` as the
+vertical.
+"""
+struct CoordinateOnly{C <: AbstractPoint}
```
imreddyTeja: Would `CoordinateOnlyGeometry` be a better name?

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
...
+A interpolator for interpolating ClimaCore fields defined on a space whose
+vertical direction is z to a space whose vertical direction is pressure.
+
+Using this struct requires loading ClimaInterpolations.jl.
+
+The approach for computing the pressure field is by:
+1. Apply aggregate with max for columns of the pressure field to ensure that
+   pressures are always monotonic.
+2. Vertical interpolation to the specified pressure coordinates for the columns
+   of the ClimaCore field of interest using the pressure field.
```
imreddyTeja: This could be clearer

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
...
+
+!!! warning "No validation of pressure-height relationship"
+    The implementation assumes pressure decreases monotonically with height. If
```
imreddyTeja, line 16: If this assumption is made why does step 1 from above apply

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
+"""
+    PressureInterpolator
+
+A interpolator for interpolating ClimaCore fields defined on a space whose
+vertical direction is z to a space whose vertical direction is pressure.
```
imreddyTeja: Would "vertical coordinate" make more sense here

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
...
+struct PressureInterpolator{
+    FIELD <: Fields.Field,
+    CENTER <: Fields.Field,
+    FACE <: Fields.Field,
+    SPACE <: Spaces.AbstractSpace,
+    LEVELS,
+}
+    """A ClimaCore.Field representing pressure. This field is defined on a space
+    with height (z) as the vertical coordinate, not pressure."""
+    pfull_field::FIELD
```
imreddyTeja: I'm confused about if this is supposed to be on a center or face space

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
...
+end
+
+"""
+    PressureInterpolator(
+        pfull_field::Fields.Field,
+        pfull_space::Spaces.AbstractSpace,
+    )
+
+Construct a `PressureInterpolator` from `pfull_field`, a pressure field and
+`pfull_space`, a space with pressure as the vertical direction.
```
imreddyTeja: The commas make this a bit hard to read

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,267 @@
...
+"""
+function PressureInterpolator(
+    pfull_field::Fields.Field,
+    pfull_space::Spaces.AbstractSpace,
+)
+    axes(pfull_field).staggering isa Grids.CellCenter || error("The staggering of the
```
imreddyTeja, line 176: I think this assumes that the space is finite difference space

src/Remapping/interpolate_pressure.jl
ph-kev: I added more description for this.

src/Remapping/interpolate_pressure.jl
ph-kev: I added a type constraint for `pfull_space`.

docs/src/remapping.md
```diff
@@ -163,3 +163,87 @@ int12 = interpolate(remapper, [field1, field2])
...
+vertical coordinate is pressure rather than height.
+
+## How it works
+
+The `PressureInterpolator` performs the following steps:
+
+1. **Ensure monotonicity**: Applies a cumulative maximum along each column to
+   ensure pressure decreases monotonically with height.
+2. **Vertical interpolation**: Interpolates field values to the specified
+   pressure coordinates using linear interpolation with flat boundary
```
dennisYatunin:
```suggestion
   pressure coordinates using linear interpolation with constant boundary
```
A more standard term you could use throughout this PR is "constant boundary conditions", or you could also use "constant extrapolation at boundaries". The word "flat" isn't very clear in this context.

docs/src/APIs/remapping_api.md
```diff
@@ -7,4 +7,12 @@ CurrentModule = ClimaCore
 ```@docs
 Remapping.interpolate_array
 Remapping.interpolate
+Remapping.PressureInterpolator
+Remapping.PressureInterpolator(pfull_field::Fields.Field, pfull_levels)
+Remapping.PressureInterpolator(pfull_field::Fields.Field, pfull_space::Union{Spaces.AbstractFiniteDifferenceSpace, Spaces.ExtrudedFiniteDifferenceSpace})
```
dennisYatunin: It would be good to combine these into a single docstring for the `PressureInterpolator` constructor, or to only have a docstring for the constructor users will typically call.

ext/ClimaCoreClimaInterpolationsExt.jl
```diff
@@ -0,0 +1,67 @@
+module ClimaCoreClimaInterpolationsExt
```
dennisYatunin: The code here does not look like it belongs in a module extension. An extension for package `X` is meant to be used in the following scenarios:
- Adding methods to functions defined in `X` that dispatch on types defined in ClimaCore
- Adding methods to functions defined in ClimaCore that dispatch on types defined in `X`

In this case, you're calling a function from ClimaInterpolations, but you're not adding methods to it or using it for dispatch. If this function is required for ClimaCore's `PressureInterpolator` to work with any ClimaCore `Field`, ClimaInterpolations should just be a dependency of ClimaCore.

ext/ClimaCoreClimaInterpolationsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+    (; scratch_center_pressure_field, scratch_face_pressure_field, pressure_levels) =
+        pfull_intp
+    scratch_pfull_array = if axes(field).staggering isa Grids.CellCenter
+        field2array(scratch_center_pressure_field)
+    else
+        field2array(scratch_face_pressure_field)
+    end
+    field_array = field2array(field)
+    dest = field2array(dest)
+    reverse!(field_array, dims = 1)
```
dennisYatunin: I'm guessing `reverse!` and `interpolate1d!` take comparable amounts of time, so calling `reverse!` here roughly doubles the time required per interpolation. Might not matter too much, but it would definitely be more efficient to keep `field_array` unchanged and instead compute `interpolate1d!` in reverse order.

More importantly, getting rid of `reverse!` would let you rename `interpolate_pressure!!` to `interpolate_pressure!` (and `interpolate_pressure!` to `interpolate_pressure`), which would be good because `!!` is not standard notation in Julia.

src/Geometry/localgeometry.jl
```diff
@@ -63,8 +63,22 @@ struct SurfaceGeometry{FT, N}
     normal::N
 end
 
+"""
+    CoordinateOnlyGeometry
+
+The necessary coordinates information defined at each node.
+
+This is currently used for constructing a space with `PPoint` as the
+vertical.
```
dennisYatunin:
```suggestion
This is currently used for constructing spaces with pressure as the vertical coordinate.
```

src/Remapping/interpolate_pressure.jl
dennisYatunin: "Coordinate" is definitely more correct than "direction" in this context. There is only one vertical direction (the one you face when you look up from the ground to space), but there are many possible coordinate systems you can use to describe movement along that direction. In this case, we are using the spatial coordinate `z [m]` and the pressure coordinate `p [Pa]`.

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,288 @@
...
+
+!!! warning "No validation of pressure-height relationship"
+    The implementation assumes pressure decreases monotonically with height. If
+    the interpolated field appears unrealistic, check for instabilities or
+    inversions in the pressure field.
+
+!!! note "Boundary conditions for vertical interpolation"
+    Vertical interpolation uses flat boundary conditions at the top and bottom
+    of the atmosphere. Interpolated values at pressure levels outside the
+    model's vertical range may be inaccurate.
```
dennisYatunin: If you're using constant (`Flat`) extrapolation, wouldn't that mean you never get values outside the model's range? The unit tests seem to confirm that this is the case. I think it would be better to use linear extrapolation, so that you could at least see when a variable has gone out of range. Though I suppose constant extrapolation is much simpler to debug, so another option is to keep it and print a warning when something is out of range.

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,288 @@
...
+"""
+    update!(pfull_intp::PressureInterpolator)
+
+Update `pfull_intp` after the pressure field has been modified.
+
+This should only be called once whenever the pressure field changes before
+performing new interpolations.
+"""
+function update!(pfull_intp::PressureInterpolator)
+    (; pfull_field, scratch_center_pressure_field, scratch_face_pressure_field) = pfull_intp
```
dennisYatunin, line 226:
```suggestion
function update!(pfull_intp::PressureInterpolator, pfull_field)
    (; scratch_center_pressure_field, scratch_face_pressure_field) = pfull_intp
```
Instead of storing a copy of the pressure field in `PressureInterpolator`, you could pass the field directly from ClimaAtmos. That way, ClimaAtmos will not need to copy pressure values into the interpolator, saving some global reads/writes. But if there's a reason it's simpler to have a local copy, you can keep it as is.

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,288 @@
...
+function _update!(
+    pfull_field::Fields.Field,
+    scratch_center_pressure_field,
+    scratch_phalf_field,
+)
+    pfull_array = Fields.field2array(pfull_field)
+    scratch_pfull_array = Fields.field2array(scratch_center_pressure_field)
+    # Pressure is decreasing for increasing z
+    reverse!(pfull_array, dims = 1)
+    accumulate!(max, scratch_pfull_array, pfull_array, dims = 1)
```
dennisYatunin: If possible, it would be better to accumulate over the original pressure values without reversing them, using `min` instead of `max` to enforce monotonicity. Running `interpolate_pressure!` with `reverse = true` will then give you roughly the same result as what you have now, though there will be some minor differences in areas of non-monotonicity.

src/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,288 @@
...
+                if Symbol(exc.f) in fns && !is_pkg_loaded(pkg)
+                    print(
+                        io,
+                        "\nImport ClimaInterpolations to enable `$(exc.f)`.";
+                    )
+                end
+            end
+        end
+    end
+end
```
dennisYatunin: Throwing a `MethodError` for a public API function if another package isn't manually loaded is very unorthodox. It would be better to automatically load the other package as a dependency in this situation.

test/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,232 @@
...
+        for space in (extruded_space, col_space)
+            space = ExtrudedCubedSphereSpace(FT;
+                z_elem = 10,
+                z_min = 0,
+                z_max = 1,
+                radius = 10,
+                h_elem = 10,
+                n_quad_points = 4,
+                staggering = Grids.CellCenter(),
+            )
```
dennisYatunin:
```suggestion
```
This looks like it was copy-pasted by accident? The same value is defined a few lines above.

test/Remapping/interpolate_pressure.jl
```diff
@@ -0,0 +1,232 @@
...
+            h_elem = 10,
+            n_quad_points = 4,
+            staggering = Grids.CellCenter(),
+        )
+        col_space = ColumnSpace(FT;
+            z_elem = 10,
+            z_min = 0,
+            z_max = 1,
+            staggering = Grids.CellCenter(),
+        )
```
dennisYatunin: Instead of redefining and looping over these spaces in each `testset`, you could also have a single outer loop that runs all the `testset`s for one space at a time.

src/Remapping/interpolate_pressure.jl
ph-kev: I think it is easier to debug when using a flat extrapolation, because if you write pressure in pressure coordinates, then you can tell when the pressure doesn't make sense at the boundaries. I can just make this into a keyword argument for the user though.

src/Remapping/interpolate_pressure.jl
ph-kev: I think it is easier for the interpolator to share a reference to the pressure field in that case.

docs/src/APIs/remapping_api.md
ph-kev: I removed the docstring for the one that takes in `pfull_field` and `pfull_space`.

src/Remapping/interpolate_pressure.jl
dennisYatunin: Ah right, looking at the pressure itself will indicate whether any clipping happened. I think it's fine to keep the current version as the default then, though a kwarg that switches to linear or an automatic warning might still be helpful to add.

PR #2431, opened 2026-01-20
Make ClimaCoreMakie release with bumped Makie compat

imreddyTeja:
Update Makie compat to 0.24. Drop support for older versions due to breaking changes.

With some extra effort, we could probably add backwards support
Update to use Makie ComputePipelines

### DIFF
--- a/lib/ClimaCoreMakie/Project.toml
+++ b/lib/ClimaCoreMakie/Project.toml
@@ -1,11 +1,11 @@
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
 name = "ClimaCoreMakie"
 uuid = "908f55d8-4145-4867-9c14-5dad1a479e4d"
-version = "0.4.6"
+version = "0.4.7"
 
 [compat]
 ClimaCore = "0.11, 0.12, 0.13, 0.14"
-Makie = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21"
+Makie = "0.24"
 julia = "1.7"
 
 [deps]
--- a/lib/ClimaCoreMakie/src/fieldcontourf.jl
+++ b/lib/ClimaCoreMakie/src/fieldcontourf.jl
@@ -16,33 +16,35 @@ These are inherited from [`Makie.tricontourf`](https://docs.makie.org/stable/exa
 - `colormap::Union{Symbol, Vector{<:Colorant}} = :viridis` sets the colormap from which the band colors are sampled.
 
 """
-@recipe(FieldContourf, field) do scene
-    attrs = Makie.Attributes(; coords = nothing)
-    return merge(attrs, Makie.default_theme(scene, Makie.Tricontourf))
+@recipe FieldContourf (field,) begin
+    coords = nothing
+    Makie.filter_attributes(
+        Makie.documented_attributes(Makie.Tricontourf);
+        exclude = (:coords,),
+    )...
 end
 
 function Makie.plot!(plot::FieldContourf)
 
-    field = plot.field
-    # Only update vertices if axes updates
-    space = lift(axes, field; ignore_equal_values = true)
-    vertices = lift(plot_vertices, space, plot.coords)
-    plot.triangulation = lift(plot_triangles_matrix, space)
-
-    xs = lift(vertices) do vertices
-        map(x -> x[1], vertices)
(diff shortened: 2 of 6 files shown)


lib/ClimaCoreMakie/Project.toml
```diff
@@ -1,11 +1,11 @@
...
 name = "ClimaCoreMakie"
 uuid = "908f55d8-4145-4867-9c14-5dad1a479e4d"
-version = "0.4.6"
+version = "0.4.7"
 
 [compat]
 ClimaCore = "0.11, 0.12, 0.13, 0.14"
-Makie = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21"
+Makie = "0.24"
 julia = "1.7"
```
ph-kev, line 9: I remembered trying to use Makie 0.24 in the past with older versions of Julia (<= 1.11.x) and it not working. Is this still the case or not?

szy21 (approved):
Looks good to me. I'm not sure if I'm the best person to review this though.

lib/ClimaCoreMakie/src/fieldheatmap.jl
```diff
@@ -30,21 +34,23 @@ Makie.plottype(::ClimaCore.Fields.SpectralElementField2D) =
...
-    triangles = lift(plot_triangles, space)
-
-    plot.color = map(vec ∘ parent, field)
-
+    input_nodes = [:field, :coords]
+    output_nodes = [:vertices, :triangles, :new_color]
+    map!(plot.attributes, input_nodes, output_nodes) do f, coords
+        space = axes(f)
+        vertices = plot_vertices(space, coords)
+        triangles = plot_triangles(space)
```
ph-kev, line 44: What is that reason for not using `lift` here?

lib/ClimaCoreMakie/src/fieldcontourf.jl
```diff
@@ -16,33 +16,35 @@ These are inherited from [`Makie.tricontourf`](https://docs.makie.org/stable/exa
 - `colormap::Union{Symbol, Vector{<:Colorant}} = :viridis` sets the colormap from which the band colors are sampled.
 
 """
-@recipe(FieldContourf, field) do scene
-    attrs = Makie.Attributes(; coords = nothing)
-    return merge(attrs, Makie.default_theme(scene, Makie.Tricontourf))
+@recipe FieldContourf (field,) begin
+    coords = nothing
```
juliasloan25, line 20: do we need to define`coords`?

lib/ClimaCoreMakie/src/fieldcontourf.jl
```diff
@@ -16,33 +16,35 @@ These are inherited from [`Makie.tricontourf`](https://docs.makie.org/stable/exa
...
+    Makie.filter_attributes(
+        Makie.documented_attributes(Makie.Tricontourf);
+        exclude = (:coords,),
```
juliasloan25, line 23: From the release notes I thought that unused attributes are filtered out automatically now (the first "Semi-breaking" bullet point [here](https://github.com/MakieOrg/Makie.jl/releases/tag/v0.24.0))

juliasloan25 (approved):
Thanks Teja! Just a couple comments

lib/ClimaCoreMakie/Project.toml
juliasloan25: In the Project in Makie v0.24, the julia compat is 1.10 so it should be okay, but I haven't tried it

lib/ClimaCoreMakie/src/fieldcontourf.jl
imreddyTeja: Yeah, I think so. `plot_vertices(space, coords)` has a different behavior when coords is nothing.

lib/ClimaCoreMakie/src/fieldcontourf.jl
imreddyTeja: I made that change in the plot function below. This `exclude` is included because merge no longer works.

lib/ClimaCoreMakie/src/fieldheatmap.jl
imreddyTeja: I believe the new version of Makie uses ComputeGraphs over observables,

PR #2432, opened 2026-01-21
Fix conservation of split divergence

dennisYatunin:
This PR updates the `SplitDivergence` operator to make it discretely conservative for discontinuous fields. The current version is already conservative for fields that have been DSSed, but it includes additional terms at element boundaries that violate conservation for discontinuous quantities (e.g., velocities reconstructions on a cubed sphere).

The first commit in this PR removes the erroneous terms at element boundaries and updates the unit tests to demonstrate improved conservation properties. The unit tests are also simplified to avoid unnecessary checks and repetitive code.

The second commit in this PR makes the implementation of `SplitDivergence` more similar to the other spectral operators, ensuring correct use of `RecursiveApply` and consistent caching/indexing. Although this makes the code somewhat harder to read, it should simplify the process of removing `RecursiveApply` in #2417, with identical changes required for `SplitDivergence` and other spectral operators. (The convoluted cache indexing will also be removed from all operators in a future PR.) In addition, the docstring for `SplitDivergence` is updated to improve formatting and correct errors.

TODO: Update `SplitDivergence` in CUDA extension.

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -625,9 +625,9 @@ end
 
 """
     split_div = SplitDivergence()
-    split_div.(u, ψ)
+    split_div.(ρu, ψ)
 
-Computes the divergence of the product `u * ψ` using a **split-form (entropy-stable)** discretization.
+Computes the divergence of the product `ρu * ψ` using a **split-form (entropy-stable)** discretization.
 
 This operator is designed for the advection of scalar quantities in conservation laws (e.g., 
 thermodynamic variables or tracers). By evaluating the divergence using a specific averaging of the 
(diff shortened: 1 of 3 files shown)


src/Operators/spectralelement.jl
```diff
@@ -644,43 +644,72 @@ of two spectrally variable fields, thereby inhibiting the growth of quadratic in
...
+    \\frac{1}{2} \\left(
+        \\psi \\nabla \\cdot (\\rho \\mathbf{u}) +
+        \\rho \\mathbf{u} \\cdot \\nabla \\psi
+    \\right)
 ```
 
-## Discrete (SBP)
-The implementation uses the generalized summation-by-parts (SBP) property, where the divergence is the inverse of the mass matrix acting on the stiffness matrix. Let \\delta V_i = w_i J_i be the physical volume element (weights w, Jacobian J) and Q be the skew-symmetric stiffness matrix. The divergence is:
+## Discrete
+The split operator is discretized by applying the strong formulation of the
```
tapios:
```suggestion
The split operator is discretely equivalent to using the strong formulation of the
```
(since we are not actually discretizing it this way)

src/Operators/spectralelement.jl
```diff
@@ -691,46 +720,49 @@ SplitDivergence() = SplitDivergence{()}()
...
-        psi = get_node(space, arg2, ij, slabidx)
-        Ju1 = local_geometry.J * Geometry.contravariant1(u, local_geometry)
-        Ju1_slab[i] = Ju1
-        psi_slab[i] = psi
+        Ju1[slab_index(i)] =
+            local_geometry.J ⊠ RecursiveApply.rmap(
+                u -> Geometry.contravariant1(u, local_geometry),
+                get_node(space, arg1, ij, slabidx),
+            )
+        psi[slab_index(i)] = get_node(space, arg2, ij, slabidx)
```
tapios, line 738: Just curious: I find these 6 lines harder to read than the pattern it replaces. What's the advantage of writing it this way?

dennisYatunin: The reason I did this was to enforce consistency with the rest of the code in this file, so future PRs that refactor RecursiveApply and DataLayouts will not be behavior changing. But I totally agree that this is harder to read and needs to be refactored soon.

The RecursiveApply functions in this file allow iteration over multiple mass fluxes and/or multiple scalar quantities within the same fused operation. I'm working on removing the RecursiveApply module entirely from ClimaCore while maintining this behavior in #2417, which should eliminate most of the complexity introduced here.

As for the convoluted indexing required by DataLayouts, I'm honestly not sure why that is used throughout this file. The original intent of DataLayouts was to rearrange values in global memory so that L1/L2 caching could be improved, but it's not clear to me how this is useful for `MArray`s that are already cached in register memory. I will explore this more carefully in a future PR that refactors the DataLayouts module. If I don't end up finding any benefit to wrapping register caches in `IF` and `IJF` DataLayouts, I will remove these wrappers from `psi`, `out`, and all other register caches in ClimaCore.

src/Operators/spectralelement.jl
dennisYatunin: I think it's fair to say that we are actually discretizing this way, following the changes in this PR. There are two other equivalent ways to write down this discretization, including the compact form used in the actual implementation, but the unit tests confirm that they are all numerically identical (up to small roundoff errors).

tapios: I maintain that "is discretely equivalent" is more accurate than "is discretized by", since the formulation that is implemented does not look at all like the formula that follows here--it's just equivalent to it.

dennisYatunin: Very well, I'll use discretely equivalent

PR #2438, opened 2026-01-27
Add Buildkite pipeline to check flagship AMIP performance

### LINKED ISSUE #2433: Automatically check AMIP performance for each ClimaCore PR before merging
Initial conversation: https://github.com/CliMA/ClimaCore.jl/pull/2425#issuecomment-3770395670

If AMIP is the most important simulation case (per @tapios) for which we're trying to improve performance, we should probably be checking the AMIP performance of each PR to ClimaCore before merging instead of using the coarse nightly AMIP runs (in the Coupler repo) to check our effectiveness.

However, this is a challenge given the complex dependency relationship. Is there a proxy measurement in here that we can use?

Thoughts @dennisYatunin @imreddyTeja @daverumph?

petebachant:
Resolves #2433

Here we add the ClimaCoupler submodule to check how each PR changes performance in the most important case, since we have determined that the simpler benchmarks in this repo are not indicative enough. I suppose we could limit this to ClimaAtmos, but if the integrated land could be impacted by ClimaCore changes, why not test the whole AMIP pipeline end-to-end?

In lieu of a submodule, we could use a Julia project with the Coupler package in it, but I'm not sure if the AMIP driver script and config files would be accessible.

This pipeline will only run for commits with `[perf]` in their commit message, so we don't tie up `clima` with extra jobs. If you need to kick it off, you can always run `git commit --allow-empty -m "Trigger build [perf]" && git push`.

## TODO

- [x] Ensure GitHub webhooks are set properly
- [x] Check performance threshold against history
- [x] Dev install ClimaCore in the appropriate environment
- [x] Get performance threshold right -- probably should be 0.08
- [x] Should this be triggered manually, i.e., once a PR passes all other checks, we can run this performance check?
- [ ] Select the appropriate queue
- [ ] Ensure compiled files are cached between runs


imreddyTeja:
I'm not sure how much can be gained by using a sysimage or precompileCI here. My understanding is that almost anything that is compiled for ClimaAtmos, ClimaLand, and ClimaCoupler, would be invalidated because each PR/commit would be using a different version of ClimaCore. Probably still worth looking into though.

I don't think this should be run on every push, but I also don't know what a good alternative would be...

petebachant:
@nefrathenrici do you think it would be smarter/possible to put this into the `central` queue? We'd want to ensure it runs on the same GPU type every time to make SYPD comparable. I'm worried about tying up `clima` with these jobs.

nefrathenrici:
> @nefrathenrici do you think it would be smarter/possible to put this into the `central` queue? We'd want to ensure it runs on the same GPU type every time to make SYPD comparable. I'm worried about tying up `clima` with these jobs.

`central` is a good option if the jobs aren't too long, we have a reserved GPU node meant for quick CI jobs. Could we limit this test to only run a few steps?

petebachant:
> I think `central` is a reasonable choice in that case. Could we run this pipeline only on merge?

IMO it would be nice to know that a PR either enhances or at least doesn't hurt performance before merging.

PR #2441, opened 2026-01-30
Remove performance check pipeline git diff step and git submodule

petebachant:
We don't need this any longer since we're triggering on commit message content. A soft fail doesn't seem to stop dependents in Buildkite pipelines anyway.

Also, apparently a submodule is incompatible with Julia package management.


.buildkite/perf/.gitignore
```diff
@@ -0,0 +1 @@
+/ClimaCoupler.jl
```
ph-kev, line 1: Is this added to the .gitignore, because you were testing it locally?

.buildkite/perf/pipeline.yml
```diff
@@ -10,31 +10,17 @@ env:
   GKSwstype: 100
   SLURM_KILL_BAD_EXIT: 1
   COUPLER_PATH: ".buildkite/perf/ClimaCoupler.jl"
+  COUPLER_COMMIT: "74e724d8d86c216b90f0bd4dbaf2ec8d0eb63c36"
   CONFIG_PATH: "$COUPLER_PATH/config/nightly_configs"
```
ph-kev, line 14: Is there a reason this particular commit was chosen?

.buildkite/perf/.gitignore
petebachant: Just wanted to make sure it doesn't get committed accidentally in CI.

.buildkite/perf/pipeline.yml
petebachant: Basically, it works. We want to keep it stable so we are sure any SYPD changes come from ClimaCore.

PR #2442, opened 2026-01-31
Delete .gitmodules

ph-kev:
This is an empty file that should be deleted. It is causing problems with `Pkg.add`.

```julia
(.buildkite) pkg> add ClimaCore#kp/pfull
    Updating git-repo `https://github.com/CliMA/ClimaCore.jl.git`
ERROR: GitError(Code:ERROR, Class:Submodule, cannot get submodules without a working tree)
```

Edit: This isn't causing the problem. I haven't rebase from main, but this is an empty file that should be removed.


ph-kev:
> Strange that Julia complains since it's an empty file, but LGTM!

I haven't rebase from main, so that was the issue. I still think we should remove it, since it is an empty file.

PR #2455, opened 2026-02-04
Simpler horizontal remapping methods for diagnostics 

akshaysridhar:
Enables simpler bilinear interpolation option (see ClimaInterpolations.jl) within the Remapping module. 

Example: 
Screenshot 2026-02-04 at 1 16 01 PM

### DIFF
--- a/docs/src/remapping.md
+++ b/docs/src/remapping.md
@@ -45,8 +45,12 @@ given `field`. To obtain such coordinates, you can call the
 functions. These functions return an `Array` with the coordinates over which
 interpolation will occur. These arrays are of type `Geometry.Point`s.
 
-By default, vertical interpolation is switched off and the `field` is evaluated
-directly on the levels.
+By default, vertical interpolation is off (field evaluated on levels). Horizontal
+interpolation: `SpectralElementRemapping()` (default; uses spectral element quadrature weights) or `BilinearRemapping()`:
+
+```julia
+interpolated_array = Remapping.interpolate(field; horizontal_method = Remapping.BilinearRemapping())
+```
 
 `ClimaCore.Remapping.interpolate` allocates new output arrays. As such, it is
 not suitable for performance-critical applications.
@@ -76,6 +80,8 @@ hcoords = [Geometry.LatLongPoint(lat, long) for long in longpts, lat in latpts]
 zcoords = [Geometry.ZPoint(z) for z in zpts]
 
 interpolated_array = interpolate(field, hcoords, zcoords)
+# Or, to use bilinear remapping without spectral element weighting:
+# interpolate(field, hcoords, zcoords; horizontal_method = Remapping.BilinearRemapping())
 ```
 The output is defined on the Cartesian product of `hcoords` with `zcoords`.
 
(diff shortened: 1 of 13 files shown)


docs/src/remapping.md
```diff
@@ -76,9 +80,14 @@ hcoords = [Geometry.LatLongPoint(lat, long) for long in longpts, lat in latpts]
 zcoords = [Geometry.ZPoint(z) for z in zpts]
 
 interpolated_array = interpolate(field, hcoords, zcoords)
+# Bilinear: interpolate(field, hcoords, zcoords; horizontal_method = Remapping.BilinearRemapping())
```
ph-kev:
```suggestion
interpolate(field, hcoords, zcoords; horizontal_method = Remapping.BilinearRemapping())
```

docs/src/remapping.md
```diff
@@ -76,9 +80,14 @@ hcoords = [Geometry.LatLongPoint(lat, long) for long in longpts, lat in latpts]
...
 ```
 The output is defined on the Cartesian product of `hcoords` with `zcoords`.
 
+#### Diagnostics and NetCDF writers
+
+Pass `horizontal_method` through when remapping for output (e.g. `Remapping.interpolate(..., horizontal_method = BilinearRemapping())` or `Remapper(..., horizontal_method = BilinearRemapping())`).
```
ph-kev: I don't think it make sense to include NetCDF writers since that's something from ClimaDiagnostics.

src/Remapping/distributed_remapping.jl
```diff
@@ -210,12 +214,21 @@ struct Remapper{
...
     colons::T11
+
+    # Horizontal remapping method (SpectralElementRemapping or BilinearRemapping)
+    horiz_method::AbstractRemappingMethod
+
+    # Bilinear only: (s, t) ∈ [0,1]² and (i, j) = lower-left node of 2-point cell.
+    local_bilinear_s::T12
+    local_bilinear_t::T13
+    local_bilinear_i::T14
+    local_bilinear_j::T15
```
ph-kev: Would it be possible to move these fields into `BilinearRemapping`? One option would be something like
```julia
struct BilinearMapping{T12, T13, T14, T15}
    local_bilinear_s::T12
    local_bilinear_t::T13
    local_bilinear_i::T14
    local_bilinear_j::T15
end

function BilinearMapping()
    BilinearMapping(nothing, nothing, nothing, nothing)
end
```

and inside the constructor for Remapping, you would reconstruct `BilinearMapping` if you want to preserve the current interface.

ph-kev (commented):
I think the tests in ClimaDiagnostics are failing because a default wasn't provided for `horizontal_method` in the `Remapper` object?

I think someone who know more about the math of interpolation should look at this too.

examples/remap_visualization.jl
```diff
@@ -0,0 +1,199 @@
...
+vertmesh = Meshes.IntervalMesh(vertdomain, nelems = 1)
+verttopo = Topologies.IntervalTopology(ClimaComms.SingletonCommsContext(device), vertmesh)
+vert_center_space = Spaces.CenterFiniteDifferenceSpace(verttopo)
+
+# --- Horizontal: spectral elements ---
+quad = Quadratures.GLL{Nq}()
+horzmesh = Meshes.RectilinearMesh(horzdomain, nelements_horz, nelements_horz)
+horztopology = Topologies.Topology2D(ClimaComms.SingletonCommsContext(device), horzmesh)
+horzspace = Spaces.SpectralElementSpace2D(horztopology, quad)
+hv_center_space = Spaces.ExtrudedFiniteDifferenceSpace(horzspace, vert_center_space)
```
imreddyTeja: Could you just use a `SpectralElementSpace2D` for this example? Or do you want to show it working on an extrudedFD space?

ext/cuda/remapping_distributed.jl
```diff
@@ -4,6 +4,293 @@ import CUDA
 using CUDA: @cuda
 import ClimaCore.Remapping: _set_interpolated_values_device!
 
+# Bilinear in (s,t) ∈ [0,1]² (local to 2-point cell; reference element is [-1,1]²):
+# (1-s)(1-t)*c11 + s*(1-t)*c21 + (1-s)*t*c12 + s*t*c22
+@inline _bilinear(c11, c21, c22, c12, s, t) =
+    (1 - s) * (1 - t) * c11 + s * (1 - t) * c21 + (1 - s) * t * c12 + s * t * c22
+
+# Linear in s ∈ [0,1] on 2-point cell (1D)
+@inline _linear(c1, c2, s) = (1 - s) * c1 + s * c2
```
imreddyTeja: Are these duplicated in src?

ext/cuda/remapping_distributed.jl
```diff
@@ -4,6 +4,293 @@ import CUDA
...
+    vert_interpolation_weights,
+    vert_bounding_indices,
+    field_values,
+)
+    num_horiz = length(local_horiz_indices)
+    num_vert = length(vert_bounding_indices)
+    num_fields = length(field_values)
+    inds = (num_horiz, num_vert, num_fields)
+    i_thread =
+        (CUDA.blockIdx().x - Int32(1)) * CUDA.blockDim().x + CUDA.threadIdx().x
```
imreddyTeja: you can use `thread_index` from cuda_utils.jl for this

ext/cuda/remapping_distributed.jl
```diff
@@ -4,6 +4,293 @@ import CUDA
...
+    1 ≤ i_thread ≤ prod(inds) || return nothing
+    (i_out, j_v, k) = CartesianIndices(map(x -> Base.OneTo(x), inds))[i_thread].I
```
imreddyTeja, line 58: This could be `kernel_indexes(i_thread, inds).I`

imreddyTeja: Ahh I see that this follows the pattern of the other remapper kernels. I think either is fine

docs/src/remapping.md
```diff
@@ -76,6 +80,7 @@ hcoords = [Geometry.LatLongPoint(lat, long) for long in longpts, lat in latpts]
 zcoords = [Geometry.ZPoint(z) for z in zpts]
 
 interpolated_array = interpolate(field, hcoords, zcoords)
+# interpolate(field, hcoords, zcoords; horizontal_method = Remapping.BilinearRemapping())
```
juliasloan25, line 84:
```suggestion
# Or, to use bilinear remapping without spectral element weighting:
# interpolate(field, hcoords, zcoords; horizontal_method = Remapping.BilinearRemapping())
```

src/Remapping/distributed_remapping.jl
```diff
@@ -257,51 +263,71 @@ Remapper(
...
+    horizontal_method::AbstractRemappingMethod = SpectralElementRemapping(),
+) = _Remapper(space; target_zcoords, target_hcoords, buffer_length, horizontal_method)
 
-# Purely vertical case
+# Purely vertical case (horizontal_method accepted for uniform API, ignored)
 Remapper(
     space::Spaces.FiniteDifferenceSpace;
     target_zcoords::AbstractArray = default_target_zcoords(space),
     buffer_length::Int = 1,
+    horizontal_method::AbstractRemappingMethod = SpectralElementRemapping(),
```
imreddyTeja, line 283: I'm OK with this, but it might also make sense to create a subtype of `AbstractRemappingMethod` for the purely vertical case

examples/remap_visualization.jl
```diff
@@ -0,0 +1,199 @@
+# Visualize bilinear and spectral interpolation remapping: slotted cylinder (Zalesak)
```
juliasloan25: This example is really nice! Could you add it as a page in the rendered docs?

src/Remapping/distributed_remapping.jl
```diff
@@ -241,6 +244,9 @@ Keyword arguments
 for interpolation. Effectively, this controls how many fields can be remapped simultaneously
 in `interpolate`. When more fields than `buffer_length` are passed, the remapper will batch
 the work in sizes of `buffer_length`.
+
+`horizontal_method`: `SpectralElementRemapping()` (default; uses spectral element quadrature weights)
+or `BilinearRemapping()` (1D: linear on 2-point cell; 2D: bilinear on 2×2 cell; GPU-compatible).
```
juliasloan25: Is the spectral element option not GPU compatible? If so that could be documented a little more clearly (e.g. in the docstring for `SpectralElementRemapping`)

src/Remapping/remapping_utils.jl
```diff
@@ -1,3 +1,38 @@
+"""
+    AbstractRemappingMethod
+
+Abstract type for horizontal remapping methods. Dispatch on concrete subtypes to avoid
+branching in interpolation and Remapper logic.
+"""
+abstract type AbstractRemappingMethod end
```
juliasloan25, line 7: can you add these new types to the Remapping API docs?

src/Remapping/distributed_remapping.jl
```diff
@@ -351,14 +377,55 @@ function _Remapper(
...
+                push!(s_arr, (ξ1 - quad_pts[i]) / (quad_pts[i + 1] - quad_pts[i]))
+                push!(t_arr, (ξ2 - quad_pts[j]) / (quad_pts[j + 1] - quad_pts[j]))
+                push!(i_arr, i)
+                push!(j_arr, j)
+            end
+            local_bilinear_s = ArrayType(s_arr)
+            local_bilinear_t = ArrayType(t_arr)
+            local_bilinear_i = ArrayType(i_arr)
+            local_bilinear_j = ArrayType(j_arr)
+            local_horiz_interpolation_weights = (
```
imreddyTeja: Is this array is unused? If so it could just be `nothing`.  It might also make sense to use `local_horiz_interpolation_weights` to store a tuple containing `(s,t,i,j)` for each point

src/Remapping/distributed_remapping.jl
```diff
@@ -351,14 +377,55 @@ function _Remapper(
...
+                (ξ1, i) in zip(ξ1s, i_arr)
+            ]
+            local_bilinear_i = ArrayType(i_arr)
+            local_bilinear_s = ArrayType(s_arr)
+            local_bilinear_t = local_bilinear_j = nothing
+            local_horiz_interpolation_weights =
+                (ArrayType(zeros(FT, length(local_target_hcoords), 0)),)
+        else
+            # 2D: bilinear on 2×2 cell
+            s_arr = FT[]
```
juliasloan25: I think it's more performant to fully allocate the arrays and fill them in the loop rather than pushing and growing their size (same for the other arrays here). I'm not sure if this is the right size for it
```suggestion
            s_arr = ArrayType{FT}(undef, length(ξs_split[1])
```

src/Remapping/distributed_remapping.jl
```diff
@@ -351,14 +377,55 @@ function _Remapper(
...
+            local_bilinear_s = ArrayType(s_arr)
+            local_bilinear_t = ArrayType(t_arr)
+            local_bilinear_i = ArrayType(i_arr)
+            local_bilinear_j = ArrayType(j_arr)
+            local_horiz_interpolation_weights = (
+                ArrayType(zeros(FT, length(local_target_hcoords), 0)),
+                ArrayType(zeros(FT, length(local_target_hcoords), 0)),
+            )
+        end
+    else
```
juliasloan25:
```suggestion
    else # SpectralElementRemapping
```

src/Remapping/distributed_remapping.jl
```diff
@@ -510,6 +617,149 @@ function _set_interpolated_values!(remapper::Remapper, fields)
...
+        fv = Fields.field_values(field)
+        @inbounds for (vindex, (A, B)) in enumerate(vert_interpolation_weights)
+            (v_lo, v_hi) = vert_bounding_indices[vindex]
+            for (out_index, h) in enumerate(local_horiz_indices)
+                i, s = local_bilinear_i[out_index], local_bilinear_s[out_index]
+                out[out_index, vindex, field_index] =
+                    A * _linear(fv[CI(i, 1, 1, v_lo, h)], fv[CI(i + 1, 1, 1, v_lo, h)], s) +
+                    B * _linear(fv[CI(i, 1, 1, v_hi, h)], fv[CI(i + 1, 1, 1, v_hi, h)], s)
+            end
+        end
```
juliasloan25, line 646: Could you add a few inline comments here? The index notation isn't super clear to me. Or a docstring might work too

ext/cuda/remapping_interpolate_array.jl
```diff
@@ -138,13 +138,18 @@ function interpolate_slab_level_kernel!(
         space = axes(field)
         FT = Spaces.undertype(space)
         Nq1, Nq2 = length(I1), length(I2)
+        quad = Spaces.quadrature_style(space)
+        Nq = Quadratures.degrees_of_freedom(quad)
+        is_bilinear = (Nq1 == 2 && Nq2 == 2)
```
juliasloan25, line 186: Could you explain this? Why do we use bilinear remapping only when Nq is 2 in both directions?

juliasloan25 (commented):
Thanks Akshay! The results look really nice. Just to clarify - the default remapping remains unchanged (spectral element), and now we have an option for bilinear remapping which can be used for e.g. diagnsotics?

src/Remapping/distributed_remapping.jl
akshaysridhar: It is, will clear up docstrings.

akshaysridhar:
> Thanks Akshay! The results look really nice. Just to clarify - the default remapping remains unchanged (spectral element), and now we have an option for bilinear remapping which can be used for e.g. diagnsotics?

The intent here was to retain the existing SEM interpolation and add a simpler option that people can invoke elsewhere if necessary. Bilinear interpolation will "smooth" the result (but will not generate spurious negative values in the output due to ringing from higher polynomial orders). For diagnostic outputs only, I'd like to make the bilinear interpolation the default choice. I expect that, where conservation is critical we would be using the `ConservativeRegridding` tools rather than this `Remapping` module.

src/Remapping/remapping_utils.jl
```diff
@@ -1,3 +1,38 @@
...
+"""
+struct BilinearRemapping{T12, T13, T14, T15} <: AbstractRemappingMethod
+    local_bilinear_s::T12
+    local_bilinear_t::T13
+    local_bilinear_i::T14
+    local_bilinear_j::T15
+end
+
+"""`BilinearRemapping()` with no arguments: method tag; Remapper constructor fills in the arrays."""
+BilinearRemapping() = BilinearRemapping(nothing, nothing, nothing, nothing)
```
imreddyTeja, line 34: Is this needed? The constructor can also be dispatched to with the types

imreddyTeja (commented):
This looks mostly good to me. I'm a bit confused by the `interpolate_array` changes though

examples/remap_visualization.jl
akshaysridhar: Running this example on just a `horzspace=SpectralElementSpace2D` would be enough.

ext/cuda/remapping_distributed.jl
akshaysridhar: Yes; now moved to utils

ext/cuda/remapping_interpolate_array.jl
akshaysridhar: I added a (cursor-assisted) sketch to try and explain the indexing - it would take the "local sub element" bounded by the nearest 4 nodes; Nq1 and Nq2 are identifiers (distinct from Nq which is the source mesh quadrature DoF) for the part of the element used in the interpolation stencil.

juliasloan25: This makes sense, thank you!

PR #2457, opened 2026-02-14
Simplify Tensor interface, avoid hard-coded projections

### LINKED ISSUE #2399: Geometry module refactor
The Geometry module needs to be refactored due to code duplication and unnecessary abstraction. To refactor, we should figure out the minimal representation/data structures needed to represent geometry, then build around this structure.

Key issues to resolve:
- Multiple redundant and inconsistent constructors for AxisTensors with different conventions/ways of doing the same thing
- Unnecessary type alias constants (`const CovariantTensor = Union{CovariantVector, Covariant2Tensor}`)
- LinearAlgebra type piracy/functionality that could be handled natively by LinearAlgebra 
- Confusing and slow-to-compile generated functions
- More code reuse: use`project` for constructors, zero-padding
- Small inefficiencies, for example, `covariant1` converts a vector just to extract one component
- `rmul_with_projection` could be rewritten to use `Base.return_type`
- Confusing terminology: Terminology should be updated to match the dycore paper.

dennisYatunin:
This PR refactors the `Geometry` module to use a much simpler interface, so that all vector/tensor operations can expressed using standard array operations instead of custom API functions. Hard-coded projections have been replaced with coordinate-independent operations, and other forms of code duplication have also been removed. This new interface is a key step toward implementing a generic convolution operator, which will allow us to simplify a lot of our current `FiniteDifferenceOperator` code.


test/runtests.jl
```diff
@@ -25,7 +25,7 @@ UnitTest("DataLayouts 2dx"                         ,"DataLayouts/data2dx.jl"),
 UnitTest("DataLayouts mapreduce"                   ,"DataLayouts/unit_mapreduce.jl"),
 UnitTest("Geometry"                                ,"Geometry/geometry.jl"),
 UnitTest("rmul_with_projection"                    ,"Geometry/rmul_with_projection.jl"),
-UnitTest("AxisTensors"                             ,"Geometry/axistensors.jl"),
+UnitTest("Tensors"                             ,"Geometry/tensors.jl"),
```
dennisYatunin:
```suggestion
UnitTest("Tensors"                                 ,"Geometry/tensors.jl"),
```

test/Spaces/opt_spaces.jl
```diff
@@ -34,14 +34,14 @@ end
...
+        test_n_failures(676,  TU.SpectralElementSpace1D, context)
+        test_n_failures(866, TU.SpectralElementSpace2D, context)
         test_n_failures(4,  TU.ColumnCenterFiniteDifferenceSpace, context)
         test_n_failures(5,  TU.ColumnFaceFiniteDifferenceSpace, context)
-        test_n_failures(1147, TU.SphereSpectralElementSpace, context)
-        test_n_failures(1146, TU.CenterExtrudedFiniteDifferenceSpace, context)
-        test_n_failures(1146, TU.FaceExtrudedFiniteDifferenceSpace, context)
+        test_n_failures(872, TU.SphereSpectralElementSpace, context)
+        test_n_failures(881, TU.CenterExtrudedFiniteDifferenceSpace, context)
+        test_n_failures(881, TU.FaceExtrudedFiniteDifferenceSpace, context)
```
dennisYatunin: Wow, nice to see these go down by so much. I was not expecting `AxisTensors` to account for a quarter of our inference failures here.

test/Operators/finitedifference/tensor.jl
```diff
@@ -62,7 +62,7 @@ for FT in (Float32, Float64)
     S_ref = fill(S_scalar, fspace)
     @test S ≈ S_ref
 
-    @test norm(S_scalar) ≈ norm(Geometry.components(S_scalar))
+    @test norm(S_scalar) ≈ norm(Geometry.parent(S_scalar))
```
dennisYatunin:
```suggestion
    @test norm(S_scalar) ≈ norm(parent(S_scalar))
```

test/MatrixFields/gpu_compat_bidiag_matrix_row.jl
```diff
@@ -38,11 +38,11 @@ fspace = Spaces.FaceExtrudedFiniteDifferenceSpace(cspace)
...
-    AxisTensor{
-        GFT,
+    Tensor{
         2,
-        Tuple{CovariantAxis{(3,)}, ContravariantAxis{(3,)}},
-        SArray{Tuple{1, 1}, GFT, 2, 1},
+        GFT,
+        Tuple{Basis{Covariant, (3,)}, Basis{Contravariant, (3,)}},
+        SMatrix{1, 1, GFT, 1},
     },
```
dennisYatunin:
```suggestion
    typeof(C3(GFT(0)) * CT3(GFT(0))'),
```
Bit easier to read if you write out the value, since `C3` and `CT3` are already defined

test/Geometry/tensor_conversion_benchmarks.jl
```diff
@@ -142,11 +145,11 @@ reference_func(::typeof(Geometry.transform)) = ref_transform
 
 # Correctness comparisons
 components(x::T) where {T <: Real} = x
-components(x) = Geometry.components(x)
+components(x) = Geometry.parent(x)
```
dennisYatunin:
```suggestion
components(x) = parent(x)
```

test/Geometry/method_info.jl
```diff
@@ -1,354 +1,385 @@
 #! format: off
 
+# Short type aliases for LocalGeometry in method tables (Option C: 6 type params)
+const LG_13_XZ{FT}    = LocalGeometry{(1,3),XZPoint{FT},FT,
```
dennisYatunin: This file would probably be much simpler to read if all these types were generated programmatically, but I suppose we can leave it as is for now.

test/Geometry/geometry.jl
```diff
@@ -125,7 +125,8 @@ end
     @test_throws Exception wᵢ * uᵏ
 
     T = uᵏ ⊗ uᵏ
-    @test uᵏ ⊗ uᵏ isa Geometry.Contravariant2Tensor
+    @test uᵏ ⊗ uᵏ isa
+          Geometry.Tensor{2, <:Any, <:Tuple{Geometry.Basis{Geometry.Contravariant}, <:Any}}
```
dennisYatunin: I don't think there's much benefit to having tests of "internal structure" like this. It doesn't affect correctness, performance, or latency, so you can just omit it.

examples/hybrid/box/bubble_3d_rhotheta.jl
```diff
@@ -253,9 +253,9 @@ function rhs!(dY, Y, _, t)
     # Horizontal momentum
     @. dρuₕ += -uvdivf2c(ρw ⊗ If(uₕ))
     Ih = Ref(
-        Geometry.Axis2Tensor(
+        Geometry.Tensor(
+            (@SMatrix [1.0 0.0; 0.0 1.0]),
```
dennisYatunin:
```suggestion
            I,
```
This can be simplified by using `I`, since it's imported from `LinearAlgebra`

examples/hybrid/plane/density_current_2d_flux_form.jl
```diff
@@ -241,10 +241,7 @@ function rhs!(dY, Y, _, t)
 
     # horizontal momentum
     Ih = Ref(
-        Geometry.Axis2Tensor(
-            (Geometry.UAxis(), Geometry.UAxis()),
-            @SMatrix [1.0]
-        ),
+        Geometry.Tensor(SMatrix{1, 1}(1.0), (Geometry.UAxis(), Geometry.UAxis())),
```
dennisYatunin:
```suggestion
        Geometry.Tensor(I, (Geometry.UAxis(), Geometry.UAxis())),
```
This can also be simplified with `I`

examples/hybrid/sphere/baroclinic_wave_rhoe.jl
```diff
@@ -32,9 +32,9 @@ function postprocessing(sol, output_dir)
...
-    anim = Plots.@animate for Y in sol.u
-        ᶜv = Geometry.UVVector.(Y.c.uₕ).components.data.:2
-        Plots.plot(ᶜv, level = 3, clim = (-6, 6))
-    end
-    Plots.mp4(anim, joinpath(output_dir, "v.mp4"), fps = 5)
+    # anim = Plots.@animate for Y in sol.u
+    #     ᶜv = Geometry.UVVector.(Y.c.uₕ).components.data.:2
+    #     Plots.plot(ᶜv, level = 3, clim = (-6, 6))
+    # end
+    # Plots.mp4(anim, joinpath(output_dir, "v.mp4"), fps = 5)
```
dennisYatunin: Does this not work anymore? I thought the new `Tensor` could still be accessed with `components.data`.

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -310,9 +310,12 @@ function implicit_equation_jacobian!(j, Y, p, δtγ, t)
...
-    g³³(gⁱʲ) = Geometry.AxisTensor(
-        (Geometry.Contravariant3Axis(), Geometry.Contravariant3Axis()),
-        Geometry.components(gⁱʲ)[end],
+    g³³(gⁱʲ) = Geometry.Tensor(
+        SMatrix{1, 1}(parent(gⁱʲ)[end]),
+        (
+            Geometry.Basis{Geometry.Contravariant, (3,)}(),
+            Geometry.Basis{Geometry.Contravariant, (3,)}(),
+        ),
     )
```
dennisYatunin:
```suggestion
    g³³(gⁱʲ) = reshape(
        gⁱʲ,
        Geometry.Contravariant3Axis(),
        Geometry.Contravariant3Axis(),
    )
```
I think this is equivalent?

ext/cuda/operators_finite_difference.jl
```diff
@@ -86,7 +86,10 @@ function Base.copyto!(
         # this block config is better for VIJFH. It is only used when the total number of
         # threads in a block is between 32 and 256 to avoid underutilization of the GPU and
         # errors due to too many registers used when the block size is too large.
-        if !Topologies.isperiodic(space) && mask isa NoMask &&
+        # TODO: the eager path crashes on pure 1D FiniteDifferenceSpaces
+        # (Ni = 1), e.g. convergence_column.jl. Restricted to extruded spaces (Ni > 1)
+        # until the 1D-column codepath through eager_copyto_stencil_kernel! is fixed.
```
dennisYatunin: 🙀 Could you add some info about what exactly is wrong with the single-column case, if you know?

ext/cuda/operators_spectral_element.jl
```diff
@@ -181,6 +181,12 @@ Base.@propagate_inbounds function resolve_shmem!(obj, ij, slabidx)
     nothing
 end
 
+# Extract the first basis of a Tensor type statically from its type parameter,
+# avoiding a runtime `axes(::Type)` call
+@inline _tensor_row_axis(
+    ::Type{<:Geometry.Tensor{N, T, B, C}},
+) where {N, T, B, C} = B.instance[1]
```
dennisYatunin: I'm confused how this is different from `first(axes(T))`? The definition of `axes(::Type{<:Tensor})` is identical to this.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+Base.show(io::IO, ::ScalarBasis) = print(io, "ScalarBasis()")
+
+no_metric_error(T1, T2) =
+    throw(DimensionMismatch("Metric is needed for change of basis: $T1 vs $T2"))
+scalar_error(T) =
+    throw(DimensionMismatch("Incompatible bases: one scalar vs $T vectors"))
+
+check_same_type(::T1, ::T2) where {T1, T2} = T1 == T2 || no_metric_error(T1, T2)
+check_same_type(::OneScalar, ::T) where {T} = T == OneScalar || scalar_error(T)
+check_same_type(::T, ::OneScalar) where {T} = T == OneScalar || scalar_error(T)
```
dennisYatunin:
```suggestion
check_same_type(::OneScalar, ::T) where {T} = scalar_error(T)
check_same_type(::T, ::OneScalar) where {T} = scalar_error(T)
```
The method below already handles the case where they're equal

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+check_same_type(::OneScalar, ::OneScalar) = true
+
+function combine_bases(b1::Basis, b2::Basis)
+    check_same_type(basis_type(b1), basis_type(b2))
+    b1 == b2 && return b1 # fast path avoids unrolled_unique union return type
+    return Basis(
+        basis_type(b1),
+        unrolled_unique((basis_vector_names(b1)..., basis_vector_names(b2)...)),
+    )
+end
```
dennisYatunin:
```suggestion
combine_bases(b::B, ::B) where {B <: Basis} = b
combine_bases(b1::Basis, b2::Basis) =
    check_same_type(basis_type(b1), basis_type(b2)) &&
    Basis(
        basis_type(b1),
        unrolled_unique((basis_vector_names(b1)..., basis_vector_names(b2)...)),
    )
```
Avoiding `check_same_type` in the fast path will probably reduce latency by a tiny bit

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+combine_bases(b1::Basis, b2::Basis, rest::Basis...) =
+    combine_bases(combine_bases(b1, b2), rest...)
```
dennisYatunin:
```suggestion
combine_bases(b::Basis, bases::Basis...) =
    unrolled_reduce(combine_bases, bases; init = b)
```
Use `unrolled_reduce` to avoid recursion limits during compilation

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+# @generated so the filter is computed at macro expansion time from the
+# Basis type parameters. A non-generated version using
+# `unrolled_filter(in(basis_vector_names(b2)), basis_vector_names(b1))`
+# hits a union return type: `Base.Fix2(in, tuple)` stores the tuple as a
+# runtime field (not a type parameter), so `unrolled_filter` cannot
+# constant-fold the predicate and the return type widens.
```
dennisYatunin: I'm pretty sure you can avoid this issue by using `unrolled_in` instead of `in`. The variable manipulations in `field_name_set.jl` are very similar and don't require additional generated functions like this.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+                             bases, $(unrolled_map(length, bases))"))
+
+# Allow UniformScaling as components for square 2-tensors (materializes into SMatrix)
+function Tensor(s::UniformScaling, bases::NTuple{2, Basis})
+    N1 = length(bases[1])
+    N2 = length(bases[2])
+    N1 == N2 ||
+        throw(DimensionMismatch("UniformScaling requires square tensor, got $(N1)×$(N2)"))
+    T = typeof(s.λ)
+    Tensor(SMatrix{N1, N2, T}(s), bases)
```
dennisYatunin, line 269: Is the conversion to an `SMatrix` necessary? Keeping the components as a `UniformScaling` would let us reduce operation counts by a factor of 4 in 2D or 9 in 3D, which seems like a good optimization to allow.

If the conversion is necessary, I'd suggest replacing "materializes" with "converted" in the comment above, since there isn't any broadcasting here.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+end
+
+Base.parent(x::Tensor) = x.components
+Base.axes(x::Tensor) = x.bases
+# Type-level axes: extracts the bases tuple from the type parameter B without
+# needing an instance. Used by MatrixFields to enumerate tensor components from
+# a type alone (e.g., when computing scalar keys for a Tensor{2} matrix entry).
+Base.axes(::Type{Tensor{N, T, B, C}}) where {N, T, B, C} = B.instance
```
dennisYatunin: This is somewhat abusing the `AbstractArray` interface, since `axes` can't be applied to any other type. Might be better to make this its own function, or call `Core.Compiler.return_type(axes, Tuple{T})` to get the result of `axes(::T)`. I'd recommend a new function, though, since the syntax for calling `return_type` is bit cumbersome.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+
+
+Base.zero(x::Tensor) = zero(typeof(x))
+Base.zero(x::Adjoint{<:Any, <:AbstractTensor}) = adjoint(zero(parent(x)))
+Base.zero(::Type{Adjoint{T, V}}) where {T, V <: AbstractTensor} = adjoint(zero(V))
+Base.one(x::Tensor) = one(typeof(x))
+
+Base.zero(::Type{Tensor{N, T, B, C}}) where {N, T, B, C} =
+    Tensor(zero(C), B.instance)
+# Adjoint{T,P} has no type-level zero; unwrap, zero the parent, re-wrap.
```
dennisYatunin: Isn't `zero(::Type{<:Adjoint})` defined a few lines above?

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+        tensor isa AbstractTensor || return tensor # return scalar values
+        return add_any_bases(tensor, x, bases) # add bases to non-scalars
+    end
+    components = f(unrolled_map(parent ∘ Base.Fix2(reshape, bases), args)...)
+    components isa AbstractArray || return components  # return scalar values
+    return Tensor(components, bases)               # add bases to non-scalars
+end
+
+Base.map(f::F, args::AbstractTensor...) where {F} =
+    reshape_and_apply_f(Base.Fix1(map, f), args...)
```
dennisYatunin:
```suggestion
Base.map(f::F, x::AbstractTensor) where {F} = apply_f(Base.Fix1(map, f), x)
Base.map(f::F, args::AbstractTensor...) where {F} =
    reshape_and_apply_f((args...) -> map(f, args...), args...)
```
`Base.Fix1` only allows more than one argument as of Julia 1.12, so you need to define a new function for this to work on earlier versions

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+
+# Use method dispatch instead of a ternary to avoid polluting the return type:
+# Julia must infer both branches of a ternary, and the AnyBasis branch recurses
+# with parent(x)::SMatrix which has no overlap_bases method, widening to Any.
+new_basis_for_product(x, y) =
+    _new_basis_for_product(axes(x, ndims(x)), axes(y, 1), x, y)
+_new_basis_for_product(ax, ay, _, _) = overlap_bases(ax, dual(ay))
+_new_basis_for_product(::AnyBasis, ::AnyBasis, x, y) =
```
dennisYatunin:
```suggestion
_new_basis_for_product(::AnyBasis, _, x, y) =
```
The second type constraint here seems unnecessary.

Also, I'm not sure I understand the issue described in the comment above. Your workaround looks reasonable, but note that recursing through two different functions (in this case, `new_basis_for_product` and `_new_basis_for_product`) is much more likely to hit a recursion limit than recursing through a single function. At least, that's what I observed in my recent `RecursiveApply` and `struct.jl` refactoring PRs.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+# Tensor with AnyBasis assigned to its axis at index n, which can be reshaped to
+# have any BasisType for that axis (and any basis_vector_names for all its axes)
+struct TensorWithAnyBasis{n, N, T, B, X <: AbstractTensor{N, T}, G <: Metric} <:
+       AbstractTensor{N, T, B}
+    x::X
+    g::G
+end
+
+function TensorWithAnyBasis{n}(x::X, g::G) where {n, X, G}
+    bases = Base.setindex(axes(x), AnyBasis(), n)
```
dennisYatunin:
```suggestion
    bases = Utilities.Unrolled.unrolled_setindex(axes(x), AnyBasis(), Val(n))
```
`Base.setindex` does not guarantee constant propagation of the index, but that can be fixed by using `unrolled_setindex` from #2488. This also applies to every subsequent use of `Base.setindex` in this file.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+#########################################
+
+Base.:+(args::AbstractTensor...) = reshape_and_apply_f(+, args...)
+Base.:-(x::AbstractTensor, y::AbstractTensor) = reshape_and_apply_f(-, x, y)
+Base.:(==)(x::AbstractTensor, y::AbstractTensor) =
+    unrolled_map(basis_type, axes(x)) == unrolled_map(basis_type, axes(y)) &&
+    reshape_and_apply_f(==, x, y)
+Base.isapprox(x::AbstractTensor, y::AbstractTensor; kwargs...) =
+    unrolled_map(basis_type, axes(x)) == unrolled_map(basis_type, axes(y)) &&
+    reshape_and_apply_f((x, y) -> isapprox(x, y; kwargs...), x, y)
```
dennisYatunin:
```suggestion
Base.:(==)(x::AbstractTensor, y::AbstractTensor) = reshape_and_apply_f(==, x, y)
Base.isapprox(x::AbstractTensor, y::AbstractTensor; kwargs...) =
    reshape_and_apply_f((x, y) -> isapprox(x, y; kwargs...), x, y)
```
Is there a reason to avoid `BasisType` conversions here? It seems reasonable to convert tensors to a common `BasisType` before checking for equality.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+    syms::NTuple{3, Symbol}, names::NTuple{N, Int}, name::Symbol,
+) where {N}
+    for n in 1:N
+        i = names[n]
+        if i <= 3 && name === syms[i]
+            return n
+        end
+    end
+    return nothing
+end
```
dennisYatunin:
```suggestion
_sym_to_index(syms, names, sym) =
    unrolled_findfirst(name -> syms[name] == sym, names)
```
Here's a simpler version of this (with `name` renamed to `sym` since it is more similar to `syms` than `names`). You can also use `===` instead of `==`, but constant propagation should make that unnecessary here.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+    # Pass through to getfield for actual struct fields
+    name === :components && return getfield(x, :components)
+    name === :bases && return getfield(x, :bases)
+    # Named component access: v.u₁, v.u², v.u, etc.
+    # Use getfield directly to avoid recursive getproperty calls, which would
+    # cause Julia to infer a union return type for b, making basis_type a runtime dispatch.
+    b = getfield(x, :bases)[1]
+    syms = _symbols(basis_type(b))
+    idx = _sym_to_index(syms, basis_vector_names(b), name)
+    isnothing(idx) ? zero(eltype(x)) : @inbounds getfield(x, :components)[idx]
```
dennisYatunin: It would be good to check if `name` is in `syms` before returning `zero(eltype(x))`. Something like `(::Covariant3Vector).w` should throw an error, but with the current version it will return `0` instead.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+    arg1::Real,
+) where {B <: Basis} = Tensor(SVector(arg1), (B.instance,))
+
+(::Type{Tensor{1, T, Tuple{B}, SVector{2, T}}} where {T})(
+    arg1::Real, arg2::Real,
+) where {B <: Basis} = Tensor(SVector(arg1, arg2), (B.instance,))
+
+(::Type{Tensor{1, T, Tuple{B}, SVector{3, T}}} where {T})(
+    arg1::Real, arg2::Real, arg3::Real,
+) where {B <: Basis} = Tensor(SVector(arg1, arg2, arg3), (B.instance,))
```
dennisYatunin:
```suggestion
(::Type{T})(args::Number...) where {T <: Tensor{1}} =
    Tensor(SVector(args...), axes(T))
```
All the type constraints here are checked by the `Tensor` struct itself, so these definitions can be made much simpler. And I don't see a need to restrict the arguments to `Real` numbers; any `Number` type should be acceptable here.

Also, isn't this equivalent to the "splatted constructors" in the `@eval` loop above? I feel like only one of these needs to be provided (preferably this version as it does not require a bunch of generated constructor names).

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+    # Check that no components were dropped
+    src_names = basis_vector_names(axes(v, 1))
+    dest_names = basis_vector_names(b)
+    T = eltype(v)
+    for n in 1:length(src_names)
+        i = src_names[n]
+        if !(i in dest_names) && parent(v)[n] != zero(T)
+            throw(InexactError(:transform, typeof(b), v))
+        end
+    end
```
dennisYatunin:
```suggestion
    unrolled_all(
        (name, value) -> name in dest_names || iszero(value),
        zip(src_names, parent(v)),
    ) || throw(InexactError(:transform, typeof(b), v))
```
This should be unrolled to get optimized out during compilation; same goes for the function below.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,707 @@
...
+"""
+function outer end
+const ⊗ = outer
+
+@inline outer(x::AbstractVector, y::AbstractVector) = x * y'
+@inline outer(x::AbstractVector, y::Number) = x * y
+@inline outer(x::AbstractVector, y) = RecursiveApply.rmap(y -> x ⊗ y, y)
+
+cross(x::UVVector, y::WVector) = UVVector(x.v * y.w, -x.u * y.w)
+cross(y::WVector, x::UVVector) = UVVector(-x.v * y.w, x.u * y.w)
```
dennisYatunin: Does this only need to be defined for `UVVector` and `WVector`? Seems a bit arbitrary compared to the other multiplication operators (`*`/`dot`/`outer`).

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,379 @@
...
+# compile time. Any method that performs a matrix-vector product must instead
+# dispatch on Tensor{1, <:Any, <:Tuple{Basis{BT, names}}} where {names} so
+# that Julia captures `names` as a type parameter and avoids runtime dispatch
+# in basis overlap/dual computations. 
+
+# Convert any Tensor{1} to the Contravariant basis of lg
+@inline _to_contravariant(v::ContravariantTensor, ::LocalGeometry) = v
+@inline _to_contravariant(
+    v::Tensor{1, <:Any, <:Tuple{Basis{Orthonormal, names}}}, lg::LocalGeometry,
+) where {names} = lg.∂ξ∂x * v
```
dennisYatunin:
```suggestion
@inline _to_contravariant(v::LocalVector, lg::LocalGeometry) = lg.∂ξ∂x * v
```
The vector aliases in `tensors.jl` would make these method definitions much easier to read

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,379 @@
...
+# directly.
+#
+# This arises in extruded geometries: the horizontal `LocalGeometry` has a 2D
+# metric (dimensions 1,2 only), so `gⁱʲ * Covariant123` yields `Contravariant12`
+# — the metric knows nothing about dimension 3. But the vertical direction is
+# orthogonal with unit arc-length by construction (reference and physical
+# coordinates coincide there), giving g^{33} = g_{33} = 1. Therefore
+# u³ = u₃ = u (all three representations are equal), and projecting a
+# `Covariant123` source to `Contravariant123` should copy component 3 as-is
+# rather than zero-filling it.
```
dennisYatunin: This complexity can be avoided by only returning 3-by-3 matrices from the `LocalGeometry`. As long as `g` covers all possible tensor components, the standard `reshape` function will be sufficient.

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,379 @@
...
+@inline function transform(
+    b::Basis{Covariant}, v::AbstractTensor{1}, lg::LocalGeometry,
+)
+    return transform(b, _to_covariant(v, lg))
+end
+@inline function transform(
+    b::Basis{Orthonormal}, v::AbstractTensor{1}, lg::LocalGeometry,
+)
+    return transform(b, _to_local(v, lg))
+end
```
dennisYatunin:
```suggestion
transform(b, v::AbstractTensor{1}, lg::LocalGeometry)
    transform(b, _to_basis_type(basis_type(b), v, lg))
```
Using less specific low-level methods (e.g., replacing `_to_contravariant`/`_to_covariant`/`_to_local` with a more generic `_to_basis_type`) will let you eliminate a lot of the code duplication in this file

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,379 @@
...
-    project(CovariantAxis{I}(), project(LocalAxis{I}(), u), local_geometry)
+# Use explicit `where {names}` on matrix-product methods so Julia captures the
+# names type parameter and can infer concrete return types (see comment at top).
+@inline ContravariantVector(u::ContravariantTensor, ::LocalGeometry) = u
+@inline ContravariantVector(
+    u::Tensor{1, <:Any, <:Tuple{Basis{Orthonormal, names}}}, lg::LocalGeometry,
+) where {names} = lg.∂ξ∂x * u
+@inline ContravariantVector(
+    u::Tensor{1, <:Any, <:Tuple{Basis{Covariant, names}}}, lg::LocalGeometry,
+) where {names} = lg.gⁱʲ * u
```
dennisYatunin:
```suggestion
ContravariantVector(u::AbstractTensor{1}, lg::LocalGeometry) = _to_contravariant(u, lg)
```
Aren't these methods equivalent to `_to_contravariant`/`_to_covariant`/`_to_local`?

src/Geometry/Geometry.jl
```diff
@@ -1,9 +1,11 @@
...
 
 using ..RecursiveApply
-import LinearAlgebra
-import UnrolledUtilities: unrolled_findfirst
-
+import LinearAlgebra: det, dot, norm, norm_sqr, cross, UniformScaling, Adjoint
+import Random
+import UnrolledUtilities: unrolled_findfirst, unrolled_map, unrolled_foreach,
+    unrolled_unique, unrolled_allunique, unrolled_in, unrolled_filter, unrolled_product,
+    unrolled_any
```
dennisYatunin: Might as well just do `using UnrolledUtilities` if you're importing this many functions

src/Geometry/localgeometry.jl
```diff
@@ -8,48 +8,68 @@ isapproxsymmetric(A::AbstractMatrix{T}; rtol = 10 * eps(T)) where {T} =
...
     ∂ξ∂x = inv(∂x∂ξ)
-    C = typeof(coordinates)
-    Jinv = inv(J)
-    gⁱʲ = ∂ξ∂x * ∂ξ∂x'
-    gᵢⱼ = ∂x∂ξ' * ∂x∂ξ
-    isapproxsymmetric(components(gⁱʲ)) || error("gⁱʲ is not symmetric.")
-    isapproxsymmetric(components(gᵢⱼ)) || error("gᵢⱼ is not symmetric.")
-    return FullLocalGeometry{I, C, FT, S}(coordinates, J, WJ, Jinv, ∂x∂ξ, ∂ξ∂x, gⁱʲ, gᵢⱼ)
+    _gᵢⱼ = ∂x∂ξ' * ∂x∂ξ
+    _gⁱʲ = ∂ξ∂x * ∂ξ∂x'
```
dennisYatunin:
```suggestion
    gᵢⱼ = ∂x∂ξ' * ∂x∂ξ
    gⁱʲ = ∂ξ∂x * ∂ξ∂x'
```
Underscores aren't necessary here, and same goes for the function below

src/Geometry/localgeometry.jl
```diff
@@ -81,163 +101,23 @@ undertype(::Type{SurfaceGeometry{FT, N}}) where {FT, N} = FT
...
-            A[1, 1] zero(FT)
-            C[1, 1] B[1, 1]
-        ]
+function blockmat(a::Tensor{2}, b::Tensor{2}, ::Nothing = nothing)
+    new_bases = (
+        combine_bases(axes(a, 1), axes(b, 1)),
+        combine_bases(axes(a, 2), axes(b, 2)),
     )
+    return reshape(a, new_bases) + reshape(b, new_bases)
 end
```
dennisYatunin: Isn't this equivalent to the automatic reshaping performed by `+`? I feel like you can just replace `blockmat` with `+` without changing any results, but maybe I'm missing something.

src/Geometry/rmul_with_projection.jl
```diff
@@ -1,21 +1,24 @@
 import LinearAlgebra: Adjoint, AdjointAbsVec
 import .RecursiveApply: rmap, rmaptype
-# import LinearAlgebra: I, UniformScaling, Adjoint, AdjointAbsVec
+
 # Types that are treated as single values when using matrix fields.
-const SingleValue = Union{Number, AxisTensor, AdjointAxisTensor}
+# AbstractCovector (Tensor{2} with ScalarBasis) is already covered by AbstractTensor.
+# Adjoint{T, <:AbstractTensor} covers the case where adjoint() returns a Julia Adjoint
+# wrapper rather than our Covector type (e.g., from composition or old codepaths).
```
dennisYatunin: I'd strongly recommend dropping support for `Adjoint{<:Any, <:AbstractTensor}`, if that's possible. The new interface is designed to ensure that all tensors are subtypes of `AbstractTensor`, so that we don't need to define duplicate code for adjoints. This simplification would let you eliminate a lot of the code that was added here and in the `MatrixFields` module.

src/Grids/spectralelement.jl
```diff
@@ -298,7 +296,11 @@ function _SpectralElementGrid2D(
     high_order_quadrature_style = Quadratures.GLL{Nq * 2}()
     high_order_Nq = Quadratures.degrees_of_freedom(high_order_quadrature_style)
 
-    LG = Geometry.FullLocalGeometry{AIdx, CoordType2D, FT, SMatrix{2, 2, FT, 4}}
+    _∂x∂ξ_bases2D = (
+        Geometry.Basis{Geometry.Orthonormal, AIdx}(),
+        Geometry.Basis{Geometry.Covariant, AIdx}(),
+    )
```
dennisYatunin:
```suggestion
```
I don't see this used anywhere

src/Operators/finitedifference.jl
```diff
@@ -4236,22 +4236,22 @@ promote_bc(bc::SetCurl{<:Integer}, ::Type{FT}) where {FT} = SetCurl(FT(bc.val))
 sconvert(::Type{T}, x::SArray{S}) where {T, S} = SArray{S, T}(x...)
 
 function promote_axis_tensor(
-    at::Geometry.AxisTensor{T, N, A, S},
+    at::Geometry.Tensor{N, T, B, S},
```
dennisYatunin:
```suggestion
    at::Geometry.Tensor,
```
Unused parameters should be avoided in type constraints (it's not great for readability, and I think it also sightly increases latency)

src/Spaces/pointspace.jl
```diff
@@ -48,9 +48,12 @@ function PointSpace(
         coord,
         FT(1.0),
         FT(1.0),
-        Geometry.AxisTensor(
-            (Geometry.LocalAxis{AIdx}(), Geometry.CovariantAxis{AIdx}()),
-            FT(1.0),
+        Geometry.Tensor(
+            SMatrix{1, 1}(FT(1.0)),
```
dennisYatunin:
```suggestion
            I,
```
If you can avoid converting `I` to an `SMatrix` when constructing a `Tensor`, using `I` here might eliminate some unnecessary floating-point multiplications.

dennisYatunin (commented):
This is looking pretty great! My main comments are that we should avoid allowing tensor adjoints to be specified in two different ways, and that a lot of the code remaining in `conversions.jl` can be simplified or eliminated. Also, it would be helpful to add detailed docstrings for `Basis`, `Metric`, `Tensor`, and `TensorWithAnyBasis`, with some examples that indicate how `reshape` works and when it gets called.

examples/hybrid/sphere/baroclinic_wave_rhoe.jl
nefrathenrici: oops, this was commented out for debugging

src/Geometry/tensors.jl
nefrathenrici: The conversion is necessary, we'd need to remove the `size(components) == length.(bases)` check from the `Tensor` constructor. I think keeping the UniformScaling in the tensor components wouldn't actually improve things since every op routes through `reshape_and_apply_f` / `reshape(x, bases)`, which materializes to `SArray` anyway

src/Geometry/tensors.jl
nefrathenrici: A Tensor doesn't carry a Metric, so we can't convert between bases here.

src/Geometry/tensors.jl
nefrathenrici: I can add a single 3D cross definition that returns a UVW, I think I just did this to match the previous implementation.

src/Geometry/tensors.jl
dennisYatunin: Ah, right, I forgot `reshape` was hardcoded to return `SArray`s. At some point in the future I'll try generalizing the `size` check and the `reshape` implementation to handle non-`SArray` components, but for now we can just leave it as is. There are only 2 or 3 places in ClimaAtmos where we could currently use this optimization, so the performance impact for AMIP would be relatively small anyway.

src/Geometry/tensors.jl
dennisYatunin: But a `TensorWithAnyBasis` would be able to do arbitrary conversions, and a regular `Tensor` would be able to add/drop basis components without changing the `BasisType`. I think it's fine to just call `reshape` here, and let it throw an error when it needs a `Metric` that wasn't provided.

nefrathenrici: Ok, I'll remove the check.

docs/src/matrix_fields.md
```diff
@@ -108,7 +108,7 @@ scalar_field_matrix
 A FieldMatrix entry can be:
 
 - A `UniformScaling`, which contains a `Number`
-- A `DiagonalMatrixRow`, which can contain either a `Number` or a tensor (represented as a `Geometry.Axis2Tensor`)
+- A `DiagonalMatrixRow`, which can contain either a `Number` or a rank-2 tensor (represented as a `Geometry.Tensor{2}`)
```
dennisYatunin: Should this say `AbstractTensor{2}` instead of `Tensor{2}`? Same for the instance of `Tensor{2}` below.

ext/cuda/column_matrix_helpers.jl
```diff
@@ -427,4 +427,4 @@ Base.@propagate_inbounds outer_or_mul(x::T1, y::T2) where {T1 <: AbstractVector,
 Base.@propagate_inbounds outer_or_mul(
     x::T1,
     y::T2,
-) where {T1 <: Geometry.AdjointAxisVector, T2 <: Geometry.Axis2Tensor} = (x * y)'
+) where {T1 <: Geometry.AbstractCovector, T2 <: Geometry.Tensor{2}} = (x * y)'
```
dennisYatunin:
```suggestion
) where {T1 <: Geometry.AbstractCovector, T2 <: Geometry.AbstractTensor{2}} = (x * y)'
```

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
-        local_geometry,
-    )
-    u¹, u³ = components(vector3)
-    return Contravariant123Vector(u¹, v, u³)
-end
+## Basis-type conversion helpers (private)
+
+# Metrics are identity-padded to full (1,2,3) shape on `LocalGeometry`, so
+# every conversion is a single matvec — names outside `lg`'s geometry `I`
+# ride the identity block of the padded matrix automatically. Same-type
```
dennisYatunin, line 5: This is a bit unclear. The `LocalGeometry` tensors now cover all available dimension names, and there's no additional padding here.

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
+    _to_basis_type(Contravariant(), u, lg)
+@inline CovariantVector(u::AbstractTensor{1}, lg::LocalGeometry) =
+    _to_basis_type(Covariant(), u, lg)
+@inline LocalVector(u::AbstractTensor{1}, lg::LocalGeometry) =
+    _to_basis_type(Orthonormal(), u, lg)
+
+## Scalar constructor for 1D vectors (e.g. WVector(1.0, lg))
+
+# 1D vector types can be constructed from a scalar + LocalGeometry.
+# The LocalGeometry is ignored — the scalar is wrapped directly.
```
dennisYatunin: Does this need to be supported? It looks redundant.

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
-    v::AxisTensor{FT,2,Tuple{LocalAxis{(1,2)},LocalAxis{(1,2)}},SMatrix{2,2,FT,4}},
-    lg::FullLocalGeometry{(1,2,3),XYZPoint{FT},FT,SMatrix{3,3,FT,9}}
-) where {FT}
-    AxisTensor(
-        (ContravariantAxis{(2,)}(), LocalAxis{(1, 2)}()),
-        @inbounds @SMatrix [
-            lg.∂ξ∂x[2, 1]*v[1, 1]+lg.∂ξ∂x[2, 2]*v[2, 1] lg.∂ξ∂x[2, 1]*v[1, 2]+lg.∂ξ∂x[2, 2]*v[2, 2]
-        ]
-    )
+for (BT, VecType, fn) in (
```
dennisYatunin: It looks like `BT` should be a symbol here, and `fn` can be replaced with `VecType`.

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
-    AxisTensor(
-        (ContravariantAxis{(3,)}(), LocalAxis{(1, 2)}()),
-        @inbounds @SMatrix [lg.∂ξ∂x[3, 3]*v[1, 1] lg.∂ξ∂x[3, 3]*v[1, 2]]
-    )
+
+## Scalar component extractors
+
+for (n, cov_sym, con_sym) in ((1, :u₁, :u¹), (2, :u₂, :u²), (3, :u₃, :u³))
+    @eval @inline $(Symbol(:covariant, n))(u::AbstractTensor{1}, lg::LocalGeometry) =
+        CovariantVector(u, lg).$cov_sym
```
dennisYatunin:
```suggestion
for n in 1:3
    @eval @inline $(Symbol(:covariant, n))(u::AbstractTensor{1}, lg::LocalGeometry) =
        project($(Symbol(:Covariant, n, :Axis))(), u, lg)[1]
```
No need for any symbols with mislabeled super/sub-scripts this way

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
+    @eval @inline $(Symbol(:contravariant, n))(u::AbstractTensor{1}, lg::LocalGeometry) =
+        project($(Symbol(:Contravariant, n, :Axis))(), u, lg)[1]
+    @eval @inline $(Symbol(:contravariant, n))(u::Tensor{2}, lg::LocalGeometry) =
```
dennisYatunin: Should this also be an `AbstractTensor{2}` instead of a `Tensor{2}`?

src/Geometry/conversions.jl
```diff
@@ -1,516 +1,140 @@
...
-#! format: on
 
+@inline Jcontravariant3(u::AbstractTensor, lg::LocalGeometry) =
+    lg.J * contravariant3(u, lg)
 
-"""
-    divergence_result_type(V)
+# required for curl-curl
+@inline covariant3(u::Contravariant3Vector, lg::LocalGeometry{(1, 2)}) =
+    contravariant3(u, lg)
```
dennisYatunin: Is this a lingering example of method-level sparsity enforcement? If so, you can drop it for now and replace it with more generic sparsity handling in the future.

src/Geometry/conversions.jl
```diff
@@ -542,60 +166,33 @@ Curl is only defined for `CovariantVector`` field input types.
 @inline curl_result_type(_, ::Type{<:CovariantVector{FT}}) where {FT} =
     ContravariantNullVector{FT}
```
dennisYatunin: You could express this more succinctly by always padding,
```julia
@inline curl_result_type(_, ::Type{<:CovariantVector{FT}}) where {FT} =
   Contravariant123Vector{FT}
```
or by using an unrolled loop over pairs of input indices,
```julia
@inline function curl_result_type(::Val{I1}, ::Type{<:CovariantVector{FT, I2}}) where {I1, I2, FT}
    index_pairs = unrolled_filter(((n1, n2),) -> n1 != n2, unrolled_product(I1, I2))
    I = unrolled_unique(unrolled_map(((n1, n2),) -> 6 - n1 - n2, index_pairs))
    return ContravariantVector{FT, I}
end
```
The second option is roughly identical to what's currently here, but without any missing cases and with a correction to the `curl_result_type(::Val{(3,)}, ::Type{Covariant3Vector{FT}})` case, which can return a `ContravariantNullVector{FT}`.

src/Geometry/conversions.jl
```diff
@@ -542,60 +166,33 @@ Curl is only defined for `CovariantVector`` field input types.
...
-    local_geometry.J * Covariant123Vector(
+## Norm and cross-product (used in broadcast.jl)
+##
+## These are metric-aware versions of norm and cross that take a LocalGeometry.
+## broadcast.jl routes `norm(field)` and `cross(field1, field2)` here so that
+## the correct geometric magnitude is computed regardless of what basis the
+## vectors are stored in. Unlike LinearAlgebra.norm / LinearAlgebra.cross, these
+## convert to the local Orthonormal frame (or Contravariant for cross) first.
+
+_norm_sqr(x, lg::LocalGeometry) = sum(x -> _norm_sqr(x, lg), x)
```
dennisYatunin, line 147: What is this method for?

src/Geometry/conversions.jl
```diff
@@ -542,60 +166,33 @@ Curl is only defined for `CovariantVector`` field input types.
...
+_norm_sqr(x::Number, ::LocalGeometry) = norm_sqr(x)
+_norm_sqr(x::AbstractArray, ::LocalGeometry) = norm_sqr(x)
+_norm_sqr(uᵢ::AbstractTensor{1}, lg::LocalGeometry) =
+    norm_sqr(parent(LocalVector(uᵢ, lg)))
+_norm_sqr(uᵢ::OrthonormalTensor, ::LocalGeometry) =
+    norm_sqr(parent(uᵢ))
```
dennisYatunin: The parent array is extracted in `norm_sqr(::AbstractTensor)`, so there's no need to do it again here.

src/Geometry/Geometry.jl
```diff
@@ -28,14 +28,8 @@ include("auto_broadcaster_methods.jl")
...
+in a `Tensor`.
 """
-Δz_metric_component(::Type{<:LatLongZPoint}) = 9
-Δz_metric_component(::Type{<:Cartesian3Point}) = 1
-Δz_metric_component(::Type{<:Cartesian13Point}) = 4
-Δz_metric_component(::Type{<:Cartesian123Point}) = 9
-Δz_metric_component(::Type{<:XYZPoint}) = 9
-Δz_metric_component(::Type{<:ZPoint}) = 1
-Δz_metric_component(::Type{<:XZPoint}) = 4
+Δz_metric_component(::Any) = 9
```
dennisYatunin: Does this still need to be defined?

src/Geometry/localgeometry.jl
```diff
@@ -1,55 +1,71 @@
...
-    gᵢⱼ::gᵢⱼT #::Axis2Tensor{FT, Tuple{CovariantAxis{I}, CovariantAxis{I}}, S}
+    "Canonical metric ∂x/∂ξ wrapped in [`Metric`](@ref). Identity-padded to
+    full (UVWAxis, Covariant123Axis) shape so a single matvec covers every
+    conversion regardless of `I`."
+    metric::Metric{Tensor{2, FT, Tuple{UVWAxis, Covariant123Axis}, SMatrix{3, 3, FT, 9}}}
+    "Contravariant metric tensor gⁱʲ. Identity-padded to full
+    (Contravariant123, Contravariant123) shape; cached as a real field so
+    Field-level property access (`Fields.local_geometry_field(space).gⁱʲ`)
+    works through DataLayouts."
+    gⁱʲ::Tensor{2, FT, Tuple{Contravariant123Axis, Contravariant123Axis}, SMatrix{3, 3, FT, 9}}
```
dennisYatunin: These fields could also be assigned generic types `metric::M` and `gⁱʲ::G`, which will make it easier to experiment with dropping or modifying them in the future.

src/Geometry/mul_with_projection.jl
```diff
@@ -42,12 +44,28 @@ needs_projection(::Type{X}, ::Type{Y}) where {X, Y} = false
...
-) where {X <: Union{AdjointAxisVector, Axis2TensorOrAdj}, Y <: AxisTensor} =
-    axes(X)[2] != Geometry.dual(axes(Y)[1])
+) where {X <: Tensor{2}, Y <: AbstractTensor} =
+    Geometry.tensor_bases(X)[2] != Geometry.dual(Geometry.tensor_bases(Y)[1])
+function needs_projection(
+    ::Type{X},
+    ::Type{Y},
+) where {X <: SingleValue, Y <: Union{Tuple, NamedTuple}}
+    X <: Number && return false
+    eltype(Y) === Any && return false
```
dennisYatunin: Is this a remnant from before the `RecursiveApply` refactor? Same for the method below.

src/Geometry/mul_with_projection.jl
```diff
@@ -1,36 +1,38 @@
...
-adjoint_type(::Type{A}) where {A} = Adjoint{eltype(A), A}
-adjoint_type(::Type{A}) where {T, S, A <: Adjoint{T, S}} = S
-
-axis1(::Type{<:Axis2Tensor{<:Any, <:Tuple{A, Any}}}) where {A} = A
-axis1(::Type{<:AdjointAxis2Tensor{<:Any, <:Tuple{Any, A}}}) where {A} = A
-
-axis2(::Type{<:Axis2Tensor{<:Any, <:Tuple{Any, A}}}) where {A} = A
-axis2(::Type{<:AdjointAxis2Tensor{<:Any, <:Tuple{A, Any}}}) where {A} = A
+basis1(::Type{<:Tensor{2, <:Any, <:Tuple{B, Any}}}) where {B} = B
+basis2(::Type{<:Tensor{2, <:Any, <:Tuple{Any, B}}}) where {B} = B
```
dennisYatunin:
```suggestion
basis1(::Type{T}) where {T <: AbstractTensor{2}} = Geometry.tensor_bases(T)[1]
basis1(::Type{T}) where {T <: AbstractTensor{2}} = Geometry.tensor_bases(T)[2]
```
Some of the other `Tensor` types below could also be replaced with `AbstractTensor`s.

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,691 @@
...
+    x ⊗ y
+
+Compute the outer product of `x` and `y`.
+"""
+function outer end
+const ⊗ = outer
+
+@inline outer(x::AbstractVector, y::AbstractVector) = x * y'
+@inline outer(x::AbstractVector, y::Number) = x * y
+@inline outer(x::AbstractVector, y) = nested_broadcast(y -> x ⊗ y, y)
```
dennisYatunin:
```suggestion
@inline outer(x, y) = x * y'
```
Both `*` and `'` already call `nested_broadcast`, and you can call `'` on a `Number` (that's just a no-op).

src/MatrixFields/single_field_solver.jl
```diff
@@ -233,8 +230,10 @@ is not necessarily commutative). So, the following are all valid combinations of
 eltype(x), eltype(A), and eltype(b):
 - Number, Number, and Number
 - SVector{N}, SMatrix{N, N}, and SVector{N}
-- AxisVector with axis A1, Axis2TensorOrAdj with axes (A2, dual(A1)), and
-  AxisVector with axis A2
+- Tensor{1} with Basis B1, Tensor{2} (or its adjoint) with bases (B2, dual(B1)),
+  and Tensor{1} with Basis B2
+- nested type (Tuple or NamedTuple), scalar type (Number, SMatrix, or
+  Tensor{2}/adjoint thereof), nested type (Tuple or NamedTuple)
```
dennisYatunin, line 236: This also looks like a relic from before the `RecursiveApply` refactor.

src/Topologies/dss_transform.jl
```diff
@@ -36,21 +36,30 @@ Base.@propagate_inbounds dss_transform(
...
+@inline dss_transform(
+    arg::Geometry.OrthonormalTensor,
+    local_geometry::Geometry.LocalGeometry,
+    weight,
+) = arg * weight
+@inline dss_transform(
+    arg::Geometry.Covariant3Vector,
+    local_geometry::Geometry.LocalGeometry,
+    weight,
+) = arg * weight
```
dennisYatunin: This case is already covered by the method below.

test/Geometry/method_info.jl
```diff
@@ -1,355 +1,369 @@
...
+    (Tensor{2,FT,Tuple{WAxis,WAxis},SMatrix{1,1,FT,1}},LG_13_XZ{FT}, 58),
+    (Tensor{2,FT,Tuple{Covariant3Axis,UAxis},SMatrix{1,1,FT,1}},LG_13_XZ{FT}, 103),
+    (Tensor{2,FT,Tuple{Covariant3Axis,WAxis},SMatrix{1,1,FT,1}},LG_13_XZ{FT}, 103),
+    (Covariant1Vector{FT},LG_13_XZ{FT}, 103),
+    (Tensor{2,FT,Tuple{WAxis,UVAxis},SMatrix{1,2,FT,2}},LG_123_XYZ{FT}, 62),
+    (Tensor{2,FT,Tuple{WAxis,WAxis},SMatrix{1,1,FT,1}},LG_123_XYZ{FT}, 58),
+    (Tensor{2,FT,Tuple{Covariant3Axis,UVAxis},SMatrix{1,2,FT,2}},LG_123_XYZ{FT}, 107),
+    (Tensor{2,FT,Tuple{Covariant3Axis,WAxis},SMatrix{1,1,FT,1}},LG_123_XYZ{FT}, 103),
     ]
 end
```
dennisYatunin: I feel like testing the padded FLOP counts here doesn't serve much of a purpose. You can leave it in if you want, but we only need tests like this for sparse tensors, which won't be introduced until a future PR.

dennisYatunin (commented):
I think this looks mergeable, save for some minor nitpicks. If you want to check this with ClimaAtmos before merging, feel free to define some aliases like `const AxisTensor = Tensor`, which I used to ensure backward compatibility when refactoring `RecursiveApply`.

docs/src/matrix_fields.md
nefrathenrici: AbstractTensor{2} is the abstract supertype, but MatrixFields only implements the indexing/sizing machinery for Tensor{2}.

src/Geometry/conversions.jl
nefrathenrici: Clarified this line.

src/Geometry/conversions.jl
nefrathenrici: I checked and this is needed in ClimaAtmos but I can simplify this to one method.

src/Geometry/conversions.jl
nefrathenrici: I changed it to always pad, thanks.

src/Geometry/conversions.jl
nefrathenrici: This is the fallback for container types, used in `src/Fields/broadcast.jl`

src/Geometry/localgeometry.jl
nefrathenrici: I think the concrete types are needed here, Datalayouts needs the full element type at compile time. We could move the type information into the LocalGeometry definition but it wouldn't simplify things. Happy to defer to you if this doesn't make sense

src/Geometry/mul_with_projection.jl
nefrathenrici: This is a remnant but I think it's still needed in the eager stencil kernel, since it operates on the direct BandMatrixRow elements instead of the broadcast layer.

test/Geometry/method_info.jl
nefrathenrici: I'll remove them, they don't make sense with the uniform padded storage.

nefrathenrici (approved):

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,716 @@
...
+# A `Components{T <: ComponentsType, names}` labels a tensor axis by the
+# variance of its components. Variance refers to the way components scale under
+# a change of generalized coordinates ξⁱ, whose two reciprocal bases they
+# multiply:
+#  - the tangent basis `eᵢ = ∂r/∂ξⁱ`, tangent to the coordinate lines: spans V,
+#    transforms linearly with the coordinates.
+#  - the dual basis `eⁱ = ∇ξⁱ`, with `eⁱ · eⱼ = δⁱⱼ`: spans V*, transforms
+#    inversely.
+
+# A tensor is named after the `ComponentsType` of its first axis:
```
imreddyTeja, line 16: Could we get rid of the names `CovariantTensor` and `ContravariantTensor`? From my understanding, a `CovariantTensor` should have all of its axis/indices be covariant, and a `ContravariantTensor` should have all of its axis/indices be contravariant. If the axis are mixed, it would be a mixed tensor. For example, see page 8 [here](https://bohr.physics.berkeley.edu/classes/h190/s13/lectures/Lecture_on_tensor_analysis.pdf)

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,716 @@
...
+julia> v = Covariant12Vector(1.0, 2.0);
+
+julia> reshape(v, (Covariant123Axis(),))         # zero-fill the missing u₃
+Tensor([1.0, 2.0, 0.0], (Covariant123Components(),))
+
+julia> reshape(v, (Covariant2Axis(),))           # drop u₁, keep u₂
+Tensor([2.0], (Covariant2Components(),))
+```
+"""
+struct Components{T <: ComponentsType, names} <: AbstractComponents end
```
imreddyTeja, line 115: I'm confused about when `names` would have a length > 1

src/Geometry/tensors.jl
```diff
@@ -0,0 +1,716 @@
...
+"""
+    outer(x, y)
+    x ⊗ y
+
+Compute the outer product of `x` and `y`.
+"""
+function outer end
+const ⊗ = outer
+
+@inline outer(x, y) = x * y'
```
imreddyTeja, line 685: should this method have type restrictions?

src/Geometry/localgeometry.jl
```diff
@@ -1,55 +1,78 @@
...
-    gⁱʲ::gⁱʲT #::Axis2Tensor{FT, Tuple{ContravariantAxis{I}, ContravariantAxis{I}}, S}
-    "Covariant metric tensor (gᵢⱼ), transforms contravariant to covariant vector components"
-    gᵢⱼ::gᵢⱼT #::Axis2Tensor{FT, Tuple{CovariantAxis{I}, CovariantAxis{I}}, S}
+    "Canonical metric ∂x/∂ξ. Identity-padded to full
+    (UVWAxis, Covariant123Axis) shape so a single matvec covers every
+    conversion regardless of `I`."
+    ∂x∂ξ::M
+    "Contravariant metric tensor gⁱʲ. Identity-padded to full
+    (Contravariant123, Contravariant123) shape."
+    gⁱʲ::G
```
imreddyTeja, line 24: What is the reasoning behind including this in the struct? Couldn't this also be calculated on the fly with ∂ξ∂x * ∂ξ∂x'? Or is it too expensive

src/Geometry/localgeometry.jl
```diff
@@ -1,55 +1,78 @@
...
+isapproxsymmetric(A::AbstractMatrix{T}; rtol = 10 * eps(T)) where {T <: AbstractFloat} =
     Base.isapprox(A, A'; rtol)
 
 """
     LocalGeometry
 
 The necessary local metric information defined at each node.
 """
-struct LocalGeometry{I, C <: AbstractPoint, FT, ∂x∂ξT, ∂ξ∂xT, gⁱʲT, gᵢⱼT}
+struct LocalGeometry{I, C <: AbstractPoint, FT, M, G}
```
imreddyTeja, line 11: Because the properties of this struct are no longer equivalent to the fields, I think it would be good to mention the properties which are not part of the struct in the docstring.

imreddyTeja (approved):
This PR makes the geometry module much more understandable to me. Thanks! I left some minor comments

imreddyTeja:
> This PR makes the geometry module much more understandable to me. Thanks! I left some minor comments

The comments I left can probably be addressed in a later PR

src/Geometry/tensors.jl
dennisYatunin: It will only have length 1 if there is a single component along that axis; `N` components will correspond to `N` different names.

src/Geometry/tensors.jl
dennisYatunin: This form should be correct for tensors of rank 0, 1, or 2, which is all we support for now.

src/Geometry/localgeometry.jl
dennisYatunin: If I recall correctly, Nat left this because a view of the `gⁱʲ` subfield is extracted from the local geometry field in ClimaAtmos. It would definitely be better to compute `gⁱʲ` from `∂ξ∂x`, but we'll need to change that part of ClimaAtmos first.

PR #2466, opened 2026-03-10
FD Operator CUDAExt

imreddyTeja:
TODO before merge:

- [x] Generic Nv support
- [x] Squash commits
- [x] Delete old copyto_stencil_64
- [x] rename files and kernel
- [x] add shmem support to auto_launch

### DIFF
--- a/ext/cuda/operators_fd_eager.jl
+++ b/ext/cuda/operators_fd_eager.jl
@@ -453,7 +453,9 @@ end
 Projects `mat2_row` onto the correct axis for multiplication with `mat1_row` if necessary, and returns the projected row.
 """
 Base.@propagate_inbounds function project_row2_for_mul(mat1_row, mat2_row, space)
-    if !ClimaCore.Geometry.needs_projection(typeof(mat1_row), typeof(mat2_row))
+    mat1_et = mat1_row isa BandMatrixRow ? eltype(mat1_row) : typeof(mat1_row)
+    mat2_et = mat2_row isa BandMatrixRow ? eltype(mat2_row) : typeof(mat2_row)
+    if !ClimaCore.Geometry.needs_projection(mat1_et, mat2_et)
         return mat2_row
     end
     v = threadIdx().x
@@ -463,7 +465,7 @@ Base.@propagate_inbounds function project_row2_for_mul(mat1_row, mat2_row, space
     h = blockIdx().z
     hidx = (i, j, h)
     project_onto =
-        ClimaCore.Geometry.recursively_find_dual_axes_for_projection(typeof(mat1_row))
+        ClimaCore.Geometry.recursively_find_dual_axes_for_projection(mat1_et)
     if space.staggering isa Spaces.CellCenter && v == CUDA.blockDim().x
         lg = new_struct(Spaces.local_geometry_type(typeof(space)))
     else
--- a/ext/cuda/column_matrix_helpers.jl
+++ b/ext/cuda/column_matrix_helpers.jl
@@ -24,10 +24,10 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd < li || v + pd > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 &&
                    (0i32 < v + mat1_row_d + half <= CUDA.blockDim().x)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d + half + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
@@ -60,10 +60,10 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd < li || v + pd > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 &&
                    (0i32 < v + mat1_row_d - half < CUDA.blockDim().x)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d - half + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
@@ -96,10 +96,10 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd < li || v + pd > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 &&
                    (0i32 < v + mat1_row_d <= CUDA.blockDim().x - 1i32)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
@@ -134,9 +134,9 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd < li || v + pd > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 && (0i32 < v + mat1_row_d <= CUDA.blockDim().x)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
@@ -169,10 +169,10 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd + half < li || v + pd + half > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 &&
                    (0i32 < v + mat1_row_d + half <= CUDA.blockDim().x)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d + half + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
@@ -205,10 +205,10 @@ Base.@propagate_inbounds function row_mul_mat!(
         if v + pd + half < li || v + pd + half > ri
             zero_entry
         else
-            UnrolledUtilities.unrolled_sum((ld1:ud1...,)) do mat1_row_d
+            UnrolledUtilities.unrolled_mapreduce(⊞, (ld1:ud1...,)) do mat1_row_d
                 if ld2 <= pd - mat1_row_d <= ud2 &&
                    (0i32 < v + mat1_row_d - half < CUDA.blockDim().x)
-                    @inbounds mat1_row[mat1_row_d] *
+                    @inbounds mat1_row[mat1_row_d] ⊠
                               matrix2[v + mat1_row_d - half + (i - 1i32) * CUDA.blockDim().x][pd - mat1_row_d]
                 else
                     zero_entry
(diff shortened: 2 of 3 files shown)


test/Operators/finitedifference/convergence_column.jl
```diff
@@ -67,7 +67,7 @@ convergence_rate(err, Δh) =
             cent_field = operator.(face_field)
             wcent_field = woperator.(face_J, face_field)
 
-            Δh[k] = Spaces.local_geometry_data(fs).J[vindex(1)]
+            Δh[k] = to_cpu(Spaces.local_geometry_data(fs).J)[vindex(1)]
```
dennisYatunin:
```suggestion
read_first_level(array) = ClimaComms.allowscalar(() -> array[vindex(1)], device)
            ...
            Δh[k] = read_first_level(Spaces.local_geometry_data(fs).J)
```
`ClimaComms.allowscalar` would be faster than copying the whole array onto the CPU

src/MatrixFields/MatrixFields.jl
```diff
@@ -111,18 +117,25 @@ const FieldOrStencilStyleType = Union{
...
+
     unrolled_reduce(args; init = field_or_broadcasted) do arg1, arg2
         arg1_isa_matrix =
-            eltype(arg1) <: BandMatrixRow || arg1 isa LazyOperatorBroadcasted
+            eltype(arg1) <: BandMatrixRow || (arg1 isa LazyOperatorBroadcasted)
+        if arg1 isa LazyOperatorBroadcasted && length(arg1.args) > 0
+            arg1_isa_matrix =
+                eltype(arg1.args[1]) <: BandMatrixRow ||
+                arg1.args[1] isa LazyOperatorBroadcasted
+        end
```
dennisYatunin: I've seen this sort of conditional variable updating hurt inference inside recursive calls. Better to have something like
```julia
arg1_isa_matrix =
    arg1 isa LazyOperatorBroadcasted && length(arg1.args) > 0 ?
    eltype(arg1.args[1]) <: BandMatrixRow || arg1.args[1] isa LazyOperatorBroadcasted :
    eltype(arg1) <: BandMatrixRow || arg1 isa LazyOperatorBroadcasted
```
But also, this looks like it should be defined recursively? Only going down one level into `arg1.args` feels a bit arbitrary.

src/MatrixFields/MatrixFields.jl
```diff
@@ -161,4 +174,16 @@ function Base.show(io::IO, field::ColumnwiseBandMatrixField)
     end
 end
 
+# TODO: move into CUDAExt
```
dennisYatunin: This could also live in `Geometry/rmul_with_projection.jl` for now

src/MatrixFields/matrix_multiplication.jl
```diff
@@ -278,6 +278,16 @@ function Operators.return_eltype(
...
+    if (
+        matrix1 isa Operators.StencilBroadcasted && matrix1.op isa FDOperatorMatrix &&
+        !(eltype(arg) <: BandMatrixRow)
+    )
+        if matrix1.op.op isa OneArgFDOperator
+            return Operators.return_eltype(matrix1.op.op, arg)
+        else
+            return Operators.return_eltype(matrix1.op.op, matrix1.args[1], arg)
+        end
+    end
```
dennisYatunin, line 291: Does the call to `rmul_return_type` below not generate the same result? I don't see why this new branch is needed.

.buildkite/pipeline.yml
```diff
@@ -651,6 +651,16 @@ steps:
           - "julia --color=yes --check-bounds=yes --project=.buildkite test/Operators/finitedifference/unit_column.jl"
           - "julia --color=yes --check-bounds=yes --project=.buildkite test/Operators/finitedifference/convergence_column.jl"
 
+      - label: "Unit: gpu column"
+        key: unit_gpu_column
+        command:
+          # - "julia --color=yes --check-bounds=yes --project=.buildkite test/Operators/finitedifference/unit_column.jl"
```
dennisYatunin: Is `unit_column.jl` not working because of JET failures, or is it running into some problem with correctness?

ext/cuda/operators_finite_difference.jl
```diff
@@ -81,7 +85,18 @@ function Base.copyto!(
...
+        if !Topologies.isperiodic(space) && mask isa NoMask && n_face_levels == 64
+                new_bc = recursively_replace_fd_ops(bc′)
+                args = (
+                    strip_space(out, space),
+                    strip_space(new_bc, space),
+                    axes(out),
+                )
+                mykr =
+                    CUDA.@cuda always_inline = true threads = (64, 4, 1) blocks =
+                        (1, Nj, Nh) shmem = 64 * 4 * 9 * 4 new_stencil_entry!(args...)
```
dennisYatunin: Is this splitting `Ni` among threads while splitting `Nj` among blocks, but then also parallelizing across both for shmem? It would be good to avoid hardcoding these values if possible.

And you can probably generalize this to arbitrary `Nv`, since we only need to ensure good performance for the flagship config; for other configs, the main priority can just be simplicity and ease of debugging. In the future, we might as well also absorb the `isperiodic` check into `new_stencil_entry!`, but you can leave it separate for now.

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+include("matmul.jl")
+
+"""
+    check_if_fits_in_shmem(x)
+
+Check if `x`, or the `eltype(x)` can fit in shared memory with the current config.
+"""
+check_if_fits_in_shmem(bc::Union{StencilBroadcasted, Broadcasted, Field}) =
+    sizeof(eltype(bc)) <= 36
+check_if_fits_in_shmem(val) = sizeof(typeof(val)) <= 36
```
dennisYatunin, line 37: Would be good to explain where `36` comes from in the docstring

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+
+`OneArgFDOperator`s with affine BCs are also replaced with `MultiplyColumnwiseBandMatrixField`s
+if all the BCs are `SetValue`s
+
+"""
+recursively_replace_fd_ops(val) = val
+
+function recursively_replace_fd_ops(
+    bc::Base.Broadcast.Broadcasted{Style, Axes, F, ARGS},
+) where {Style, Axes, F, ARGS}
```
dennisYatunin:
```suggestion
function replace_fd_ops(bc::Base.Broadcast.Broadcasted)
```
The type signatures of these `recursively_replace_fd_ops` methods can be simplified. Also, all of the functions here are recursive, so you can drop that from the name.

ext/cuda/operators_fd_eager.jl
```diff
@@ -0,0 +1,543 @@
...
+            recursively_replace_fd_ops(bc.args[1]),
+        )
+        new_args = (opmat, recursively_replace_fd_ops(bc.args[end]))
+        newop = MultiplyColumnwiseBandMatrixField()
+        return StencilBroadcasted{Style, typeof(newop), typeof(new_args), Axes, Work}(
+            newop,
+            new_args,
+            bc.axes,
+            bc.work,
+        )
```
dennisYatunin, line 97:
```suggestion
        opmat = FDOperatorMatrix(bc.op)
        new_op = MultiplyColumnwiseBandMatrixField()
        new_args = (Base.Broadcast.broadcasted(opmat, bc.args[1]), bc.args[end])
        new_bc = StencilBroadcasted{Style}(new_op, new_args, bc.axes, bc.work)
        return recursively_replace_fd_ops(new_bc)
```
These recursive calls can also be a bit less repetitive, and the type constructors can use fewer parameters.

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+"""
+Base.@propagate_inbounds function calc_level_val(
+    bc::BC,
+    space,
+) where {BC <: Base.Broadcast.Broadcasted}
+    i = threadIdx().y
+    j = blockIdx().y
+    v = threadIdx().x
+    h = blockIdx().z
+    @inline @inbounds resolved_args = UnrolledUtilities.unrolled_map(
```
dennisYatunin:
```suggestion
    resolved_args = UnrolledUtilities.unrolled_map(
```
Calling a `@propagate_inbounds` function with `@inbounds` is already equivalent to puting `@inbounds @inline` in front of every line, so those additional annotations can be removed

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+
+"""
+    calc_level_val(val::T, space)
+
+If `val` is not a `Broadcasted`, `StencilBroadcasted`, or `Field`, just return `val`.
+If it is a `Ref`, return `val[]`. If it is a one element tuple, return the element.
+"""
+Base.@propagate_inbounds calc_level_val(val::T, space) where {T <: Ref} = val[]
+Base.@propagate_inbounds calc_level_val(val::T, space) where {V, T <: Tuple{V}} =
+    val[Int32(1)]
```
dennisYatunin:
```suggestion
Base.@propagate_inbounds calc_level_val(val::Tuple{<:Any}, space) = first(val)
```
`first` and `last` have slightly lower latencies than `getindex`

ext/cuda/operators_fd_eager.jl
```diff
@@ -0,0 +1,543 @@
...
+        reconstruct_placeholder_space(axes(bc.args[Int32(2)]), space)
+
+    mat2_row = @inline @inbounds calc_level_val(bc.args[Int32(2)], mat2_space)
+    mat1_row = @inline @inbounds calc_level_val(bc.args[Int32(1)], mat1_space)
+    # project before placing in shared memory to avoid projecting multiple times
+    mat2_row_converted =
+        @inline @inbounds project_row2_for_mul(mat1_row, mat2_row, mat2_space)
+    # It should be possible to use static shared memory here, but it allocates new shared memory
+    # for each layer of recursion
+    CUDA.sync_threads()
```
dennisYatunin: What is the purpose of this first synchronization? The second one is to ensure that every level sees the same values in `mat2`, but there are no matrix values being synchronized here.

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+    # project before placing in shared memory to avoid projecting multiple times
+    mat2_row_converted =
+        @inline @inbounds project_row2_for_mul(mat1_row, mat2_row, mat2_space)
+    # It should be possible to use static shared memory here, but it allocates new shared memory
+    # for each layer of recursion
+    CUDA.sync_threads()
+    # it should be possible to use a multi dim shared array here as well, but it seems to
+    # cause some weird issues with the indexing, so I'm just using a 1D array and indexing manually
+    mat2 = CUDA.CuDynamicSharedArray(typeof(mat2_row_converted), Int32(256))
+    @inbounds mat2[v  + (i - 1) * 64] = mat2_row_converted
```
dennisYatunin:
```suggestion
    mat2[v  + (i - Int32(1)) * Int32(64)] = mat2_row_converted
```
Lost some `Int32`s here. Also, to improve readability, you can use `CUDA.i32`, as in `(i - 1i32) * 64i32`

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+    space,
+) where {
+    S,
+    BC <:
+    StencilBroadcasted{S, <:FDOperatorMatrix},
+}
+    op = bc.op.op
+    args = bc.args
+    val = @inline @inbounds get_op_row(op, args, space)
+    CUDA.sync_warp()
```
dennisYatunin: What is the purpose of this synchronization?

ext/cuda/newmm.jl
```diff
@@ -0,0 +1,543 @@
...
+    v = threadIdx().x
+    i = threadIdx().y
+    j = blockIdx().y
+    v = threadIdx().x
+    h = blockIdx().z
+    hidx = (i, j, h)
+    project_onto =
+        ClimaCore.Geometry.recursively_find_dual_axes_for_projection(typeof(mat1_row))
+    if space.staggering isa Spaces.CellCenter && v == Int32(64)
+        lg = rzero(Spaces.local_geometry_type(typeof(space)))
```
dennisYatunin:
```suggestion
        lg = new_struct(Spaces.local_geometry_type(typeof(space)))
```
You can avoid all these calls to `rzero` (which come which a decently large latency penalty) by using something like
```julia
@generated new_struct(::Type{T}) where {T} = Expr(:new, :T)
```

ext/cuda/matmul.jl
```diff
@@ -0,0 +1,451 @@
...
+            init = zero_entry,
+        ) do mat1_row_d
+            if (Int32(0) < v + mat1_row_d + half <= Int32(64))
+                @inbounds outer_or_mul(mat1_row[mat1_row_d], matrix2[v + mat1_row_d + half+ (i - Int32(1)) * Int32(64)])
+            else
+                zero_entry
+            end
+        end
+    end
+end
```
dennisYatunin, line 325:
```suggestion
@inline function row_mul_vec!(::Type{P}, mat_row, vec, shape) where {P}
    v_mat = threadIdx().x
    i = threadIdx().y
    zero_entry = rzero(P)
    ld, ud = MatrixFields.outer_diagonals(typeof(mat_row))
    d_offset = shape == FaceToCenter() ? half : shape == CenterToFace() ? -half : 0
    return UnrolledUtilities.unrolled_mapreduce(⊞, ld:ud; init = zero_entry) do d
        v_vec = v_mat + d + d_offset
        Int32(1) <= v_vec <= Spaces.nlevels(axes(vec)) || return zero_entry
        @inbounds outer_or_mul(mat_row[d], vec[v_vec + (i - Int32(1)) * Int32(64)])
    end
end
```
I think this method covers all 4 cases of matrix-vector multiplication. And it should be straightforward to extend to matrix-matrix multiplication, letting you get rid of all the code duplication in this file.

dennisYatunin (approved):

src/MatrixFields/MatrixFields.jl
imreddyTeja: I think that would make a dependency loop because MatrixFields depends on Geometry already

ext/cuda/operators_fd_eager.jl
imreddyTeja: To ensure that any potential shmem use for the recursion is complete

ext/cuda/newmm.jl
imreddyTeja: I'll remove this. I thought it might help the warp reconverge after diverging in `get_op_row`

.buildkite/pipeline.yml
imreddyTeja: now enabled

src/MatrixFields/matrix_multiplication.jl
imreddyTeja: Not in the case with divgrad of a vec

PR #2482, opened 2026-04-06
Add space-based CUDA kernel fusion for `copyto`

petebachant:
This is a 13% bump in SYPD for the prog EDMF 1M Atmos config (no land).

[Kernel analysis](https://github.com/petebachant/climaatmos-opt/blob/14d2b34ddf6b0f8d288496a002306bf9c46afde2/notebooks/compare-kernels.ipynb):

image


petebachant:
@dennisYatunin this may be interesting to you. When running ClimaAtmos on its own, this change speeds things up significantly, but when run with the coupler with the same exact Atmos config, we hit the fallback condition, i.e., the compiler fails to fuse the kernels.

dennisYatunin:
Wow, yeah, that is very interesting! Running the same function from ClimaCoupler should only add a couple of stack frames on top of running it in ClimaAtmos, but I suppose this shows that your example is right on the edge of triggering a compiler heuristic and de-optimizing.

I think the simplest way to avoid this would be to directly call the ClimaAtmos implicit solver from the AMIP driver, forcing it to be compiled efficiently before it gets called inside ClimaCoupler's deeper stacktrace. That's definitely not a pattern we want to use frequently, and understanding how to avoid these compiler heuristics altogether would be a much more sustainable solution. But at least while we're still figuring things out you can use it, especially if that leads to such a big performance improvement.

PR #2484, opened 2026-04-07
[REPLACED] More efficient solver for TridiagonalMatrixFields

Mikolaj-A-Kowalski:
@petebachant @imreddyTeja @sjavis @AdelekeBankole 

In  this PR we will aim to contribute a more efficient implementation of the tridiagonal matrix solver to that effect we:
- split the `multiple_field_solve!` to isolate tridiagonal cases 
- redirect tridiagonal cases to a specialised solver

At the moment the specialised solver is significantly faster (e.g. when tested on L40 single solve reduces from ~200ms to 90ms [compere.tar.gz](https://github.com/user-attachments/files/26542935/compere.tar.gz)). Once we trigger `[perf]` tag we will see how it performs in the complete simulation :crossed_fingers: 

This is still draft since the solver is not good enough from the point of the view of the launch configuration. We launch a block per column and a thread for each vertical level. This works well if the number of levels is close to multiple of `32` (like AMIP), but it is not stable enough.  In this PR we will try to provide and test some alternatives. At the moment I see two options we can try: 

**Stitch Matrixes together on load to shared memory**   
This would basically use the existing (quite fast) shmem solver, but to balance the load, load multiple columns into shmem to solve them as a single matrix (as far as I know some rows with 0 off-diagonals at the matrix 'stiches' should not cause numerical problems...) . Here the problem would be though that since the 1st (or last) element of the off-diagonal would need to be patched to 0 (since it may not be zero, currently having "don't care" status) on load  to shared memory. 

Since the loading of data currently accounts for at least 50% of the solver runtime extra logic on load may not be performant. We shell see.

**Shared Memory Parallel Thomas** 
Since we expect the vertical tridiagonal matrixes to be small we can solve them in batches of 32. Have a single wrap per block and allow each of the threads to solve full matrix with the Parallel Thomas.    

### TODO

When the PR is ready go through the checklist:
- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/).
- [ ] Unit tests are included.
- [ ] Code is exercised in an integration test.
- [ ] Documentation has been added/updated.

### DIFF
--- a/ext/cuda/matrix_fields_single_field_solve.jl
+++ b/ext/cuda/matrix_fields_single_field_solve.jl
@@ -7,7 +7,7 @@ import ClimaCore.Fields
 import ClimaCore.Spaces
 import ClimaCore.Topologies
 import ClimaCore.MatrixFields
-import ClimaCore.DataLayouts: vindex
+import ClimaCore.DataLayouts: vindex, universal_size
 import ClimaCore.MatrixFields: single_field_solve!
 import ClimaCore.MatrixFields: _single_field_solve!
 import ClimaCore.MatrixFields: band_matrix_solve!, unzip_tuple_field_values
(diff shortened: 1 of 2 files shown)


Mikolaj-A-Kowalski:
Replaced by #2486.

PR #2486, opened 2026-04-09
More efficient solver for TridiagonalMatrixFields

### LINKED ISSUE #2401: Attempt to optimize `run_field_matrix_solver!` for CUDA with a parallel algorithm
This shows up as the top kernel in the "flagship" config (with the exception of 0M instead of 1M microphysics):

Image

Above is from `results/nsys/baseline.nsys-rep`, commit a0bb502 of [this repo](https://github.com/petebachant/clima-gpu-profiling.git).

I've tried a few modifications, which are summarized [here](https://github.com/petebachant/clima-gpu-profiling?tab=readme-ov-file#experiment-results). So far, none have sped up the simulation--only improved registers per thread and occupancy metrics.

Mikolaj-A-Kowalski:
Replaces #2484 (since we need PR from inside the repo to run the CI `perf` jobs. the description is copied verbatim)

Closes #2401

@petebachant @imreddyTeja @sjavis @AdelekeBankole 

In  this PR we will aim to contribute a more efficient implementation of the tridiagonal matrix solver to that effect we:
- split the `multiple_field_solve!` to isolate tridiagonal cases 
- redirect tridiagonal cases to a specialised solver

At the moment the specialised solver is significantly faster (e.g. when tested on L40 single solve reduces from ~200ms to 90ms [compere.tar.gz](https://github.com/user-attachments/files/26542935/compere.tar.gz)). Once we trigger `[perf]` tag we will see how it performs in the complete simulation :crossed_fingers: 

This is still a draft since the solver is not good enough from the point of view of the launch configuration. We launch a block per column and a thread for each vertical level. This works well if the number of levels is close to multiple of `32` (like AMIP), but it is not stable enough.  In this PR we will try to provide and test some alternatives. At the moment I see two options we can try: 

**Stitch Matrices together on load to shared memory**   
This would basically use the existing (quite fast) shmem solver, but to balance the load, load multiple columns into shmem to solve them as a single matrix (as far as I know some rows with 0 off-diagonals at the matrix 'stiches' should not cause numerical problems...). Here the problem would be though that since the 1st (or last) element of the off-diagonal would need to be patched to 0 (since it may not be zero, currently having "don't care" status) on load  to shared memory. 

Since the loading of data currently accounts for at least 50% of the solver runtime extra logic on load may not be performant. We shall see.

**Shared Memory Parallel Thomas** 
Since we expect the vertical tridiagonal matrices to be small we can solve them in batches of 32. Have a single wrap per block and allow each of the threads to solve full matrix with the Parallel Thomas.    

### TODO

When the PR is ready go through the checklist:
- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/).
- [x] Unit tests are included. (new solver is covered by existing unit tests)
- [x] Code is exercised in an integration test.
- [x] Documentation has been added/updated.

### DIFF
--- a/ext/cuda/matrix_fields_single_field_solve.jl
+++ b/ext/cuda/matrix_fields_single_field_solve.jl
@@ -13,7 +13,18 @@ import ClimaCore.MatrixFields: _single_field_solve!
 import ClimaCore.MatrixFields: band_matrix_solve!, unzip_tuple_field_values
 
 function single_field_solve!(device::ClimaComms.CUDADevice, cache, x, A, b)
-    Ni, Nj, _, _, Nh = size(Fields.field_values(A))
+
+    Ni, Nj, _, Nv, Nh = size(Fields.field_values(A))
+
+    # Tridiagonal solvers are handled by special implementation
+    # The special solver is limited in Nv by the number of threads per block
+    # hence it cannot be used for very large matrices.
+    # 512 should run on most GPUs
+    if eltype(A) <: MatrixFields.TridiagonalMatrixRow && Nv <= 512
+        single_field_solve_tridiagonal!(cache, x, A, b)
+        return
+    end
+
     us = UniversalSize(Fields.field_values(A))
     mask = Spaces.get_mask(axes(x))
     cart_inds = cartesian_indices_columnwise(us)
--- a/ext/cuda/matrix_fields_single_field_solve.jl
+++ b/ext/cuda/matrix_fields_single_field_solve.jl
@@ -262,14 +262,14 @@ function tridiag_pcr_kernel!(
 
         # Compute elimination factors
         @inbounds begin
-            k1 = (i > stride) ? -local_ai ⊠ inv(s_b[i_minus]) : zero(eltype(a))
-            k2 = (i <= Nv - stride) ? -local_ci ⊠ inv(s_b[i_plus]) : zero(eltype(a))
+            k1 = (i > stride) ? -local_ai * inv(s_b[i_minus]) : zero(eltype(a))
+            k2 = (i <= Nv - stride) ? -local_ci * inv(s_b[i_plus]) : zero(eltype(a))
 
             # Update coefficients
-            local_ai = k1 ⊠ s_a[i_minus]
-            local_bi = local_bi ⊞ k1 ⊠ s_c[i_minus] ⊞ k2 ⊠ s_a[i_plus]
-            local_ci = k2 ⊠ s_c[i_plus]
-            local_di = local_di ⊞ k1 ⊠ s_d[i_minus] ⊞ k2 ⊠ s_d[i_plus]
+            local_ai = k1 * s_a[i_minus]
+            local_bi = local_bi + k1 * s_c[i_minus] + k2 * s_a[i_plus]
+            local_ci = k2 * s_c[i_plus]
+            local_di = local_di + k1 * s_d[i_minus] + k2 * s_d[i_plus]
         end
 
         CUDA.sync_threads()
@@ -287,7 +287,7 @@ function tridiag_pcr_kernel!(
     end
 
     #  Final solve into x
-    @inbounds x[idx] = inv(s_b[i]) ⊠ s_d[i]
+    @inbounds x[idx] = inv(s_b[i]) * s_d[i]
     return nothing
 end
 
--- a/ext/cuda/matrix_fields_single_field_solve.jl
+++ b/ext/cuda/matrix_fields_single_field_solve.jl
@@ -7,7 +7,7 @@ import ClimaCore.Fields
 import ClimaCore.Spaces
 import ClimaCore.Topologies
 import ClimaCore.MatrixFields
-import ClimaCore.DataLayouts: vindex
+import ClimaCore.DataLayouts: vindex, universal_size
 import ClimaCore.MatrixFields: single_field_solve!
 import ClimaCore.MatrixFields: _single_field_solve!
 import ClimaCore.MatrixFields: band_matrix_solve!, unzip_tuple_field_values
@@ -208,5 +208,118 @@ function band_matrix_solve_local_mem!(
     @inbounds for v in 1:Nv
         x[vindex(v)] = inv(A₀[vindex(v)]) * b[vindex(v)]
     end
+    return nothing
+end
+
+
+function tridiag_pcr_kernel!(
+    x, a, b, c, d, ::Val{Nv}, ::Val{n_iter},
+) where {Nv, n_iter}
+    (idx_i, idx_j, idx_h) = blockIdx()
+    i = threadIdx().x
+    if i > Nv
+        return nothing
+    end
+
+    s_a = CUDA.CuStaticSharedArray(eltype(a), Nv)
+    s_b = CUDA.CuStaticSharedArray(eltype(b), Nv)
+    s_c = CUDA.CuStaticSharedArray(eltype(c), Nv)
+    s_d = CUDA.CuStaticSharedArray(eltype(d), Nv)
+
+    idx = CartesianIndex(idx_i, idx_j, 1, i, idx_h)
+
+    # Load into shared memory
+    @inbounds begin
+        local_ai = a[idx]
+        local_bi = b[idx]
+        local_ci = c[idx]
+        local_di = d[idx]
+
+        s_a[i] = local_ai
+        s_b[i] = local_bi
+        s_c[i] = local_ci
+        s_d[i] = local_di
+    end
+    CUDA.sync_threads()
+
+    # PCR iterations
+    stride = 1
+
+    for _ in 1:n_iter
+        i_minus = max(i - stride, 1)
+        i_plus = min(i + stride, Nv)
+
+        # Compute elimination factors
+        @inbounds begin
+            k1 = (i > stride) ? -local_ai ⊠ inv(s_b[i_minus]) : zero(eltype(a))
+            k2 = (i <= Nv - stride) ? -local_ci ⊠ inv(s_b[i_plus]) : zero(eltype(a))
+
+            # Update coefficients
+            local_ai = k1 ⊠ s_a[i_minus]
+            local_bi = local_bi ⊞ k1 ⊠ s_c[i_minus] ⊞ k2 ⊠ s_a[i_plus]
+            local_ci = k2 ⊠ s_c[i_plus]
+            local_di = local_di ⊞ k1 ⊠ s_d[i_minus] ⊞ k2 ⊠ s_d[i_plus]
+        end
+
+        CUDA.sync_threads()
+
+        # Copy back for next iteration
+        @inbounds begin
+            s_a[i] = local_ai
+            s_b[i] = local_bi
+            s_c[i] = local_ci
+            s_d[i] = local_di
+        end
+
+        CUDA.sync_threads()
+        stride *= 2
+    end
+
+    #  Final solve into x
+    @inbounds x[idx] = inv(s_b[i]) ⊠ s_d[i]
+    return nothing
+end
+
+
+"""
+    single_field_solve_tridiagonal!(cache, x, A, b)
+
+Specialized solver for the tridiagonal MatrixField. Solves each column in
+parallel launching Nv threads per block where Nv is the number of vertical levels.
+Works best if Nv is multiple of 32. There is an upper limit on the size of Nv
+due to resource limits of the GPU (register and shared memory usage). For
+A100 it is 1024, but may differ depending on the hardware.
+"""
+function single_field_solve_tridiagonal!(cache, x, A, b)
+
+    device = ClimaComms.device(x)
+    device isa ClimaComms.CUDADevice || error("This solver supports only CUDA devices.")
+
+    eltype(A) <: MatrixFields.TridiagonalMatrixRow || error(
+        "This function expects a tridiagonal matrix field, but got a field with element type $(eltype(A))",
+    )
+
+    # Get field dimensions
+    Ni, Nj, _, Nv, Nh = universal_size(Fields.field_values(A))
+
+    # Prepare data
+    Aⱼs = unzip_tuple_field_values(Fields.field_values(A.entries))
+    A₋₁, A₀, A₊₁ = Aⱼs
+    x_data = Fields.field_values(x)
+    b_data = Fields.field_values(b)
+
+    # Solve
+    threads_per_block = Nv
+    n_iter = ceil(Int, log2(Nv))
+    args = (x_data, A₋₁, A₀, A₊₁, b_data, Val(Nv), Val(n_iter))
+
+    auto_launch!(
+        tridiag_pcr_kernel!,
+        args;
+        threads_s = (threads_per_block,),
+        blocks_s = (Ni, Nj, Nh),
+    )
+
+    call_post_op_callback() && post_op_callback(x, device, cache, x, A, b)
     return nothing
 end
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@ ClimaCore.jl Release Notes
 main
 -------
 
+- Add a specialized shared memory-based tridiagonal solver that uses PCR algorithm for CUDA backend. [2486](https://github.com/CliMA/ClimaCore.jl/pull/2486) 
+
 v0.14.51
 -------
 
--- a/src/MatrixFields/field_matrix_solver.jl
+++ b/src/MatrixFields/field_matrix_solver.jl
@@ -279,9 +279,15 @@ function run_field_matrix_solver!(
     case1 = length(names) == 1
     case2 = all(name -> cheap_inv(A[name, name]), names.values)
     case3 = any(name -> cheap_inv(A[name, name]), names.values)
+
+    # Direct all TridiagonalMatrixRow cases to `single_field_solve` path so
+    # they can be redirected to a specialised solver
+    # TODO: Group multiple Tridiagonals to the special 'multiple_field' solver
+    case4 = any(name -> eltype(A[name, name]) <: TridiagonalMatrixRow, names.values)
+
     # TODO: remove case3 and implement _single_field_solve_diag_matrix_row!
     #       in multiple_field_solve!
-    if case1 || case2 || case3
+    if case1 || case2 || case3 || case4
         foreach(names) do name
             single_field_solve!(cache[name], x[name], A[name, name], b[name])
         end
--- a/src/MatrixFields/field_matrix_solver.jl
+++ b/src/MatrixFields/field_matrix_solver.jl
@@ -216,7 +216,7 @@ we might want to parallelize it in the future).
 If `Aₙₙ` is a diagonal matrix, the equation `Aₙₙ * xₙ = bₙ` is solved by making a
 single pass over the data, setting each `xₙ[i]` to `inv(Aₙₙ[i, i]) * bₙ[i]`.
 
-Otherwise, the equation `Aₙₙ * xₙ = bₙ` is solved using Gaussian elimination
+Otherwise, on a CPU, the equation `Aₙₙ * xₙ = bₙ` is solved using Gaussian elimination
 (without pivoting), which makes two passes over the data. This is currently only
 implemented for tri-diagonal and penta-diagonal matrices `Aₙₙ`. In Gaussian
 elimination, `Aₙₙ` is effectively factorized into the product `Lₙ * Dₙ * Uₙ`,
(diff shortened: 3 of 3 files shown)


file
petebachant, line 1: Committed by mistake?

Mikolaj-A-Kowalski: Sorry. Committed on purpose, but in a commit designed to be dropped before merging. Just wanted to make one to trigger the `[perf]` pipeline.

petebachant:
Looks like the e2e performance pipeline is failing:

```
ERROR: LoadError: InvalidIRError: compiling MethodInstance for ClimaCoreCUDAExt.tridiag_pcr_kernel!(::ClimaCore.DataLayouts.VIJFH{…}, ::ClimaCore.DataLayouts.VIJFH{…}, ::ClimaCore.DataLayouts.VIJFH{…}, ::ClimaCore.DataLayouts.VIJFH{…}, ::ClimaCore.DataLayouts.VIJFH{…}, ::Val{…}, ::Val{…}) resulted in invalid LLVM IR
```

Do you have materials committed to your dev repo with all the scripts, configs, etc., you used to test this during development to generate that attached report? Maybe this only happens when running with the Coupler? I've seen strange idiosyncracies where kernels sped up for Atmos don't improve when run with the Coupler.

For this PR we should definitely have a unit test for the new solver, ideally using the same case(s) as the existing solver to check numerical values are similar enough.

Mikolaj-A-Kowalski:
> Do you have materials committed to your dev repo with all the scripts, configs, etc., you used to test this during development to generate that attached report? Maybe this only happens when running with the Coupler? I've seen strange idiosyncracies where kernels sped up for Atmos don't improve when run with the Coupler.

Sadly it happens only during the AMIP run. I was able to reproduce that locally on our side, but still straggling through compilation times and julia crashes to get an `Infiltrator` into the function and see what types it is given to make it fail. I suspect it is just something simple that a MatrixField with some odd `eltype` is getting passed to the solver.

We have been iterating the solver on a [testbed](https://github.com/Cambridge-ICCS/clima-gpu-profiling/blob/2f6591e2cc1dad2a0602db193da0be4b2e9167b7/TridiagTestbed/benchmark_tridiag_solve.jl).

file
petebachant: No worries. For that you can use `git commit --allow-empty -m "Trigger [perf] pipeline"` to push up an empty commit.

Mikolaj-A-Kowalski: Gotcha ;-) Have to admit I missed the `--allow-empty` option.

Mikolaj-A-Kowalski:
It seem that a crash is caused by a Field with elements of `ClimaCore.Geometry.Covariant3Vector` being fed to the solver. I am trying to determine if it is a valid use case (probably not...?) or some bug in the redirect that we are doing.  

EDIT: So I was not quite 100% correct. The `ClimaCore.Geometry.Covariant3Vector` is provided for the RHS (and the result `x`), not the tridiagonal matrix itself. It is still nice Float32.

imreddyTeja:
> > Do you have materials committed to your dev repo with all the scripts, configs, etc., you used to test this during development to generate that attached report? Maybe this only happens when running with the Coupler? I've seen strange idiosyncracies where kernels sped up for Atmos don't improve when run with the Coupler.
> 
> Sadly it happens only during the AMIP run. I was able to reproduce that locally on our side, but still straggling through compilation times and julia crashes to get an `Infiltrator` into the function and see what types it is given to make it fail. I suspect it is just something simple that a MatrixField with some odd `eltype` is getting passed to the solver.
> 
> We have been iterating the solver on a [testbed](https://github.com/Cambridge-ICCS/clima-gpu-profiling/blob/2f6591e2cc1dad2a0602db193da0be4b2e9167b7/TridiagTestbed/benchmark_tridiag_solve.jl).

If you aren't already, I think it would be helpful to use `test/MatrixFields/field_matrix_solvers.jl`. It doesn't cover every use case, but it should replicate a few common atmos setups.

Mikolaj-A-Kowalski:
> If you aren't already, I think it would be helpful to use test/MatrixFields/field_matrix_solvers.jl. It doesn't cover every use case, but it should replicate a few common atmos setups.

It was a good shout.

The problem seems to be that we did not handle properly the cases where the components of the matrix and vectors are tensors and co/contra-variant vectors. I am not fully certain if I interpret things correctly, but this is why we should have used the 'boxed' operators and `RecursiveApply` functionalities?

Mikolaj-A-Kowalski:
So by boxing the operators seems to solve the problem with failing end-to-end run. 

Basically we missed that not only elements of the field may be the `AxisTensor`s, but also the RHS `b` may be a tuple and the solution needs to be applied element-wise. 

Now the CI seems to run happily and AMIP run is reporting healthy speedup. I do, however, have a couple of worries to resolve. 

First is that there is a lot of tests failing in the CI (although the failures seem to be ignored when determining the status of the overall job). I assume these are tests known to fail?    

Below I have attached the Nsys reports for the before and after this PR. However, in the after I cannot seem to find the matrix solver kernels. It is quite scary. The unit tests do pass and locally I have verified that the solver runs correctly. Also I presume that something would have failed if e.g. due to some bug the solver was not called at all? Hence I guess it is something happening with the kernel renaming scheme?  

[reports.tar.gz](https://github.com/user-attachments/files/26900648/reports.tar.gz)

Mikolaj-A-Kowalski:
Ok. I have verified in REPL that the solver does get run. I have tried to use the `@debug` logging to check what kernel name it should be assigned with and it should be `run_field_matrix_solver__FILE_ClimaCore_jl_src_MatrixFields_field_matrix_solver_jl_LXXX`. 

Looking into the `Kernel summary` in a spreadsheet the kernels are still there.  

The only suspicious thing is that the performance improvement is much, much better than we could have expected from `ncu` results. Feels to good to be true.  We drop from over 100ms to ~1ms. It feels odd.

 

[summary.ods](https://github.com/user-attachments/files/26903757/summary.ods)

ext/cuda/matrix_fields_single_field_solve.jl
```diff
@@ -7,13 +7,20 @@ import ClimaCore.Fields
...
+import ClimaCore.DataLayouts: vindex, universal_size
 import ClimaCore.MatrixFields: single_field_solve!
 import ClimaCore.MatrixFields: _single_field_solve!
 import ClimaCore.MatrixFields: band_matrix_solve!, unzip_tuple_field_values
 import ClimaCore.RecursiveApply: ⊠, ⊞, ⊟, rmap, rzero, rdiv
 
 function single_field_solve!(device::ClimaComms.CUDADevice, cache, x, A, b)
+
+    # Tridiagonal solvers are handled by special implementation
+    if eltype(A) <: MatrixFields.TridiagonalMatrixRow
```
imreddyTeja: Should a similar catch for tridiagonal solvers be added to `generated_single_field_solve!` or `multiple_field_solve!`?

ext/cuda/matrix_fields_single_field_solve.jl
```diff
@@ -211,3 +218,114 @@ function band_matrix_solve_local_mem!(
...
+    Ni, Nj, _, Nv, Nh = universal_size(Fields.field_values(A))
+
+    # Prepare data
+    Aⱼs = unzip_tuple_field_values(Fields.field_values(A.entries))
+    A₋₁, A₀, A₊₁ = Aⱼs
+    x_data = Fields.field_values(x)
+    b_data = Fields.field_values(b)
+
+    # Solve
+    threads_per_block = min(Nv, 256)
```
imreddyTeja: What happens if Nv > 256?

ext/cuda/matrix_fields_single_field_solve.jl
```diff
@@ -211,3 +218,114 @@ function band_matrix_solve_local_mem!(
...
+    n_iter = ceil(Int, log2(Nv))
+    args = (x_data, A₋₁, A₀, A₊₁, b_data, Val(Nv), Val(n_iter))
+
+    auto_launch!(
+        tridiag_pcr_kernel!,
+        args;
+        threads_s = (threads_per_block,),
```
imreddyTeja, line 330: This is probably not a big issue because most long simulations run with 64 vertical faces, but wouldn't this result in a theoretical occupancy of 50% when Nv==32?

ext/cuda/matrix_fields_single_field_solve.jl
```diff
@@ -211,3 +218,114 @@ function band_matrix_solve_local_mem!(
     end
     return nothing
 end
+
+
+function tridiag_pcr_kernel!(
+    x, a, b, c, d, ::Val{n}, ::Val{n_iter},
```
imreddyTeja:
```suggestion
    x, a, b, c, d, ::Val{Nv}, ::Val{n_iter},
```
I think it would be follow the pattern of using Nv to refer to the number of vertical levels

ext/cuda/matrix_fields_single_field_solve.jl
imreddyTeja: Ahh I just saw the TODO for this

imreddyTeja (commented):
This looks mostly good to me. I'm having trouble opening the summary.ods file you attached, but I will run some quick benchmarks locally to verify the speedup. It would be good to run the nightly pipeline, or at least a simulatiojn from it, with this before merging.

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: Bad things... 

This is a good catch! This line stayed there by accident. It should be removed.

Ultimately the only problem that may happen is that the vertical size may exceed the maximum number of threads per block that can be supported (for a given register usage) and then the kernel will fail to launch. 

Ultimately if we will want a proper error handling on that (with `CUDA.maxthreads`), but an annoying part is that to call it we need an access to the compiled kernel so it may need to happen inside the `auto_launch!`.

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: Exactly! This is a biggest drawback of the kernel as it is. When the `Nv` is 33 basically we end up with a 50% effective occupancy. This is the worst case scenario.  But I have to admit that it is annoyingly difficult to avoid...

As I mentioned in the description I have tried to improve it, by 'concatenating'  multiple tridiagonal systems into a bigger system (for a test basically loaded all matrixes in a single horizontal element at once (i.j- dimensions)) and solve it instead using the same solver. This would avoid the "trailing masked threads" problem basically by averaging. But in the end it did not work. (You can see the attempt [here](https://github.com/CliMA/ClimaCore.jl/commit/ac9bb24cdde8617ffb8578d3a246425f77ff389f))

Basically what happens is that since more threads participate in the synchronisation they spend more time for their brethren to catch up. In addition there are extra PCR steps due to a larger system. For the worst case scenario of `Nv=33` the concatenated solver was as fast like the one in the PR. In general it was significantly slower ( I looked in a standalone "testbed" though. Not in AMIP) 

What we still need to try is to do Thomas with single matrix per thread but in shared memory. 

However there is also a possibility that switching to the shared memory decreased the fraction of the time spent in the solver to a point where we perhaps don't need to worry about the occupancy. Basically the change from Nv of 32 to Nv 33 would increase the runtime of the solver by a factor of about 2, but if is is small enough fraction of time it will have small overall effect on performance. I still need to test that though (the `nsys` results indicate that tridiagonal solutions are a very small fraction of the time, but I am not sure I count them all due to the kernel renaming, I have made `nsys` and `ncu` runs in AMIP without the renaming to verify but did not process the results yet).

imreddyTeja:
I did some quick local testing with the `moist_dycore_prognostic_edmf_prognostic` matrix solver test ran at a 64 vertical faces and a higher horizontal resolution. The ldiv! time before was 28ms, and it is 11ms after :rocket:

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: To update on the above. I had a look at the nsys run without the kernel renaming and it seems that the GPU time in the tridiagonal solver is still rather large (3.4% percent) [no_rename.tar.gz](https://github.com/user-attachments/files/27125362/no_rename.tar.gz). This implies that something have gone wrong in the kernel renaming in the original report... 

3.4% is a large number, sufficiently large to worry about the loss of the occupancy.

The only thing I am thinking is that it may not be trivial to change it so perhaps it would be better to  address it in a separate, follow-up PR>

Mikolaj-A-Kowalski:
>  It would be good to run the nightly pipeline, or at least a simulatiojn from it, with this before merging. 

Is it something I can do by pushing a commit with a right annotation or clicking something somewhere on buildkite? 

Edit: or opening a new branch in the `ClimaCoupler.jl`?

imreddyTeja:
Downstream test in ClimaAtmos [here](https://buildkite.com/clima/climaatmos-ci/builds/29444)

Mikolaj-A-Kowalski:
I think this PR is now ready for the proper review. There is a couple of issues I would like some guidance on though: 

- **Updating online documentation**: should we mention something about the default use of the PCR solver on CUDA back-end in the online docs ( `docs/src/matrix_fields.md`)? On one hand it may be good idea to bring attention to the sub-optimal performance for `Nv` that are not multiples of 32. On the other it may be too much of the implementation detail. 
- **Integration tests**: the cechklist mentiones verifing that the integration tests are in place. The solver is run as the part of the unit tests in `test/MatrixFields/field_matrix_solvers.jl` and inside AMIP. Is this sufficient or should I add anything?

The commit history needs to be cleaned up. I plan to do it after we finish iterating on the PR (I don't want GitHub to invalidate the conversations as the result of force pushes)

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: Should be addressed in 3aea99fe79811a2140be50a9651adb7283d7b6e6

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: I have removed the line in 3aea99fe79811a2140be50a9651adb7283d7b6e6 

If the size is too large to handle. `CUDA.jl` will emit its standard error that the number of threads is too large. 

If we want I can also add a check inside the `auto_launch!` so `ClimaCore.jl` emits a message that may be made more descriptive.

imreddyTeja:
**Updating online documentation**: I think it would be a good idea to put this in the docs. Maybe [`ClimaCore.MatrixFields.BlockDiagonalSolve`] (https://clima.github.io/ClimaCore.jl/dev/matrix_fields/#ClimaCore.MatrixFields.BlockDiagonalSolve) would be an appropriate place? We should definitely mention that different algorithms are used for solving the tridiagonal case on gpu vs cpu. I don't think we need to put the `Nv` should be a multiple of 32 in the docstring, but it's up to you. At the very least, it would be good to mention it in a comment. I was planning on documenting the general advice for choosing (Nv, Ni, Nj, Nh) values when I add docs for the eager stencil evaluation.  

 **Integration tests** - I think the test coverage for the solver is sufficient.

ext/cuda/matrix_fields_single_field_solve.jl
imreddyTeja: I have a fix for that. I will try to open a PR for it by end of the PST day

ext/cuda/matrix_fields_single_field_solve.jl
imreddyTeja: Should we also check that `Nv <= 256` here, and fallback to the previous implementation otherwise? This would keep support for large `Nv`s

imreddyTeja (commented):
After updating the docs, I think this is ready to go. Could you also add an entry to NEWS.md?

ext/cuda/matrix_fields_single_field_solve.jl
Mikolaj-A-Kowalski: I have added the redirect in d6c2910a4126e56fe54a1dcc08cea52423d2e3e3 

I used the threshold of 512 since at the moment the kernel consumes only 30 registers so it supports large number of threads per block.

Mikolaj-A-Kowalski:
> Updating online documentation: I think it would be a good idea to put this in the docs. Maybe [ClimaCore.MatrixFields.BlockDiagonalSolve] (https://clima.github.io/ClimaCore.jl/dev/matrix_fields/#ClimaCore.MatrixFields.BlockDiagonalSolve) would be an appropriate place? We should definitely mention that different algorithms are used for solving the tridiagonal case on gpu vs cpu. I don't think we need to put the Nv should be a multiple of 32 in the docstring, but it's up to you. At the very least, it would be good to mention it in a comment. I was planning on documenting the general advice for choosing (Nv, Ni, Nj, Nh) values when I add docs for the eager stencil evaluation. 

I have added extra documentation in c7ce231f9350aa2ea744a790cc27ac596b3a4e62

> After updating the docs, I think this is ready to go. Could you also add an entry to NEWS.md? 

Added in c50f7806ae5773c7242ae275f73062ec3718c236 (although causes a merge conflict)

If we are broadly happy with the changes I will proceed to rebase on `main` and cleanup the history.

imreddyTeja:
> > Updating online documentation: I think it would be a good idea to put this in the docs. Maybe [ClimaCore.MatrixFields.BlockDiagonalSolve] (https://clima.github.io/ClimaCore.jl/dev/matrix_fields/#ClimaCore.MatrixFields.BlockDiagonalSolve) would be an appropriate place? We should definitely mention that different algorithms are used for solving the tridiagonal case on gpu vs cpu. I don't think we need to put the Nv should be a multiple of 32 in the docstring, but it's up to you. At the very least, it would be good to mention it in a comment. I was planning on documenting the general advice for choosing (Nv, Ni, Nj, Nh) values when I add docs for the eager stencil evaluation.
> 
> I have added extra documentation in [c7ce231](https://github.com/CliMA/ClimaCore.jl/commit/c7ce231f9350aa2ea744a790cc27ac596b3a4e62)
> 
> > After updating the docs, I think this is ready to go. Could you also add an entry to NEWS.md?
> 
> Added in [c50f780](https://github.com/CliMA/ClimaCore.jl/commit/c50f7806ae5773c7242ae275f73062ec3718c236) (although causes a merge conflict)
> 
> If we are broadly happy with the changes I will proceed to rebase on `main` and cleanup the history.

yes please. This looks great!

Mikolaj-A-Kowalski:
All the conflicts should be resolved and it is ready to go when the test pass ;-)

imreddyTeja (approved):
this should be a nice speed up. thank you!

imreddyTeja:
@dennisYatunin Do you expect there to be any changes needed because #2417 was merged in?

Mikolaj-A-Kowalski:
> @dennisYatunin Do you expect there to be any changes needed because https://github.com/CliMA/ClimaCore.jl/pull/2417 was merged in? 

I have rebased on the current `main` and the major change that was needed to get the tests to pass was to 'unbox' all operators again. It is my understanding that they got deprecated?

I have also triggered another end-to-end performance check to see if there is any difference after #2417

imreddyTeja:
> > @dennisYatunin Do you expect there to be any changes needed because https://github.com/CliMA/ClimaCore.jl/pull/2417 was merged in? 
> 
> 
> 
> I have rebased on the current `main` and the major change that was needed to get the tests to pass was to 'unbox' all operators again. It is my understanding that they got deprecated?
> 
> 
> 
> I have also triggered another end-to-end performance check to see if there is any difference after #2417

I checked in with him briefly yesterday. All the changes in #2417 should be backwards compatible. I believe you are correct that we should 'unbox' the operators, as they are now deprecated.

Mikolaj-A-Kowalski:
Also are the failures in the `KinematicDriver.jl` caused by this PR? 
I had a quick glance but and they seem to complain about additional Host allocations, but it is not obvious to be if the functions that gets called include any MatrixField solvers or whether they run on CPU or GPU?

imreddyTeja:
> Also are the failures in the `KinematicDriver.jl` caused by this PR? I had a quick glance but and they seem to complain about additional Host allocations, but it is not obvious to be if the functions that gets called include any MatrixField solvers or whether they run on CPU or GPU?

Those allocations are caused by #2417. The downstream tests has been failing on main with the same number of allocations since that pr was merged. I believe all the tests that we run through GitHub actions are cpu only.

imreddyTeja:
If you think this is good to go, I think this should be merged

Mikolaj-A-Kowalski:
> If you think this is good to go, I think this should be merged 

I think it is. I assume I should just squash and merge?

imreddyTeja:
> > If you think this is good to go, I think this should be merged 
> 
> 
> 
> I think it is. I assume I should just squash and merge? 

Yes. Either squash and merge or use the rebase option on the web interface. There isn't a preferred way as far as I know.

PR #2488, opened 2026-04-09
Add missing `@propagate_inbounds` and optimized form of `setindex`

dennisYatunin:
## Purpose

This PR fixes two issues I discovered while working on #2417:
- When recursive unrolling was replaced with `unrolled_map` in #2413, several `@propagate_inbounds` annotations where accidentally dropped, since neither `unrolled_map` nor the anonymous functions passed to it were annotated with `@propagate_inbounds`. This has led to unnecessary bounds checking during certain broadcast operations, which can be seen in the CI flame graphs for ClimaCore and ClimaAtmos.
- In #2464, I implemented several functions in terms of `Base.setindex`, including the performance-critical `struct_field_view` and `struct_indices`. It turns out that `Base.setindex` does not always inline the index value, leading to unnecessary index comparisons and bounds checks during runtime. These were also visible in the CI flame graphs, though they were not as significant as the first issue.

## Content

Three new functions are added to a sub-module of `Utilities`: `unrolled_setindex`, `unrolled_insert`, and `unrolled_map_with_inbounds`. These will eventually be generalized and moved to `UnrolledUtilities.jl`, but for now they can be explicitly imported from `ClimaCore.Utilities.Unrolled`. I expect we'll need to define other similar functions in the process of refactoring `DataLayouts`/`Operators`, after which we can update the `UnrolledUtilities` API to include them.

---

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -5,6 +5,33 @@ import UnrolledUtilities: unrolled_map
 include("plushalf.jl")
 include("cache.jl")
 
+module Unrolled # TODO: Move all of these functions into UnrolledUtilities.jl
+
+# Alternative to Base.setindex with guaranteed constant propagation
+@inline unrolled_setindex(x::Tuple, value, ::Val{i}) where {i} =
+    ntuple(n -> n == i ? value : x[n], Val(length(x)))
+
+# Analogue of insert! that follows the same pattern as unrolled_setindex
+@inline unrolled_insert(x::Tuple, value, ::Val{i}) where {i} =
+    ntuple(n -> n == i ? value : x[n < i ? n : n - 1], Val(length(x) + 1))
+
+# Same as UnrolledUtilities.unrolled_map, but annotated with @propagate_inbounds
+@generated unrolled_map_with_inbounds(f, x::NTuple{N, Any}) where {N} = quote
+    Base.@_propagate_inbounds_meta
+    return Base.Cartesian.@ntuple $N n -> f(x[n])
+end
+
+# Remove each function's recursion limit for better type inference on Julia 1.10
+if hasfield(Method, :recursion_relation)
+    for f in (unrolled_setindex, unrolled_insert, unrolled_map_with_inbounds)
+        for m in methods(f)
+            m.recursion_relation = Returns(true)
+        end
+    end
+end
+
+end
+
 """
     cart_ind(n::NTuple, i::Integer)
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,5 +1,5 @@
 # Toplevel interface functions for recurisve broadcast expressions
-import UnrolledUtilities: unrolled_map
+import ..Utilities.Unrolled: unrolled_map_with_inbounds
 
 """
     slab(data::AbstractData, h::Integer)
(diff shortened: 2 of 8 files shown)


src/interface.jl
```diff
@@ -13,13 +13,11 @@ function slab end
 Base.@propagate_inbounds slab(x, inds...) = x
 Base.@propagate_inbounds slab(tup::Tuple, inds...) = slab_args(tup, inds...)
 
-# Recursively call slab() on broadcast arguments in a way that is statically reducible by the optimizer
-# see Base.Broadcast.preprocess_args
-Base.@propagate_inbounds function slab_args(args::Tuple, inds...)
-    unrolled_map(args) do arg
+Base.@propagate_inbounds slab_args(args::Tuple, inds...) =
+    unrolled_map_with_inbounds(args) do arg
+        Base.@_propagate_inbounds_meta
```
imreddyTeja, line 18: Why is `  Base.@_propagate_inbounds_meta` needed here if it's already in `unrolled_map_with_inbounds`?

imreddyTeja (approved):
Do you have evidence that these changes make inbounds propagate as expected? Same question for the inlining of the index

src/Utilities/Utilities.jl
```diff
@@ -5,6 +5,33 @@ import UnrolledUtilities: unrolled_map
...
+@inline unrolled_setindex(x::Tuple, value, ::Val{i}) where {i} =
+    ntuple(n -> n == i ? value : x[n], Val(length(x)))
+
+# Analogue of insert! that follows the same pattern as unrolled_setindex
+@inline unrolled_insert(x::Tuple, value, ::Val{i}) where {i} =
+    ntuple(n -> n == i ? value : x[n < i ? n : n - 1], Val(length(x) + 1))
+
+# Same as UnrolledUtilities.unrolled_map, but annotated with @propagate_inbounds
+@generated unrolled_map_with_inbounds(f, x::NTuple{N, Any}) where {N} = quote
+    Base.@_propagate_inbounds_meta
```
ph-kev, line 20: Why can't we do
```suggestion
    @inline @propagate_inbounds
```
instead?

src/interface.jl
dennisYatunin: Because the anonymous function also needs to `@propagate_inbounds`; otherwise the inbounds propagation will terminate between `unrolled_map_with_inbounds` and the call to `slab`.

src/Utilities/Utilities.jl
dennisYatunin: `@propagate_inbounds` can only be used in front of regular function definitions, not with anonymous functions or generated functions. Also, `@propagate_inbounds` implies `@inline`, so that would be superfluous.

dennisYatunin:
@imreddyTeja Here are screenshots of the "Analysis: Flamegraph Profile" job from ClimaCore's CI that illustrate the difference; the first screenshot is from the latest [main branch](https://buildkite.com/clima/climacore-ci/builds/6625), while the second screenshot is from [this branch](https://buildkite.com/clima/climacore-ci/builds/6640):
Screenshot 2026-04-09 at 7 23 27 PM
Screenshot 2026-04-09 at 7 23 57 PM

PR #2490, opened 2026-04-13
Add gpu latency regression test

imreddyTeja:
Adds a test for kernel latency from the cuda ext. This does not include the time between the end of the CUDA launch api call end the actual start of the kernel.

### DIFF
--- a/test/gpu/latency_benchmarks.jl
+++ b/test/gpu/latency_benchmarks.jl
@@ -1,0 +1,65 @@
+using ClimaCore
+using ClimaCore.CommonSpaces
+import ClimaComms
+using Test
+using CUDA
+using BenchmarkTools
+import LazyBroadcast: lazy
+
+
+# the timings for these benchmark are taken using the central cluster
+@testset "benchmarks time to kernel launch" begin
+    # test to catch regressions and improvement to kernel launch time from ClimaCoreCUDAExt
+    # after the inital compilation
+    ext = Base.get_extension(ClimaCore, :ClimaCoreCUDAExt)
+    @assert !isnothing(ext) # cuda must be loaded to test this extension
+    space = ExtrudedCubedSphereSpace(Float32;
+        z_elem = 10,
+        z_min = 0,
+        z_max = 1,
+        radius = 10,
+        h_elem = 10,
+        n_quad_points = 4,
+        staggering = CellCenter(),
+    )
+
+    scalar_field_1 = fill(1.0f0, space)
+    scalar_field_2 = fill(1.0f0, space)
+    # basic expression
+    # intentionally benchmark without a sync between each trial
+    # CUDA.synchronize()
+    latency = median(@benchmark $scalar_field_1 .= $scalar_field_1 .+ $scalar_field_2).time
+    # update this value if the kernel launch time changes significantly and it is expected
+    baseline_latency = 20500
+    @test latency ≈ baseline_latency atol = 3000
+    percent_change_latency =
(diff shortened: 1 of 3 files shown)


test/gpu/latency_benchmarks.jl
```diff
@@ -0,0 +1,50 @@
...
+        staggering = CellCenter(),
+    )
+
+    scalar_field_1 = fill(1.0f0, space)
+    scalar_field_2 = fill(1.0f0, space)
+    # basic expression
+    # intentionally benchmark without a sync
+    latency = median(@benchmark $scalar_field_1 .= $scalar_field_1 .+ $scalar_field_2).time
+    @test latency ≈ 20500 atol = 3000
+    # update this value if the kernel launch time changes significantly and it is expected
```
petebachant, line 32: I would usually expect this comment above the line where it's defined

test/gpu/latency_benchmarks.jl
```diff
@@ -0,0 +1,50 @@
...
+    @test latency ≈ 22500 atol = 3000
+    # update this value if the kernel launch time changes significantly and it is expected
+
+    # nested lazy broadcast
+    lazy_sum_1 = @. lazy(scalar_field_1 + scalar_field_2)
+    lazy_sum_2 = @. lazy(lazy_sum_1 + lazy_sum_1)
+    lazy_sum_3 = @. lazy(lazy_sum_2 + lazy_sum_2)
+    latency = median(@benchmark $scalar_field_1 .= $lazy_sum_3).time
+    @test latency ≈ 29000 atol = 3000
+    # update this value if the kernel launch time changes significantly and it is expected
```
petebachant, line 59: Same comment here about being above the definition line

test/gpu/latency_benchmarks.jl
```diff
@@ -0,0 +1,50 @@
...
+        n_quad_points = 4,
+        staggering = CellCenter(),
+    )
+
+    scalar_field_1 = fill(1.0f0, space)
+    scalar_field_2 = fill(1.0f0, space)
+    # basic expression
+    # intentionally benchmark without a sync
+    latency = median(@benchmark $scalar_field_1 .= $scalar_field_1 .+ $scalar_field_2).time
+    @test latency ≈ 20500 atol = 3000
```
petebachant: Should we print something here in the logs like the absolute latency, difference from baseline, and percent change, so we'll know if we've made an improvement and by how much, so we can reset the baseline?

petebachant (approved):
Overall LGTM with a few inline comments. Additionally:

1. Does it make sense to sync the GPU right before each benchmark?
2. Should we measure allocations as well?
3. Has this latency ever been identified as a problem, i.e., has there ever been a change merged in that blew it up and dropped some simulation's SYPD significantly? Just curious.

imreddyTeja:
> Overall LGTM with a few inline comments. Additionally:
> 
> 1. Does it make sense to sync the GPU right before each benchmark?
> 2. Should we measure allocations as well?
> 3. Has this latency ever been identified as a problem, i.e., has there ever been a change merged in that blew it up and dropped some simulation's SYPD significantly? Just curious.

I thought I replied to this earlier, but I guess I never submitted the comment.

1. Yes, this is probably a good idea and I added this.
2. I'm not sure. Launching a kernel allocates from CUDA.jl though. 
3. This is a problem for ClimaLand, where the launch latency of most kernels is longer than the kernel themselves, so the latency is not completely hidden. I have another PR that improves latency, but I thought this should be merged first.

PR #2495, opened 2026-04-24
allow CUDA.jl v6

juliasloan25:
Allow the latest CUDA.jl. According to the release notes, this will have ideally no breaking changes.

I want to update the buildkite manifest to use CUDA v6 before I merge this, and right now it's blocked because CTS doesn't have that in its compat - see https://github.com/CliMA/ClimaTimeSteppers.jl/pull/410


juliasloan25:
this change is already in main

PR #2500, opened 2026-04-29
Only call `cuda_synchronize` when it's truly necessary

### LINKED ISSUE #2498: Don't synchronize GPUs in `dss!` if only running on one
Something like:

```julia
function weighted_dss_start!(
    data::DSSTypes2D,
    space::...,
    dss_buffer::DSSBuffer,
)
    Quadratures.requires_dss(quadrature_style(space)) || return nothing
    sizeof(eltype(data)) > 0 || return nothing
    device = ClimaComms.device(topology(space))
    weighted_dss_prepare!(data, space, dss_buffer)
    # Only synchronize if there is actually MPI communication to do
    isempty(dss_buffer.perimeter_elems) || cuda_synchronize(device; blocking = true)
    ClimaComms.start(dss_buffer.graph_context)
    return nothing
end
```

or detect necessity with:

```julia
ClimaComms.nprocs(ClimaComms.context(topology(space))) > 1 && cuda_synchronize(device; blocking = true)
```

petebachant:
Resolves #2498 

This speeds up `nsys` profiling on `clima` by about ~10% but doesn't affect SYPD.


imreddyTeja (approved):
looks good to me, other than the one spelling comment left by copilot. Thanks!

petebachant:
> looks good to me, other than the one spelling comment left by copilot. Thanks!

Can you think of any other reasons these syncs might have been necessary that would not be covered by the tests?

imreddyTeja:
> > looks good to me, other than the one spelling comment left by copilot. Thanks!
> 
> Can you think of any other reasons these syncs might have been necessary that would not be covered by the tests?

I can't, but I am not very familiar with the DSS code.

src/Fields/Fields.jl
```diff
@@ -464,7 +464,14 @@ function Spaces.weighted_dss!(
         )
     end
 
-    cuda_synchronize(device; blocking = true)
+    needs_sync =
+        dss_buffer1 isa Topologies.DSSBuffer &&
+        !isempty(dss_buffer1.perimeter_elems) ||
```
dennisYatunin, line 469: When is the `perimeter_elems` array empty? Isn't there always a boundary between elements when we have a `DSSBuffer`, or do we allocate empty buffers even when they aren't needed? I'm curious which examples showed a speedup when profiling.

petebachant:
Closing because the SYPD gain calculation from profiling is a bit misleading, and this provided no benefit on an actual simulation.

PR #2504, opened 2026-05-01
Use MacOS runners only on PR merges

ph-kev:
This PR is identical to https://github.com/CliMA/ClimaCoupler.jl/pull/1722, https://github.com/CliMA/ClimaAtmos.jl/pull/4287, and https://github.com/CliMA/ClimaLand.jl/pull/1738. There is a concurrency limit of 5 MacOS runners for the entire organization.


imreddyTeja (approved):
thanks!

PR #2521, opened 2026-06-02
Tr/rm bcs

imreddyTeja:
- Removed `SetValue` from `GradientC2F`, `DivergenceC2F`, `CurlC2F`, `UpwindBiasedProductC2F`
  - Removed `SetGradient` from `InterpolateC2F` and `WeightedInterpolateC2F`
  - Removed `AdvectionC2C`, `AdvectionF2F`, `FluxCorrectionF2F`, and `FluxCorrectionC2C` operators

The `step` example is now merged with the advection example.

TODO: Add a compat for ClimaCore to KinematicDrivers.jl


akshaysridhar:
Thanks @imreddyTeja. I looked over the artifacts (and updated docstrings) in the `examples/column` and `hybrid/plane + box` sections the outcomes are reasonable. @dennisYatunin can comment on the `operator_matrices.jl` changes. `BorisBook` FCT operators are not used anywhere in ClimaAtmos.jl. `FluxCorrection<>` ops are also unused in ClimaAtmos.jl - the related examples (density current / rising bubble) could be written with prescribed constant diffusion to simplify them if needed.

test/MatrixFields/operator_matrices.jl
```diff
@@ -247,206 +229,63 @@ end
...
-            )
-        ),
-    )
-
-    # TODO: This case's reference function takes too long to compile on both
-    # CPUs and GPUs (more than half an hour), as of Julia 1.9. This might be
-    # happening because of excessive inlining---aside from *, all other finite
-    # difference operators use @propagate_inbounds. So, the reference function
-    # is currently disabled, although the test does pass when it is enabled.
-    test_field_broadcast(;
```
dennisYatunin, line 399: The tests deleted here were somewhat useful for tracking the limits of Julia's compiler, especially this last one. Any chance you could keep the preexisting tests, and just replace the now-defunct `ᶜadvect_matrix(ᶠuvw)` with some similar matrix?

test/MatrixFields/operator_matrices.jl
```diff
@@ -247,206 +229,63 @@ end
     end
 
     test_field_broadcast(;
-        test_name = "product of six operator matrices",
+        test_name = "product of four operator matrices",
```
dennisYatunin: And these tests were all meant to exercise the recursion limit, which might not be hit after only four multiplication arguments. Could you replace the deprecated operators with similar matrices instead of removing them?

test/Operators/finitedifference/conv_advection_diffusion1d.jl
```diff
@@ -61,21 +61,26 @@ convergence_rate(err, Δh) =
...
-            A = Operators.AdvectionC2C(
-                bottom = bc_vb,
-                top = Operators.Extrapolate(),
-            )
-            gradc2f = Operators.GradientC2F(; bottom = bc_vb, top = bc_gt)
+            top_center_left_biased_grad =
+                Geometry.Covariant3Vector.(
+                    Fields.level(T, Fields.nlevels(T)) .-
+                    Fields.level(T, Fields.nlevels(T) - 1)
+                )
```
dennisYatunin, line 77: Is this equivalent to `∇gaussian` at the second-to-last level? Might be simpler to write it that way if it is.

test/Operators/finitedifference/unit_column.jl
```diff
@@ -351,9 +329,195 @@ end
...
+        @test cpu_parent(new)[2:n] ≈ ref
+    end
+
+    @testset "FluxCorrectionC2C, Extrapolate" begin
+        # A(v,θ)[i] = |v³[i+1/2]| ∂θ⁺ - |v³[i-1/2]| ∂θ⁻, where `Extrapolate`
+        # drops the term outside of the boundary (zero flux through the face)
+        ref = map(1:n) do i
+            ⁺ = i == n ? zero(FT) : abs(w³ᶠ[i + 1]) * (tᶜ[i + 1] - tᶜ[i])
+            ⁻ = i == 1 ? zero(FT) : abs(w³ᶠ[i]) * (tᶜ[i] - tᶜ[i - 1])
+            ⁺ - ⁻
```
dennisYatunin: These variable names are a tad hard to read

test/Operators/hybrid/unit_2d.jl
```diff
@@ -297,7 +288,7 @@ end
     ccoords = Fields.coordinate_field(hv_center_space)
     fcoords = Fields.coordinate_field(hv_face_space)
     fcoords_1 = Fields.level(fcoords, ClimaCore.Utilities.half)
-    curl_bcfield₁ = Geometry.Covariant1Vector.(0.0 .* fcoords_1.z)
+    curl_bcfield₁ = Geometry.Contravariant2Vector(2.0)
```
dennisYatunin: This variable should be renamed, since it now has the same vector type as `curl_bcfield²`.

docs/src/examples.md
```diff
@@ -79,25 +79,29 @@ The example code solves the equation for 4 different tendencies with the followi
   where ``\partial`` is the [`face-to-center divergence`](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.DivergenceF2C) and $UB$ is the [`center-to-face upwind biased product`](https://clima.github.io/ClimaCore.jl/dev/operators/#ClimaCore.Operators.UpwindBiasedProductC2F) operator.
 - Tendency 2:
 
-  $$D = \partial(UB) + \textrm{fcc}(v, \theta),$$  
+  $$D = \partial(UB) - \textrm{fcc}(v, \theta),$$
```
dennisYatunin, line 82: This looks like the opposite of the `deformation_flow.jl` example, which has `UB + FCC` rather than `UB - FCC`. Where did you get this sign flip from?

dennisYatunin (approved):
Great simplification! Few minor comments about changes to tests and examples, but looks great otherwise.

docs/src/examples.md
imreddyTeja: This makes it match the example code. If I flip the sign in the example code, the outputs become garbage, so I'm assuming the sign in the example code is correct. My understanding might be incorrect, but I think most of the advection operators return a flux. The AdvectionC2C/F2F and FluxCorrectionC2C/F2F return the actual tendency.

test/Operators/finitedifference/conv_advection_diffusion1d.jl
imreddyTeja: Yeah it is, but then I would need to project the value from a WVector into a Covariant3Vector. If you think that is a more understandable approach, I can make that change.

Issue #2514, opened 2026-05-22
Enable generic kernel fusion


dennisYatunin:
## Purpose

As of #2466 and #2486, all stencil operations in ClimaCore are parallelized using similar CUDA kernels, with one column assigned to each block of threads, subdivided into one point per thread, and with shared memory used to propagate data between points. This now leaves us with five types of kernels in ClimaCore:
- Pointwise operations—one point per thread, each block handles points that are stored consecutively (block size chosen to maximize occupancy); i.e.,
  - Applying a function to a `DataLayout` or `Field` (`f.(args...)` for a non-`AbstractOperator` `f`)
- Stencil operations—one point per thread, each block handles points in a column, with intermediate values cached in shared memory; i.e.,
  - Applying a predefined stencil (`op.(args...)` for a `FiniteDifferenceOperator` `op`)
  - Applying a generic stencil (`op .* arg` for a `ColumnwiseBandMatrixField` `op`)
  - Inverting a generic stencil (`single_field_solve!`)
  - Inverting multiple generic stencils (`multiple_field_solve!`)
- Spectral element operations—one point per thread, each block handles `Nvt` slabs that are stored consecutively (`Nvt` chosen so block size is roughly 256), with intermediate values cached in shared memory; i.e.,
  - Applying a predefined spectral element operator (`op.(arg)` for a `SpectralElementOperator` `op`)
- Inter-element exchanges—one perimeter point per thread, each block handles perimeter points that are stored consecutively (block size is roughly 256), with intermediate values cached in global memory; i.e.,
  - Applying DSS to a `DataLayout` (`dss!`) or `Field` (`weighted_dss!`), using a `DSSBuffer` as a cache
- Upward aggregations—one column per thread, each block handles columns that are stored consecutively (block size chosen to maximize occupancy); i.e.,
  - Reducing each column to one value (`column_reduce!`)
  - Accumulating values at each level along a column (`column_accumulate!`)

To minimize global memory reads, which currently use up the majority of our compute time, we can fuse sequential pointwise and stencil operations, and we can also fuse sequential pointwise and spectral element operations. If there are no unforeseen roadblocks, we could achieve the following without needing to modify the types of kernels we launch:
1. The implicit solve can be fused into a single kernel, since it only involves pointwise and stencil operations.
    - This would singlehandedly eliminate most (~80%) of the reads that occur in each timestep.
[...]

Issue #2468, opened 2026-03-17
ClimaCore refactoring plan for discussion


tapios:
# ClimaCore Refactoring Plan

## Purpose

ClimaCore.jl has ~35k lines of source and ~33k lines of tests. The three largest modules (Operators (7.5k), MatrixFields (5.2k), and DataLayouts (4.2k)) account for 48% of the source code and contain substantial structural duplication. This plan targets a ~20% source reduction, making the code easier to extend, maintain, and optimize for performance.

The refactoring is organized into 6 phases. Each phase is independently shippable and has a clear "done" criterion. The phases are designed so that later ones benefit from earlier ones but do not require them.

This is meant as a starting point for discussion, not yet as a task list. It's likely I'm missing things, so please edit this as needed. 

---

## Phase 1: DataLayouts — Parameterized Layouts

**Target**: `src/DataLayouts/` (4,231 lines → ~1,800 lines)

### Problem

14 DataLayout types are defined, but they are all permutations of 5 universal index dimensions `(i, j, f, v, h)`. The FH/HF variants (e.g., IJFH vs IJHF, VIFH vs VIHF) are identical except for whether the field dimension `f` precedes or follows the element dimension `h` in memory. This creates systematic duplication:

- 9 `slab()` implementations differing only in which dimension is dropped
- 11 `column()` implementations differing only in which dimensions are extracted
- 5 `level()` implementations differing only in slice position
- `fill.jl`, `copyto.jl`, `mapreduce.jl`, `broadcast.jl` all re-implement the same loops per layout
- `to_data_specific()` and `to_data_specific_field()` encode 22 manual index permutations
- `field_dim()`, `h_dim()`, `type_params()`, `union_all()` each have 12-15 one-liner definitions

The existing code already shows awareness of this: `to_data_specific` encodes permutations as tuple selections, `field_dim`/`h_dim` return integer positions, and `singleton` types provide a dispatch key. The infrastructure for parameterization is half-built. It may be that @dennisYatunin's current work is already addressing this.

### Possible Approach

**Step 1.1 — Encode layout as a permutation tuple.**

Define a trait that maps each layout singleton to its index permutation:

```julia
# Universal index order: (i, j, f, v, h)
# Storage permutation for IJFH: array dims are (i, j, f, h) → universal indices (1, 2, 3, 5)
@inline storage_axes(::IJFHSingleton) = (1, 2, 3, 5)  # i, j, f, h
[...]

PR #2522, opened 2026-06-03
Unify DataLayouts and simplify API

dennisYatunin:
## Purpose

This PR completes the first task outlined in #2468, unifying most of the structs in the `DataLayouts` module and simplifying their API to eliminate code duplication. This is a prerequisite for #2514.

All loops and reductions over layouts now run through two scope-dispatched primitives, `foreach_slice` and `reduce_points`, which work identically on CPUs (including nested multithreading) and GPUs (as single kernels). The rewrite is verified to compile for GPUs without runtime dispatch, and hot paths match the performance of `main` on CPUs.

Tested against the ClimaAtmos CI pipeline in [this build](https://buildkite.com/clima/climaatmos-ci/builds/31003) (only a single MSE test failure due to roundoff error).

## Content

- Rename `AbstractData` to `DataLayout`, and make it a subtype of `AbstractArray`
- Simplify all layouts to four structs: `DataF`, `VIJHWithF`, `VIH1`, and `IH1JH2`
  - Enable consistent treatment of static arrays (`MArray`s or cuda shmem arrays) by `similar(::DataLayout, ...)`
  - Allow any parent array dimension to be treated as an `F` axis
  - Make `getindex` and `setindex!` match standard `AbstractArray` behavior
  - Propagate linear indexing through the `IndexStyle`, instead of separate structs like `NonExtrudedBroadcasted`
- Replace dispatch based on parent arrays with a distinct layer of abstraction
  - Introduce the `DataScope` singleton, a more versatile generalization of a `ClimaComms.AbstractDevice`
  - Propagate this singleton through all operations on `DataLayout`s, including broadcasts and reductions
  - Define rules for combining `DataScope`s, as well as rules for partitioning into smaller `DataScope`s
  - Automatically divide column/slab/level/point views among `DataScope` partitions
- Define a minimal set of communication primitives through `DataScope` dispatch: `foreach_slice` and `reduce_points`
  - Implement the loops in `fill!`, `copyto!`, and `fused_copyto!` through the `foreach_point` primitive (a simple wrapper for `foreach_slice`)
  - Enable column masking in `reduce` and `fused_copyto!`
  - Make `fused_copyto!` treat all layouts consistently (and actually fuse the `copyto!` loops every time it is called)
- Simplify all `BroadcastStyle`s to one concrete struct, still called `DataStyle`
  - Use `DataStyle` to propagate the layout type, ignoring `DataF` layouts when combining distinct types
  - Limit layout-specific broadcasting behavior to four methods of `shape_params`
- Make the broadcasting implementation consistent with the default methods for `AbstractArray`s in `Base`
  - Replicate the optimizations for broadcasting over `Ref`s, and for simple identity function broadcasts
  - Formalize how single-valued `Tuple`s in broadcasts are treated as `Ref`s
  - Have multi-valued `Tuple`s fall back to the default behavior for `AbstractArray`s instead of erroring
- Guarantee that the new primitives compile for GPUs without runtime dispatch
  - Add `Utilities.stable_view`, which constructs inference-stable `SubArray`s for all slice and property views (GPUArrays replaces contiguous `CuArray` views with uninferrable derived arrays), eliminating reshaped views on GPUs altogether
  - Keep 128-bit integers and runtime-built error strings out of kernel argument types and kernel code, which LLVM cannot compile to PTX
- Keep old code loadable during the transition
  - Add `DataLayouts/deprecated.jl` with exported aliases for `AbstractData`, `IJFH`, and `IJHF` (downstream packages and registered satellite packages still reference them)
  - Read legacy layout shapes from existing HDF5 files by reshaping them explicitly

## Roadmap

A suggested order for reviewing the changes:

1. **Core API definitions** — the layout types, struct storage, and scopes:
   `src/DataLayouts/DataLayouts.jl`, `src/DataLayouts/struct_storage.jl`, `src/DataLayouts/scopes.jl`, `src/DataLayouts/deprecated.jl`
2. **Indexing, broadcasting, and loop primitives** — how values are accessed and iterated:
   `src/DataLayouts/indexing.jl`, `src/DataLayouts/broadcast.jl`, `src/DataLayouts/masks.jl`, `src/DataLayouts/loops.jl`, `src/Utilities/safe_mapreduce.jl`, `src/Utilities/Utilities.jl`
3. **GPU implementations of the primitives**:
   `ext/cuda/scopes.jl`, `ext/cuda/loops.jl`, `ext/cuda/cuda_utils.jl`, `ext/cuda/data_layouts.jl`, `ext/cuda/adapt.jl`
4. **Name and indexing updates in consumers** — mostly mechanical (4-D `(v, i, j, h)` indices, 5-D parent arrays, `shape_params`):
   `src/Fields/`, `src/Spaces/`, `src/Grids/`, `src/Operators/`, `src/MatrixFields/`, `src/Limiters/`, `src/Topologies/dss*.jl`, `src/Remapping/`, `src/InputOutput/`, remaining `ext/cuda/` files, `lib/ClimaCoreMakie`, `lib/ClimaCorePlots`, `examples/`
5. **New and updated tests**:
   `test/DataLayouts/unit_loops.jl`, `test/Utilities/unit_stable_view.jl`, updated `test/DataLayouts/`, `test/Fields/`, `test/Spaces/`, `test/Operators/`, and GPU expectation updates (`test/gpu/latency_benchmarks.jl`, `test/Spaces/opt_spaces.jl`, `test/Operators/finitedifference/opt_examples.jl`)

---


src/DataLayouts/loops.jl
```diff
@@ -0,0 +1,145 @@
...
+    end
+end
+
+# Make copyto! handle single-element Tuples the same way as Refs in broadcasts.
+@inline function Base.copyto!(
+    dest::DataLayout,
+    bc::Broadcast.Broadcasted{Broadcast.Style{Tuple}},
+    mask...,
+)
+    isone(length(bc)) || Broadcast.throwdm(axes(dest), axes(bc))
```
Mikolaj-A-Kowalski: I have a stupid question. I have been poking at this PR to understand how the DataLayouts are changing, but I have encountered a bug where an error triggers here. 
```
using ClimaCore


dl = ClimaCore.DataLayouts.VIJFH{Float32, 63, 4, 4, 16}(Array{Float32})
dl .= 1
```
dispatches to here and since `length(bc)` is not one it triggers the domain mismatch(?) error despite the axes matching. 

I was trying to understand what is going on but ultimately I couldn’t figure out the intention behind wrapping of the value in `fill!` inside a tuple: 

The line later in the file
```
@inline Base.fill!(dest::DataLayout, value) = dest .= (value,)
```

If I remove the method above, my example runs fine, but I presume it breaks on DataLayouts with elements other than floating point numbers.

dennisYatunin: Yes, this was implemented incorrectly. I was combining several pieces from Base's `copyto!` implementation and didn't check whether `fill!` ended up working.

The main difficulty here was that Base drops the `BroadcastStyle` before calling `copyto!`, but I needed the information contained in that style. This method has now been updated to avoid dropping the style for single-element tuples, so that they can be passed to a specialized `copyto!` method for `MaybeLazyDataLayout`. The style is still dropped for multi-element tuples (as we cannot treat these like `Ref`s), which now causes `copyto!` to use the generic method from Base. This fallback behavior is what you observed when you removed the `fill!` method in your test—it should work like any `AbstractArray` broadcast, but it doesn't try to parallelize the computation over multiple threads like the `copyto!` method for `MaybeLazyDataLayout`.

I was also initially trying to implement `fill!` through `copyto!`, so that I could minimize the amount of code duplication. After staring at Base's implementation of `copyto!` some more, though, I've decided that the performance benefits of having a separate method for `fill!` outweigh the increase in code size. Besides, now that I have a working version of `foreach_point`, it only takes a few lines to implement `fill!`.

src/DataLayouts/DataLayouts.jl
```diff
@@ -1,2233 +1,549 @@
...
+    VIJHWithF{T, Nv, Ni, Nj, Nh, F, default_scope_type()}(args...)
+function VIJHWithF{T, Nv, Ni, Nj, Nh, F, S}(
+    ::Type{A},
+    Nh_dynamic = Nh,
+) where {T, Nv, Ni, Nj, Nh, F, S, A}
+    check_Nh_dynamic(Nh_dynamic)
+    Nf = num_basetypes(eltype(A), T)
+    array = similar(A, add_f_dim((Nv, Ni, Nj, Nh_dynamic), Nf, Val(F))...)
+    return VIJHWithF{T, Nv, Ni, Nj, Nh, F, S}(array)
 end
```
Mikolaj-A-Kowalski, line 392: I have a bit of a semantic question regarding this part. 

Say I run with `CLIMACOMMS_DEVICE=CUDA` and execute the following code: 
```julia
using ClimaCore
using CUDA

dl = ClimaCore.DataLayouts.VIJFH{Float32, 63, 4, 4, 16}(Array{Float32})
@info  ClimaCore.DataScope(dl)
```
I will get `ClimaCore.DataLayouts.ExternalDeviceScope{ClimaComms.CUDADevice}`. 

I have to say that it is not expected. When running the default  CPUSingleThread we get `ThisThread` scope. 


I have two questions: 
1. Is allowing the construction of DataLayouts without explicitly specifying the scope a good idea? I feel it may not be that easy to infer it automatically. 
2. In the above example, what is the "correct" scope of the DataLayout? Can it be determined?

dennisYatunin: Thanks, this was also a bug! I was originally hoping to avoid "array-based dispatch" altogether, so I chose the default scope based on the default `AbstractDevice` from ClimaComms. But, as your example shows, this is incorrect when the input array has not been allocated on the default `AbstractDevice`.

I've fixed this by replacing `device_scope` with `array_scope`, so that the default scope is now chosen based on the input array type. This is similar to the old `parent_array_type` and `device_dispatch` functions, but with two key differences:
- It supports multi-threaded CPUs through a check for `isone(Threads.nthreads())`.
- It will only be used to autofill the `DataScope` of a new `DataLayout`, not to regulate multiple dispatch across different parts of ClimaCore.

The only reason I'm allowing a `DataLayout` to be constructed without an explicit `DataScope` is to simplify our unit tests. If this weren't possible, every unit test would need to check whether it has access to a multi-threaded CPU or a CUDA device. Since that information can be inferred from the input array type, there is no need to add so much verbosity to the unit tests.

Ideally, this functionality will not get used outside of testing or debugging. Every other function in this redesigned API (`similar`, `view`, `column`, etc.) automatically propagates the `DataScope`s of preexisting `DataLayout`s. Aside from testing and debugging, there shouldn't be many other situations where a `DataLayout` has to be constructed from scratch.

Mikolaj-A-Kowalski:
I am running into another issue (it may be know however) when I try do amassment to  a DataLayout: 
```julia
dl = ClimaCore.DataLayouts.VIJFH{Float32, 63, 4, 4, 16}(Array{Float32})
dl .= Float64(4)
``` 

It seems to be related to  the fact that the RHS does not get converted to the `eltype` of the DataLayout.

dennisYatunin:
> I am running into another issue (it may be know however) when I try do amassment to a DataLayout:
> 
> ```julia
> dl = ClimaCore.DataLayouts.VIJFH{Float32, 63, 4, 4, 16}(Array{Float32})
> dl .= Float64(4)
> ```
> 
> It seems to be related to the fact that the RHS does not get converted to the `eltype` of the DataLayout.

Thanks, I missed the conversion in `Base.setindex!(::DataLayout, ...)`. Definitely need to keep that around, since `set_struct!` automatically bitcasts its inputs and uses `@inbounds`.

Mikolaj-A-Kowalski:
When looking at this PR, one question also jumped to my mind, which is what kind of the scopes we expect to exist when we are done. 

At the moment there seems to be a CPU hierarchy: `ThisThreadPool` -> `ThisThread` 
and GPU hierarchy `ExternalDevice` -> `ThisKernelGrid` -> `ThisKernelBlock` -> `ThisThread` 

Is it more-or-less fixed or is it still work in progress?

dennisYatunin:
> When looking at this PR, one question also jumped to my mind, which is what kind of the scopes we expect to exist when we are done.
> 
> At the moment there seems to be a CPU hierarchy: `ThisThreadPool` -> `ThisThread` and GPU hierarchy `ExternalDevice` -> `ThisKernelGrid` -> `ThisKernelBlock` -> `ThisThread`
> 
> Is it more-or-less fixed or is it still work in progress?

The CPU hierarchy is fixed, and I doubt it will ever need to get more complicated. I was initially hoping to keep the GPU hierarchy simple as well, but I had to add a bit more complexity to match ClimaCore's current behavior.

Aside from some naming simplifications (`ExternalDevice` is now `ThisHost`, `ThisKernelGrid` is now `ThisKernel`, and `ThisKernelBlock` is now `ThisBlock`), I've introduced a new layer between blocks and threads called `ThisSubBlock{N}`. The type parameter `N` denotes the number of threads, which can be any value between 2 and 32 (with 32 corresponding to one warp).

When processing slice views in parallel across `ThisKernel`, the scope assigned to each slice will be determined according to its size:
- For single-point views, `foreach_point` will assign `ThisThread` to each point.
- For columns with 64 levels, `foreach_column` will assign `ThisBlock` to each column.
- For slabs with 16 nodal points, `foreach_slab` will assign `ThisSubBlock{16}` to each slab.

This will match the current implementation of `copyto!` in ClimaCore's CUDA extension:
- For `Broadcasted` expressions, `copyto!` assigns one thread to each point.
- For `StencilBroadcasted` expressions with 64 levels, `copyto!` assigns one block to each column.
- For `SpectralBroadcasted` expressions with 16 nodal points, `copyto!` assigns one half-warp to each slab.
  - In particular, `spectral_partition` assigns one block to `Ni * Nj * Nvthreads` points, where `Nvthreads` is roughly `n_max_threads ÷ (Ni * Nj)`. When `Ni * Nj` is 16, this is equivalent to assigning one half-warp to each slab, and grouping multiple half-warps into blocks of size `n_max_threads`.

To minimize the amount of code we'll need in the CUDA extension, I've handled all of these cases in a single `foreach_slice` primitive, which `foreach_point`/`foreach_level`/`foreach_slab`/`foreach_column` just call with a particular slice operator (`view`/`level`/`slab`/`column`, respectively). The most important part of `foreach_slice` is the `slice_subscope(scope, op, args...)` function, which is defined as follows:
- "By default, this is the smallest subset of `scope` that does not require any thread to process more than one point from the largest slice returned by `op`. When no such subset is available, the largest subset is used in order to minimize the number of points per thread."

If the default `slice_subscope` is not optimal in some kernels, it should be straightforward to override for special cases by adding new methods to the CUDA extension, which can be specialized on the current scope, the slice operator, or the types of `DataLayout`/`LazyDataLayout` arguments. This could potentially be helpful for optimizing fused broadcast expressions with significant register pressure or shared memory requirements. Aside from any such special cases, though, we'll only need to add CUDA extension code for launching kernels from `ThisHost`, and we'll be able to eliminate a lot of our current code duplication.

Let me know if any of this can be changed to further simplify things for you, @Mikolaj-A-Kowalski.

.claude/skills/fix-ci/SKILL.md
```diff
@@ -0,0 +1,208 @@
+---
```
tapios: This file probably is better at home in DeveloperGuides (from where we can vendor it to the different repos). I suggest you make this general (it almost is, the ClimaCore mentions are inessential) and start a skills directory in DeveloperGuides.

.claude/skills/fix-ci/test_compilation.jl
```diff
@@ -0,0 +1,959 @@
+"""
```
tapios: Same as above: what is generalizable to other repos is better at home in a skills dir in DeveloperGuides

ext/cuda/scopes.jl
```diff
@@ -0,0 +1,142 @@
...
+    max_subblocks = MAX_WARPS_PER_BLOCK * DataLayouts.num_subscopes(scope, ThisWarp())
+    array = DataLayouts.scoped_static_array(ThisBlock(), T, (dims..., max_subblocks))
+    subblock_index = DataLayouts.subscope_rank(scope, ThisBlock())
+    return @inbounds view(array, ntuple(Returns(:), Val(length(dims)))..., subblock_index)
+end
+
+# The last sub-block in a block may be only partially filled, so its active thread
+# count is computed from the block's total. CUDA.active_mask is unusable here: since
+# Volta's independent thread scheduling, it is not guaranteed to be consistent across
+# lanes of a warp (see https://stackoverflow.com/questions/54055195).
```
tapios: Do we need the active_mask comment? It appears to refer to old hardware.

tapios (commented):
This looks good as far as I can tell. It's nice to see the simplifications realized here.

I will send a list of detailed code comments from an audit separately, which identified two apparent reduction bugs.

tapios:
One more thing: Please add a smoke test for the DG Bickley jet (just run it a few steps). I checked it works with this branch, but numerical_flux.jl is not covered by tests; we want to make sure the DG part remains maintained (and soon build out that option).

.claude/skills/fix-ci/SKILL.md
dennisYatunin: The file has been generalized, cleared of jargon, passed through several rounds of adversarial critique by Fable, and moved to https://github.com/CliMA/DeveloperGuides/blob/main/.claude/skills/fix-ci/SKILL.md.

.claude/skills/fix-ci/test_compilation.jl
dennisYatunin: The file has been moved to https://github.com/CliMA/DeveloperGuides/blob/main/.claude/skills/fix-ci/test_compilation.jl

ext/cuda/scopes.jl
dennisYatunin: This comment is here to avoid the common mistake (e.g., made by Claude and almost missed by me) of using `CUDA.active_mask` to compute the number of threads that were launched. This would have been fine with older versions of CUDA, but now this function returns the number of threads that happen to be synchronized, rather than the number of threads that were launched. The fact that this isn't clarified in the documentation for CUDA.jl only adds to the confusion.

The bit about Volta was irrelevant though, so I've removed it.

src/DataLayouts/DataLayouts.jl
```diff
@@ -1,2233 +1,550 @@
...
 """
-    struct UniversalSize{Ni, Nj, Nv} end
-    UniversalSize(data::AbstractData)
+    DataLayout{T, N, F, S, A}
 
-A struct containing static dimensions (except `Nh`),
-universal to all datalayouts:
+An `N`-dimensional `AbstractArray` containing values of type `T`, stored in a
+parent array of type `A` whose memory layout is determined by the layout's type.
+Every value is identified by up to four indices—a vertical level `v`, horizontal
```
imreddyTeja:
```suggestion
Every value is identified by up to four indices — a vertical level `v`, horizontal
```

src/DataLayouts/indexing.jl
```diff
@@ -0,0 +1,140 @@
...
+) = view(arg, CartesianIndex(index1, index2, indices...))
+
+# Only construct slice views when necessary.
+@propagate_inbounds level(arg::IndexableData, v) =
+    all_ones(vijh_params(arg).Nv) ? arg : level_view(arg, v)
+@propagate_inbounds slab(arg::IndexableData, v, h) =
+    all_ones(vijh_params(arg).Nv, vijh_params(arg).Nh) ? arg : slab_view(arg, v, h)
+@propagate_inbounds column(arg::IndexableData, i, j, h) =
+    all_ones(vijh_params(arg).Ni, vijh_params(arg).Nj, vijh_params(arg).Nh) ? arg :
+    column_view(arg, i, j, h)
```
imreddyTeja: I'm not sure if the behavior when `all_ones` is true is a good idea. This makes it more difficult to catch incorrect indexing as a user

src/DataLayouts/masks.jl
```diff
@@ -0,0 +1,85 @@
...
+        i_map[n] = index[2]
+        j_map[n] = index[3]
+        h_map[n] = index[4]
+        n += 1
+    end
+    fill!(mask.N, n - 1)
+    if !(mask.i_map isa Array)
+        copyto!(mask.i_map, i_map)
+        copyto!(mask.j_map, j_map)
+        copyto!(mask.h_map, h_map)
```
imreddyTeja: I don't think this will function correctly on a cpu. I believe `Array` returns a copy

imreddyTeja (commented):
Have you looked at gpu benchmarks compared to main? It looks like the end to end test is a bit slower

ext/cuda/scopes.jl
```diff
@@ -0,0 +1,165 @@
...
+        minor = CUDA.attribute(device, CUDA.DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR)
+        throw(ArgumentError("Compute Capability $major.$minor is not supported"))
+    end
+    device_assumptions_checked[] = true
+    return nothing
+end
+
+@inline x_component((; x, y, z)) =
+    isone(y) && isone(z) ? x :
+    throw(ArgumentError("y and z dimensions in launch configuration are not supported"))
```
ph-kev: Why are the y and z dimensions in the launch configuration not supported?

src/DataLayouts/broadcast.jl
```diff
@@ -1,610 +1,179 @@
...
-} = true
-isascalar(
-    bc::NonExtrudedBroadcasted{Style},
-) where {
-    Style <:
-    Union{Base.Broadcast.AbstractArrayStyle{0}, Base.Broadcast.Style{Tuple}},
-} = true
-isascalar(bc) = false
+const MaybeLazyDataLayout = Union{DataLayout, LazyDataLayout}
+const MaybeFusedDataLayoutBroadcast = Union{LazyDataLayout, FusedMultiBroadcast}
```
ph-kev, line 85: I think this might be a better name?

```suggestion
const LazyOrFusedDataLayoutBroadcast = Union{LazyDataLayout, FusedMultiBroadcast}
```

src/DataLayouts/scopes.jl
```diff
@@ -0,0 +1,275 @@
...
+        num_subscopes(subscope, scope)
+    view_range =
+        strided_access(scope) ? (rank:n:length(indices)) :
+        (length(indices) * (rank - 1) ÷ n + 1):(length(indices) * rank ÷ n)
+    return subscope_index_view(scope, indices, view_range)
+end
+
+# Return a view by default, so iterating over it is as efficient as iterating
+# over the original indices. GPU scopes need to override this, since a strided
+# view of CartesianIndices is a ReshapedArray with GPU-incompilable bounds checks.
```
ph-kev:
```suggestion
# view of CartesianIndices is a ReshapedArray with GPU-incompatible bounds checks.
```

src/Utilities/safe_mapreduce.jl
```diff
@@ -0,0 +1,57 @@
...
+"""
+    safe_mapreduce(f, op, itr; [init])
+
+Analogue of `Base.mapreduce(f, op, itr; [init])` for an indexable collection
+`itr`, with the additional guarantee that it can be compiled in GPU kernels.
+Unlike `Base.mapreduce`, this never reaches the empty-collection error path,
+which builds an error string that cannot be compiled for a GPU; when `init` is
+not provided, `itr` is assumed to be non-empty. Without `init` the reduction is
+pairwise, so its roundoff error grows logarithmically rather than linearly with
+`length(itr)`; with `init` it is a sequential left fold seeded by `init`.
```
ph-kev: Could you add a sentence about the associativity of the reduction being dependent on whether `init` is initialized or not and maybe that the associativity of the reduction is not guaranteed.

imreddyTeja:
I checked the performance of this PR at commit bd53967e3a3e4d258939c3b4d201b0daa28def12 vs main. Here are my findings:

From profiling prognostic edmf with 1M microphysics in the coupler:
1. Overall, each step takes about 1.25x longer
2. There is a negligible speed-up to spectral operations
3. There is no difference in finite difference operator performance. This is expected.
4. the total time spent in Pointwise operations is ~1.5x greater than before. This is where the slowdown comes from.

When looking at simple point wise expressions in isolation, I found:
1. This branch is always faster or the same as main when not using a VIJFH with V==64. 
2. when using a VIJFH with V==64, this branch is slower than main, and uses more registers. This could be from the "tail effect", or because the specialized kernel for VIJFH_64 uses the block/grid indices for indexing.

The slowdown is not ideal, but I do not think it is enough of a degradation to prevent merging this. The "tail effect" issue should be an easy fix, and any potential differences in indexing computation should be less impactful for non-trivial point-wise kernels. At the moment, over half the point wise kernels in ClimaAtmos take less than 25 microseconds. That is short enough to where a few integer divisions will be noticeable. Hopefully, fusing more point wise operations will make this a non-problem.

ext/cuda/scopes.jl
```diff
@@ -0,0 +1,165 @@
...
+    )
+        major = CUDA.attribute(device, CUDA.DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR)
+        minor = CUDA.attribute(device, CUDA.DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR)
+        throw(ArgumentError("Compute Capability $major.$minor is not supported"))
+    end
+    device_assumptions_checked[] = true
+    return nothing
+end
+
+@inline x_component((; x, y, z)) =
```
imreddyTeja: nsight compute claims that this is causing high register usage

src/DataLayouts/indexing.jl
dennisYatunin: I've kept the special case for singleton dimensions to improve latency/inference, but I added a `@boundscheck` to avoid the issue you mentioned.

src/DataLayouts/masks.jl
dennisYatunin: Good catch! `Array` is now only called when the input is not already an `Array`.

ext/cuda/scopes.jl
dennisYatunin: This is a limitation of the current design, since nonlinear grid/block dimensions make it much harder to implement automatic coarsening (looping over multiple indices per thread) and automatic launch config adjustment (using the largest possible blocks while making sure they can all run at the same time). Support for `y` and `z` dimensions can be added in a future PR, but only if the measurable performance benefits outweigh the increase in complexity.

src/DataLayouts/broadcast.jl
dennisYatunin: I actually used that name in my first draft of this PR. I ended up replacing it with the current name because `LazyDataLayout` and `FusedMultiBroadcast` are both "lazy" objects, so calling the union "lazy or fused" is a bit confusing.

ext/cuda/scopes.jl
dennisYatunin: The checks for `isone(y)` and `isone(z)` have been dropped. I added a warning to the docstrings of `ThisKernel` and `ThisBlock` to make sure users are aware of the assumption.

NEWS.md
```diff
@@ -7,6 +7,16 @@ main
 v0.14.55
 -------
 
+- Refactor DataLayouts module [2522](https://github.com/CliMA/ClimaCore.jl/pull/2522)
```
imreddyTeja, line 7: nit: This should be under main.

imreddyTeja (approved):

PR #2525, opened 2026-06-10
Add space with multiple columns

ph-kev:
This PR adds a space that can supports multiple columns. This is needed to complete https://github.com/CliMA/ClimaLand.jl/issues/1723.

This PR have been tested with the `FluxnetSimulation`s in ClimaLand and it produces identical results with the current single column used for the `FluxnetSimulation`.

The remapping feature will not be done in this PR. For the purpose of calibration, this is not needed because the data is extracted from the ClimaCore fields rather than from the NetCDF files produced from the diagnostics.

Note that most of the code in this PR have been written with Claude.

### DIFF
--- a/ext/cuda/adapt.jl
+++ b/ext/cuda/adapt.jl
@@ -7,7 +7,7 @@ Adapt.adapt_structure(
     grid::Grids.ExtrudedFiniteDifferenceGrid,
 ) = Grids.DeviceExtrudedFiniteDifferenceGrid(
     Adapt.adapt(to, Grids.vertical_topology(grid)),
-    Adapt.adapt(to, grid.horizontal_grid.quadrature_style),
+    Adapt.adapt(to, Grids.quadrature_style(grid)),
     Adapt.adapt(to, grid.global_geometry),
     Adapt.adapt(to, grid.center_local_geometry),
     Adapt.adapt(to, grid.face_local_geometry),
--- a/src/CommonGrids/CommonGrids.jl
+++ b/src/CommonGrids/CommonGrids.jl
@@ -67,7 +67,12 @@ grid = ExtrudedCubedSphereGrid(;
 module CommonGrids
 
 export ExtrudedCubedSphereGrid,
-    CubedSphereGrid, ColumnGrid, Box3DGrid, SliceXZGrid, RectangleXYGrid
+    CubedSphereGrid,
+    ColumnGrid,
+    Box3DGrid,
+    SliceXZGrid,
+    RectangleXYGrid,
+    PointColumnEnsembleGrid
 
 import ClimaComms
 import ..DataLayouts,
@@ -699,6 +704,74 @@ function RectangleXYGrid(
         enable_bubble,
         enable_mask,
     )
+end
+
+"""
+    PointColumnEnsembleGrid(
+        ::Type{<:AbstractFloat}; # defaults to Float64
+        points::AbstractVector{Geometry.LatLongPoint{FT}},
+        z_elem::Integer,
+        z_min::Real,
+        z_max::Real,
+        radius::Real,
+        device::ClimaComms.AbstractDevice = ClimaComms.device(),
+        stretch::Meshes.StretchingRule = Meshes.Uniform(),
+        z_mesh::Meshes.IntervalMesh = DefaultZMesh(FT; z_min, z_max, z_elem, stretch),
+    )
+
+A convenience constructor that builds an
+[`Grids.ExtrudedFiniteDifferenceGrid`](@ref) for N independent columns at
+arbitrary (lat, lon) locations on a sphere, given:
+
+ - `FT` the floating-point type (defaults to `Float64`) [`Float32`, `Float64`],
+ - `points` a vector of `Geometry.LatLongPoint` specifying each column
+   location,
+ - `z_elem` the number of z-points,
+ - `z_min` the domain minimum along the z-direction,
+ - `z_max` the domain maximum along the z-direction,
+ - `radius` the radius of the sphere,
+ - `device` the `ClimaComms.device`,
+ - `stretch` the mesh `Meshes.StretchingRule` (defaults to
+   [`Meshes.Uniform`](@ref)),
+ - `z_mesh` the vertical mesh, defaults to an `Meshes.IntervalMesh` along `z`
+   with given `stretch`.
+
+There is no horizontal connectivity between columns. Horizontal operators are
+not supported. Use [`ClimaCore.Fields.bycolumn`](@ref) to iterate over columns.
+
+# Example usage
+
+```julia
+using ClimaCore.CommonGrids, ClimaCore.Geometry
+points = [LatLongPoint(0.0, 0.0), LatLongPoint(10.0, 20.0), LatLongPoint(-5.0, 90.0)]
+grid = PointColumnEnsembleGrid(;
+    points  = points,
+    z_elem  = 10,
+    z_min   = 0,
+    z_max   = 10_000,
+    radius  = 6.371229e6,
+)
+```
+"""
+PointColumnEnsembleGrid(; kwargs...) = PointColumnEnsembleGrid(Float64; kwargs...)
+function PointColumnEnsembleGrid(
+    ::Type{FT};
+    points::AbstractVector{Geometry.LatLongPoint{FT}},
+    z_elem::Integer,
+    z_min::Real,
+    z_max::Real,
+    radius::Real = FT(6.371229e6),
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    stretch::Meshes.StretchingRule = Meshes.Uniform(),
+    z_mesh::Meshes.IntervalMesh = DefaultZMesh(FT; z_min, z_max, z_elem, stretch),
+) where {FT}
+    h_grid = Grids.PointCloudGrid(points; radius, device)
+    z_topology = Topologies.IntervalTopology(
+        ClimaComms.SingletonCommsContext(device),
+        z_mesh,
+    )
+    z_grid = Grids.FiniteDifferenceGrid(z_topology)
+    return Grids.ExtrudedFiniteDifferenceGrid(h_grid, z_grid)
 end
 
 end # module
--- a/src/Grids/extruded.jl
+++ b/src/Grids/extruded.jl
@@ -45,7 +45,7 @@ local_geometry_type(
 ) where {H, V, A, GG, CLG, FLG} = eltype(CLG) # calls eltype from DataLayouts
 
 function ExtrudedFiniteDifferenceGrid(
-    horizontal_grid::Union{SpectralElementGrid1D, SpectralElementGrid2D},
+    horizontal_grid::AbstractSpectralElementGrid,
     vertical_grid::FiniteDifferenceGrid,
     hypsography::HypsographyAdaption = Flat();
     deep = false,
@@ -70,7 +70,7 @@ end
 
 # memoized constructor
 function ExtrudedFiniteDifferenceGrid(
-    horizontal_grid::Union{SpectralElementGrid1D, SpectralElementGrid2D},
+    horizontal_grid::AbstractSpectralElementGrid,
     vertical_grid::FiniteDifferenceGrid,
     hypsography::HypsographyAdaption,
     global_geometry::Geometry.AbstractGlobalGeometry,
@@ -96,7 +96,7 @@ end
 
 # Non-memoized constructor. Should not generally be called, but can be defined for other Hypsography types
 function _ExtrudedFiniteDifferenceGrid(
-    horizontal_grid::Union{SpectralElementGrid1D, SpectralElementGrid2D},
+    horizontal_grid::AbstractSpectralElementGrid,
     vertical_grid::FiniteDifferenceGrid,
     hypsography::Flat,
     global_geometry::Geometry.AbstractGlobalGeometry,
@@ -125,6 +125,11 @@ function _ExtrudedFiniteDifferenceGrid(
 end
 
 topology(grid::ExtrudedFiniteDifferenceGrid) = topology(grid.horizontal_grid)
+
+ClimaComms.context(grid::ExtrudedFiniteDifferenceGrid) =
+    ClimaComms.context(grid.horizontal_grid)
+ClimaComms.device(grid::ExtrudedFiniteDifferenceGrid) =
+    ClimaComms.device(grid.horizontal_grid)
 
 vertical_topology(grid::ExtrudedFiniteDifferenceGrid) =
     topology(grid.vertical_grid)
@@ -157,9 +162,8 @@ struct DeviceExtrudedFiniteDifferenceGrid{VT, Q, GG, CLG, FLG} <:
     face_local_geometry::FLG
 end
 
-# Specialize to allow on-device call of `device` for `DeviceExtrudedFiniteDifferenceGrid`
-ClimaComms.device(grid::DeviceExtrudedFiniteDifferenceGrid) =
-    ClimaComms.device(vertical_topology(grid))
+ClimaComms.device(::DeviceExtrudedFiniteDifferenceGrid) = DeviceSideDevice()
+ClimaComms.context(::DeviceExtrudedFiniteDifferenceGrid) = DeviceSideContext()
 
 local_geometry_type(
     ::Type{DeviceExtrudedFiniteDifferenceGrid{VT, Q, GG, CLG, FLG}},
@@ -180,3 +184,16 @@ const ExtrudedRectilinearSpectralElementGrid3D =
     ExtrudedFiniteDifferenceGrid{<:RectilinearSpectralElementGrid2D}
 const ExtrudedCubedSphereSpectralElementGrid3D =
     ExtrudedFiniteDifferenceGrid{<:CubedSphereSpectralElementGrid2D}
+const ExtrudedPointCloudGrid = ExtrudedFiniteDifferenceGrid{<:PointCloudGrid}
+
+# The show method for `AbstractGrid` calls `topology(grid)`, which errors for
+# a point-cloud horizontal grid, so print the point-cloud fields directly
+function Base.show(io::IO, grid::ExtrudedPointCloudGrid)
+    indent = get(io, :indent, 0)
+    iio = IOContext(io, :indent => indent + 2)
+    println(io, nameof(typeof(grid)), ":")
+    print_pointcloud_horizontal(iio, grid.horizontal_grid, indent)
+    println(iio)
+    println(iio, " "^(indent + 2), "vertical:")
+    print(iio, " "^(indent + 4), "mesh: ", vertical_topology(grid).mesh)
+end
--- a/src/Grids/pointcloud.jl
+++ b/src/Grids/pointcloud.jl
@@ -1,0 +1,145 @@
+
+"""
+    PointCloudGrid(
+        context :: ClimaComms.AbstractCommsContext,
+        points  :: AbstractVector{Geometry.LatLongPoint{FT}},
+    )
+
+A horizontal grid consisting of N arbitrary, disconnected (lat, long) locations
+on a sphere. There is no connectivity between columns; no spectral element
+basis, DSS, or horizontal operators are supported on this grid.
+
+This is the horizontal component used by a "point cloud" extruded space (N
+independent columns at user-chosen sphere locations).
+
+The `local_geometry` is stored as a `VIJFH{LG, 1, 1, 1, N}` data layout, with
+each of the `N` locations represented by an element with one nodal point. Based
+on the [metric
+tensor](https://en.wikipedia.org/wiki/Metric_tensor#The_round_metric_on_a_sphere)
+of a sphere, the horizontal Jacobian `∂x∂ξ` is given by the diagonal matrix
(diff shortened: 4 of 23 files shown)


test/TestUtilities/TestUtilities.jl
```diff
@@ -164,6 +165,26 @@ function FaceExtrudedFiniteDifferenceSpace(::Type{FT}; kwargs...) where {FT}
     return Spaces.FaceExtrudedFiniteDifferenceSpace(cspace)
 end
 
+function PointColumnEnsembleSpace(::Type{FT}; context, zelem = 10, kwargs...) where {FT}
+    staggering = Spaces.CellCenter()
+    lats = FT.([0.0, 1.0, 2.0])
+    longs = FT.([3.0, 4.0, 5.0])
+    points = [Geometry.LatLongPoint(lat, long) for (lat, long) in zip(lats, longs)]
+    z_min = 10.0
+    z_max = -10.0
```
nefrathenrici: These should probably be flipped

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,118 @@
...
+    deg2rad = FT(π) / 180
+
+    for (h, pt) in enumerate(points)
+        # Sphere metric: arc length per degree in each coordinate direction.
+        # ∂x∂ξ is diagonal: (R·π/180) in the lat direction,
+        #                    (R·cosd(lat)·π/180) in the lon direction.
+        s_lat = FT(radius) * deg2rad
+        s_lon = FT(radius) * deg2rad * cosd(pt.lat)
+        J = s_lat * s_lon   # det of diagonal Jacobian
+        ∂x∂ξ_mat = SMatrix{2, 2, FT, 4}(s_lat, zero(FT), zero(FT), s_lon)
```
nefrathenrici:
```suggestion
        ∂x∂ξ_mat = SMatrix{2, 2, FT, 4}(0, s_lat, s_lon, 0)
```
The axes need to be flipped. SMatrix enforces float type so I `0` is fine here.

src/Spaces/extruded.jl
```diff
@@ -277,3 +276,55 @@ const ExtrudedRectilinearSpectralElementSpace3D = ExtrudedFiniteDifferenceSpace{
 const ExtrudedCubedSphereSpectralElementSpace3D = ExtrudedFiniteDifferenceSpace{
     <:Grids.ExtrudedCubedSphereSpectralElementGrid3D,
 }
+
+# PointCloudSpace: ExtrudedFiniteDifferenceSpace backed by a PointCloudGrid.
+# N independent columns at arbitrary (lat, lon) locations; no horizontal operators.
+const PointCloudSpace = ExtrudedFiniteDifferenceSpace{
```
nefrathenrici: Why is a `PointCloudSpace` needed? Won't we just use `PointColumnEnsembleSpace` to construct the space? This would let you delete the `bycolumn` and other methods for `PointCloudSpace`

nefrathenrici (commented):
This looks good, I only saw a few immediate correctness issues.

ext/cuda/adapt.jl
```diff
@@ -13,6 +13,19 @@ Adapt.adapt_structure(
     Adapt.adapt(to, grid.face_local_geometry),
 )
 
+# PointCloudGrid has no quadrature_style field; pass nothing so that FD
+# operators on PointColumnEnsembleSpace work on CUDA.
+Adapt.adapt_structure(
```
imreddyTeja: Would it be better to combine this with the above method
```
Adapt.adapt_structure(
    to::CUDA.KernelAdaptor,
    grid::Grids.ExtrudedFiniteDifferenceGrid,
) = Grids.DeviceExtrudedFiniteDifferenceGrid(
    Adapt.adapt(to, Grids.vertical_topology(grid)),
    hasproperty(grid, :horizontal_grid.quadrature_style), Adapt.adapt(to, grid.horizontal_grid.quadrature_style),
    Adapt.adapt(to, grid.global_geometry),
    Adapt.adapt(to, grid.center_local_geometry),
    Adapt.adapt(to, grid.face_local_geometry),
)

```

src/Grids/extruded.jl
```diff
@@ -45,7 +45,7 @@ local_geometry_type(
 ) where {H, V, A, GG, CLG, FLG} = eltype(CLG) # calls eltype from DataLayouts
 
 function ExtrudedFiniteDifferenceGrid(
-    horizontal_grid::Union{SpectralElementGrid1D, SpectralElementGrid2D},
+    horizontal_grid::Union{SpectralElementGrid1D, SpectralElementGrid2D, PointCloudGrid},
```
imreddyTeja: This union appears a lot. Could it just be a const or abstract type?

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,118 @@
...
+
+This is the horizontal component used by a "point cloud" extruded space (N independent
+columns at user-chosen sphere locations).
+
+The `local_geometry` is stored as an `IFH{LG, 1, N}` data layout — one node per
+"element" (`Ni = 1`), N "elements" (`Nh = N`). The horizontal Jacobian `∂x∂ξ` is the
+diagonal matrix `diag(R·π/180, R·cosd(lat)·π/180)` reflecting the sphere metric at
+each point, and `J = R²·cosd(lat)·(π/180)²`.
+"""
+struct PointCloudGrid{
```
imreddyTeja, line 23: I think it might be better to add a field for topology, and make it always nothing.

imreddyTeja (approved):

src/CommonGrids/CommonGrids.jl
```diff
@@ -706,4 +711,77 @@ function RectangleXYGrid(
...
+    z_min::Real,
+    z_max::Real,
+    radius::Real = 6.371229e6,
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.SingletonCommsContext(device),
+    stretch::Meshes.StretchingRule = Meshes.Uniform(),
+    z_mesh::Meshes.IntervalMesh = DefaultZMesh(FT; z_min, z_max, z_elem, stretch),
+) where {FT}
+    @assert context isa ClimaComms.SingletonCommsContext "PointColumnEnsembleGrid only supports SingletonCommsContext."
+    @assert ClimaComms.device(context) == device "The given device and context device do not match."
```
dennisYatunin:
```suggestion
    context = ClimaComms.SingletonCommsContext(device)
```
No need to provide a keyword argument for `context` if it can only have one possible value.

src/Grids/extruded.jl
dennisYatunin: Yeah aren't these all subtypes of `AbstractSpectralElementGrid`? Could just use that instead.

src/Grids/extruded.jl
```diff
@@ -126,6 +126,13 @@ end
 
 topology(grid::ExtrudedFiniteDifferenceGrid) = topology(grid.horizontal_grid)
 
+# For PointCloudGrid there is no horizontal topology; route context/device
+# through the horizontal grid directly (which stores its own context).
+ClimaComms.context(grid::ExtrudedFiniteDifferenceGrid{<:PointCloudGrid}) =
+    ClimaComms.context(grid.horizontal_grid)
+ClimaComms.device(grid::ExtrudedFiniteDifferenceGrid{<:PointCloudGrid}) =
+    ClimaComms.device(grid.horizontal_grid)
```
dennisYatunin:
```suggestion
ClimaComms.context(grid::ExtrudedFiniteDifferenceGrid) =
    ClimaComms.context(grid.horizontal_grid)
ClimaComms.device(grid::ExtrudedFiniteDifferenceGrid) =
    ClimaComms.device(grid.horizontal_grid)
```
This could also be the default for every `ExtrudedFiniteDifferenceGrid`. Special cases for such specific subtypes are generally harder to maintain.

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,118 @@
...
+sphere. There is no connectivity between columns; no spectral element basis, DSS, or
+horizontal operators are supported on this grid.
+
+This is the horizontal component used by a "point cloud" extruded space (N independent
+columns at user-chosen sphere locations).
+
+The `local_geometry` is stored as an `IFH{LG, 1, N}` data layout — one node per
+"element" (`Ni = 1`), N "elements" (`Nh = N`). The horizontal Jacobian `∂x∂ξ` is the
+diagonal matrix `diag(R·π/180, R·cosd(lat)·π/180)` reflecting the sphere metric at
+each point, and `J = R²·cosd(lat)·(π/180)²`.
```
dennisYatunin:
```suggestion
The `local_geometry` is stored as an `IFH{LG, 1, N}` data layout, with each of
the `N` locations represented by an element with one nodal point. Based on the
[metric tensor](https://en.wikipedia.org/wiki/Metric_tensor#The_round_metric_on_a_sphere)
of a sphere, the horizontal Jacobian `∂x∂ξ` is given by the diagonal matrix
`diag(R·π/180, R·cosd(lat)·π/180)`, with the determinant `J = R²·cosd(lat)·(π/180)²`.
```
Wording here was a bit unclear, and I thought it would be good to add a link to some reference.

src/Grids/pointcloud.jl
dennisYatunin: This could simplify all the `topology(grid)` methods, but subsequent operations on the result of `topology` would still be invalid. We could make some sort of `NoTopology` struct, or add methods to topology-related functions so they can handle `Nothing`, but that feels roughly equivalent to what this PR is doing anyway. So I don't think adding a field for the topology would be very useful.

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,118 @@
...
+
+local_geometry_type(::Type{PointCloudGrid{C, GG, LG}}) where {C, GG, LG} =
+    eltype(LG)
+
+ClimaComms.context(grid::PointCloudGrid) = grid.context
+ClimaComms.device(grid::PointCloudGrid) = ClimaComms.device(grid.context)
+
+# No topology — return nothing. Callers that need a topology (e.g. DSS) should
+# not be called on a PointCloudGrid.
+topology(::PointCloudGrid) = nothing
```
dennisYatunin: Could this throw an error instead? Otherwise, users will see a `MethodError` about how something does not work when the topology is `Nothing`, rather than a more informative message about how they should not access the topology of a `PointCloudGrid`.

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,118 @@
...
+`local_geometry` are set from the sphere geometry at each point:
+`∂x∂ξ = diag(R·π/180, R·cosd(lat)·π/180)`, `J = R²·cosd(lat)·(π/180)²`.
+"""
+function PointCloudGrid(
+    points::AbstractVector{Geometry.LatLongPoint{FT}};
+    radius::Real,
+    device::ClimaComms.AbstractDevice = ClimaComms.device(),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.SingletonCommsContext(device),
+) where {FT}
+    @assert context isa ClimaComms.SingletonCommsContext "PointCloudGrid only supports SingletonCommsContext."
```
dennisYatunin:
```suggestion
    context = ClimaComms.SingletonCommsContext(device)
```
As in `CommonGrids`, you can avoid the keyword argument if it only accepts one possible value.

src/Spaces/multicolumn.jl
```diff
@@ -0,0 +1,209 @@
...
+    PointCloudLevelSpace
+
+A horizontal space of N independent points (lat, lon, z), produced by taking a
+vertical level of a [`MultiColumnFiniteDifferenceSpace`](@ref).  This is the
+N-column analogue of [`PointSpace`](@ref), which is the single-column level
+space.
+
+`local_geometry` holds an `IFH{LG, 1, N}` data layout.
+"""
+struct PointCloudLevelSpace{
```
dennisYatunin: Can this be called a `PointCloudSpace` instead? The name of every other `___Space` type matches that of its corresponding `___Grid` type.

src/Spaces/multicolumn.jl
```diff
@@ -0,0 +1,209 @@
...
+    size(local_geometry_data(space), 4)  # Nv dim of VIFH
+
+horizontal_space(space::MultiColumnFiniteDifferenceSpace) =
+    level(MultiColumnFiniteDifferenceSpace(grid(space), CellCenter()), 1)
+
+# No DSS / mask machinery needed.
+get_mask(space::PointCloudLevelSpace) = DataLayouts.NoMask()
+get_mask(space::MultiColumnFiniteDifferenceSpace) = DataLayouts.NoMask()
+set_mask!(fn, space::MultiColumnFiniteDifferenceSpace) =
+    set_mask!(fn, grid(space).horizontal_grid)
```
dennisYatunin: Shouldn't this do nothing if `get_mask` always returns `NoMask`?

src/Spaces/multicolumn.jl
```diff
@@ -0,0 +1,209 @@
...
+# PointSpace field can be broadcast with a CenterFiniteDifferenceSpace field.
+function issubspace(
+    level_space::PointCloudLevelSpace,
+    full_space::MultiColumnFiniteDifferenceSpace,
+)
+    hgrid = grid(full_space).horizontal_grid
+    # local_geometry of the level space is an IFH slice of the full VIFH;
+    # both share the same horizontal grid data array.
+    return size(level_space.local_geometry, 5) == size(
+        Grids.local_geometry_data(hgrid, nothing), 5)
```
dennisYatunin: Could this check the `local_geometry` itself, instead of just its `size`? Maybe something like `level_space.local_geometry === Grids.local_geometry_data(hgrid, nothing)`, if that's possible.

dennisYatunin (approved):
This looks great! It has a fair bit of code duplication, both in terms of method definitions and `const` type declarations, but that's a higher-level design issue that can be fixed separately. I left a few nitpicks about readability and error handling, but overall it looks ready to merge in.

src/Grids/pointcloud.jl
ph-kev: I tested this and it wasn't giving the same answer when comparing it against the ClimaLand single column run.

ph-kev: I tested this again and they are identical (up to some tolerance). I am not sure why I was getting different answers before.

src/CommonGrids/CommonGrids.jl
```diff
@@ -706,4 +711,72 @@ function RectangleXYGrid(
...
+```
+"""
+PointColumnEnsembleGrid(; kwargs...) = PointColumnEnsembleGrid(Float64; kwargs...)
+function PointColumnEnsembleGrid(
+    ::Type{FT};
+    points::AbstractVector{Geometry.LatLongPoint{FT}},
+    z_elem::Integer,
+    z_min::Real,
+    z_max::Real,
+    radius::Real = 6.371229e6,
```
imreddyTeja: Should this be wrapped in an `FT` like above?

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,121 @@
...
+    deg2rad = FT(π) / 180
+
+    for (h, pt) in enumerate(points)
+        # Sphere metric: arc length per degree in each coordinate direction.
+        # ∂x∂ξ is diagonal: (R·π/180) in the lat direction,
+        #                    (R·cosd(lat)·π/180) in the lon direction.
+        s_lat = FT(radius) * deg2rad
+        s_lon = FT(radius) * deg2rad * cosd(pt.lat)
+        J = s_lat * s_lon   # det of diagonal Jacobian
+        ∂x∂ξ_mat = SMatrix{2, 2, FT, 4}(zero(FT), s_lat, s_lon, zero(FT))
```
imreddyTeja: Where does this behavior come from. It seems different than both a true single-column and an extruded spectral element. I think this would change the result of integrals

src/Spaces/multicolumn.jl
```diff
@@ -0,0 +1,211 @@
...
+# field (IFH) across a full multi-column field (VIFH), analogous to
+# issubspace(::SpectralElementSpace2D{<:LevelGrid}, ::ExtrudedFiniteDifferenceSpace).
+function issubspace(
+    level_space::PointCloudSpace,
+    full_space::MultiColumnFiniteDifferenceSpace,
+)
+    return level_space.full_grid === grid(full_space)
+end
+function issubspace(space1::PointCloudSpace, space2::PointCloudSpace)
+    return space1.full_grid === space2.full_grid
```
imreddyTeja: should there be another check after this to make sure the lgs are equal?

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,121 @@
...
+        radius  :: Real,
+        device  :: ClimaComms.AbstractDevice = ClimaComms.device(),
+    )
+
+Convenience constructor: build a `PointCloudGrid` from a vector of
+`LatLongPoint`s and a sphere `radius`. The horizontal metric terms in
+`local_geometry` are set from the sphere geometry at each point:
+`∂x∂ξ = diag(R·π/180, R·cosd(lat)·π/180)`, `J = R²·cosd(lat)·(π/180)²`.
+"""
+function PointCloudGrid(
```
imreddyTeja, line 62: Should this replicate the caching behavior of the other grids? I think that lets the brpadcasting axes check use `===`

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,121 @@
...
+        s_lon = FT(radius) * deg2rad * cosd(pt.lat)
+        J = s_lat * s_lon   # det of diagonal Jacobian
+        ∂x∂ξ_mat = SMatrix{2, 2, FT, 4}(zero(FT), s_lat, s_lon, zero(FT))
+
+        lg_slab = slab(local_geometry, h)
+        lg_slab[slab_index(1)] = Geometry.LocalGeometry(
+            pt,
+            J,
+            J,   # WJ — unit quadrature weight × J
+            Geometry.AxisTensor(∂x∂ξ_axes, ∂x∂ξ_mat),
```
imreddyTeja: Isn't it possible for `∂x∂ξ_mat` to be un-invertible?

src/Grids/pointcloud.jl
```diff
@@ -0,0 +1,121 @@
...
+
+    DA = ClimaComms.array_type(device)
+    return PointCloudGrid(
+        context,
+        global_geometry,
+        DataLayouts.rebuild(local_geometry, DA),
+    )
+end
+
+function Base.show(io::IO, grid::PointCloudGrid)
```
imreddyTeja, line 140: I think you need one of these for the extruded grid that wraps the point cloud grid

imreddyTeja (commented):
I left some minor comments.

src/Grids/pointcloud.jl
nefrathenrici: I think it should be 
```suggestion
        # Sphere metric: arc length per degree in each coordinate direction.
        # ∂x∂ξ is diagonal: (R·π/180) in the lat direction,
        #                    (R·cosd(lat)·π/180) in the lon direction.
        s_lat = FT(radius) * deg2rad
        s_lon = FT(radius) * deg2rad * cosd(pt.lat)
        J = s_lat * s_lon   # det of diagonal Jacobian
        ∂x∂ξ_mat = SMatrix{2, 2, FT, 4}(s_lon, zero(FT), zero(FT), s_lat)
```

```
[ s_lon 0 ]
[ 0 s_lat ]
```
I'd also prefer not to use `AxisTensor`, since that is deprecated code.

src/Grids/pointcloud.jl
ph-kev: It is possible if you chose a point at the poles (latitude = 90 degrees). I added a check for this and throw an error if this is possible.

ext/cuda/adapt.jl
```diff
@@ -7,6 +7,7 @@ Adapt.adapt_structure(
     grid::Grids.ExtrudedFiniteDifferenceGrid,
 ) = Grids.DeviceExtrudedFiniteDifferenceGrid(
     Adapt.adapt(to, Grids.vertical_topology(grid)),
+    grid.horizontal_grid isa Grids.PointCloudGrid ? nothing :
```
dennisYatunin: Does `Adapt.@adapt_structure PointCloudGrid` in the `Grids` module not handle this case properly?

ph-kev: I don't think so, since
```
Adapt.adapt_structure(
    to::CUDA.KernelAdaptor,
    grid::Grids.ExtrudedFiniteDifferenceGrid,
)
```
is called if you have an `ExtrudedFiniteDifferenceGrid` with a `PointCloudGrid`. I could add quadrature style as a field in the grid that is always nothing if you don't like the ternary.

dennisYatunin: Ah, I think I see what you mean. Maybe a simpler move would be to use `quadrature_style(grid.horizontal_grid)`, where
```julia
quadrature_style(::PointCloudGrid) = nothing
```
But you can also leave it as is if you prefer.

dennisYatunin: Adding this method for `quadrature_style` would also probably get rid of the need for your `ExtrudedPointCloudGrid` `show` method.

ph-kev: We still need the show method since `topology` throws an error message for a `PointCloudGrid`.

PR #2528, opened 2026-06-18
Missing method for point space field matrix solves

kmdeck:
The land has an off-diagonal block in the Jacobian which, in a column, corresponds to \partialY\partialX where both Y and X are defined on the same point space. Without defining this method, the linear equation solver fails.


imreddyTeja (approved):
It would be good to add a test for this

kmdeck:
I can try and write a test for this but I dont really understand it, @dennisYatunin had suggested the fix.

It had been failing in climaland when we were doing the linear solve, related to an off diagonal block of the Jacobian where both variables were defined on point spaces

imreddyTeja:
> I can try and write a test for this but I dont really understand it, @dennisYatunin had suggested the fix.
> 
> It had been failing in climaland when we were doing the linear solve, related to an off diagonal block of the Jacobian where both variables were defined on point spaces

It's probably fine to not add a test for this. ClimaCore tests probably don't cover a lot of `PointSpace` functionality already

PR #2544, opened 2026-07-09
FD operator refactor 

### LINKED ISSUE #2418: Refactor Matrix multiplication

### LINKED ISSUE #2449: Reduce warp divergence with boundary conditions
Investigate penalty of boundary conditions, and reduce divergence

### LINKED ISSUE #2478: Operator Matrices applied to a Broadcast with type args
## Describe the bug
When using CUDA, an operator matrix cannot be applied to a broadcast that has an arg that is a Type. I believe this is because the matrix vector multiplication method calls eltype on the broadcast, which is not inferable with `CUDA.CuRefType`(s)

## To Reproduce
cfield.x is a scalar field
```julia
@. ᶠinterp(cfield.x^2 * eps(FT))
```

imreddyTeja:
Disclosure: I used claude fable to assist with the pull request, especially for adding test cases, fixing edge case compilation issues, and writing docstrings. 

Transforms finite difference operators into their operator matrix equivalent through broadcast expression manipulation. Deletes `stencil_interior`, `stencil_left_boundary`, and `stencil_right_boundary` methods for operators with an operator matrix equivalent.

With the exception of  manually constructed operator matrices for `GradientF2C` and `DivergenceF2C` using the `SetValue` boundary condition, all specialized `SetValue`, `SetGradient`, `SetCurl`, and `SetDivergence` methods are deleted.

Non-extrapolate boundary conditions are now applied as a `SetBoundary` to the input or output.

Fixes the return type of gradient of a vector.

Fixes the return type of divergence of a 2-Tensor.

Modifies the boundary conditions ofadvection operators in a breaking way.
Adds to the `conv_column.jl` test. See plots [here](https://caltech.box.com/s/dq1ggh6r21lm0qxfmgfnjt5oci9y5fnm)

Modifies all non linear advection operators to use a single pathway.

Extend eager evaluation to all domains with n_levels < 256, expressions with large element types, and periodic topologies, and masked domains.

closes #2478
closes #2449
closes #2418

## Changes overview

1. **Eager FD evaluation** (`ext/cuda/operators_fd_eager.jl`): one x-thread per
   face level, one column per y-slot; each sub-expression of a broadcast is
   evaluated once per level and cached in dynamic shared memory instead of being
   re-evaluated per stencil offset. Column indices are decomposed with
   `CartesianIndices` whose `Ni`/`Nj` extents are compile-time constants read
   off the layout type (`vijh_params`), so the per-thread `divrem` uses
   fixed divisors (an earlier draft used runtime divisors and was 1.7x slower).
2. **Linear operators become matrix multiplies** inside the eager kernel: a
   `GradientF2C(θ)`-style node is rewritten to
   `FDOperatorMatrix ⋅ θ` (`MultiplyColumnwiseBandMatrixField`), with the
   operator's boundary rows folded into in-range columns
   (`MatrixFields/operator_matrices.jl`). Value-fixing BCs route through a
   generalized `SetBoundaryOperator` that gained projecting conditions
   (`SetGradient`/`SetCurl`/`SetDivergence`) so a rewritten broadcast can
   reapply the original operator's BCs.
3. **Advection operators** (upwind, FCT, TVD, LinVanLeer) become
   `AdvectionOperator`s / `NonLinearAdvectionOperator`s sharing one stencil and
   one eager shmem handler (measured 6.1x on nested FCT expressions).
4. **Boundary conditions simplified**: the one-sided advection BCs
   (`FirstOrderOneSided`, `ThirdOrderOneSided`) are replaced by a parametric
   **`Extrapolate{N}`** (N = 0, 1, 2 ghost-point extrapolation order); custom
   callable BCs are removed. Several operators/BC combinations are deleted
   outright (`AdvectionC2C/F2F`, `FluxCorrectionC2C/F2F`,
   `UpwindBiasedGradient`, `LeftBiased3rdOrder*`/`RightBiased3rdOrder*`;
   `Extrapolate` removed from `GradientF2C`/`DivergenceF2C`, `SetGradient`
   added to `GradientF2C`).

## Notable design choices and behavior changes

- **Missing BC now use finite zero boundary rows** (not main's loud NaN). This is
  deliberate and documented in NEWS, but interacts with the open BC-name
  finding below.

- **Ghost-padding advection boundaries**: all faces use the interior stencil;
  out-of-range centers are clamped then replaced by the BC's extrapolation.
  At the boundary face the extrapolation order caps at 1 (only 2 in-range
  points); order separation is only observable at the one-in face.

## Math review

- changes to the examples, and particularly the advection convergence tests in `test/Operators/finitedifference/conv_column.jl` should be looked at by someone.
- Naming of the modified `Extrapolate`


test/Operators/finitedifference/conv_column.jl
```diff
@@ -364,101 +437,598 @@ end
...
+        meshes = [
+            (;
+                stretch_name = "uniform",
+                stretch = Meshes.Uniform(),
+                rates = [(0.5, 0.05), (0.5, 0.05), (0.5, 0.05)],
+            ),
+            (;
+                stretch_name = "stretched",
+                stretch = Meshes.ExponentialStretching(1.0),
+                rates = [(0.68, 0.1), (0.17, 0.1), (1.13, 0.1)],
```
tapios: Risky to pin these (noise-dominated) rates for stretched meshes (especially for the extreme stretching here, where elements can occupy a significant fraction of the domain). Better to test with more elements. Also looks like convergence rate is based on smallest grid spacing, but error is probably dominated by large mesh spacings (at the top). Consider reworking this to be more meaningful.

test/Operators/finitedifference/conv_column.jl
```diff
@@ -364,101 +437,598 @@ end
...
+    # inflow at the boundaries and a scalar whose zero at the walls is simple,
+    # the O(Δz) one-in ghost error of Extrapolate{0} collapses the tendency
+    # rate to ~0.5 near the boundary and the flux rate to ~1.5, while
+    # Extrapolate{1} keeps them at ~2. Extrapolate{2} is indistinguishable
+    # from Extrapolate{1} here, so it is not tested: at the boundary face the
+    # order is capped at 1 (the two are bitwise identical there, and that face
+    # is excluded from the flux error anyway), and at the one-in face sin's
+    # inflection at its boundary zero (c''(±π) = 0) degenerates
+    # Extrapolate{1}'s O(Δz²) ghost error to O(Δz³), the same order as
+    # Extrapolate{2}'s. A scalar with a simple zero and nonzero curvature at
```
tapios: Would be good to add a non-degenerate test that covers Extrapolate{2}, and use that for Extrapolate{1} too (to see the generic 1.5-order convergence)

test/Operators/finitedifference/conv_column.jl
```diff
@@ -501,72 +1078,35 @@ end
...
-
-    # Upwind3rdOrderBiasedProductC2F conv, with f(z) = sin(z), w(z) = cos(z)
-    @test err_adv_wc[3] ≤ err_adv_wc[2] ≤ err_adv_wc[1] ≤ 4e-3
-    @test conv_adv_wc[1] ≈ 2 atol = 0.2
-    @test conv_adv_wc[2] ≈ 2 atol = 0.1
-    @test conv_adv_wc[3] ≈ 2 atol = 0.1
-end
-
-@testset "Upwind3rdOrderBiasedProductC2F + DivergenceF2C on (uniform and stretched) non-periodic mesh, with FirstOrderOneSided + DivergenceF2C SetValue BCs, constant w" begin
+@testset "Simple FCT: lin combination of UpwindBiasedProductC2F + Upwind3rdOrderBiasedProductC2F on (uniform and stretched) non-periodic mesh, finite-volume-averaged initial condition" begin
```
tapios: This set looks redundant now

test/Operators/finitedifference/conv_column.jl
tapios: Better to test for real convergence with a stretching that is not as extreme.

tapios (commented):
I left some suggestions for improving tests. I implemented them (plus a few more) in https://github.com/CliMA/ClimaCore.jl/commit/51b43ae9b5c7c7360941b778161252e4fc3036f6 (the branch sits on top of yours). Testing this, Claude surfaced a dynamic dispatch issue, fixed in https://github.com/CliMA/ClimaCore.jl/commit/df7e8a2efb55f50b332e8ebf68f80777a9f8364b (I am less confident about these changes than the test changes in the other commit).

If this makes sense to you, merge it into your branch. 

I think there also is one other issue that may cause the BoundsErrors in the tests: In src/MatrixFields/matrix_multiplication.jl (l. 345-352), each corner type clips only its own end of the band but leaves ld/ud untouched. As in the GPU path, both ends chould be clamped.

tapios:
Can you remove the downstream KinematicDriver tests, please? KD is no longer maintained.

tapios:
While I think the KD downstream test should be removed, it points to one real issue: `DivergenceF2C(top = Extrapolate())` is no longer available. The extrapolation boundary condition for outflow (e.g., precipitation leaving the domain) is legitimate, and we will need it somewhere. It would be good to enable that.

src/Operators/finitedifference.jl
```diff
@@ -98,6 +104,27 @@ strip_space(bc::AbstractBoundaryCondition, parent_space) =
     NullBoundaryCondition()
 
 This is used as a placeholder when no other boundary condition can be applied.
+
+Wherever an operator needs a boundary row for this condition (that is, wherever
+[`boundary_width`](@ref) is nonzero for it), the value produced there is a placeholder
+rather than a meaningful result, and which placeholder depends on how the operator is
+evaluated:
+
+  - an operator that is rewritten into an operator matrix multiply gets a zero boundary
```
tapios: This seems a bit risky: If the BC is missing, we now return a zero instead of NaN. May make debugging hard if that happens by accident. Can we guard this somewhere (e.g., with an error)?

tapios (approved):
Looks good to me. Consider the comment about the outflow boundary conditions (to old Extrapolate) in the divergence. It would be good to make clear how to emulate this behavior now (and how to get outflow BCs).

src/Operators/finitedifference.jl
imreddyTeja: I'll switch it to NaNs, unless you would prefer an error to be thrown.

tapios: NaN is good. Thank you!

dennisYatunin: `NaN`s will be tricky if we want to stay consistent with the `BandMatrixRow` coefficients---those can't have `NaN`s in them because `+` will always propagate them into the output. An error might be more useful from a debugging perspective.

imreddyTeja: I changed the behavior so the operator matrix rows themselves contains NaNs. That should be the same as the behavior of the old non operator matrix versions.

src/Operators/finitedifference.jl
```diff
@@ -3477,83 +2216,310 @@ end
...
+
+`x` must have a scalar eltype. Each boundary value may be a number or a
+`Field` on the corresponding boundary level of `x`'s space; a boundary value
+that is already an [`AbstractBoundaryCondition`](@ref) (e.g. a `SetGradient`)
+is instead applied as given, so a Dirichlet value on one boundary can be
+combined with an explicit condition on the other; and a boundary without a
+prescribed value is computed as by `GradientC2F` without a boundary condition
+there. The result is materialized on the face space.
+"""
+function gradient_c2f_dirichlet(x; boundary_values...)
```
dennisYatunin: The `_dirichlet` behavior can be moved into the corresponding operator constructors. That way, you would still have just one version of each operator, and auto-convert Dirichlet BCs into our standard BCs during operator construction.

examples/hybrid/plane/density_current_2d_invariant_rhoe.jl
```diff
@@ -20,7 +20,8 @@ import ClimaCore:
     Spaces,
     Quadratures,
     Fields,
-    Operators
+    Operators,
+    Utilities
```
dennisYatunin:
```suggestion
    Operators
```
The new import is unused.

examples/hybrid/flux_correction_utils.jl
```diff
@@ -0,0 +1,83 @@
...
+
+# A diffusive flux correction for center- and face-valued quantities, written in
+# terms of the gradient operators. Its diffusivity is |velocity| Δz, which is
+# the form of numerical diffusion first-order upwinding introduces. Neither
+# version carries boundary values of its own: the center-valued one passes no
+# correction flux through the boundary faces, and the face-valued one takes its
+# outer gradient there to be the one-sided difference with the inner gradient
+# set to zero outside of the domain.
+
+function add_flux_correction_c2c(d_, velocity, quantity)
```
dennisYatunin:
```suggestion
function add_flux_correction_c2c!(d_, velocity, quantity)
```
This and `add_flux_correction_f2f` should have exclamation marks in their names.

examples/hybrid/flux_correction_utils.jl
```diff
@@ -0,0 +1,83 @@
...
+    # The gradient on each boundary face is ±`inner_grad` at the adjacent cell
+    # center, which is the one-sided difference with `inner_grad` taken as zero
+    # outside of the domain.
+    top_gradient_extrapolate = Operators.SetGradient(
+        Geometry.outer.(
+            (Geometry.Covariant3Vector(-1),),
+            Fields.Field(
+                Fields.field_values(Fields.level(inner_grad, n_levels)),
+                top_level_space,
+            ),
```
dennisYatunin: If you allow `AbstractBoundaryCondition` to store a lazy `Broadcasted` instead of a `Field`, this sort of clunkiness would be unnecessary.

ext/cuda/operators_finite_difference.jl
```diff
@@ -33,32 +33,78 @@ function Base.copyto!(
 
     fspace = Spaces.face_space(space)
     n_face_levels = Spaces.nlevels(fspace)
+    high_resolution = !(n_face_levels ≤ 256)
+    # https://github.com/JuliaGPU/CUDA.jl/issues/2672
+    max_shmem = CUDA.attribute(
+        device(),
+        CUDA.DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK,
+    )
```
dennisYatunin:
```suggestion
    max_shmem = device_attributes().max_shmem_per_block
```
Use the cached attributes to reduce host-side latency.

src/Geometry/Geometry.jl
```diff
@@ -23,6 +23,36 @@ include("conversions.jl")
...
+construction. The cost is a frozen method table: the generator cannot see
+`_dual_axes_for_projection` methods defined after it, including any added by
+downstream packages.
+
+This is defined here, rather than next to `_dual_axes_for_projection`, because a
+generator may only call methods that already exist when it is defined, and the
+last `_dual_axes_for_projection` method is added in `auto_broadcaster_methods.jl`.
+"""
+@generated recursively_find_dual_axes_for_projection(::Type{X}) where {X} =
+    QuoteNode(_dual_axes_for_projection(X))
```
dennisYatunin: This generated function can probably be avoided if you use `unrolled_map`/`unrolled_all`/`unrolled_allequal` instead of `map`/`all`/`allequal`, as well as `fieldtype_vals` instead of `fieldtypes`. But if that takes too much effort you can just leave it as is.

test/Operators/finitedifference/unit_upwind_schemes.jl
```diff
@@ -82,3 +122,678 @@ import ClimaCore:
...
+# the condition's extrapolation from the in-range interior points of the
+# stencil (the order is reduced at the boundary face itself, where only 2
+# interior points are in range). The reference for every face, boundary faces
+# included, is therefore the operator's own pointwise function applied to
+# hand-clamped (or hand-extrapolated) stencil values.
+
+# Ghost-point reconstructions are no longer user-supplied callables; this is
+# used to check that custom callables are rejected.
+struct CustomGhost end
+(::CustomGhost)(closest, second_closest) = 3 * closest - 2 * second_closest
```
dennisYatunin: Is this functionality that needs to be tested? Unless there's some particular failure mode this is catching, it looks like a generic anti-duck-typing test, which isn't very Julia-flavored.

test/Operators/finitedifference/unit_upwind_schemes.jl
```diff
@@ -82,3 +122,678 @@ import ClimaCore:
...
+            end
+        end
+
+        # Extrapolate on Upwind3rdOrderBiasedProductC2F evaluates the interior
+        # stencil with extrapolated ghost points: at the boundary face itself
+        # both ghost points share the extrapolation from the 2 in-range
+        # interior points (so the order is reduced to at most 1 there), and at
+        # the face one in from the boundary the single ghost point is
+        # extrapolated from the 3 in-range points.
+        upwind3rd(v, a⁻⁻, a⁻, a⁺, a⁺⁺) =
```
dennisYatunin: Are the `upwind` and `upwind3rd` unit tests here distinct from those in `unit_operator_matrices.jl`? If the upwinding `BandMatrixRow` coefficients are already known to be correct, and if matrix multiplication has its own unit tests, we don't need an additional set of unit tests for the upwinding outputs.

ext/cuda/operators_fd_eager.jl
```diff
@@ -506,30 +771,73 @@ end
...
+) = ClimaCore.Utilities.AutoBroadcaster(
+    project_or_map(axes_per_component, lg, ClimaCore.Utilities.unwrap(y)),
+)
+# A single axis projects every tensor leaf below it (a container, like a
+# BandMatrixRow, maps the whole projection over its entries).
+Base.@propagate_inbounds project_or_map(axis, lg, y) =
+    map(Base.Fix1(recursively_project, (axis, lg)), y)
+@inline project_or_map(axis, lg, y::Number) = y
+Base.@propagate_inbounds project_or_map(axis, lg, y::AbstractTensor) =
+    @inbounds @inline project(axis, y, lg)
```
dennisYatunin:
```suggestion
@inline project_or_map(axis, lg, y::AbstractTensor) = project(axis, y, lg)
```
Why does this need an `@inbounds`? None of the input arguments are indices.

src/Geometry/mul_with_projection.jl
dennisYatunin, line 18: Is this machinery still used anywhere following your changes?

src/MatrixFields/operator_matrices.jl
```diff
@@ -332,41 +603,155 @@ op_matrix_first_row(op, bc, space, idx, hidx, args...) =
...
+#
+# The row must constant-fold during inference: these stencil branches are compiled
+# into every broadcast over the operator matrix (including the matrix-multiply
+# kernels, whose inference budget is already tight), and any runtime NaN-row
+# construction there -- even a fully type-stable one -- makes the surrounding kernels
+# type-unstable. `@assume_effects :foldable` licenses concrete evaluation of this
+# call during inference, so it reduces to a constant just like `rzero`; the function
+# only turns type arguments into an isbits value, so the asserted effects hold (do
+# not add reads of runtime state here).
+Base.@assume_effects :foldable @inline nan_boundary_row(
```
dennisYatunin: Instead of the `@assume_effects` incantation, you could probably put an `@inline` or something on the `convert` method for `BandMatrixRow`. That would fix the underlying constant-folding issue, rather than just this special case.

src/MatrixFields/operator_matrices.jl
```diff
@@ -593,195 +921,156 @@ Base.@propagate_inbounds function op_matrix_interior_row(
...
+# trailing zeros of `extrapolate_weights`'s 3-tuple keep them untouched here,
+# so that the opposite boundary's fold (applied before or after this one) can
+# zero them and fold their coefficients itself. `extrapolate_weights` returns
+# 3 weights, which is enough for any row with up to 4 entries (a row with at
+# least 1 ghost entry has at most 3 others).
+@inline function fold_extrapolate_row_left(
+    row::BandMatrixRow{ld, bw},
+    bc::Operators.Extrapolate,
+    nghost,
+    nother = 0,
```
dennisYatunin: We probably want to make sure that `nghost` and `nother` are constant-propagated into here, to minimize the runtime cost of the `ntuple` constructors. This may require wrapping them in `Val`s when you call `fold_extrapolate_row_left`/`fold_extrapolate_row_right`.

dennisYatunin (approved):
Excellent refactor; this will make it much more straightforward to rewrite the stencil operators in terms of `foreach_column`. My only major suggestion is to turn the `_dirichlet` operator functions back into standard operator constructors: the functional forms are syntactically different and may be confusing for new users, and the distinction between materialized and unmaterialized functional forms is also rather clunky.

src/MatrixFields/operator_matrices.jl
imreddyTeja: That did the trick. thanks!

test/Operators/finitedifference/unit_upwind_schemes.jl
imreddyTeja: The only distinction is that these test the auto conversion into operator matrices. I can remove them if you think that is un-needed

examples/hybrid/flux_correction_utils.jl
imreddyTeja: That already worked... I think I was doing this because I was following the pattern used for dirichlet bcs.

src/Geometry/Geometry.jl
imreddyTeja: I didn't realize `fieldtype_vals` existed already. I think that will let me remove the generated function

test/Operators/finitedifference/unit_upwind_schemes.jl
dennisYatunin: Yeah I think that functionality is effectively tested elsewhere as well, since it's the same for all linear operators. The tests in this file can be limited to the nonlinear operators, since they aren't exercised elsewhere.

src/Geometry/Geometry.jl
imreddyTeja: Yep, that fixes it!

src/Geometry/mul_with_projection.jl
imreddyTeja: Yes. The cpu matmul code still uses this because it is lazily evaluating expressions. It can be removed after we switch to `foreach_column`

PR #2545, opened 2026-07-11
Unify row_mul_mat! and row_mul_vec! into single methods

imreddyTeja:
> [!NOTE]
> The commit in this PR was authored by Claude (Claude Code), working on behalf of @imreddyTeja.

## Summary

`ext/cuda/column_matrix_helpers.jl` contained 8 nearly identical `row_mul_mat!` methods (one per pair of matrix staggering shapes) and 4 nearly identical `row_mul_vec!` methods. This PR replaces them with **one method each**, shrinking the file from 431 to 119 lines (−376/+65).

The per-shape index arithmetic is factored into three small compile-time shape-trait helpers:

- `column_slot(v, d, shape)` — converts entry `d` of the matrix row computed by thread `v` into the shared-memory slot of the value it multiplies (`v + d`, with `±half` corrections for the mixed-staggering shapes)
- `n_column_slots(shape)` — number of valid slots in the matrix's column space (`blockDim().x` for face columns, one fewer for center columns)
- `product_shape(shape1, shape2)` — the shape of `matrix1 * matrix2`, used for the product-row bounds check in `row_mul_mat!`

All shapes are singleton types, so everything still resolves at compile time.

One deliberate behavior nuance: in the two CenterToFace-product cases, the old outer bounds check (`v + pd + half < 1`) admitted product column 0, which lies outside the matrix. The unified slot-based check zeroes that entry instead, matching the CPU reference `multiply_matrix_at_index` in `src/MatrixFields/matrix_multiplication.jl`.

## Performance

**This change is performance neutral.** The unified helpers compile down to essentially the same machine code as the 12 hand-written methods: generated PTX is bit-identical for the same-staggered shape combinations (after normalizing session-dependent pointer constants and label numbering), and the mixed-staggering combinations differ only by register renumbering and equivalent bounds computations (e.g. `< blockDim` vs `<= blockDim - 1`), with equal or slightly lower register counts. Measured A100 kernel timings across all 12 shape combinations plus a chained product are unchanged (within run-to-run noise).

## Verification (on an A100)

- All `test/MatrixFields/matrix_fields_broadcasting` scripts (scalar 1–17, non-scalar 1–5) produce identical results before and after the refactor.
- Per-case GPU kernel timings across all 12 shape combinations plus a chained product are unchanged (within run-to-run noise).
- Generated PTX is bit-identical for the same-staggered shape combinations; the mixed-staggering combinations differ only by register renumbering and equivalent bounds computations (e.g. `< blockDim` vs `<= blockDim - 1`), with equal or slightly lower register counts.

https://claude.ai/code/session_011oirTEPPBEj5BWwhgYKMzn

### DIFF
--- a/ext/cuda/column_matrix_helpers.jl
+++ b/ext/cuda/column_matrix_helpers.jl
@@ -1,415 +1,92 @@
-# row_mul_mat! handles banded matrix * banded matrix. There are 8 methods, but they all have the
-# same structure, so we they could be written as a single method.
-# The others can be obtained by copy-pasting and changing the indices appropriately.
-# Note that these are all specialized for CUDA.blockDim().x faces , so the indices are hardcoded.
+# Helpers for multiplying a single row of a banded matrix with a banded matrix or vector
+# stored in shared memory. These are specialized for columns with CUDA.blockDim().x face
+# levels, computed by threads `v = 1:CUDA.blockDim().x` (one column per `threadIdx().y`).
+#
+
+# Number of valid slots in the column space of a matrix with the given shape.
+@inline n_column_slots(::Union{FaceToCenter, FaceToFace}) = CUDA.blockDim().x
+@inline n_column_slots(::Union{CenterToFace, CenterToCenter}) =
+    CUDA.blockDim().x - 1i32
+
+# Shape of `matrix1 * matrix2`: its rows match `matrix1`'s rows and its columns match
+# `matrix2`'s columns.
+@inline product_shape(
+    ::Union{FaceToCenter, CenterToCenter},
+    ::Union{CenterToFace, CenterToCenter},
+) = CenterToCenter()
+@inline product_shape(
+    ::Union{FaceToCenter, CenterToCenter},
+    ::Union{FaceToCenter, FaceToFace},
+) = FaceToCenter()
+@inline product_shape(
+    ::Union{CenterToFace, FaceToFace},
+    ::Union{CenterToFace, CenterToCenter},
+) = CenterToFace()
+@inline product_shape(
+    ::Union{CenterToFace, FaceToFace},
+    ::Union{FaceToCenter, FaceToFace},
+) = FaceToFace()
+
+# row_mul_mat! handles banded matrix * banded matrix. Entries of the product row whose
+# column index lies outside the product matrix are set to zero, matching
+# `multiply_matrix_at_index` in `src/MatrixFields/matrix_multiplication.jl`.
 Base.@propagate_inbounds function row_mul_mat!(
     ::Type{P},
     mat1_row,
     matrix2,
-    ::FaceToCenter,
-    ::CenterToFace,
+    shape1::MatrixFields.AbstractMatrixShape,
+    shape2::MatrixFields.AbstractMatrixShape,
 ) where {P}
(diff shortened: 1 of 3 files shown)


ext/cuda/column_matrix_helpers.jl
```diff
@@ -1,250 +1,73 @@
...
-    ld2, ud2 = MatrixFields.outer_diagonals(mat2_eltype)
-    pd1, pd2 = MatrixFields.outer_diagonals(prod_eltype)
+# `column_slot(v, d, shape)` is the shared-memory slot of the value that entry `d` of the
+# matrix row computed by thread `v` multiplies, for a matrix with the given `shape`: the row
+# index of thread `v` is `v` for center rows and `v - half` for face rows, the column index
+# is `row index + d`, and the slot of a column index is `index + half` for face columns and
+# `index` for center columns.
+Base.@propagate_inbounds column_slot(v, d, ::Union{CenterToCenter, FaceToFace}) = v + d
+Base.@propagate_inbounds column_slot(v, d, ::FaceToCenter) = v + d + half
+Base.@propagate_inbounds column_slot(v, d, ::CenterToFace) = v + d - half
```
dennisYatunin:
```suggestion
Base.@propagate_inbounds column_slot(v, d, shape) = MatrixFields.band_matrix_d(v + d, shape)
```

ext/cuda/column_matrix_helpers.jl
```diff
@@ -1,250 +1,73 @@
...
 
-    li = 1i32
-    ri = CUDA.blockDim().x
+# Number of valid slots in the column space of a matrix with the given shape.
+@inline n_column_slots(::Union{FaceToCenter, FaceToFace}) = CUDA.blockDim().x
+@inline n_column_slots(::Union{CenterToFace, CenterToCenter}) =
+    CUDA.blockDim().x - 1i32
```
dennisYatunin, line 9: This could instead be written as `MatrixFields.band_matrix_n_cols(nlevels(...), shape)`, but you can leave it as is if you prefer.

ext/cuda/column_matrix_helpers.jl
```diff
@@ -1,250 +1,73 @@
...
+    ld2, ud2 = MatrixFields.outer_diagonals(eltype(matrix2))
+    pd1, pd2 = MatrixFields.outer_diagonals(P)
+    prod_shape = product_shape(shape1, shape2)
+    mat2_offset = (block_col_idx - 1i32) * CUDA.blockDim().x
+    zero_entry = zero(eltype(P))
     prod_entries = UnrolledUtilities.unrolled_map((pd1:pd2...,)) do pd
-        if v + pd + half < li || v + pd + half > ri
-            zero_entry
+        prod_slot = column_slot(v, pd, prod_shape)
+        if prod_slot < 1i32 || prod_slot > n_column_slots(prod_shape)
```
dennisYatunin: It would be good to use the opposite condition here (`if 0i32 < prod_slot <= n_column_slots(prod_shape)`) for consistency with the equivalent conditions on lines 68 and 98.

dennisYatunin (approved):
This looks equivalent to the version in `src`, so it should be straightforward to get rid of the code duplication now. Thanks!

PR #2548, opened 2026-07-13
Remove columnwise! operator and its tests

### LINKED ISSUE #2520: `columnwise!` is unused outside of the tests
We can probably delete this

imreddyTeja:
## Summary
- Removes the `columnwise!` operator implementation (`src/Operators/columnwise.jl`) and its CUDA extension (`ext/cuda/operators_columnwise.jl`)
- Removes the associated unit test (`test/Operators/finitedifference/unit_columnwise.jl`)
- Removes the corresponding `include`s from `src/Operators/Operators.jl` and `ext/ClimaCoreCUDAExt.jl`
- Removes the "Unit: columnwise" and "Unit: gpu columnwise" buildkite CI jobs
closes #2520
## Test plan
- [x] `using ClimaCore` loads cleanly with no errors
- [x] Confirmed no remaining references to `columnwise!` or the deleted files elsewhere in the repo


dennisYatunin (approved):
I think we'll want something like `unit_columnwise.jl` as a performance test for kernel fusion, but if it's too much work to rewrite using `foreach_column` then just remove it for now.

imreddyTeja:
> I think we'll want something like `unit_columnwise.jl` as a performance test for kernel fusion, but if it's too much work to rewrite using `foreach_column` then just remove it for now.

I left the test and deleted the `columnwise!` usage. I removed it from the buildkite pipeline, but we can add it back once it is compatible with `foreach_column`

PR #2549, opened 2026-07-14
Standardize README and refresh copyright

tapios:
## Summary
- Reformat the badge table to the shared CliMA layout (Documentation / Version / License / Tests / Code Coverage / Downloads / DOI)
- Add stable-docs, JuliaHub version, and License badges; drop the separate "Docs Build" row; normalize the Downloads badge
- Rewrite the body (it was *very* stale): concise intro, feature list, runnable Quick Example, documentation links, ecosystem integration, and a Contributing section (dev-guides + AGENTS.md). Drops the outdated blurb and the WeakValueDicts legacy note.
- Refresh LICENSE and NOTICE copyright to `2021-2026 California Institute of Technology (Climate Modeling Alliance)`
- Adds new logo

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2021-2026 California Institute of Technology (Climate Modeling Alliance)
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/NOTICE
+++ b/NOTICE
@@ -1,4 +1,4 @@
-Copyright (c) 2022, the California Institute of Technology (Caltech)
+Copyright 2021-2026 California Institute of Technology (Climate Modeling Alliance)
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


imreddyTeja (approved):
Thank you. Is this something we need to do for other repos?

tapios:
> Thank you. Is this something we need to do for other repos?

Yes, that would be good. I've started going through a few. If you could take on some more, that would help.

imreddyTeja:
> > Thank you. Is this something we need to do for other repos?
> 
> Yes, that would be good. I've started going through a few. If you could take on some more, that would help.

Will do. Where is the "CliMA layout " for READMEs specified?

tapios:
> > > Thank you. Is this something we need to do for other repos?
> > 
> > 
> > Yes, that would be good. I've started going through a few. If you could take on some more, that would help.
> 
> Will do. Where is the "CliMA layout " for READMEs specified?

Just use what we have as a model (e.g., in ClimaCore, Thermodynamics, SurfaceFluxes, Rootsolvers, Insolation, ...). There's no hard-and-fast rule. For the table, some repos have GH CI, some also Buildkite. Some of DOIs, some don't. We don't have to be rigid. But we should strive for the READMEs (and the docs too) to become as helpful to new users and people who are just curious as possible.

Issue #2542, opened 2026-07-08
macOS GHA runners not running


ph-kev:
In https://github.com/CliMA/ClimaCore.jl/pull/2504, the GHA was changed so that only macOS runners run on commits on the main branch. Since unit tests are not ran on the main branch, this means that there are no tests that run on macOS.

tapios:
Closed by #2550.

PR #2550, opened 2026-07-14
Run Unit Tests on pushes to main and pin CI badge to main

tapios:
## Summary
- The `Unit Tests` workflow only triggered on `pull_request`, so the README CI badge showed the status of the latest PR run on any branch — currently a failing run from an unrelated PR branch.
- Add a `push` trigger for `main` so the workflow runs on the default branch, and move the `tags: '*'` filter to `push` (tag filters have no effect under `pull_request`).
- Pin the README badge and its link to `?branch=main` so it tracks main only.

### DIFF
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -1,10 +1,13 @@
 name: Unit Tests
 
 on:
-  pull_request:
+  push:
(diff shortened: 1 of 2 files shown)


imreddyTeja:
This closes #2542

PR #2556, opened 2026-07-26
Unify stong/weak forms with type parameter dispatch

tapios:
`Divergence`, `Gradient`, and `Curl` now take a second type parameter (`StrongForm`/`WeakForm`) that factors out the three form-dependent differences (transposed-and-negated derivative matrix, W-weighted input, J vs WJ rescale). The twelve strong/weak `apply_operator` methods reduce to six; `WeakDivergence`, `WeakGradient`, and `WeakCurl` become backwards-compatibility aliases collected in `src/Operators/deprecated.jl` for later removal.

* Results are bitwise identical to the previous implementations, except for the strong divergence (differs by order of operations)
* All public names and constructors unchanged

### DIFF
--- a/src/Operators/spectralelement.jl
+++ b/src/Operators/spectralelement.jl
@@ -112,7 +112,8 @@ computation; they differ only in three form-dependent factors:
  - whether the argument is weighted by the quadrature weights `W` or by the
    Jacobian factor; see `form_weighted_arg` and `form_jacobian`,
  - whether the result is rescaled by `J` or by `WJ`; see
-   `form_jacobian_rescale`.
+   `form_jacobian_rescale`, and by `W` or not at all; see
+   `form_weight_rescale`.
 
 Operators with strong/weak variants carry a `FormType` as their second type
 parameter, e.g. `Divergence{I, StrongForm}`, with the weak variant available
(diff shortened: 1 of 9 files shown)


tapios:
Needs #2557 before merging, to fix race condition in Datalayouts/scopes.jl.

src/Operators/spectralelement.jl
```diff
@@ -1155,14 +1068,28 @@ Note that unused dimensions will be dropped: e.g. the 2D curl of a
...
+# Form-dependent factors: the strong curl is εⁱʲᵏ J⁻¹ Dⱼ uₖ, while the weak
+# curl is -εⁱʲᵏ (WJ)⁻¹ Dⱼᵀ (W uₖ), with the transpose and sign of the
+# derivative matrix carried by form_deriv_entry.
+@inline curl_form_arg_scale(::StrongForm, local_geometry, x) = x
+@inline curl_form_arg_scale(::WeakForm, local_geometry, x) =
+    (local_geometry.WJ * local_geometry.invJ) * x
+@inline curl_form_rescale(::StrongForm, x, local_geometry) =
+    x * local_geometry.invJ
+@inline curl_form_rescale(::WeakForm, x, local_geometry) =
+    x / local_geometry.WJ
```
dennisYatunin: `curl_form_arg_scale` looks identical to `grad_form_arg_scale`, and `curl_form_rescale` looks identical to `div_form_rescale`. Maybe you could unify and rename these functions?

dennisYatunin (approved):
Thanks for simplifying this! I only have one small comment about code duplication; looks great otherwise.

src/Operators/spectralelement.jl
tapios: Good idea. Thank you!

PR #2557, opened 2026-07-27
Fix spurious "Nested loops over ThisThreadPool" error

tapios:
`jl_in_threaded_region` is a process-global flag, but `parallelize_over(f, ::ThisThreadPool)` read it twice. If an unrelated task entered a threaded region between the two reads, the first read said "not in a region" and the second said "in a region", leading to an error without actual nesting.

This affects any code that broadcasts `Field`s from several concurrent tasks. The `plane/inertial_gravity_wave` example does that via `ThreadsX.mapreduce`, and fails intermittently (PR #2556, build 7117).

### CPU Multithreading & Scope Fixes
- **Scope Resolution**: Loops resolve their scope once up front with `resolved_scope`, and thread rank comes from task-local storage (`:climacore_pool_threads`) rather than global state.
- **Thread Pool Partitioning**: Concurrent loops divide the CPU pool dynamically, each taking an equal share of threads instead of letting the first loop claim all threads. This is ~2.6x faster with two concurrent tasks on 8 threads.
- **Explicit Scope & Task Spawning**: Explicit `ThisThreadPool()` calls now resolve properly. Pool threads are spawned explicitly into `:default` with `@spawn :default` (protecting against execution inside `:interactive` tasks), and parallel worker errors are collected and thrown via `CompositeException`.

### GPU Performance Optimizations (`ClimaCoreCUDAExt`)
- **Device Attribute Caching**: Cache invariant CUDA device attributes (`MAX_THREADS_PER_BLOCK`, `WARP_SIZE`, `SM_COUNT`, etc.) to eliminate repeated C driver calls (`cuDeviceGetAttribute`) on every launch. Reduces host-side broadcast launch latency by ~54x (from ~477 μs down to ~8.8 μs per launch on an A100).
- **Occupancy Query Caching**: Cache invariant kernel launch configurations (`CUDA.launch_configuration`) in an `IdDict` to eliminate repeated C driver occupancy calculations (`cuOccupancyMaxPotentialBlockSize`) on every broadcast and reduction launch. Saves an additional ~2 μs of host CPU launch overhead per kernel call (~4.2% faster reductions).
- **Task-Local Reduction Buffers**: Intermediate GPU reduction arrays in `reduce_points` are cached in task-local storage to avoid allocating device memory (`similar`) during repeated scalar reductions.
- **Warp Alignment**: Ensure thread block sizes for slice loops are rounded to warp multiples (`32`) to prevent warp divergence.

### CI & Workflow Improvements
- **Documentation Deploy Concurrency**: Added a `concurrency` group (`documentation-deploy`) to `.github/workflows/docs.yml` to queue documentation preview deployments sequentially, preventing git `--force-with-lease` rejections (`! [rejected] HEAD -> gh-pages (stale info)`) when multiple CI jobs run concurrently.

### DIFF
--- a/ext/cuda/cuda_utils.jl
+++ b/ext/cuda/cuda_utils.jl
@@ -4,6 +4,27 @@ import ClimaCore.DataLayouts
 
 const reported_stats = Dict()
 const kernel_names = IdDict()
+# CUDA.launch_configuration runs an occupancy calculation, which costs far more than the
+# rest of a launch, and its result only depends on the kernel. The blocks and threads are
+# copied into a concrete type of our own, so that the cache stays type stable no matter what
+# CUDA.launch_configuration returns, and so that a lookup does not dynamically dispatch.
+const LaunchConfiguration = @NamedTuple{blocks::Int, threads::Int}
+const OCCUPANCY_CACHE = IdDict{Any, LaunchConfiguration}()
+const OCCUPANCY_CACHE_LOCK = ReentrantLock()
+
+# The cache is only read and written while the lock is held, since an IdDict cannot be read
+# while it is being rehashed. The occupancy calculation runs under the lock as well: it only
+# happens once per kernel, and holding the lock keeps two threads from both running it.
+cached_launch_configuration(fun) =
+    lock(OCCUPANCY_CACHE_LOCK) do
+        get!(OCCUPANCY_CACHE, fun) do
+            config = CUDA.launch_configuration(fun)
+            LaunchConfiguration((
+                Int(config.blocks),
+                Int(config.threads),
+            ))
+        end
+    end::LaunchConfiguration
 
 collect_kernel_stats() = false
 
@@ -172,7 +193,7 @@ function auto_launch!(
             # Note: `name = nothing` here will revert to default behavior
             kernel = CUDA.@cuda name = kernel_name always_inline = true launch =
                 false f!(args...)
-            config = CUDA.launch_configuration(kernel.fun)
+            config = cached_launch_configuration(kernel.fun)
             threads = min(nitems, config.threads)
             blocks = cld(nitems, threads)
             kernel(args...; threads, blocks) # This knows to use always_inline from above.
@@ -189,7 +210,7 @@ function auto_launch!(
         # occursin("single_field_solve_kernel", string(nameof(F!))) || return nothing
         if !haskey(reported_stats, key)
             kernel = CUDA.@cuda always_inline = true launch = false f!(args...)
-            config = CUDA.launch_configuration(kernel.fun)
+            config = cached_launch_configuration(kernel.fun)
             threads = isnothing(nitems) ? nothing : min(nitems, config.threads)
             blocks = isnothing(nitems) ? nothing : cld(nitems, threads)
             # For now, let's just collect info, later we can benchmark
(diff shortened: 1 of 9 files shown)


src/DataLayouts/scopes.jl
```diff
@@ -125,6 +125,20 @@ of this instruction is not necessarily parallelized over all available threads.
...
+[`DataScope`](@ref) that will run a loop over `scope`. This is smaller than `scope` when
+`scope` cannot be parallelized over at the moment, such as when a CPU thread pool is
+already running a loop launched somewhere else.
+
+Loops must resolve their scope before reading [`num_threads`](@ref) or
+[`thread_rank`](@ref), and must then use the resolved scope everywhere, so that the number
+of threads used to divide up a loop's indices cannot change while the loop is running. By
+default, every scope is already resolved.
+"""
+resolved_scope(scope) = scope
```
dennisYatunin: This is a pretty important correction to the interface; might be good to put it earlier in the file and mention it in the `DataScope` docstring. It should definitely be put before `num_threads` and `thread_rank`, since it needs to be called before them.

ext/cuda/loops.jl
```diff
@@ -7,7 +33,14 @@ function DataLayouts.foreach_slice(::ThisHost, op::O, f::F, args...; kwargs...)
...
         max_slice_points = maximum(Base.Fix1(DataLayouts.num_slice_points, op), args)
-        threads = min(threads_via_occupancy(kernel, args), max_slice_points)
+        unaligned_threads = min(threads_via_occupancy(kernel, args), max_slice_points)
+        # Round down to a whole number of warps, so that no warp is only partly filled.
+        # Rounding up instead would launch more threads than a slice has points, which
+        # breaks the assumption that every thread of a reduction has a value to reduce.
+        threads =
+            unaligned_threads > THREADS_PER_WARP ?
+            fld(unaligned_threads, THREADS_PER_WARP) * THREADS_PER_WARP :
+            unaligned_threads
```
dennisYatunin: While the value of `num_threads` for a sub-block (e.g., a warp) can overcount the number of launched threads, the value actually used in `reduce_points` and `shuffle_reduce` is `num_active_threads`, which does not get rounded up to the nearest warp. So it should be safe for the kernel launcher to internally round up the value of `unaligned_threads`, since the extra threads don't participate in any reduction operations.

Was there some test you ran that showed a flaw in this reasoning?

ext/cuda/loops.jl
```diff
@@ -1,3 +1,29 @@
+# View of a per-task buffer that holds each block's result during a two-stage reduction.
+# Allocating a new device array for every reduction is slower than the reduction itself, so
+# one buffer per element type is kept for the lifetime of the task that reduces. The buffer
+# only ever grows, and reductions on a task run one after another, so no two live views of
+# it overlap.
+function task_reduction_buffer(::Type{T}, required_length::Int) where {T}
```
dennisYatunin: This is a great optimization, and I don't see why the same argument wouldn't also apply to CPU arrays. Any chance you could generalize this to all `:climacore_reduction_buffers` and move it into `src`?

Here's one way to generalize it: add an optional `buffer` flag to `scoped_array` (default value `false`), and use this `task_local_storage` implementation when the flag is `true`. Then, you could just change `empty_results = DataLayouts.scoped_array(ThisHost(), T, 0)` in this file and `results = scoped_array(scope, T, num_threads(scope))` in `src` to use `scoped_array(...; buffer = true)`, and not need to add any GPU-specific behavior.

ext/cuda/cuda_utils.jl
```diff
@@ -232,10 +247,50 @@ end
...
+    threads_per_warp = 0,
+    max_threads_per_SM = 0,
+    SM_count = 0,
+)
+const CACHED_DEVICE_ATTRIBUTES =
+    Ref{typeof(UNCACHED_DEVICE_ATTRIBUTES)}(UNCACHED_DEVICE_ATTRIBUTES)
+
+function cached_device_attributes()
+    attributes = CACHED_DEVICE_ATTRIBUTES[]
+    all(!iszero, values(attributes)) && return attributes
```
dennisYatunin: Would this be any simpler if you gated it with a lock instead of `all(!iszero, values(attributes))`?

src/DataLayouts/loops.jl
```diff
@@ -75,11 +75,37 @@ argument. A [`DataMask`](@ref), which by default is set to [`NoMask`](@ref), may
 also be used to skip over a particular subset of slices.
 """
 @inline foreach_slice(op::O, f::F, args...; mask = NoMask()) where {O, F} =
-    foreach_slice(DataScope(args...), op, f, args...; mask)
+    foreach_resolved_slice(DataScope(args...), op, f, args...; mask)
```
dennisYatunin: Instead of defining a new `foreach_resolved_slice` function and specializing on `ThisThreadPool`, I think you can just add a method of `foreach_slice` for `ThisThreadPool`. Same for `reduce_points`/`reduce_resolved_points`. Would be good to avoid more intermediate functions if possible.

dennisYatunin (approved):

src/DataLayouts/scopes.jl
tapios: - Added `resolved_scope` to the list of `DataScope` methods in the main docstring.
- Moved `resolved_scope(scope) = scope` and its docstring to appear before `partition` (and before `num_threads` / `thread_rank`).

ext/cuda/loops.jl
tapios: Rounded warp sizes up with `cld` in kernel launcher. Also updated the comment to explain that because sub-block reductions rely on `num_active_threads`, rounding up to complete warps is safe and ensures no warp is only partially populated.

ext/cuda/loops.jl
tapios: Good point. Generalized reduction buffer caching to CPU and GPU arrays: Moved `task_reduction_buffer` into `src/DataLayouts/scopes.jl` and generalized it to store buffers for any array type `A` and element type `T`. Also added an optional `buffer = false` kwarg to `scoped_array` and replaced explicit calls to `task_reduction_buffer` by calls to `DataLayouts.scoped_array(ThisHost(), T, ...; buffer = true)`.

ext/cuda/cuda_utils.jl
tapios: Yes, that works. Implemented.

src/DataLayouts/loops.jl
tapios: I like this. Eliminated `foreach_resolved_slice` and `reduce_resolved_points` in favor of the suggested approach.

src/DataLayouts/loops.jl
```diff
@@ -77,9 +79,35 @@ also be used to skip over a particular subset of slices.
 @inline foreach_slice(op::O, f::F, args...; mask = NoMask()) where {O, F} =
     foreach_slice(DataScope(args...), op, f, args...; mask)
 
+# A thread pool has to be resolved before looping over it, and given back afterward.
+@inline function foreach_slice(
+    pool::ThisThreadPool,
+    op::O,
+    f::F,
+    args...;
+    mask = NoMask(),
```
dennisYatunin: `foreach_slice` already sets `mask = NoMask()` in its top-level method. There shouldn't be a need to set `mask = NoMask()` in every other method like this.

ext/cuda/loops.jl
tapios: I get it. I reverted this change back to main.

src/DataLayouts/loops.jl
tapios: Ok, removed default.

tapios: Also updated docstring to be clearer.

PR #2558, opened 2026-07-28
Fix tail effect on GPUs

dennisYatunin:
This PR extends `max_resident_blocks` (renamed to `max_active_blocks`) so that it accounts for register pressure and shared memory requirements, fixing the "tail effect" problem that was introduced in #2522 for 64-level pointwise kernels (see [this comment](https://github.com/CliMA/ClimaCore.jl/pull/2522#issuecomment-5063662876) for details). This PR also improves the launch configuration caching, leading to a 5%--20% drop in kernel launch latency, and it exposes all occupancy limits that the CUDA API can identify internally.

### DIFF
--- a/ext/cuda/cuda_utils.jl
+++ b/ext/cuda/cuda_utils.jl
@@ -1,30 +1,275 @@
 import CUDA
 import ClimaCore.Fields
 import ClimaCore.DataLayouts
+import ClimaCore.Utilities
+
+function uncached_device_attributes()
+    device = CUDA.device()
+    get_attr(code) = Int(CUDA.attribute(device, code))
+    sm_version = (
+        get_attr(CUDA.DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR),
+        get_attr(CUDA.DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR),
+    )
+    sm_version[1] > 12 && throw(ArgumentError("Missing device attributes for \
+                                               Compute Capability $(sm_version[1])"))
+    return (;
+        threads_per_warp = get_attr(CUDA.DEVICE_ATTRIBUTE_WARP_SIZE),
+
+        # kernel launch configuration limits
+        max_block_dim_x = get_attr(CUDA.DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_X),
+        max_block_dim_y = get_attr(CUDA.DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Y),
+        max_block_dim_z = get_attr(CUDA.DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Z),
+        max_grid_dim_x = get_attr(CUDA.DEVICE_ATTRIBUTE_MAX_GRID_DIM_X),
+        max_grid_dim_y = get_attr(CUDA.DEVICE_ATTRIBUTE_MAX_GRID_DIM_Y),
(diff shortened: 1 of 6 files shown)


imreddyTeja:
This does fix the tail effect, but it does not give the 50% speedup that ncu thought it would. It still says the average active warps increased by 1. It does give a small speed up. I'm a bit concerned about how this will impact ClimaLand though. I'd guess the extra dictionary usage might impact latency

ext/cuda/cuda_utils.jl
```diff
@@ -1,30 +1,245 @@
...
+Collection of hardware properties from the CUDA API and CUDA Runtime Headers,
+obtained through calls to `CUDA.attribute`.
+
+Attributes are obtained by querying the C driver, which adds measurable latency
+if done more than once, so all attributes are cached after the first call. An
+`Atomic` initialization flag is used to ensure that every thread either sees a
+fully written cache or sets the cache itself. Any thread can set the cache, as
+ClimaComms only targets one device from threads launched by the same process.
+"""
+function device_attributes()
```
imreddyTeja, line 71: Would it be useful to set the device attributes in an __init__ that runs during extension loading?

ext/cuda/cuda_utils.jl
```diff
@@ -1,30 +1,266 @@
...
+
+Computing a configuration requires querying the C driver for the kernel's
+register pressure and shared memory requirements, which adds measurable latency
+if done more than once per kernel, so the result is cached in an `IdDict`. A
+lock is used to prevent multiple threads from updating the cache simultaneously,
+since an `IdDict` should never be read as it is being rehashed.
+"""
+function launch_configuration(f::F, args, config_args...; strict = true) where {F}
+    cu_function = (CUDA.@cuda always_inline = true launch = false f(args...)).fun
+    lock(LAUNCH_CONFIGURATION_CACHE_LOCK)
```
imreddyTeja, line 255: why is this needed?

ext/cuda/cuda_utils.jl
```diff
@@ -1,30 +1,266 @@
...
+        return get!(LAUNCH_CONFIGURATION_CACHE, (cu_function, config_args..., strict)) do
+            uncached_launch_configuration(cu_function, config_args...; strict)
+        end
+    finally
+        unlock(LAUNCH_CONFIGURATION_CACHE_LOCK)
+    end
+end
+
+threads_via_occupancy(f::F, args) where {F} = launch_configuration(f, args).threads
+# function config_via_occupancy(f::F, nitems, args) where {F}
```
imreddyTeja: this can be deleted

imreddyTeja (approved):

ext/cuda/cuda_utils.jl
dennisYatunin: This would prevent us from running CUDA compilation tests on CPUs that are not connected to any GPUs, which the current version allows (`unchached_device_attributes` has an inferable return type regardless of whether a GPU is available).

ext/cuda/cuda_utils.jl
dennisYatunin: `Dict` and `IdDict` are not thread-safe, so if multiple threads are running ClimaCore in parallel and call `launch_configuration` at the same time, we need a lock to avoid race conditions (similar to [`ThreadSafeDicts.jl`](https://github.com/wherrera10/ThreadSafeDicts.jl/blob/master/src/ThreadSafeDicts.jl)).

Issue #2554, opened 2026-07-23
Dycore toward v1


tapios:
# ClimaCore Refactoring Plan, Part II: Toward v1.0 (follow-up on #2468)

## Purpose

#2468 laid out a six-phase refactoring plan. Phase 1 (DataLayouts unification) is complete in #2522: 14 layout types → 4, `DataLayout <: AbstractArray` with universal `(v, i, j, h)` indexing and a device-generic `DataScope`/loop abstraction replacing the per-layout CPU/CUDA kernel families.

This issue tracks the remaining phases and defines what "ClimaCore v1.0" means.

## v1.0 release criteria

1. **API**: exported surface finalized (post Phase 4 alias cleanup); deprecation shims removed or documented; no exported name we are not prepared to support through v1.x.
2. **Correctness**: GPU numerical-correctness tests (CPU-vs-GPU comparison) for all operator families; MPI DSS tests at ≥2 ranks pass on CPU and GPU.
3. **Performance**: performance benchmarks at parity or better vs. the last v0.14.x release; allocation budgets restored to 0 or tracked as explicit issues.
4. **Docs**: Phase 7 complete, including a migration guide from v0.14.x.
5. **Downstream**: ClimaAtmos, ClimaLand, ClimaCoupler, ClimaDiagnostics, ClimaUtilities, KinematicDriver pass against a v1.0.

## Phase 1 carry-over (from the #2522 review) [@dennisYatunin]

After #2522 merges:

- [x] Allow fusion of Field broadcasts on top of DataLayout broadcasts [PR](https://github.com/Clima/ClimaCore.jl/pull/2571).
- [x] New release and downstream compat PRs: ClimaAtmos `test/grids.jl`, ClimaUtilities `TempestRegridderExt`, ClimaLand `variable_types.jl`.
- [x] Restore compressed-map masked execution for generic + fused broadcasts on GPU (inactive columns currently occupy threads; fused broadcasts ignore masks) + a masked-space GPU benchmark.

---

## Phase 2: Finite Difference Operators — Stencil Abstraction

**Target**: `src/Operators/finitedifference.jl`, `ext/cuda/operators_finite_difference.jl`, `ext/cuda/operators_fd_*`, `ext/cuda/column_matrix_helpers.jl` (6,390 lines → ~3,000 lines)

### Problem

53 operator struct types, 80+ boundary-condition method implementations. The dominant duplication pattern:

[...]

PR #2562, opened 2026-07-31
Release v0.15.0

dennisYatunin:
Bump ClimaCore to v0.15.0, and bump the patch versions of all `lib` packages after updating their `compat` entries. This release includes the DataLayouts API refactor, which changes all field layouts to `VIJHWithF` and enables pointwise kernel fusion.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -24,12 +24,13 @@ jobs:
           version: '1.10'
       - name: Install dependencies
         run: >
-          julia --project=docs -e 'using Pkg; Pkg.develop(path=".");
+          julia --project=docs -e 'using Pkg;
           Pkg.develop(path="lib/ClimaCoreVTK");
           Pkg.develop(path="lib/ClimaCoreTempestRemap");
           Pkg.develop(path="lib/ClimaCoreMakie");
           Pkg.develop(path="lib/ClimaCorePlots");
           Pkg.develop(path="lib/ClimaCoreSpectra");
+          Pkg.develop(path=".");
           Pkg.instantiate(;verbose=true)'
       - name: Build and deploy
         env:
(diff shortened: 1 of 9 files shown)


nefrathenrici (approved):

.github/workflows/docs.yml
```diff
@@ -24,12 +24,13 @@ jobs:
...
         run: >
-          julia --project=docs -e 'using Pkg; Pkg.develop(path=".");
+          julia --project=docs -e 'using Pkg;
           Pkg.develop(path="lib/ClimaCoreVTK");
           Pkg.develop(path="lib/ClimaCoreTempestRemap");
           Pkg.develop(path="lib/ClimaCoreMakie");
           Pkg.develop(path="lib/ClimaCorePlots");
           Pkg.develop(path="lib/ClimaCoreSpectra");
+          Pkg.develop(path=".");
           Pkg.instantiate(;verbose=true)'
```
ph-kev, line 34: You can develop all the packages at the same time instead of doing it one by one since you can pass a vector of strings to `Pkg.develop`.

imreddyTeja (approved):
It might be good to highlight the breaking changes in NEWS.md or the release notes

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ClimaCore"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
 authors = ["CliMA Contributors <clima-software@caltech.edu>"]
-version = "0.14.55"
+version = "0.15.0"
```
ph-kev, line 5: If a breaking release is going to made, do we need to keep the files for keeping backward functionality around?

dennisYatunin: We're planning to remove those during Phase 4 of #2554. They are kept for now to avoid requiring too many sudden changes in downstream packages.

.github/workflows/docs.yml
dennisYatunin: I tried passing them as a vector of strings, but couldn't get it to work. Calling `Pkg.develop(["lib/ClimaCoreVTK", ..., "."])` threw an error where it suggested I use `path = ...`, but calling `Pkg.develop(path = ["lib/ClimaCoreVTK", ..., "."])` threw another error where it said I can't pass a vector to `path`. But let me know if there's some other way I should do this.

PR #2565, opened 2026-08-04
Adds `Fields.fieldvector2array!` and `Fields.array2fieldvector!`

tapios:
## Purpose

Enable `fieldvectors` in Krylov workspaces

## Changes 

-  Added `Fields.fieldvector2array!` and `Fields.array2fieldvector!`,
  allocation-free, GPU-safe block-wise copies between a `FieldVector` and a
  flat vector. Together with `Krylov.ktypeof(::FieldVector)` (defined in
  `KrylovExt`), these let iterative Krylov solvers keep their workspace vectors on the
  GPU while model code continues to operate on `FieldVector`s.
- Fixed `zero(::FieldVector)` to preserve scalar (`ScalarWrapper`) components
  instead of turning them into 0-dimensional `Array`s, and fixed
  `ClimaComms.array_type(::FieldVector)` (and hence
  `Krylov.ktypeof(::FieldVector)`) for `FieldVector`s with plain-array or
  scalar components, which previously threw a `MethodError`.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -415,25 +415,22 @@ end
 
 """
     fieldvector2array!(array, fv)
-    array2fieldvector!(fv, array)
 
-Copy between a `FieldVector` `fv` and a flat `AbstractVector` `array` of the
-same length, without allocating or scalar indexing: each component block is
-copied with a single array-level `copyto!`, so `FieldVector`s backed by GPU
-arrays are supported (including mixed cases, where the flat array and some
-components live on different devices). Entries are ordered as in the
-`FieldVector`'s own linear indexing: component blocks in order, each in the
-linear order of its backing array.
+Copy the entries of the `FieldVector` `fv` into the flat `AbstractVector`
+`array` of the same length, without allocating or scalar indexing: each
+component block is copied with a single array-level `copyto!`, so
+`FieldVector`s backed by GPU arrays are supported (including mixed cases,
+where `array` and some components live on different devices). Entries are
+ordered as in the `FieldVector`'s own linear indexing: component blocks in
+order, each in the linear order of its backing array.
 
 Scalar (`ScalarWrapper`) components are written to `array` with a `fill!` on a
-one-element view, which is GPU-safe; copying them back out of `array` requires
-a scalar read, so `array2fieldvector!` only supports scalar components when
-`array` is a CPU array (a GPU-backed `array` throws a scalar-indexing error
-rather than performing a hidden synchronizing transfer).
+one-element view, which is GPU-safe.
 
 Intended for interfacing with libraries that operate on flat vectors, such as
 the Krylov.jl workspace vectors given by `Krylov.ktypeof(::FieldVector)` (see
-`KrylovExt`).
+`KrylovExt`). See [`array2fieldvector!`](@ref) for the inverse copy and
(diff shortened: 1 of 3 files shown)


src/Fields/fieldvector.jl
```diff
@@ -409,10 +413,103 @@ function LinearAlgebra.norm(x::FieldVector)
     sqrt(LinearAlgebra.norm_sqr(x))
 end
 
+"""
+    fieldvector2array!(array, fv)
+    array2fieldvector!(fv, array)
+
```
imreddyTeja, line 418: Nit: I think it would be good to make separate docstrings for both of these. 

There is no documentation for field vectors in the public docs at the moment, but splitting the doctoring up will make it easier to add later

src/Fields/fieldvector.jl
```diff
@@ -409,10 +413,103 @@ function LinearAlgebra.norm(x::FieldVector)
...
+one-element view, which is GPU-safe; copying them back out of `array` requires
+a scalar read, so `array2fieldvector!` only supports scalar components when
+`array` is a CPU array (a GPU-backed `array` throws a scalar-indexing error
+rather than performing a hidden synchronizing transfer).
+
+Intended for interfacing with libraries that operate on flat vectors, such as
+the Krylov.jl workspace vectors given by `Krylov.ktypeof(::FieldVector)` (see
+`KrylovExt`).
+"""
+function fieldvector2array!(array::AbstractVector, fv::FieldVector)
```
imreddyTeja, line 435: Would adding the allocating versions of these functions be useful to anyone? If so, it should be easy to add allocating versions that just call the in place versions

src/Fields/fieldvector.jl
```diff
@@ -409,10 +413,103 @@ function LinearAlgebra.norm(x::FieldVector)
...
+end
+
 import ClimaComms
 
-ClimaComms.array_type(x::FieldVector) =
-    promote_type(unrolled_map(ClimaComms.array_type, _values(x))...)
+function ClimaComms.array_type(x::FieldVector)
+    # ScalarWrapper components hold CPU scalars regardless of where the other
+    # components live, so they do not participate in the promotion.
+    arrays = unrolled_filter(!Base.Fix2(isa, ScalarWrapper), Tuple(_values(x)))
```
imreddyTeja: I think this might need to be recursive check. 
`Y = ClimaCore.Fields.FieldVector(f = f, c = (a = 1.0, b = 2.0));`
would trype to promote (Array, CuArray), which does not return a CuArray

imreddyTeja:
```
julia> Y = ClimaCore.Fields.FieldVector(f = f_field, c = (a = 1.0, b = 2.0));

julia> ClimaComms.array_type(f_field)
CUDA.CuArray

julia> ClimaComms.array_type(Y)
DenseArray
```

src/Fields/fieldvector.jl
tapios: Done.

src/Fields/fieldvector.jl
tapios: Good point. Fixed.

src/Fields/fieldvector.jl
tapios: Good idea. Done.

imreddyTeja (approved):

PR #2567, opened 2026-08-05
Simplify FieldVector flat-array conversion helpers

tapios:
Follow-on on #2565:

Replace the hand-rolled first/tail recursion in `_blocks2array!` and `_array2blocks!` with `unrolled_reduce` folds over the entry offset, and give `_array2block(s)!` the same `(array, offset, block)` argument order as `_block(s)2array!`. Unwrap wrapped arrays (e.g. views) through parent in `_array_type` instead of reporting the wrapper type.

- Fixes `_array_type` for wrapped arrays: views/reshapes now report the underlying storage type (e.g. CuArray) instead of the wrapper (SubArray → promotes to AbstractArray).
- Makes `length(::FieldVector)` allocation-free: the `AbstractArray` fallback computes length through `axes`, whose `blockedrange` is not inferrable for nested `FieldVectors` and allocated on every call. Conversions are now allocation-free for nested and scalar components (previously a few hundred bytes per call).
- Removes an inference cliff in the converters at large component counts: the hand-rolled recursion allocates ~20 KB per call at 64 same-type components, while the unrolled_reduce form stays allocation-free; both are allocation-free at realistic sizes.

### DIFF
--- a/src/Fields/fieldvector.jl
+++ b/src/Fields/fieldvector.jl
@@ -104,6 +104,16 @@ BlockArrays.blockaxes(fv::FieldVector) =
 Base.axes(fv::FieldVector) =
     (BlockArrays.blockedrange(map(length ∘ backing_array, Tuple(_values(fv)))),)
 
+# The AbstractArray fallback computes length from axes, whose blockedrange is
+# not inferrable for nested FieldVectors and allocates on every call; sum the
+# block lengths directly instead (length recurses into nested FieldVectors,
+# whose backing_array is the FieldVector itself).
+Base.length(fv::FieldVector) = unrolled_reduce(
+    (n, value) -> n + length(backing_array(value)),
+    Tuple(_values(fv)),
+    0,
+)
+
 Base.@propagate_inbounds Base.getindex(
     fv::FieldVector,
     block::BlockArrays.Block{1},
@@ -464,7 +474,7 @@ function array2fieldvector!(fv::FieldVector, array::AbstractVector)
              length $(length(fv))",
         ),
     )
-    _array2blocks!(Tuple(_values(fv)), array, 0)
+    _array2blocks!(array, 0, Tuple(_values(fv)))
     return fv
 end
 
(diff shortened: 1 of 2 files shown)


imreddyTeja:
I'm still seeing allocations:
```
julia> Y = ClimaCore.Fields.FieldVector(f = (a = f_field, b = f_field), c = (a = 1.0, b = 2.0));

julia> function wrap(array, Y)
       ClimaCore.Fields.fieldvector2array!(array, Y)
       return
       end
wrap (generic function with 1 method)

julia> array = ones(FT, length(Y));

julia> @allocated wrap(array, Y)
57016

julia> @allocated wrap(array, Y)
320

julia> Y = ClimaCore.Fields.FieldVector(f = (a = f_field, b = f_field));

julia> array = ones(FT, length(Y));

julia> @allocated wrap(array, Y)
1486512

julia> @allocated wrap(array, Y)
0
```

tapios:
> I'm still seeing allocations:
> 
> ```
> julia> Y = ClimaCore.Fields.FieldVector(f = (a = f_field, b = f_field), c = (a = 1.0, b = 2.0));
> 
> julia> function wrap(array, Y)
>        ClimaCore.Fields.fieldvector2array!(array, Y)
>        return
>        end
> wrap (generic function with 1 method)
> 
> julia> array = ones(FT, length(Y));
> 
> julia> @allocated wrap(array, Y)
> 57016
> 
> julia> @allocated wrap(array, Y)
> 320
> 
> julia> Y = ClimaCore.Fields.FieldVector(f = (a = f_field, b = f_field));
> 
> julia> array = ones(FT, length(Y));
> 
> julia> @allocated wrap(array, Y)
> 1486512
> 
> julia> @allocated wrap(array, Y)
> 0
> ```

Good catch. It was a pre-existing problem on main: `FieldVector` had no length method, so the `AbstractArray` fallback computes it through `axes`. BlockArrays.blockedrange(map(length ∘ backing_array, ...))` then is uninferrable with nested components. 

Fix is  a dedicated `Base.length(::FieldVector)` method in `fieldvector.jl`, now added.

(`axes(fv)` itself still allocates for nested `FieldVectors`--should probably be fixed separately.)

imreddyTeja (approved):

PR #2568, opened 2026-08-05
Optimize GPU DataLayout indexing with linear striding and singleton broadcasting

tapios:
### Summary

Improve performance of pointwise kernels by improving or eliminating per-point index
arithmetic and broadcast error paths that can trap on the GPU.

- **Linear multi-slab striding**: single-field property views (`data.ρ`) become linearly
  indexable via a constant-stride formula (one integer division per point instead of a full
  Cartesian decomposition), with allocation-free struct access in kernels.
- **Singleton broadcasting in point loops**: 0-dimensional and singleton-dimension layouts
  (e.g. surface fields) now broadcast like ordinary expressions. `each_slice_index` no longer 
  contains a kernel-reachable throw (the previous `DimensionMismatch` compiled into a device 
  trap that surfaced as `CUDA error: misaligned address (716)`). Genuine mismatches are rejected on the host. `IndexStyle` of a broadcast expression is linear only when its layouts share a shape,
  fixing silently wrong reductions over mixed-shape lazy expressions.
- **Linear point views**: `view_point_struct` reuses `array_and_index_args`, so the views
  built at every point of every broadcast argument use strided linear indices instead of
  converting linear → Cartesian (3 divisions per argument per point).
- **Single-launch reductions**: `reduce_points` on the GPU is one kernel launch using the
  last-block-done pattern, with the block results folded in registers by one warp (no
  second launch, no shared memory, so arbitrarily wide element types reduce at full
  bandwidth (`NTuple{16, Float64}`: +32%)). CPU reductions keep their documented pairwise
  accuracy and `@simd` vectorization.
- **Compacted masked loops**: `IJHMask` point and column loops iterate precomputed active
  indices instead of filtering every index against the mask, with coalesced access on GPUs.
  Masked launches are configured from unmasked upper bounds, so they no longer synchronize
  the stream to read the active count.

Adds `test/DataLayouts/benchmark_loops.jl` (CPU + GPU pipeline steps) covering broadcasts
and reductions over layouts, property views, and equal-/mixed-shape lazy expressions, plus
unit tests for singleton broadcasting and the host-side mismatch error.

CPU performance is at parity or better (broadcasts ~1.7× faster from unrolled slice
construction; mixed-shape lazy reductions ~25% faster); results are unchanged except where
mixed-shape reductions were previously wrong. Modest GPU performance gains.

---
## GPU Microbenchmarks: `ts/tail_effect` vs `main`

**Hardware**: NVIDIA A100-SXM4-40GB · **Julia**: 1.11.9  
**Prognostic size**: `Nv = 63, Nq = 4, Nh = 5400` (~5.4M points)

*(Updated with `max_waves=1` explicitly applied)*

### Pointwise kernel bandwidth (`dest .= a .+ 2 .* b`)

| Layout | `main` | This PR | Change |
|:---|:---:|:---:|:---:|
| 1-component `VIJFH` | 122.4 µs (994 GB/s) | **118.3 µs (1029 GB/s)** | +3.5% faster |
| Property-view `VIJFH` | 126.4 µs (963 GB/s) | **122.0 µs (997 GB/s)** | +3.6% faster |
| Raw `CuArray` reference | 119.4 µs (1019 GB/s) | 120.2 µs (1012 GB/s) | — |
| 1-comp overhead vs `CuArray` | 1.02× | **0.98×** | overhead eliminated |
| Prop-view overhead vs `CuArray` | 1.06× | **1.02×** | −4 pp |

### Scalar reduction (`sum(field)`)

| Domain size | `main` | This PR | Speedup |
|:---|:---:|:---:|:---:|
| `Nh = 50` (~50k points) | **47.8 µs** | 49.6 µs | 0.96× |
| `Nh = 500` (~504k points) | 55.1 µs | 55.0 µs | par |
| `Nh = 5400` (~5.4M points) | 90.1 µs | **87.5 µs** | **1.03×** |

Property-view reductions (`sum(field.a)`):

| Domain size | `main` | This PR | Speedup |
|:---|:---:|:---:|:---:|
| `Nh = 50` | **48.4 µs** | 49.7 µs | 0.97× |
| `Nh = 500` | 58.5 µs | **54.8 µs** | **1.07×** |
| `Nh = 5400` | 95.5 µs | **92.9 µs** | **1.03×** |

### Wide multi-component reduction bandwidth

`sum` of `NTuple{N, Float64}` at prognostic size. A100 `CuArray` ceiling ≈ 1019 GB/s.

| Eltype | `main` | This PR | Change |
|:---|:---:|:---:|:---:|
| `NTuple{1}` | 444 GB/s | **450 GB/s** | +1% |
| `NTuple{2}` | 617 GB/s | **670 GB/s** | **+9%** |
| `NTuple{4}` | 768 GB/s | **890 GB/s** | **+16%** |
| `NTuple{8}` | 936 GB/s | **1039 GB/s** | **+11%** |
| `NTuple{16}` | 1023 GB/s | **1119 GB/s** | **+9%** |

### Multi-component reduction time vs. raw `CUDA.mapreduce`

| Components | `main` | This PR | vs. `main` | vs. `CUDA.mapreduce` |
|:---|:---:|:---:|:---:|:---:|
| `Nf = 1` | 92.2 µs | **89.9 µs** | **1.03×** | 1.86× |
| `Nf = 2` | 131.8 µs | **122.1 µs** | **1.08×** | 2.22× |
| `Nf = 4` | 211.2 µs | **183.2 µs** | **1.15×** | 2.60× |

(`CUDA.mapreduce` on raw parent: ~167.4 µs, ~270.7 µs, ~476.0 µs for `Nf` = 1, 2, 4)

### Lazy broadcast reductions

| Expression | `main` | This PR | Speedup |
|:---|:---:|:---:|:---:|
| `sum(a .+ b)` (equal shape) | 145.2 µs | **129.1 µs** | **1.12×** |
| `sum(a .+ surface)` (mixed shape) | 197.8 µs | **187.2 µs** | **1.06×** |

### Host-side kernel launch latency (tiny problem, isolates dispatch cost)

| Kernel | `main` | This PR | Notes |
|:---|:---:|:---:|:---|
| Small broadcast | 18.1 µs | 18.7 µs | par |
| Small fill | 16.3 µs | 17.6 µs | par |
| Small `sum` (first call) | **39.3 µs** | 46.1 µs | +7 µs: task-local sync-counter init |
| Raw `CuArray` broadcast | 17.9 µs | 18.6 µs | par |

### Register Count Reduction (via CUDA.registers)

By simplifying the struct_index / parent_offset accessors and using CompactDeviceView, we successfully reduced register pressure in the primary GPU kernels.

main: 72 – 80 registers
branch: 64 registers
Reduction kernels also dropped from 46 to 40 registers.

ClimaAtmos with config/perf_configs/bm_aquaplanet_progedmf_1m.yml is essentially unchanged.


test/DataLayouts/unit_loops.jl
```diff
@@ -113,6 +113,50 @@ end
...
+    device = ClimaComms.device()
+    volume = test_data(device, Float64, 1, 10)
+    surface = test_data(device, Float64, 1, 1) # statically singleton Nv
+    point = DataLayouts.DataF{Float64}(device_array(device, rand(1)))
+    dest = test_data(device, Float64, 1, 10)
+
+    # Singleton dimensions inside broadcast expressions expand like Base's.
+    dest .= volume .+ surface
+    @test parent(dest) ==
+          device_array(device, Array(parent(volume)) .+ Array(parent(surface)))
```
dennisYatunin:
```suggestion
    @test parent(dest) == parent(volume) .+ parent(surface)
```
Is there a need to go back and forth between the host and device array types? Same for the other tests below.

test/DataLayouts/unit_loops.jl
```diff
@@ -113,6 +113,50 @@ end
...
+        Array(parent(volume)) .+ Array(parent(surface)) .+ Array(parent(point))[]
+    @test parent(dest) == device_array(device, reference)
+
+    # Reductions over expressions with mixed shapes require Cartesian indices,
+    # which Broadcast.newindex projects onto singleton dimensions; expressions
+    # whose layouts all share a shape permit linear indices.
+    mixed_bc = Base.broadcasted(+, volume, surface)
+    @test Base.IndexStyle(mixed_bc) == IndexCartesian()
+    @test sum(identity, mixed_bc) ==
+          sum(Array(parent(volume)) .+ Array(parent(surface)))
```
dennisYatunin:
```suggestion
    @test parent(Base.materialize(mixed_bc)) == parent(volume) .+ parent(surface)
```
Better to check the entire parent array instead of just its sum.

test/DataLayouts/cuda.jl
```diff
@@ -1,9 +1,17 @@
...
+using CUDA
+import ClimaCore
 import ClimaCore: slab
 import ClimaCore.DataLayouts: VIJFH
 ClimaComms.@import_required_backends
 
+# Cross-check every launch configuration computed by this test against CUDA's
+# occupancy API. The check is disabled by default so that a tie-breaking
+# difference on an untested device degrades performance instead of crashing.
+Base.get_extension(ClimaCore, :ClimaCoreCUDAExt).VALIDATE_LAUNCH_CONFIGURATIONS[] = true
```
dennisYatunin: If this needs to be exposed for testing, I'd recommend moving it into `src`. Global constants in extensions are typically meant to be private, and we might need a similar flag for non-CUDA devices if we ever get around to using those.

test/DataLayouts/cuda.jl
```diff
@@ -1,9 +1,17 @@
+ENV["CLIMACOMMS_DEVICE"] = "CUDA"
```
dennisYatunin: Shouldn't this be part of the CI pipeline, rather than hard-coded into the test file?

src/Utilities/safe_mapreduce.jl
```diff
@@ -19,15 +19,14 @@ When `init` is available, the reduction is a sequential left fold seeded by
 grows logarithmically rather than linearly with respect to length. Sequential
 reductions use `@simd` loops, meaning that associativity is not guaranteed.
 """
-Base.@propagate_inbounds function safe_mapreduce(
+Base.@propagate_inbounds @inline function safe_mapreduce(
```
dennisYatunin:
```suggestion
Base.@propagate_inbounds function safe_mapreduce(
```
Since `@propagate_inbounds` implies `@inline`, the new annotation here is redundant. Same for `mapreduce_pairwise` below, and also some methods defined elsewhere in this PR.

src/Utilities/safe_mapreduce.jl
```diff
@@ -19,15 +19,14 @@ When `init` is available, the reduction is a sequential left fold seeded by
...
     f::F,
     op::O,
     itr;
     init = NoInit(),
 ) where {F, O}
     first_index, last_index = firstindex(itr), lastindex(itr)
     if init isa NoInit
-        @assert first_index <= last_index # itr must be nonempty if init is missing
```
dennisYatunin, line 30: Is this assertion incorrect? Or did you need to remove it for compilation reasons?

src/DataLayouts/struct_storage.jl
```diff
@@ -115,8 +117,9 @@ end
 @inline single_index(index, ::Val{Nf}) where {Nf} =
     isone(Nf) ? Tuple(index) : throw(ArgumentError("F axis is required unless Nf = 1"))
 
-@inline struct_index(i, array) = i
-@inline struct_indices(array, ::Val{Nf}) where {Nf} = (Base.OneTo(Nf),)
+@inline struct_index(i, array) = iszero(ndims(array)) ? CartesianIndex() : i
+@inline struct_indices(array, ::Val{Nf}) where {Nf} =
+    iszero(ndims(array)) ? (CartesianIndex(),) : (Base.OneTo(Nf),)
```
dennisYatunin: Do the original versions of these not work for zero-dimensional arrays? I've tested them using `DataF` layouts without any problems so far. Shouldn't make any difference for performance, but I'd expect the original versions to have slightly lower latency.

src/DataLayouts/struct_storage.jl
```diff
@@ -184,16 +187,36 @@ julia> set_struct!(zeros(Int64, 3, 4), (Int32(2), Int32(0), Int128(1)), 2, 3)
...
 ```
 """
-@inline function set_struct!(array, value::T, index...) where {T}
-    Nf = num_basetypes(eltype(array), T)
-    @boundscheck checkbounds(array, struct_indices(array, Val(Nf), index...)...)
-    entries = bitcast_struct(NTuple{Nf, eltype(array)}, value)
-    unrolled_foreach(enumerate(entries)) do (i, entry)
-        @inbounds array[struct_index(i, array, index...)] = entry
-    end
+@propagate_inbounds function set_struct!(
```
dennisYatunin, line 189:
```suggestion
@inline function set_struct!(
```
No need for a `@propagate_inbounds` annotation here, since you're using `@inbounds` in the function body. Same for `set_struct_entries!` and `get_struct` below.

src/DataLayouts/scopes.jl
```diff
@@ -71,41 +71,42 @@ partition(scope) = scope
 Checks if one [`DataScope`](@ref) is equal to another scope, or if it is a
 [`partition`](@ref) of the scope, or a partition of a partition, and so on.
 """
-is_subscope(subscope, scope) =
+@inline is_subscope(subscope, scope) =
     subscope == scope ||
     partition(scope) != scope && is_subscope(subscope, partition(scope))
 
-num_subscopes(subscope, scope) =
+@inline @noinline throw_invalid_subscope() = throw(ArgumentError("Invalid subscope"))
```
dennisYatunin: I'm guessing this was meant to be `@noinline`, rather than `@inline @noinline`? But also, was there something wrong with the more informative `invalid_subscope_string` error message? If the issue was inlining, you could just wrap `throw(ArgumentError(invalid_subscope_string(subscope, scope)))` in a `@noinline`, rather than scrapping the error message altogether.

src/DataLayouts/scopes.jl
```diff
@@ -367,14 +368,28 @@ of the scope gets a nonempty chunk whenever there are at least as many indices
...
+) where {S <: DataScope} = indices
+Base.@propagate_inbounds @inline function subscope_indices(::ThisThread, scope, indices)
+    scope == ThisThread() && return indices
+    rank = thread_rank(scope)
+    n = num_threads(scope)
+    view_range =
+        strided_access(scope) ? (rank:n:length(indices)) :
+        (length(indices) * (rank - 1) ÷ n + 1):(length(indices) * rank ÷ n)
+    return subscope_index_view(scope, indices, view_range)
+end
```
dennisYatunin: Are these not equivalent to the method below?

src/DataLayouts/masks.jl
```diff
@@ -85,3 +85,35 @@ Check whether a [`DataMask`](@ref) marks the point at some index as active.
...
+    (; i_map, j_map, h_map) = inds.mask
+    real_n = inds.indices[n]
+    @inbounds CartesianIndex(i_map[real_n], j_map[real_n], h_map[real_n])
+end
+Adapt.@adapt_structure ActiveColumnIndices
+
+struct ActivePointIndices{M, V} <: AbstractVector{CartesianIndex{4}}
+    mask::M
+    indices::V
+    Nv::Int
```
dennisYatunin: `Nv` should probably be a static parameter, so the compiler can replace it with a `SignedMultiplicativeInverse` in `divrem` below. If it's a dynamic value, `divrem` has to be compiled into an actual division operation.

src/DataLayouts/loops.jl
```diff
@@ -15,20 +15,31 @@ end
...
+@inline each_maskable_slice_index(_, mask::IJHMask, ::typeof(column), args...) =
+    ActiveColumnIndices(mask)
+@inline each_maskable_slice_index(_, mask::IJHMask, ::typeof(view), args...) =
+    ActivePointIndices(mask, size(first(args), 1))
+
+# Every valid mask and slice operator combination has indexable slice indices:
+# NoMask uses the full index ranges, and IJHMask (which only supports column
+# and view slices) uses compacted active indices, so no combination requires
+# filtering the indices with a per-slice mask lookup.
+@noinline throw_invalid_slice_mask() = throw(ArgumentError("Invalid slice mask"))
```
dennisYatunin: Could this also keep the more informative `invalid_mask_string` error message? You can keep it wrapped in `@noinline` if that's needed for compilation, but the message itself shouldn't make a difference.

src/DataLayouts/loops.jl
```diff
@@ -138,14 +171,74 @@ end
...
+    slice_arg(op, subscope, index, first(args)),
+    slice_args(op, subscope, index, Base.tail(args))...,
+)
+
+@inline call_f(f::F, (a1,)::Tuple{Any}) where {F} = f(a1)
+@inline call_f(f::F, (a1, a2)::Tuple{Any, Any}) where {F} = f(a1, a2)
+@inline call_f(f::F, (a1, a2, a3)::Tuple{Any, Any, Any}) where {F} = f(a1, a2, a3)
+@inline call_f(f::F, (a1, a2, a3, a4)::Tuple{Any, Any, Any, Any}) where {F} =
+    f(a1, a2, a3, a4)
+@inline call_f(f::F, slices::Tuple) where {F} = f(slices...)
```
dennisYatunin: Use `UnrolledUtilities` to avoid hard-coded loop unrolling like this.

src/DataLayouts/loops.jl
```diff
@@ -254,33 +347,64 @@ end
...
+# the slice index machinery uses for single-point views.
 @inline reduce_points(::ThisThread, op::O, arg; mask, init...) where {O} =
     if mask == NoMask()
         indices = @inbounds subscope_indices(ThisThread(), DataScope(arg), eachindex(arg))
         safe_mapreduce(index -> (@inbounds arg[index]), op, indices; init...)
     else
         indices = subscope_slice_indices(ThisThread(), DataScope(arg), mask, view, arg)
-        mapreduce(index -> (@inbounds arg[index]), op, indices; init...)
+        safe_mapreduce(index -> (@inbounds arg[index]), op, indices; init...)
     end
```
dennisYatunin, line 358: With this change, you should be able to drop the first branch, since it is now equivalent to the second branch.

src/DataLayouts/loops.jl
```diff
@@ -254,33 +347,64 @@ end
...
+            @inbounds dest_column[v] = value
+            v += 1
+        end
     end
+
+# Left fold over the levels of a column (right fold when flipped), seeded by
+# the init value when one is given and by the first reduced level otherwise.
+# Written as a plain while loop, rather than as a call to reduce over the
+# column slice, so that it compiles in GPU kernels, where each column is
+# reduced by a single thread.
```
dennisYatunin: If `reduce` isn't compiling in GPU kernels, there must be something wrong with the `reduce_points` implementation. Fused kernels should support calls to `reduce`, so it would be better to fix any compilation issues there, instead of writing a custom alternative wherever we need to do reductions.

src/DataLayouts/loops.jl
```diff
@@ -365,9 +489,12 @@ end
...
 # reducing them. Remove all AutoBroadcaster wrappers after obtaining the result.
-function Base.reduce(op::O, arg::MaybeLazyDataLayout; kwargs...) where {O}
-    reducible = arg isa LazyDataLayout ? Broadcast.instantiate : Broadcast.broadcastable
-    result = drop_auto_broadcasters(reduce_points(op, reducible(arg); kwargs...))
+@inline function Base.reduce(op::O, arg::MaybeLazyDataLayout; kwargs...) where {O}
+    result = if arg isa LazyDataLayout
+        drop_auto_broadcasters(reduce_points(op, Broadcast.instantiate(arg); kwargs...))
+    else
+        drop_auto_broadcasters(reduce_points(op, Broadcast.broadcastable(arg); kwargs...))
+    end
```
dennisYatunin: How is this different from what was already here? It looks like the same thing, just more verbose.

src/DataLayouts/indexing.jl
```diff
@@ -1,34 +1,79 @@
...
+    (1 <= F <= length(params)) || return :(false)
+    params[F] <: AbstractUnitRange || return :(false)
+    for i in 1:(F - 1)
+        params[i] <: Base.Slice || return :(false)
+    end
+    for i in (F + 1):length(params)
+        params[i] <: Base.Slice || return :(false)
+    end
+    return :(true)
+end
```
dennisYatunin: This would probably be simpler if written with `unrolled_all`, instead of a `@generated` function.

src/DataLayouts/indexing.jl
```diff
@@ -153,24 +234,70 @@ end
     view(arg, CartesianIndex(indices...))
 
 # Reduce latency by only constructing slice views when necessary.
-@propagate_inbounds function level(arg::IndexableData, v)
-    (; Nv) = vijh_params(arg)
-    Nv == 1 || return level_view(arg, v)
+@propagate_inbounds function level(arg::VIJHWithF{<:Any, 1}, v)
```
dennisYatunin: All of these layout-specific boundschecking methods were exactly what the refactor was trying to eliminate. Any chance you could keep these methods layout-agnostic, rather than defining a new method per layout?

src/DataLayouts/DataLayouts.jl
```diff
@@ -525,18 +547,71 @@ end
...
+    i,
+    j,
+    h,
+) where {T, Ni, Nj, Nh, S}
     (h2, h1) = fldmod(h - 1, size(data, 1) ÷ Ni) .+ 1
     array = stable_view(parent(data), Ni * (h1 - 1) .+ (i:i), Nj * (h2 - 1) .+ (j:j))
-    return rebuild(data, array; Ni = 1, Nj = 1, Nh = 1)
+    return IH1JH2{T, 1, 1, 1, S}(array)
 end
+@inline reassign(data::DataF{T, S, A}, scope::NewScope) where {T, S, A, NewScope} =
```
dennisYatunin: Was there something wrong with the layout-agnostic version of `reassign`? Its original definition was relatively simple, so I'm not sure why you needed to replace it with these hard-coded methods.

src/DataLayouts/broadcast.jl
```diff
@@ -155,14 +166,14 @@ Replaces each of the [`layout_args`](@ref) in a broadcast expression with
 @propagate_inbounds function modify_args(f::F, bc::LazyDataLayout, f_args...) where {F}
     modified_args = unrolled_map_with_inbounds(bc.args) do arg
         Base.@_propagate_inbounds_meta
-        arg isa MaybeLazyDataLayout ? f(arg, f_args...) : arg
+        modify_arg(f, arg, f_args...)
```
dennisYatunin: Isn't this equivalent to what was already here? The `Base.@_propagate_inbounds_meta` already ensures that `@inbounds` is propagated into `f`.

ext/cuda/loops.jl
```diff
@@ -1,4 +1,11 @@
-function DataLayouts.foreach_slice(::ThisHost, op::O, f::F, args...; kwargs...) where {O, F}
+function DataLayouts.foreach_slice(
+    ::ThisHost,
+    op::O,
+    f::F,
+    args...;
+    max_waves = nothing,
```
dennisYatunin: The top-level method of `foreach_slice` restricts the keyword arguments to just `mask`, so this `max_waves` argument can never be used. Probably best to leave it as an internal parameter, instead of exposing it as a keyword argument.

ext/cuda/loops.jl
```diff
@@ -8,44 +15,124 @@ function DataLayouts.foreach_slice(::ThisHost, op::O, f::F, args...; kwargs...)
...
         result = DataLayouts.reduce_points(ThisBlock(), op, arg; kernel_kwargs...)
+        block_idx = DataLayouts.partition_rank(ThisKernel())
         if is_first_thread_in(ThisBlock())
-            @inbounds results[DataLayouts.partition_rank(ThisKernel())] = result
+            @inbounds results[block_idx] = result
+        end
+
+        # Make this block's result visible to the whole device before its
+        # counter increment, so that the last block observes every result.
+        CUDA.threadfence()
```
dennisYatunin, line 67: Since this is introducing a block-level barrier, the occupancy computation will need to be updated. It's not necessary to do that in this PR, but at least leave a TODO above `max_active_blocks` so we don't forget.

dennisYatunin (approved):
These are some really great performance improvements! I'm a bit concerned about he proliferation of manually unrolled methods and layout-specific methods, but we can always simplify those later. I've commented on the places where the increased verbosity seems most unnecessary; the rest looks good enough to merge in as is.

test/DataLayouts/unit_loops.jl
tapios: Done. The assertions now compare parent arrays directly `(parent(dest) == parent(volume) .+ parent(surface))`, and the same for the cases below it.

test/DataLayouts/unit_loops.jl
tapios: Done

test/DataLayouts/cuda.jl
tapios: Done. It now lives in `DataLayouts` with a docstring, so tests don't reach into extension, and a future non-CUDA device extension can share it.

test/DataLayouts/cuda.jl
tapios: Done. Removed from the test file, added as env: on the Buildkite step.

src/Utilities/safe_mapreduce.jl
tapios: Done. Also replace in other places where I'd added the redundant pair.

src/DataLayouts/struct_storage.jl
tapios: I applied, then reversed. This suggestion turned out to be wrong (causing CI errors)

tapios: The @inbounds covers the element access, but there's also a @boundscheck checkbounds(...) on the first line. @boundscheck blocks are only elided when the caller is in an @inbounds region and the function being called is @propagate_inbounds. With plain @inline, that bounds check ran on every point.

src/Utilities/safe_mapreduce.jl
tapios: Neither; collateral from debugging. Restored. It compiles fine in kernels, as it did on main.

src/DataLayouts/struct_storage.jl
tapios: You were right, they work. The special cases were unnecessary; reverted to the originals.

src/DataLayouts/scopes.jl
tapios: Both correct. It's now a plain `@noinline throw_invalid_subscope(subscope, scope)` that throws the original generated `invalid_subscope_string` message.

src/DataLayouts/scopes.jl
tapios: Yes, indeed. Collapsed the 4 methods into 1.

src/DataLayouts/loops.jl
tapios: Done, wrapped in @noinline.

src/DataLayouts/loops.jl
tapios: Done differently, and it's faster than either version. Routing through unrolled_map + Base.Fix1 cost 2× on CPU broadcasts (3670 vs 1846 µs); the closure and apply layer don't fully disappear here. But a plain two-method tuple recursion beats both (tested).

src/DataLayouts/loops.jl
tapios: Agreed and done. Once `reduce_points` was restored to `safe_mapreduce`, main's `fill!(dest, reduce(op, ...))` form works again. `column_fold` is deleted and `column_reduce!` is back to main's version

src/DataLayouts/indexing.jl
tapios: Done. Rewritten with fieldtypes + unrolled_all.

src/DataLayouts/masks.jl
tapios: Done. `ActivePointIndices{Nv}` now carries it as a type parameter

src/DataLayouts/loops.jl
tapios: Done

src/DataLayouts/loops.jl
tapios: Correct. Back to the ternary.

src/DataLayouts/indexing.jl
tapios: Done. Restored main's vijh_params-based versions

src/DataLayouts/DataLayouts.jl
tapios: Nothing. The generic was still there underneath; the four explicit methods were just shadowing it. Deleted.

src/DataLayouts/broadcast.jl
tapios: Yes. Restored the inline form.

ext/cuda/loops.jl
tapios: Updated and verified occupancy computation

ext/cuda/loops.jl
tapios: Agreed. It's now a local variable at the top of the `ThisHost` method rather than a keyword argument,

dennisYatunin: It's already `nothing` by default, so you shouldn't need it at the top either.

PR #2570, opened 2026-08-06
Test suite consolidation, dual-precision expansion, and CI stabilization.

tapios:
**Consolidation & taxonomy**
- Make `test/runtests.jl` the source of truth: register all standalone and GPU test files with a `tier` (unit/inference/conv/allocs/smoke/gpu/misc), a `subsystem`, and a `slow` flag; match filename prefixes to tiers (`unit_`, `inference_`, `conv_`, …); document the conventions in `test/README.md`.
- Rebuild the Buildkite pipeline around the harness: one job per subsystem on CPU and GPU, plus one per tier, replacing the single-process steps that duplicated them.
- Keep the two-core GitHub runners inside their budget. GHA and Downgrade skip `slow` tests and the tiers that repeat their cases or run end-to-end (conv/inference/allocs/smoke), which Buildkite covers. Downgrade also drops `:inference`, whose JET counts are pinned to the Manifest and drift under forced-oldest dependencies. Unit tests go from hitting the 80-minute timeout to ~45 minutes.
- Add CG/DG integration smoke tests (Bickley jet, column advection, solid-body rotation, baroclinic wave) and DG interface-flux and stability property tests.
- Probe all four divergence forms (strong, weak, split, DG) with vector spherical harmonics, whose divergence is pointwise-exact at every wavenumber: post-DSS max-norm accuracy up to the element resolution limit, the pre-DSS inter-element jump, and its O(h^(Nq−1)) refinement rate (measured 2.82 against a design rate of 3 for Nq = 4).
- Sweep Float32 and Float64 across DataLayouts, CommonSpaces, Spaces, and Operators; parameterize the shared 2D/3D test utilities by `FT`.

**Fixes**
- Fix the Bickley jet DG flux: physical velocity components were labeled contravariant, so the volume term disagreed with the interface Rusanov flux by the metric factor π/4. Both forms now build from one physical flux function.
- Fix GPU kernel compilation in the bidiag-matrix-row test (`Base.Fix1` holding a `Type` is not a bitstype) and allocation-test failures that came from measurement rather than allocation: an inline `@allocated f(...)` depends on the caller's inlining budget, so it now runs from `@noinline` wrappers.
- Merge `:opt` into `:inference`; JET's `@test_opt` is a type-stability check, so the two tiers tested the same thing. The tier runs under both `--check-bounds` settings, since JET reads the optimized IR and a bounds check that fails to compile under only one setting would otherwise go unnoticed.
- Set `CLIMACOMMS_DEVICE=CUDA` on two "cuda" MPI jobs that ran on CPU; drop a duplicate Integrals perf step; add assertions to three tests that checked only that nothing threw; promote the advection-convergence job from soft to hard.
- Restore GPU matrix-field broadcasting coverage: a CPU umbrella over all 22 cases in the harness, and a Buildkite matrix running each case as its own GPU job. Several need minutes of ptxas apiece, and sharing a process starves the assembler, which is SIGKILLed partway through `test_scalar_15`. A README maps each `test_scalar_*`/`test_non_scalar_*` filename to the expression it tests.
- Instrument coverage only on the job that uploads it. `if: ${{ a }} && ${{ b }}` interpolates each `${{ }}` separately and leaves the string `"true && false"`, which is truthy either way, so codecov uploaded from every job in the matrix. Instrumentation also bypasses the precompiled pkgimages, making an instrumented run ~2.5x longer, and is what killed the 1.11 jobs near the end of the suite: with coverage off it passes the test that killed it (172 tests, 30.0 s), with coverage on it is killed ~16 s in, and 1.10 passes either way.
- Raise the RootSolvers and ClimaComms compat floors; reformat for JuliaFormatter v2.10.1.

**Known soft failures**: three GPU matrix-field broadcasting cases (`test_scalar_14`, `test_scalar_15`, `test_non_scalar_2`) and 2-rank GPU distributed remapping.


tapios:
@akshaysridhar I tagged you for review. Please take a look at the DG tests and let me know if there's anything we should add right now. 

The Bickley jet in DG has errors; fixed in #2576

test/runtests.jl
```diff
@@ -7,113 +7,201 @@ include("tabulated_tests.jl")
...
+    UnitTest("DebugOnly"                               ,"Utilities/unit_debug_only.jl"; tier = :unit, subsystem = :utilities),
+
+    # Domains
+    UnitTest("Domains"                                 ,"Domains/unit_domains.jl"; tier = :unit, subsystem = :domains),
+
+    # Geometry
+    UnitTest("Geometry"                                ,"Geometry/unit_geometry.jl"; tier = :unit, subsystem = :geometry),
+    UnitTest("mul_with_projection"                     ,"Geometry/unit_mul_with_projection.jl"; tier = :unit, subsystem = :geometry),
+    UnitTest("Tensors"                                 ,"Geometry/unit_tensors.jl"; tier = :unit, subsystem = :geometry),
+
```
imreddyTeja: nit: trailing whitespace

.buildkite/pipeline.yml
```diff
@@ -31,158 +34,155 @@ steps:
...
+        agents:
+          slurm_mem: 20G
 
-      - label: "Unit: kernel renaming"
-        key: kernel_renaming
+  # GPU counterpart of the curated-suite group: run the harness on a GPU agent
+  # with TEST_TIER=gpu, which selects the :gpu tier (the CPU-vs-GPU correctness  
+  # tests). Multi-GPU / MPI (srun) and GPU-variant steps are not harness-runnable 
+  # and remain below.
+  - group: "Unit: curated suite (GPU)"
```
imreddyTeja, line 129: This change skips some test files that used to be exercised on a GPU. Was that intentional? For example: integrals.jl, the mass borrowing limiter tests, and matrix_multiplication_recursion

test/gpu/kernel_latency_regression.jl
```diff
@@ -1,3 +1,7 @@
+#=
+julia --project=.buildkite
+using Revise; include(joinpath("test", "gpu", "kernel_latency_regression.jl"))
+=#
```
imreddyTeja: Do people use this commented out way to run the tests?

test/Operators/spectralelement/conv_sphere_hyperdiffusion_vec.jl
```diff
@@ -21,18 +21,18 @@ include("utils_sphere_hyperdiffusion_vec.jl")
...
     VSH_cov = Geometry.transform.(Ref(Geometry.Covariant12Axis()), VSH_local)
 
-    # hyperdiffusion operator
+    # Hyperdiffusion operator
     ∇⁴VSH_cov = ∇⁴(VSH_cov)
     ∇⁴VSH_local = Geometry.transform.(Ref(Geometry.UVAxis()), ∇⁴VSH_cov)
 
-    # exact solution
+    # Exact solution
     ∇⁴VSH_exact = @. l^2 * (l + 1)^2 * VSH_local
```
imreddyTeja: This PR deletes `convergence_sphere_hyperdiffusion_vec.jl`. Should convergence tests be added here?

test/MatrixFields/matrix_field_test_utils.jl
```diff
@@ -1,14 +1,58 @@
-using Test
-using JET
+if !@isdefined(USING_JET)
```
imreddyTeja, line 1: I really like this change. I think it would be good to add a one time warning if the fallback fake JET module is used. I find it fairly common for JET to be in my enviornment, but fails to load because of some compatibility issues

test/README.md
```diff
@@ -1,44 +1,209 @@
...
+   invisible.
+6. Measure allocations from a dedicated `@noinline` helper, never inline:
+
+   ```julia
+   @noinline my_op_allocs(dest, x) = @allocated my_op!(dest, x)
+   @test my_op_allocs(dest, x) == 0        # yes
+   @test (@allocated my_op!(dest, x)) == 0 # no: result depends on how much of
+                                           # the caller's inlining budget the
+                                           # surrounding code consumed
+   ```
```
imreddyTeja: This suggested is not followed in many of the tests

imreddyTeja (commented):
I left some minor comments. I'll take another look ith fresher eyes later. The only two concerns I have are
1. The comment I left about the gpu tests
2. It looks like github ci is significantly slower after this change. Is this expected?

tapios:
> I left some minor comments. I'll take another look ith fresher eyes later. The only two concerns I have are
> 
> 1. The comment I left about the gpu tests
> 2. It looks like github ci is significantly slower after this change. Is this expected?

Thanks so much for the careful review!

test/runtests.jl
tapios: Fixed

test/gpu/kernel_latency_regression.jl
tapios: This is a preexisting pattern on this repo (in something like ~150 files). I suppose it's helpful for REPL iteration, but I'd also be happy to remove it (it would touch a lot of files). 

What do you think, @imreddyTeja, @dennisYatunin?

test/Operators/spectralelement/conv_sphere_hyperdiffusion_vec.jl
tapios: Great catch! Thanks. The file was renamed, but the actual convergence test accidentally deleted. Restored the convergence test.

test/MatrixFields/matrix_field_test_utils.jl
tapios: Good idea. Implemented.

test/README.md
tapios: Fair point. Problems occurred where the measured call was inlined into a large caller. Most remaining inline `@allocated` sites already call a small named function, where the risk is lower. Rather than changing those (which does not seem necessary), i softened the README guidance here.

tapios:
> > I left some minor comments. I'll take another look ith fresher eyes later. The only two concerns I have are
> > 
> > 1. The comment I left about the gpu tests
> > 2. It looks like github ci is significantly slower after this change. Is this expected?
> 
> Thanks so much for the careful review!

Re GH ci: it is slower because there are more tests there now (e.g., `operator_matrices.jl` and `field_matrix_solvers.jl` are enabled, which main had commented out as "too long"). Some run on GH and buildkite. It would be easy to drop the redundant GH tests. The downside is that the codecov coverage (derived from GH) would decrease. 

For now, I set `TEST_EXCLUDE_TIER: "conv,opt,allocs,smoke"` for the GH workflows (those are on buildkite)

.buildkite/pipeline.yml
tapios: Good point. There were fewer GPU tests on the branch (that previously just ran on GPU). Fixed by adding a GPU subsystem matrix that mirrors the one for CPU, and including previous GPU tests (plus a few new ones).

test/gpu/kernel_latency_regression.jl
imreddyTeja: I personally have never used any of these. I have `Revise` in my startup already, and it is pretty straight forwards to run any of the tests from the `.buildkite` env.

I haven't seen this pattern anywhere else either, so if it is unused, we probably shouldn't propagate it. 

@akshaysridhar Do you find these useful?

.buildkite/pipeline.yml
```diff
@@ -1568,10 +571,44 @@ steps:
...
+      # across CI hardware (e.g. A100 vs P100). The gates are one-sided (only a
+      # slowdown fails), but still soft_fail: the @info lines log the latency
+      # and percent-change trend without failing the build on hardware
+      # differences. Re-measure the baselines on the CI GPU to promote this
+      # back to a hard gate.
+      - label: "Perf: GPU kernel launch latency"
+        key: "gpu_kernel_launch_latency"
+        retry:
+          automatic: true
+        soft_fail: true
```
imreddyTeja: I added this "launch latency test" when I worked on the PR that cut down launch latency (targeting climaland) by removing the redundent adapt.

In hindsight, this test is pretty flaky. Feel free to delete it. I can't think of a more reliable way to check kernel launch latency in ci, so maybe it can be deleted and replaced with a note in the contributor docs about checking launch latency when making changes in the cuda ext.

test/gpu/kernel_latency_regression.jl
tapios: Ok. I'll take it out. I didn't really see the point of it either.

.buildkite/pipeline.yml
tapios: Ok. I removed the test and added a comment to the docs.

test/Fields/opt_fields.jl
```diff
@@ -19,6 +15,11 @@ import ClimaCore:
     Geometry,
     Quadratures
 
+if ClimaComms.device() isa ClimaComms.CUDADevice
+    @info "Skipping CPU allocation tests on GPU"
+    exit(0)
```
imreddyTeja: Should this fully exit the program? If someone tries to run the `opt` test syite on a gpu, then all the tests after this one in `runtests` will silently get skipped

test/Operators/spectralelement/unit_sphere_dg_fluxes.jl
```diff
@@ -0,0 +1,59 @@
...
+
+# DG interface-flux consistency on the cubed-sphere spectral element space:
+# the jump penalty must vanish pointwise on continuous (single-valued) fields,
+# and applying it must not allocate. Global conservation and the weak/strong
+# (SBP) divergence identity for the central flux are covered — for perturbed,
+# non-constant states, on both the sphere and a plane — in
+# unit_dg_stability.jl.
+
+@testset "Cubed-Sphere DG Interface Fluxes" begin
+    ClimaComms.device() isa ClimaComms.CUDADevice ? CUDA.allowscalar(true) : nothing
```
imreddyTeja: nit: this is a pattern that already existed in the tests, but I believe that changes to `CUDA.allowscalar` leak from testsets.
```
julia> using CUDA

julia> using Test

julia> a = CUDA.ones(100);

julia> a[1]
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] errorscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:151
 [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:124
 [4] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:112
 [5] getindex(A::CuArray{Float32, 1, CUDA.DeviceMemory}, I::Int64)
   @ GPUArrays ~/.julia/packages/GPUArrays/uOPpi/src/host/indexing.jl:50
 [6] top-level scope
   @ REPL[4]:1
 [7] top-level scope
   @ none:1

julia> @testset "aaa" begin
        CUDA.allowscalar(true)
        a[1]
        end
┌ Warning: It's not recommended to use allowscalar([true]) to allow scalar indexing.
│ Instead, use `allowscalar() do end` or `@allowscalar` to denote exactly which operations can use scalar operations.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:184
Test Summary: |Time
aaa           | None  0.3s
Test.DefaultTestSet("aaa", Any[], 0, false, false, true, 1.786378339320963e9, 1.786378339653642e9, false, "REPL[5]")

julia> a[1]
1.0f0
```

test/MatrixFields/matrix_field_test_utils.jl
```diff
@@ -34,8 +99,10 @@ macro test_all(expression)
         # TODO: Some operations have an unelided view from getproperty
         # (48 bytes); whether the compiler elides it depends on how much
         # of its inference budget is used up.
-        @test (@allocated test_func()) ≤ 48 # allocations
-        @test_opt test_func()               # type instabilities
+        if !(ClimaComms.device() isa ClimaComms.CUDADevice)
```
imreddyTeja: Why is this now skipped when using CUDA? I don't see any new uses of the macro

test/Fields/opt_fields.jl
tapios: Yes, good point. Changed it to using the meta = :cpu_only filter.

test/Operators/spectralelement/unit_sphere_dg_fluxes.jl
tapios: Good point. Changed to using `ClimaComms.allowscalar(device) do ... end` (and here the global call was redundant anyway)

test/MatrixFields/matrix_field_test_utils.jl
tapios: Good catch; the guard is wrong, and I've reverted the relevant part.

test/Operators/spectralelement/unit_overintegration.jl
```diff
@@ -0,0 +1,69 @@
...
+        interp_op = Operators.Interpolate(space_high)
+        restrict_op = Operators.Restrict(space_std)
+
+        # 1. Exact polynomial round-trip identity:
+        # For any field in the standard space, Interpolate -> Restrict recovers the field
+        f_std = @. sin(coords_std.x) * cos(coords_std.y)
+        f_high = interp_op.(f_std)
+        f_recovered = restrict_op.(f_high)
+
+        tol = FT == Float32 ? 0.05 : 0.05
```
akshaysridhar:
```suggestion
        tol = FT(0.05)
```

test/Operators/spectralelement/utils_dg.jl
akshaysridhar, line 1: Thanks @tapios; The general structure of the test setup mirrors what I was experimenting with in my branches (compatibility with the other numerical fluxes is also reasonable - for now the `utils_dg.jl` file is enough to introduce and maintain the new test types; I've run these checks against the other flux formulations introduced in the `Operators` module of my working branches; some nit-picky comments above about test tolerance float-types, otherwise this is very helpful towards building out the full entropy stable flux configurations.

tapios: Thank you!

.github/workflows/UnitTests.yml
```diff
@@ -49,6 +49,18 @@ jobs:
...
+          # This job exists to check portability across Julia versions and
+          # operating systems, on one two-core runner per version. Buildkite is
+          # the exhaustive gate: it shards the same suite across parallel
+          # agents and runs everything excluded here.
+          #
+          # The convergence, optimization, allocation, and smoke tiers each run
+          # their code at several resolutions, twice, or end-to-end.
+          TEST_EXCLUDE_TIER: "conv,opt,allocs,smoke"
+          # `slow` tests cost minutes apiece on this runner (see the UnitTest
+          # docstring in test/tabulated_tests.jl for what qualifies).
```
imreddyTeja: I think this should also be applied to the downgrade test workflow

.buildkite/pipeline.yml
```diff
@@ -31,174 +34,272 @@ steps:
...
-        agents:
-          slurm_gpus: 1
-
-      - label: "Unit: compiler stress regression"
-        key: unit_compiler_stress_regression
+          - "julia --color=yes --project=.buildkite test/runtests.jl"
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+          TEST_SUBSYSTEM: "{{matrix.subsystem}}"
+          TEST_EXCLUDE_TIER: "conv,opt,allocs,gpu,smoke"
```
imreddyTeja: Is skipping convergence tests on GPU intentional? I'm OK with it, I just wanted to make sure it is an intentional deviation from main

tapios: Yes. Don't need GPUs for convergence checks, and the GPU build was becoming too large.

imreddyTeja (approved):
thank you!

test/gpu/latency_benchmarks.jl
```diff
@@ -1,66 +0,0 @@
-using ClimaCore
```
dennisYatunin, line 1: I found these latency benchmarks quite helpful while working on my last few PRs, particularly #2558. I'm fine with dropping the strict `@test` requirements, but would it be okay to keep the benchmarks themselves in CI? They haven't exhibited much variability (no more than 10--20%) in the builds I've been looking at.

src/DataLayouts/loops.jl
```diff
@@ -294,11 +294,28 @@ Use [`foreach_column`](@ref) to combine the levels of each column of `arg` with
...
+    if flip
+        foreach_column(dest, arg; mask) do dest_column, arg_column
+            fill!(dest_column, reduce(op, reverse(arg_column); init...))
+        end
+    else
+        foreach_column(dest, arg; mask) do dest_column, arg_column
+            fill!(dest_column, reduce(op, arg_column; init...))
+        end
     end
+end
```
dennisYatunin, line 314: Here's a less verbose alternative, which makes use of the fact that Julia is always willing to do constant propagation through at least one stack frame:

```suggestion
@inline function column_reduce!(
    op::O,
    dest,
    arg;
    mask = NoMask(),
    flip = false,
    init...,
) where {O}
    flip_val = Val(flip) # Force constant propagation of boolean flag.
    foreach_column(dest, arg; mask) do dest_column, arg_column
        maybe_reverse = flip_val == Val(true) ? reverse : identity
        fill!(dest_column, reduce(op, reverse(arg_column); init...))
    end
end
```

src/InputOutput/writers.jl
```diff
@@ -578,12 +578,13 @@ function write!(
 end
 
 function write_plain_array!(group, array::AbstractArray, name::AbstractString)
-    nd = ndims(array)
-    dims = size(array)
+    array_cpu = Array(array)
```
dennisYatunin: This can be changed to avoid an unnecessary allocation+copy when running on CUPs:
```suggestion
    array_cpu = array isa Array ? array : Array(array)
```

test/deprecations.jl
```diff
@@ -57,76 +53,84 @@ end
...
-    @test z_bnds[1, :] == [f_coords[v, 1, 1, 1].z for v in 1:z_elem]
-    @test z_bnds[2, :] == [f_coords[v + 1, 1, 1, 1].z for v in 1:z_elem]
+        # Broadcasting parent arrays of column data into vectors, and linearly
+        # indexing them in vertical order (TempestRemap's def_space_coord for
+        # finite difference spaces).
+        z = zeros(FT, z_elem)
+        z .= Array(parent(c_coords))
+        ClimaComms.allowscalar(ClimaComms.device()) do
+            @test z == [c_coords[v, 1, 1, 1].z for v in 1:z_elem]
+        end
```
dennisYatunin:
```suggestion
        @test z == [c_coords[v, 1, 1, 1].z for v in 1:z_elem]
```
Is this not already inside `allowscalar`? Same below.

test/tabulated_tests.jl
```diff
@@ -1,15 +1,27 @@
 using Test
-using PrettyTables
-using StatsBase: countmap
+const HAVE_PRETTY_TABLES = try
+    using PrettyTables
+    true
+catch
+    false
+end
```
dennisYatunin: PrettyTables is in the `Project.toml`, so won't this always be `true`?

test/README.md
```diff
@@ -1,44 +1,233 @@
...
+| `:misc`      | —             | Quality gates that don't fit above (Aqua, deprecations).                   |
+
+## Cost
+
+`tier` says what a test checks; `slow = true` says what it costs. A slow test
+takes minutes rather than seconds, usually from compiling many layout or space
+specializations, or from building large spaces.
+
+Marking a test slow does not reduce coverage anywhere it matters: Buildkite
+shards the suite across parallel agents and runs everything. The marker is
```
dennisYatunin: The word "shards" is used quite a lot throughout this PR. Consider replacing it with something less jargony, like "parallelizes" or "splits".

test/runtests.jl
```diff
@@ -1,120 +1,217 @@
...
-UnitTest("Operators - extruded 3dbox space ops"    ,"Operators/hybrid/extruded_3dbox_cuda.jl";meta=:gpu_only),
-UnitTest("Fields - CUDA mapreduce"                 ,"Fields/reduction_cuda.jl";meta=:gpu_only),
+    # DataLayouts
+    UnitTest("DataLayouts get_struct"                  ,"DataLayouts/unit_struct.jl"; tier = :unit, subsystem = :datalayouts),
+    UnitTest("DataLayouts slice indexing & setindex!"  ,"DataLayouts/unit_indexing.jl"; tier = :unit, subsystem = :datalayouts),
+    UnitTest("DataLayouts scopes"                      ,"DataLayouts/unit_scopes.jl"; tier = :unit, subsystem = :datalayouts),
+    UnitTest("DataLayouts loops"                       ,"DataLayouts/unit_loops.jl"; tier = :unit, subsystem = :datalayouts),
+    UnitTest("PlusHalf"                                ,"Utilities/unit_plushalf.jl"; tier = :unit, subsystem = :utilities),
+    UnitTest("Stable views"                            ,"Utilities/unit_stable_view.jl"; tier = :unit, subsystem = :utilities),
+    UnitTest("AutoBroadcaster"                         ,"Utilities/unit_auto_broadcaster.jl"; tier = :unit, subsystem = :utilities),
```
dennisYatunin: Maybe group the Utilities tests separately from DataLayouts?

test/README.md
```diff
@@ -1,44 +1,233 @@
...
+Every test carries a `tier`. The tier controls what runs in the default suite
+and what is opt-in.
+
+| tier         | prefix        | what it is                                                                 |
+|--------------|---------------|----------------------------------------------------------------------------|
+| `:unit`      | `unit_`       | Fast, high-coverage correctness tests (`Float64`, and `Float32` where cheap). The default run. |
+| `:inference` | `inference_`  | Type-stability / `@inferred` / JET regression gates. Kept separate because inference checks are slower. |
+| `:allocs`    | `allocs_`     | Runtime-allocation regression tests (warm-up + `@allocated == 0`). Requires running code twice, so separate from `:unit`. |
+| `:conv`      | `conv_`       | Convergence tests. Run the code at several resolutions to verify a theoretical convergence rate. |
+| `:opt`       | `opt_`        | Optimization / JET `@test_opt` / flop-counting checks. Run with default bounds checking in CI (JET's optimized-IR checks depend on the bounds flag). |
```
dennisYatunin: The `:opt` and `:inference` tiers should probably be unified. The only reason we use JET's `@test_opt` is to identify problems with type inference. Also, we should probably run inference tests with both `--check-bounds=yes` and `--check-bounds=no`; we don't want to end up in a situation where out-of-bounds errors cannot be detected on GPUs due to IRErrors during compilation.

dennisYatunin (approved):
Really happy with all the changes here; this'll be a huge improvement for CI. My only major suggestion is to unify the `:opt` and `:inference` tiers, as they are essentially testing the same thing.

test/gpu/latency_benchmarks.jl
tapios: Restored. @tests stripped, baselines reported

src/DataLayouts/loops.jl
tapios: Now uses `Val` (but with a bug above fixed (calling `reverse` conditionally).

test/tabulated_tests.jl
tapios: Dropped try ... catch

test/runtests.jl
tapios: Now have a Utilities group.

test/README.md
tapios: Done

test/README.md
tapios: Done both: :inference and :opt are now under :inference. Run 2 CPU jobs with both --check-bounds settings

test/deprecations.jl
tapios: redundant blocks removed

PR #40, opened 2021-06-11
Finite Difference Vertical Discretization and Operators

jakebolewski:
Co-authored-by: simonbyrne <simonbyrne@gmail.com>
Co-authored-by: jakebolewski <jakebolewski@gmail.com>

PR #2571, opened 2026-08-07
Allow Field views to be used in foreach_slice

dennisYatunin:
This PR allows `Field` arguments to participate in pointwise kernel fusion through the `foreach_slice` function that was added in #2522, and it bumps the version to 0.15.1.

Without these changes, pointwise kernel fusion requires every `DataLayout` to be extracted from its `Field`, which has two downsides: the resulting code is unnecessarily verbose (e.g., [this example](https://github.com/CliMA/ClimaAtmos.jl/commit/2d8470488f458263126dffebcb8ba90efd031d4d) in ClimaAtmos), and all `local_geometry` operations need to be handled manually (instead of being handled by the `Field` broadcasting machinery). In addition to adding the relevant methods for pointwise `Field` views, this PR ensures that all types of views (point, level, slab, and column) are consistent across the `DataLayouts`, `Spaces`, and `Fields` modules.

This PR also fixes several miscellaneous bugs from version 0.15.0:
- The `max_threads_per_block` and `max_blocks` arguments were flipped on one line in `launch_configuration` (#2558). The argument order is now correct.
- Loops generated by `foreach_point` permitted arguments with different sizes, extruding singleton axes as in `Broadcasted` expressions (#2568). All `foreach_slice` loops now require consistent slice indices, matching the `eachrow`/`eachcol` functions from `Base`, and singleton axes are only extruded when broadcasting.
- Subscopes for `view`, `level`, and `column` were hardcoded to `ThisThread` in `foreach_slice` (#2568). All subscopes are now determined by a single method for `slice_subscope`.
- Slices without statically inferable sizes were passed into fused kernels from `ThisHost` by `foreach_slice` (#2522). Only slices that support dynamic cache allocation can now participate in kernel fusion.
- Both left-hand-side and right-hand-side layouts were checked for consistency when determining `FusedMultiBroadcast` attributes (#2522). Only left-hand-side layouts are now checked; right-hand-side layouts can be extruded and are allowed to have different size attributes.
- Equality checks of `DataLayout` arguments returned false negatives for parent arrays with different dimensionalities (#2522). Mismatched parent arrays are now flattened with `stable_view` before they are compared.
- Equality checks generated `Tuple`/`NamedTuple` results when using `AutoBroadcaster` wrappers (#2522). Reductions are now performed over `all ∘ ==` instead of just `==`, so that equality checks always return booleans.

These changes were tested in ClimaAtmos CI [build 31222](https://buildkite.com/clima/climaatmos-ci/builds/31222). The only failing jobs were small MSE changes and OOM errors during allocation profiling, which can be avoided by selecting different broadcast expressions for kernel fusion.

### DIFF
--- a/src/DataLayouts/indexing.jl
+++ b/src/DataLayouts/indexing.jl
@@ -40,9 +40,16 @@ Base.IndexStyle(::Type{D}) where {D <: DataLayout} =
 # layout types among the args, which exceeds inference's union-splitting
 # limits for heterogeneous broadcast expressions and makes the shape checks
 # dynamic.
+# Layouts are collected recursively, so that a singleton layout nested inside
+# a broadcast expression is seen by the shape checks: a nested broadcast's own
+# merged shape_params would hide it, and a linear index does not project onto
+# singleton dimensions, so reading such a layout at a linear index would go
+# out of range.
 @inline get_non_point_arg_tuple(arg) = is_non_point_arg(arg) ? (arg,) : ()
+@inline get_non_point_arg_tuple(bc::MaybeFusedDataLayoutBroadcast) =
+    unrolled_flatmap(get_non_point_arg_tuple, layout_args(bc))
 
-@inline function equal_layout_shapes(args::Tuple)
+@inline function equal_layout_shapes(args)
     non_point_args = unrolled_flatmap(get_non_point_arg_tuple, args)
     return unrolled_allequal(layout_type, non_point_args) &&
            unrolled_allequal(shape_params, non_point_args)
@@ -55,17 +62,12 @@ end
 @inline Base.IndexStyle(bc::MaybeFusedDataLayoutBroadcast) =
     equal_layout_shapes(layout_args(bc)) ? IndexLinear() : IndexCartesian()
 
-@inline index_style_layouts(::Tuple{}) = IndexLinear()
-@inline index_style_layouts(args::Tuple{Any}) = IndexStyle(first(args))
-@inline index_style_layouts(args::Tuple) =
-    equal_layout_shapes(args) ?
-    unrolled_mapreduce(IndexStyle, IndexStyle, args) : IndexCartesian()
-
 # Allow linear indexing if all DataLayouts in an expression have the same shape.
 # Add DataLayout-only methods to avoid ambiguities with AbstractArray methods.
 for T in (:IndexableData, :DataLayout)
     @eval @inline Base.IndexStyle(arg1::$T, arg2::$T, args::$T...) =
-        index_style_layouts((arg1, arg2, args...))
+        equal_layout_shapes((arg1, arg2, args...)) ?
+        unrolled_mapreduce(IndexStyle, IndexStyle, (arg1, arg2, args...)) : IndexCartesian()
 
     @eval @inline Base.eachindex(arg::$T, args::$T...) =
         eachindex(IndexStyle(arg, args...), arg, args...)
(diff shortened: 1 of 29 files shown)


tapios:
Is this tested with atmos?

src/Spaces/Spaces.jl
```diff
@@ -84,7 +83,7 @@ global_geometry(space::AbstractSpace) = global_geometry(grid(space))
 space(refspace::AbstractSpace, staggering::Staggering) =
     space(grid(refspace), staggering)
 
-issubspace(::AbstractSpace, ::AbstractSpace) = false
+issubspace(subspace::AbstractSpace, space::AbstractSpace) = grid(subspace) === grid(space)
```
imreddyTeja: I don't think all spaces have a grid  or a grid method defined

src/DataLayouts/loops.jl
```diff
@@ -407,10 +407,11 @@ end
 @inline Base.:(==)(arg1::DataLayout, arg2::DataLayout; mask = NoMask()) =
     size(arg1) == size(arg2) && (
         mask == NoMask() &&
-        eltype(arg1) == eltype(arg2) &&
-        (Base.ispacked(eltype(arg1)) && Base.ispacked(eltype(arg2))) &&
-        (layout_type(arg1) == layout_type(arg2) && f_dim(arg1) == f_dim(arg2)) ?
-        parent(arg1) == parent(arg2) : mapreduce(==, &, arg1, arg2; mask, init = true)
+        (eltype(arg1) == eltype(arg2) && Base.ispacked(eltype(arg1))) &&
+        (layout_type(arg1) == layout_type(arg2) && f_dim(arg1) == f_dim(arg2)) &&
+        ndims(parent(arg1)) == ndims(parent(arg2)) ?
```
imreddyTeja: should this be size instead of ndims?

src/Spaces/Spaces.jl
```diff
@@ -95,6 +94,19 @@ coordinates_data(grid::Grids.AbstractGrid) =
 coordinates_data(staggering, grid::Grids.AbstractGrid) =
     local_geometry_data(staggering, grid).coordinates
 
+staggered_level_index(_, v::PlusHalf) = v
+staggered_level_index(space, v::Integer) = staggering(space) == CellFace() ? v - half : v
+
+integer_level_index(_, v::Integer) = v
```
imreddyTeja, line 117: Should this have validation on the space as well?

src/Spaces/Spaces.jl
dennisYatunin: I updated this to `subspace === space`, since the spaces are only identical if their grids are identical.

src/Spaces/Spaces.jl
dennisYatunin: I updated `staggered_level_index(_, ::PlusHalf)` to throw an error when a half-integer index is used to access a `CellCenter` space. There's no need for a similar check in `integer_level_index(_, ::Integer)`, since integer indices can now be used to access any space.

src/DataLayouts/loops.jl
dennisYatunin: The only example of automatic reshaping by a `DataLayout` is how `DataF` always treats its parent array as an `AbstractVector`. Comparing `ndims` is sufficient to catch this, and `parent(arg1) == parent(arg2)` already compares the parent sizes, so there's no need to replace `ndims` with `size`. Moreover, the `size` of a parent is generally not known during compilation, so using `size` here would hurt latency by forcing us to compile both branches.

src/Fields/Fields.jl
```diff
@@ -153,26 +158,50 @@ ClimaComms.array_type(field::Field) =
     Field(getproperty(field_values(field), name), axes(field))
 
 Base.eltype(::Type{<:Field{V}}) where {V} = eltype(V)
-Base.parent(field::Field) = parent(field_values(field))
 
-# to play nice with DifferentialEquations; may want to revisit this
-# https://github.com/SciML/SciMLBase.jl/blob/697bd0c0c7365e77fa311f2d32eade70f43a8d50/src/solutions/ode_solutions.jl#L31
-Base.size(field::Field) = ()
-Base.length(field::Field) = 1
```
tapios, line 174: This changes the semantics of `length` and `size`, from the previous SciML compatible version. Is there any risk that this causes problems, e.g., in ClimaTimeSteppers?

src/DataLayouts/indexing.jl
```diff
@@ -128,8 +128,8 @@ end
 
 # Represent every single-point DataLayout view using a zero-dimensional DataF.
 @propagate_inbounds Base.view(data::DataLayout, index::PointIndex) =
-    is_single_point(data, index) ? data :
     DataF{eltype(data), typeof(DataScope(data))}(
+        is_single_point(data, index) ? parent(data) :
```
tapios: Doesn't this always route through `check_basetype` (and its runtime checks) even for a single point? If so, maybe better to send `is_single_point` through an unchecked constructor.

src/DataLayouts/loops.jl
```diff
@@ -407,10 +404,11 @@ end
...
+        parent(arg1) == parent(arg2) :
+        mapreduce(all ∘ ==, &, arg1, arg2; mask, init = true)
```
tapios: This and similar uses of `∘ ==` may be problematic. The ComposedFunction pattern was causing the kernel failures removed in #2568

src/Fields/Fields.jl
dennisYatunin: ClimaTimeSteppers only operates on `FieldVector` arguments, never the internal `Field` components. These definitions of `size` and `length` have been around since #40, before `FieldVector` was even defined. Updating them should have no impact on downstream code.

src/DataLayouts/indexing.jl
dennisYatunin: I'm using the `DataF` constructor to reshape the parent into a 1-dimensional vector (previously `vec(array)`, now `stable_view(array, :)`). I suppose I could add an additional check for `is_single_point_(data, index) && isone(ndims(parent(data)))`, but that probably wouldn't make any difference for performance; `check_basetype` should be elided out regardless, as its result is statically inferable.

src/DataLayouts/loops.jl
dennisYatunin: Yes, I saw the issue you ran into with `eltype ∘ parent_type` in the original definition of `parent_type`. The problem is specific to recursive functions---since `Base` does not disable the recursion limit for `∘`, inference can fail when using nested data structures. Using `∘` in non-recursive functions like the one here should be safe. In this case, `all ∘ ==` should be equivalent to the more verbose `(x, y) -> all(x == y)`; it might even have slightly lower latency as it reuses the same struct instead of compiling a new one every time.

imreddyTeja (approved):

PR #2572, opened 2026-08-07
Fix failing MPI tests

nefrathenrici:
This PR fixes the library issues in the MPI tests and updates them to work with the recent datalayouts refactor. The newer climacommon uses julia 1.12 which loads its binary libcurl.so.4on startup. The manifest uses 1.10 and pins HDF5_jll, which expects a symbol version tag CURL_4 , but the binary provides CURL_OPENSSL_4 so it errors.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -195,8 +195,7 @@ steps:
         agents:
            slum_ntasks: 2
            slurm_gpus: 2
-           modules: mpiwrapper/2024_05_27 climacommon/2026_02_18
-        soft_fail: true # remove this after library issues are fixed
+           modules: mpiwrapper/2024_05_27 climacommon/2024_10_10
 
   - group: "Unit: Geometry"
     steps:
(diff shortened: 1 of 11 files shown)


tapios (approved):

.buildkite/pipeline.yml
```diff
@@ -195,8 +195,7 @@ steps:
         agents:
            slum_ntasks: 2
```
tapios:
```suggestion
           slurm_ntasks: 2
```

.buildkite/pipeline.yml
tapios, line 349: CLIMACOMMS_DEVICE: "CUDA" missing (trailing CUDA argv  above is dead I think)

tapios, line 632: CLIMACOMMS_DEVICE: "CUDA" missing

tapios (approved):
Thank you! Would be good to add NEWS.md entry.

PR #2574, opened 2026-08-08
Makes memory access patterns in remapping kernels more robust

tapios:
This PR optimizes GPU memory access patterns in the remapping kernels to ensure coalesced reads and writes:

- **`remapping_distributed.jl`:** Reordered `CartesianIndices` iteration to ensure coalesced global memory reads.
- **`remapping_interpolate_array.jl`:** Replaced in-loop global memory accumulations with explicit thread-local register accumulation in `interpolate_slab_kernel!`. 

This currently makes no performance difference in benchmarks, apparently because the compiler already optimizes these patterns (e.g., hoisting the memory accumulations into a register automatically). But these changes guarantee coalesced memory traffic and protect performance in case future compiler heuristic changes or loop complexity increases.

### DIFF
--- a/ext/cuda/remapping_distributed.jl
+++ b/ext/cuda/remapping_distributed.jl
@@ -52,10 +52,14 @@ function set_interpolated_values_linear_2d_kernel!(
     num_horiz = length(local_horiz_indices)
     num_vert = length(vert_bounding_indices)
     num_fields = length(field_values)
-    inds = (num_horiz, num_vert, num_fields)
+    # The vertical index varies fastest across threads: adjacent threads then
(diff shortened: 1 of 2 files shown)


imreddyTeja (approved):
Looks good to me. I checked the performance before and after the changes to `remapping_interpolate_array` and found it reduces the interpolation kernel's duration by about 50%

tapios:
> Looks good to me. I checked the performance before and after the changes to `remapping_interpolate_array` and found it reduces the interpolation kernel's duration by about 50%

Thank you! Yes, that's consistent with what I saw, but it made no measurable difference on atmos simulations.

PR #2575, opened 2026-08-09
Fix GPU InvalidIRError from inference widening in broadcast argument traversal

tapios:
## Problem

PR #2572  broke ClimaAtmos AMIP runs on GPU: compiling the fused cloud-fraction
broadcast kernel failed with `InvalidIRError` (`gpu_gc_pool_alloc`,
`jl_f__apply_iterate`), with the stacktrace pointing into `unrolled_filter`.

## Root cause

`unrolled_filter` (UnrolledUtilities v0.1.9) pushes items into a tuple
accumulator whose type changes at every step. For wide, complexly typed
broadcast expressions (fused EDMFX broadcasts), inference's recursion-widening
heuristics widen the accumulator, and the resulting dynamic dispatch and heap
allocation cannot be compiled for GPUs. The trigger is an inference budget,
which is why unit tests passed while AMIP failed.

## Changes

- Kernel-critical traversals build no tuples at all: `each_maskable_slice_index`
  and `equal_layout_shapes` compare each non-point layout against a reference
  layout, and `f_dim` reduces with a nothing-or-in
- Remaining kernel-reachable `unrolled_filter` calls (`layout_args`,
  `axes_at_colons`, `overlap_components`) use `unrolled_flatmap`, which maps
  each item to an empty or singleton tuple, avoidies.
- Scope errors are singleton exception types (`InvalidSubscopeError`,
  `NonOverlappingScopesError`) whose messages interpolate the scope types in
  `showerror` on the host only.
- Tests: `test/DataLayouts/opt_layout_args.jl` runs JET `@test_opt` over wide
  broadcasts and a fused multi-broadcast mixing `DataF` and `VIJFH` args (the
  AMIP failure shape); `test/aqua.jl` forbids `unrlit`
  in kernel-reachable directories. New Buildkite CPU and 1-GPU jobs run them.

## Companion change

UnrolledUtilities [PR 33](https://github.com/CliMA/UnrolledUtilities.jl/pull/33) fixes the root cause upstream by reimplementing `unrolled_filter`/`unrolled_split`/`unrolled_unique` via
`unrolled_flatmap`. This PR does not depend on it; all affected suites pass
against both v0.1.9 and the patched v0.1.10. Once v0.1.10 is registered, the
guardrail test can be dropped.


tapios:
@szy21 I have tested this with an AMIP setup locally; it runs. It would be useful though to also time it relative to what we had.

szy21:
It’s 4-5% faster than current main.

ext/cuda/scopes.jl
```diff
@@ -243,12 +250,12 @@ end
     ::typeof(view),
     args...,
 )
-    all_layouts = unrolled_flatten(unrolled_map(get_all_layouts, args))
     # 0-dimensional layouts (e.g. DataF args of fused broadcasts) broadcast to
     # every point, so they do not participate in the size or linearity checks.
-    layouts = unrolled_filter(DataLayouts.is_non_point_arg, all_layouts)
+    layouts = unrolled_flatmap(get_all_non_point_layouts, args)
```
dennisYatunin, line 255: A lot of these changes can be simplified to
```julia
@inline unrolled_filter(f::F, itr) = unrolled_flatmap(item -> f(item) ? (item,) : (), itr)
```
You can just add that to the `Utilities.Unrolled` module, if you want to merge this in before CliMA/UnrolledUtilities.jl#33

src/DataLayouts/scopes.jl
```diff
@@ -46,12 +46,30 @@ DataScope(::Type{<:Base.ReshapedArray{<:Any, <:Any, A}}) where {A} = DataScope(A
 # Infer parent types of other AbstractArrays (constant-folding not guaranteed).
 DataScope(::Type{A}) where {A <: AbstractArray} = DataScope(return_type(parent, Tuple{A}))
 
+# Scope errors are singleton exception types whose messages are only generated
+# by showerror on the host. Interpolating the scope types into a message where
+# the error is thrown would require calling a @generated function, which cannot
+# always be compiled for GPUs.
```
dennisYatunin, line 52: Oh wow, this is a rather frightening observation. If you can get a unit test out of this at some point, it would probably be quite helpful for improving `UnrolledUtilities.jl`.

dennisYatunin (approved):
Thanks for fixing these issues! CliMA/UnrolledUtilities.jl#33 will simplify this, but you can merge this in first if you'd like.

I'm curious to see if there's a simpler solution to the error message IRErrors than defining a new struct per message. Maybe something like the `print` function where we pass a collection of substrings and they all get concatenated? We can look into that later.

tapios:
> Thanks for fixing these issues! [CliMA/UnrolledUtilities.jl#33](https://github.com/CliMA/UnrolledUtilities.jl/pull/33) will simplify this, but you can merge this in first if you'd like.
> 
> I'm curious to see if there's a simpler solution to the error message IRErrors than defining a new struct per message. Maybe something like the `print` function where we pass a collection of substrings and they all get concatenated? We can look into that later.

Agreed. There may be a nicer design, e.g., a `KernelError{parts}` singleton parameterized by a tuple of Symbols and Types. I made a note to myself and will revisit this later.

ext/cuda/scopes.jl
tapios: Considered and not done, for 3 reasons. (a) The new guardrail test in test/aqua.jl forbids the `unrolled_filter` on code lines that are kernel-reachable so the accumulator-based UU v0.1.9 version can't sneak back in. (b) The biggest converted site can't go back to plain filter anyway: `each_maskable_slice_index` needs the recursive collection through nested broadcasts (`get_all_non_point_layouts`), and `f_dim`'s nothing-max reduce avoids tuple building. (c) The clean end-state I envision is different anyway: once UU with the kernel-safe filter is registered, ClimaCore can revert wholesale to upstream `unrolled_filter` and delete the guard.

src/DataLayouts/scopes.jl
tapios: Yes, it'll be good to have a clear unit tests. Will do later.

PR #2576, opened 2026-08-09
Examples: consolidate, correct, and make CI assert the physics

tapios:
**Reorganization.** Group examples by geometry (`column/`, `plane/`, `sphere/`,
`hybrid/{box,plane,sphere}`); delete the Held–Suarez cases, the
potential-temperature and internal-energy variants (the total-energy form
remains), duplicated cases, and dead scripts — 28 files deleted, net −4,200
lines. Merge the CUDA shallow-water fork into `sphere/shallow_water.jl`; share
the 2D plane setup in `plane_utils.jl`, the figure-link helper in
`example_utils.jl`, and the bubble constants in `bubble_parameters.jl`; sweep
all four flux constraints in one column limiter example. Consolidate
repetitive Buildkite steps into matrices, and point figure links at
repository-root paths so they resolve in the build logs.

**Dependency removal.** Examples, tests, and the introduction tutorial
time-step with ClimaTimeSteppers 0.10; the OrdinaryDiffEq/SciML dependencies
are gone, which drops 27 packages from the Buildkite environment. The SciML
compatibility layer goes with them — breaking for anyone passing a `Field` to
an OrdinaryDiffEq integrator, recorded in NEWS.md — and the two time-stepping
tests join the curated suite under `Pkg.test`.

**Physics fixes.** Terrain-following surface flow in the mountain and channel
cases, applied between stages rather than by mutating the stage state; the
correct Exner exponent in the mountain initial density (cv/R, not cp/R — the
state was out of hydrostatic balance by a factor of π(z)); a consistent
constant-N² Exner profile in the Agnesi case; pressure rather than density in
the isothermal channel's vertical energy flux; in-place mass-matrix
normalization of the DG tendency; degrees→radians in the sphere limiter's
slot-width test.

**Assertions.** Each example now asserts what it demonstrates: exact solutions
(column cases), conservation and stability (jets, channels, mountains), bounds
(limiters), bracketed rise speeds (bubbles), unchanged fields (balanced/at-rest
cases), with measured tolerances noted in place, and runs stopped before the
configuration becomes invalid (e.g., bubble hitting lid). Limiter error norms
are normalized by the initial tracer's norm rather than pointwise, and
assertions are rank-safe under MPI: root-only diagnostics are guarded, and
peak velocities are reduced across ranks.


test/Operators/hybrid/simulation_3d.jl
```diff
@@ -27,8 +30,13 @@ device = ClimaComms.device()
...
     Δt = 0.01
-    prob = ODEProblem(rhs!, U, (0.0, 2π))
-    sol = solve(prob, SSPRK33(), dt = Δt)
+    prob = CTS.ODEProblem(
+        CTS.ClimaODEFunction(; T_exp! = rhs!),
+        U,
+        (0.0, 2π),
+        nothing,
+    )
+    sol = CTS.solve(prob, CTS.ExplicitAlgorithm(CTS.SSP33ShuOsher()), dt = Δt)
```
imreddyTeja, line 40: `solve` calls `init` on `prob`. The CTS version of this does not copy `U` and modifies it in place. I remember running into this problem earlier. If I run this exact test, but make `rhs` set `dudt` to 1.0, it still passes.

imreddyTeja: An easy fix is just copying `U` before passing it to CTS

test/Operators/hybrid/simulation_3d.jl
```diff
@@ -27,8 +30,13 @@ device = ClimaComms.device()
     rhs!(dudt, U, nothing, 0.0)
```
imreddyTeja: Is `dudt` used anywhere after this?

examples/hybrid/box/bubble_3d_invariant_rhoe.jl
imreddyTeja, line 1: `bubble_3d_invariant_rhoe.jl` explicitally says it is the counterpart of this example. It would be good to make them more similar/unify them if possible. Some things I noticed:  
1. include for "example_utils.jl" at top vs bottom of script
2. params as global consts vs in a PhysicalParameters struct. I also think the physical consts are set to the same values
3. Does the flux form intenrionally use a different polynomial degree for the guadrature?
4. `Φ` vs `geopotential`

examples/hybrid/plane/schar_mountain.jl
```diff
@@ -0,0 +1,498 @@
...
+    gⁱʲ =
+        Fields.level(
+            Fields.local_geometry_field(hv_face_space),
+            ClimaCore.Utilities.half,
+        ).gⁱʲ
+    g33 = gⁱʲ.components.data.:9
+    u₃_bc = Geometry.Covariant3Vector.(-1 .* u₁_bc ./ g33)  # fw = -g^31 cuₕ/ g^33
+    apply_boundary_w =
+        Operators.SetBoundaryOperator(bottom = Operators.SetValue(u₃_bc))
+    @. fw = apply_boundary_w(fw)
```
imreddyTeja: This line modifies the `Y` directly. Is that OK to do during stage assemply?

examples/hybrid/box/bubble_3d_flux_form_rhoe.jl
```diff
@@ -0,0 +1,358 @@
...
+include(joinpath(@__DIR__, "../..", "example_utils.jl")) # linkfig
+
+linkfig(
+    relpath(joinpath(path, "energy.png"), joinpath(@__DIR__, "../..")),
+    "Total Energy",
+)
+linkfig(
+    relpath(joinpath(path, "mass.png"), joinpath(@__DIR__, "../..")),
+    "Mass",
+)
```
imreddyTeja, line 364: I don't think these are working correctly. When I checked the buildkite build, I see
```
[Image artifact 'hybrid/box/output/bubble_3d_flux_form_rhoe/energy.png' not found]
[Image artifact 'hybrid/box/output/bubble_3d_flux_form_rhoe/mass.png' not found]
```

.buildkite/Project.toml
```diff
@@ -46,7 +43,6 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
-SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
```
imreddyTeja, line 49: the main README still says the package is SciML compatible. Should that be updated?

imreddyTeja: I think there are a few other mentions of SciML or OrdinaryDiffEq

examples/hybrid/plane/schar_mountain.jl
```diff
@@ -0,0 +1,498 @@
...
+    Ic2f = Operators.InterpolateC2F(
+        bottom = Operators.Extrapolate(),
+        top = Operators.Extrapolate(),
+    )
+
+    f_upwind_product1 = Operators.UpwindBiasedProductC2F()
+    f_upwind_product3 = Operators.Upwind3rdOrderBiasedProductC2F(
+        bottom = Operators.FirstOrderOneSided(),
+        top = Operators.FirstOrderOneSided(),
+    )
```
imreddyTeja: The lines that use these operators are commented out. Should they be removed?

test/Operators/hybrid/simulation_3d.jl
tapios: Good catch. Removed the variable and the call.

examples/hybrid/box/bubble_3d_invariant_rhoe.jl
tapios: Good points. I unified the cases more with shared helpers. The flux form reads the shared defaults into constants rather than threading a parameter value. The polynomial order difference (4 vs 3) I suspect was incidental, not intentional; I made it  3 throughout.

examples/hybrid/plane/schar_mountain.jl
tapios: You're right that it isn't safe; the tendency modified the stage value the integrator was still combining. I moved it to the dss! projections. It leads to minor differences in test outcomes.

examples/hybrid/box/bubble_3d_flux_form_rhoe.jl
tapios: Wrong output path where these were generated; I fixed it.

.buildkite/Project.toml
tapios: ClimaCore still uses SciML in src/Fields/compat_diffeq.jl; we should remove this separately (added it to our roadmap). In the meantime, I removed this from the README and other places.

tapios: I removed the SciML compat shims; no one seems to be using them. However, this makes this branch breaking, and I'll leave a note.

examples/hybrid/plane/schar_mountain.jl
tapios: Removed, along with the commented-out lines that were their only "use"

.buildkite/Project.toml
imreddyTeja: I think Kevin needs to make a patch release with the multiple column space to continue his work. It might be easiest to remove sciml shims after the patch

test/Operators/hybrid/simulation_3d.jl
tapios: Ah right. I had fixed the same problem in a few other places but missed this one. Thank you!

.buildkite/Project.toml
tapios: Can Kevin make that patch release now, before this lands? (@ph-kev)

ph-kev: The work with the multiple column space already landed in ClimaCore v0.15.1.

imreddyTeja: The other concern I have is that a specific CMIP config is getting a bounds check error. I'm working on a fix now, but I'm happy to just make a back-port release.

src/Fields/fieldvector.jl
```diff
@@ -65,13 +64,16 @@ wrap(x::NamedTuple) = FieldVector(; pairs(x)...)
 This is called when calling `getproperty` on a `FieldVector` property of element
 type `T`.
 """
+recursive_bottom_eltype(field::Field) =
+    recursive_bottom_eltype(parent(field))
+
 unwrap(x) = x
```
imreddyTeja: So the docstring can be found automatically:
```suggestion
This is called when calling `getproperty` on a `FieldVector` property of element
type `T`.
"""
unwrap(x) = x

recursive_bottom_eltype(field::Field) =
    recursive_bottom_eltype(parent(field))
```

imreddyTeja (approved):

src/Fields/compat_diffeq.jl
```diff
@@ -1,52 +0,0 @@
...
-
-# This one is annoying
-#  https://github.com/SciML/OrdinaryDiffEq.jl/blob/181dcf265351ed3c02437c89a8d2af3f6967fa85/src/initdt.jl#L80
-Base.any(f, field::Field) = any(f, parent(field))
-
-Base.muladd(x, field::Field) = muladd.(x, field)
-
-Base.similar(field::F, ::Type{F}) where {F <: Field} = similar(field)
-
-Base.vec(field::Field) = vec(parent(field))
```
imreddyTeja, line 17: It might be a good idea to keep these. Removing `Base.any` is causing failures in downstream tests

tapios: Restored the base methods.

src/Fields/fieldvector.jl
tapios: Changed as suggested.

.buildkite/Manifest.toml
```diff
@@ -1,23 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.11"
+julia_version = "1.10.10"
```
dennisYatunin: Was this downgraded on purpose?

examples/column/advect_diffusion.jl
```diff
@@ -10,96 +19,107 @@ import ClimaCore:
...
-    saveat = collect(t₀:(10000 * Δt):t₁),
+    saveat = collect(t_start:FT(0.5):t_end),
     progress = true,
     progress_message = (dt, u, p, t) -> t,
 );
 
+using Test
+using LinearAlgebra: norm
+z_centers = vec(parent(Fields.coordinate_field(cspace).z))
+let
```
dennisYatunin: The `let` block is a bit unorthodox, both here and in other examples. If you're just trying to group the tests into one block, a `@testset` is more standard.

examples/column/limited_advection.jl
```diff
@@ -0,0 +1,187 @@
...
+        # The pulse is translated, not reshaped: the volume-weighted RMS
+        # error stays under 0.25, about half the pulse's own RMS amplitude
+        # of 0.45, and the zero-flux boundaries make the flux-form divergence
+        # conserve mass exactly.
+        @test norm(q_final .- q_analytic) ≤ 0.25
+        # 30 eps rather than 10: the GPU reduction sums in a different
+        # order and lands at ~12 eps (measured 2.8e-15 against 2.2e-15).
+        @test abs(sum(q_final) - sum(q_init)) / sum(q_init) ≤ 30eps()
+
+        if !(device isa ClimaComms.CUDADevice)
```
dennisYatunin: You can keep CUDA support by using `ClimaCore.to_device(ClimaComms.CPUSingleThreaded(), q_final)`.

examples/hybrid/plane/schar_mountain.jl
```diff
@@ -0,0 +1,499 @@
...
+hv_center_space, hv_face_space = hvspace_2D(
+    (-Lx / 2, Lx / 2),
+    (0, Lz);
+    xelem = nx,
+    zelem = nz,
+    npoly = np,
+    warp_fn = warp_schar,
+)
+
+Φ(z) = grav * z
```
dennisYatunin: I think you swapped the `Φ` character with `geopotential` elsewhere? Don't care much either way, but good to be consistent. Also, the `𝒩` could probably be replaced with a more standard `N`.

src/Fields/Fields.jl
```diff
@@ -176,6 +177,12 @@ Base.IndexStyle(::Type{<:Field{V}}) where {V} = IndexStyle(V)
 for f in (:parent, :size, :length, :ndims)
     @eval Base.$f(field::Field) = $f(field_values(field))
 end
+
+# Scalar reductions and views that delegate to the backing array. Generic and
+# downstream code (NaN checks, plotting) relies on these.
+Base.any(f, field::Field) = any(f, parent(field))
+Base.similar(field::F, ::Type{F}) where {F <: Field} = similar(field)
+Base.vec(field::Field) = vec(parent(field))
```
dennisYatunin:
```suggestion
Base.any(f, field::Field) = any(f, field_values(field))
Base.similar(field::F, ::Type{F}) where {F <: Field} = similar(field)
Base.vec(field::Field) = vec(field_values(field))
```
The `DataLayout` would be more correct than the backing array here.

src/Fields/fieldvector.jl
```diff
@@ -66,12 +65,15 @@ This is called when calling `getproperty` on a `FieldVector` property of element
 type `T`.
 """
 unwrap(x) = x
+
+recursive_bottom_eltype(field::Field) =
+    recursive_bottom_eltype(parent(field))
```
dennisYatunin, line 74:
```suggestion
    recursive_bottom_eltype(field_values(field))
```

dennisYatunin (approved):

.buildkite/Manifest.toml
tapios: Not on purpose, happened in a local re-resolve. Reverted.

examples/column/advect_diffusion.jl
tapios: Converted throughout.

examples/column/limited_advection.jl
tapios: Done

examples/hybrid/plane/schar_mountain.jl
tapios: Changed to plain N throughout. Kept `ᶜΦ` in staggered configurations to stick with model conventions.

src/Fields/Fields.jl
tapios: Applied for `vec`. Kept `parent` for `any`: it seems that on a struct-valued field (e.g. `Covariant3Vector` entries), `any(isnan, field_values(f))` throws `MethodError` (because `isnan` has no method for the entries the DataLayout iterates?).

src/Fields/fieldvector.jl
tapios: Tried it; it breaks `test/Fields/unit_field.jl` ("FieldVector basetype replacement and deepcopy"). Left a comment.

PR #2577, opened 2026-08-10
Update ClimaCoreTempestRemap compats

imreddyTeja:
It is no longer compatible with older ClimaCore versions because `DataLayouts.DataLayout` is only defined in the latest CC release. It also bumps the lower Julia and ClimaComms bounds to match that of the latest ClimaCore release

This should fix the Julia registration issues [here](https://github.com/JuliaRegistries/General/pull/164015)

### DIFF
--- a/lib/ClimaCoreTempestRemap/Project.toml
+++ b/lib/ClimaCoreTempestRemap/Project.toml
@@ -15,8 +15,8 @@ PkgVersion = "eebad327-c553-4316-9ea0-9fa01ccd7688"
 TempestRemap_jll = "8573a8c5-1df0-515e-a024-abad257ee284"
 
 [compat]
-ClimaComms = "0.5, 0.6"
-ClimaCore = "0.11, 0.12, 0.13, 0.14, 0.15"
+ClimaComms = "0.6"
+ClimaCore = "0.15"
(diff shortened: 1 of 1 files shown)


ph-kev (approved):
Do we need to do the same to any of the other `lib` packages?

imreddyTeja:
> Do we need to do the same to any of the other `lib` packages?

Good point. The other `lib` packages were able to be registered successfully. I briefly looked at `ClimaCorePlots` and `ClimaCoreMakie`, and they should have also had the lower compats increased. The Julia registry ci just didn't notice because it was able to successfully load the package. So they have already been released with incorrect lower compats...
What do you think should be done?

ph-kev:
> > Do we need to do the same to any of the other `lib` packages?
> 
> Good point. The other `lib` packages were able to be registered successfully. I briefly looked at `ClimaCorePlots` and `ClimaCoreMakie`, and they should have also had the lower compats increased. The Julia registry ci just didn't notice because it was able to successfully load the package. So they have already been released with incorrect lower compats... What do you think should be done?

The right thing to do is to open a PR in https://github.com/JuliaRegistries/General that fix the compat. See https://github.com/JuliaRegistries/General/pull/163449 as an example. It is up to you if you want to put in the effort to do this though.

PR #2578, opened 2026-08-10
Use a single Pkg.develop for CI

ph-kev:
This PR updates the CI to use a single `Pkg.develop` instead of multiple `Pkg.develop`. Each `Pkg.develop` results in resolving and updating the manifest so minimizing the number of times that `Pkg.develop` is called would speed up this up. It also makes the order of that packages are developed in not matter anymore since all of the developed packages are considered all at once.

### DIFF
--- a/.buildkite/perf/pipeline.yml
+++ b/.buildkite/perf/pipeline.yml
@@ -25,8 +25,11 @@ steps:
       - echo "--- Instantiate AMIP env"
       # Instantiate and dev install ClimaCore
       - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'
-      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.develop(path=".")'
-      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.develop(path="lib/ClimaCoreMakie")'
+      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg;
+        Pkg.develop([
+          PackageSpec(path="."),
+          PackageSpec(path="lib/ClimaCoreMakie"),
+        ])'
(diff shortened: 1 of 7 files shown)


.buildkite/perf/pipeline.yml
```diff
@@ -25,8 +25,11 @@ steps:
       - echo "--- Instantiate AMIP env"
       # Instantiate and dev install ClimaCore
       - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'
-      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.develop(path=".")'
-      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg; Pkg.develop(path="lib/ClimaCoreMakie")'
+      - julia --project=$COUPLER_PATH/experiments/AMIP/ -e 'using Pkg;
+        Pkg.develop([
+          PackageSpec(path="."),
```
imreddyTeja, line 30: What does using `PackageSpec` do?

imreddyTeja: Unrelated, but why does the performance pipeline dev `lib/ClimaCoreMakie`, but not any of the other `lib` packages?

imreddyTeja (approved):

.buildkite/perf/pipeline.yml
ph-kev: For the documentation of `Pkg.PackageSpec`, "A PackageSpec is a representation of a package with various metadata". As far as I am aware, it is a struct that describes a package like the name, version, path, etc. We need it since we can't pass in paths to `Pkg.develop` otherwise.

For the performance pipeline, it dev `lib/ClimaCoreMakie` because it broke the performance pipeline when @dennisYatunin was testing it before the release, but I don't think we need it anymore. I can remove it.

ph-kev: Wait, I think we might need it still since ClimaCoupler isn't updated to the latest ClimaCore.

PR #2582, opened 2026-08-12
Split downstream ClimaAtmos tests up

imreddyTeja:
ClimaAtmos.jl's test suite uses too much memory to run in a single gha process. I believe this is because the compiled acode and inference cache grows to over the max runner memory.

I don't know why, but I could only get it to work if each test group is run in its own step.

This does create more coupling between atmos and core, and is very repetitive, but at least we will be able to run the downstream atmos test.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -50,12 +50,62 @@ jobs:
       # from the package itself would cause the aqua test to fail. Instead, we install
       # TestEnv into the base environment, activate the test env for the downstream package,
       # and dev ClimaCore before running the test suite from there.
-      - if: (matrix.package != 'ClimaCoupler.jl')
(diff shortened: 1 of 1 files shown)


tapios (approved):
Looks fine to me. It'll become a bit brittle if the atmos test groups change, but I'm not sure there's an alternative to this. (Would it be possible to integrate the atmos buildkite pipeline and its grouping directly here?)

ph-kev (approved):

imreddyTeja:
> Looks fine to me. It'll become a bit brittle if the atmos test groups change, but I'm not sure there's an alternative to this. (Would it be possible to integrate the atmos buildkite pipeline and its grouping directly here?)

That should be possible, and more useful in my opinion. I usually manually run Atmos buildkite ci on my ClimCore PRs. 
I'll chat with @ph-kev about how to implement this.

imreddyTeja:
Closed in favor of #2585

PR #2586, opened 2026-08-14
Fix indexing of reduced-dimension fields in GPU finite difference ker…

imreddyTeja:
Fields that are missing one of an extruded space's dimensions read out of bounds when broadcast into a finite difference stencil on a GPU. A *level* field has no vertical dimension and a *column* field has no horizontal dimensions; both hold a single value along the dimensions they are missing and are meant to be broadcast across them.

```julia
space = TU.FaceExtrudedFiniteDifferenceSpace(Float64; zelem = 39, helem = 12)
a = zeros(space)
b = similar(ClimaCore.Fields.level(a, 1))          # Nv == 1
@. ClimaCore.Operators.GradientF2C()(a + b)        # reads b at v = 1:40
```

Under `--check-bounds=yes` this throws a `BoundsError` from inside the kernel, surfaced at the next synchronization. Without forced bounds checks the same read runs unchecked, past the end of the level field's data. The CPU path is unaffected.

## Root cause

Two independent defects, which is why the bug survives disabling eager evaluation.

**1. `Spaces.level` silently returned the whole extruded space on device-side spaces.**

Before a launch, `strip_space` replaces a level field's space with a `LevelPlaceholderSpace`, and the kernel rebuilds it with

```julia
reconstruct_placeholder_space(::LevelPlaceholderSpace, parent_space) =
    Spaces.level(parent_space, left_idx(parent_space)) # extract arbitrary level
```

Inside a kernel, `parent_space`'s grid is a `Grids.DeviceExtrudedFiniteDifferenceGrid`, but the `level` methods were typed on `ExtrudedFiniteDifferenceSpace2D`/`3D`, whose aliases require a host `Grids.ExtrudedFiniteDifferenceGrid`. No method matched, so the call fell through to the generic identity fallback `level(x, inds...) = x` in `src/interface.jl` and returned the extruded space unchanged:

| | host | device |
|---|---|---|
| `Spaces.level(space, 1)` | `SpectralElementSpace2D` | `ExtrudedFiniteDifferenceSpace` |
| `Operators.vidx(space, 69 + half)` | `1` | **`70`** |

`getidx` then read `field_data[70, i, j, h]` from data with `Nv == 1`. This affects both `copyto_stencil_kernel!` and `eager_copyto_stencil_kernel!`, since both reconstruct argument spaces this way.

**2. The eager kernel's `calc_level_val` read every field at the thread's own indices.**

Even with a correct level space, `calc_level_val(::Field, space)` indexed `data[v, i, j, h]` with the thread's `v, i, j, h` for every field leaf. So the eager kernel read a level field at every vertical index, and a column field at every horizontal index.

## Fix

* `src/Spaces/extruded.jl` — add `DeviceExtrudedFiniteDifferenceSpace2D`/`3D` aliases and extend both `level` methods to cover them. The aliases key on the directions spanned by the coordinates of the local geometry rather than on the layout's dimensionality: device grids do not store their horizontal grid, and grids extruded from both 1D and 2D horizontal grids use a 4-dimensional `VIJHWithF` (a `SliceXZSpace` just has `Nj == 1`). The coordinates do distinguish them — `LocalGeometry{(1, 3), XZPoint}` versus `LocalGeometry{(1, 2, 3), LatLongZPoint}`.
* `ext/cuda/operators_fd_eager.jl` — read the dimensions that a field is missing at index 1, matching `Operators.vidx` and `Operators.hindices`.

`Grids.column` is typed on the abstract grid type, so `Spaces.column` never had defect 1; only the eager kernel mishandled column fields.

## Origin

Defect 1 was introduced by #1987 (`Add LevelPlaceholderSpace`, v0.14.15), which was the first code to call `Spaces.level` on a device-side space. Before it, a level field's space fell through the generic `placeholder_space` fallback and reached the kernel concrete, so `vidx` returned 1. Defect 2 came in with #2466 (`FD Operator CUDAExt`), which added the eager kernel.

## Tests

Two regression tests, both of which fail on `main`:

* `test/Spaces/unit_spaces.jl` — "slicing device-side extruded spaces" covers the root cause directly, so it stays guarded if the finite difference kernels are rewritten. It checks a cubed-sphere space and an XZ slice, since only the slice distinguishes the two alias schemes. Without the fix it reports `ExtrudedFiniteDifferenceSpace === SpectralElementSpace2D` and `nlevels == 11` instead of `1`.
* `test/Operators/finitedifference/broadcasting_edge_cases.jl` — "Reduced-dimension arguments of a finite difference stencil" covers the behaviour end to end, with level and column arguments at `z_elem ∈ (10, 20)`, which with `Nq = 2` exercises the generic and eager kernels respectively. Without the fix, 3 of its 4 assertions error: the level case on both kernels, the column case on the eager kernel only.

The tests compare type names rather than using `isa` on a device-side space, because `show` of one calls `Topologies.print_context`, which has no `DeviceSideContext` method — a failing `isa` test errors while displaying its own operand instead of reporting the mismatch.

## Verification

Run on an A100 with `--check-bounds=yes`, on the `.buildkite` project.

Results match the CPU: `GradientF2C` bitwise identical, `DivergenceF2C` to ~1e-13 relative.

| Suite | Result |
|---|---|
| `Spaces/unit_spaces.jl` | pass (8/8 in the new testset) |
| `Operators/finitedifference/broadcasting_edge_cases.jl` | pass (4/4 new, 2/2 existing) |
| `unit_column.jl`, `unit_upwind_schemes.jl`, `unit_tensor.jl`, `unit_boundary_symmetry.jl` | pass |
| `MatrixFields/matrix_fields_broadcasting/test_scalar_1..13, 16, 17` | pass |
| `MatrixFields/matrix_fields_broadcasting/test_non_scalar_1..5` | pass |

`Fields/unit_field.jl` and `Operators/integrals.jl` were not run: `JET` was unavailable in the local environment. They do not touch the changed code paths.

## Follow-ups (not in this PR)

* `Spaces.slab` and `Spaces.horizontal_space` have the same host-only-alias shape and are plausibly wrong on device-side spaces in the same way. Nothing here exercises them.
* `Topologies.print_context` has no `DeviceSideContext` method, so a device-side extruded space cannot be shown.
* The identity fallbacks `level(x, inds...) = x` and `column(x, inds...) = x` in `src/interface.jl` are what turned a missing method into silently wrong indexing. Narrowing them would surface this class of bug loudly.

https://claude.ai/code/session_0157GSNKuMFgkQWhbFf8zD1P

### DIFF
--- a/src/Spaces/extruded.jl
+++ b/src/Spaces/extruded.jl
@@ -179,10 +179,26 @@ issubspace(subspace::FiniteDifferenceSpace, space::ExtrudedFiniteDifferenceSpace
     grid(subspace) === grid(space).vertical_grid ||
     (grid(subspace) isa Grids.ColumnGrid && grid(subspace).full_grid === grid(space))
 
-Base.@propagate_inbounds level(space::ExtrudedFiniteDifferenceSpace2D, v) =
-    SpectralElementSpace1D(level(grid(space), staggered_level_index(space, v)))
-Base.@propagate_inbounds level(space::ExtrudedFiniteDifferenceSpace3D, v) =
-    SpectralElementSpace2D(level(grid(space), staggered_level_index(space, v)))
+# This must also cover device-side spaces, which appear in place of their host
+# counterparts inside GPU kernels (see reconstruct_placeholder_space). Without a
+# method that matches, `level` falls back to the generic identity method, which
+# silently returns the extruded space itself. Device-side grids do not store
+# their horizontal grid, so the aliases above cannot tell how many horizontal
+# dimensions such a space spans; the directions spanned by the coordinates of
+# its local geometry, which host and device spaces share, stand in for that.
+Base.@propagate_inbounds level(space::ExtrudedFiniteDifferenceSpace, v) =
+    _level_space(
+        eltype(local_geometry_data(space)),
+        level(grid(space), staggered_level_index(space, v)),
+    )
+_level_space(::Type{<:Geometry.LocalGeometry{(1, 2, 3)}}, level_grid) =
+    SpectralElementSpace2D(level_grid)
+_level_space(
+    ::Type{
+        <:Union{Geometry.LocalGeometry{(1, 3)}, Geometry.LocalGeometry{(2, 3)}},
+    },
+    level_grid,
+) = SpectralElementSpace1D(level_grid)
 
 Base.@propagate_inbounds slab(space::ExtrudedFiniteDifferenceSpace, v, h) =
     SpectralElementSpaceSlab(
--- a/test/Spaces/unit_spaces.jl
+++ b/test/Spaces/unit_spaces.jl
@@ -324,6 +324,58 @@ end
 
 end
 
+# `level` and `column` have generic identity fallbacks, so a slice method that is
+# typed only on host grids does not error on a device-side space: it silently
+# returns the space it was handed. GPU kernels slice device-side spaces (see
+# Operators.reconstruct_placeholder_space), and a whole extruded space standing in
+# for one of its levels indexes a level field as if it spanned every level.
+@testset "slicing device-side extruded spaces" begin
+    FT = Float64
+    # a space with a 2D horizontal grid, and one with a 1D horizontal grid
+    spaces = (
+        ExtrudedCubedSphereSpace(
+            FT;
+            z_elem = 10,
+            z_min = 0,
+            z_max = 1,
+            radius = 10,
+            h_elem = 2,
+            n_quad_points = 2,
+            staggering = CellFace(),
+        ),
+        SliceXZSpace(
+            FT;
+            z_elem = 10,
+            x_min = 0,
+            x_max = 1,
+            z_min = 0,
+            z_max = 1,
+            periodic_x = false,
+            x_elem = 2,
+            n_quad_points = 2,
+            staggering = CellFace(),
+        ),
+    )
+    for space in spaces
+        expected =
+            space isa Spaces.ExtrudedFiniteDifferenceSpace3D ?
+            Spaces.SpectralElementSpace2D : Spaces.SpectralElementSpace1D
+        @test Spaces.level(space, 1) isa expected
+
+        @static if on_gpu
+            adapted_space = Adapt.adapt(CUDA.KernelAdaptor(), space)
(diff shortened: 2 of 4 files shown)


src/Spaces/extruded.jl
```diff
@@ -122,6 +122,34 @@ const FaceExtrudedFiniteDifferenceSpace3D = FaceExtrudedFiniteDifferenceSpace{
     <:Grids.ExtrudedFiniteDifferenceGrid{<:Grids.SpectralElementGrid2D},
 }
 
+# Device-side extruded grids do not store their horizontal grid, so the aliases
+# above cannot tell how many horizontal dimensions they span. The directions
+# spanned by the coordinates of their local geometry stand in for that: both
+# store their local geometry in a 4-dimensional layout, but a grid extruded from
+# a 1D horizontal grid spans one horizontal direction and the vertical one,
+# while a grid extruded from a 2D horizontal grid spans all three.
```
dennisYatunin: Does this not also apply to CPU-side spaces? I think you could use this instead of the preexisting aliases. 

Alternatively, it might be simpler to just check `eltype(local_geometry_data(space))` in the `level` function, instead of specializing on any of these verbose aliases.

test/Spaces/unit_spaces.jl
```diff
@@ -324,6 +324,62 @@ end
...
+        @test Spaces.level(space, 1) isa expected
+
+        @static if on_gpu
+            adapted_space = Adapt.adapt(CUDA.KernelAdaptor(), space)
+            # Compare type names rather than using isa: `show` of a device-side
+            # space calls Topologies.print_context, which has no method for a
+            # DeviceSideContext, so a failing isa test would error while trying
+            # to display the space it was given instead of reporting the type.
+            adapted_level = Spaces.level(adapted_space, 1)
+            @test nameof(typeof(adapted_level)) === nameof(expected)
```
dennisYatunin:
```suggestion
            @test adapted_level isa expected
```
This can be written more succinctly; same goes for the other `nameof` test below.

dennisYatunin (approved):
Good catch! I wonder if we even need the unsafe fallbacks for `level`/`column`/`slab` anymore; a lot of code that relies on them was removed in #2522 and #2571.

Issue #2485, opened 2026-04-09
Support Robin boundary conditions for finite difference derivative operators


valentinaschueller:
**Is your feature request related to a problem? Please describe.**
The finite difference derivative operators support `SetValue` and `SetGradient` boundary operators.
These allow to solve PDE problems with Neumann and Dirichlet boundary conditions (BC), see for example [this tutorial](https://clima.github.io/ClimaCore.jl/dev/tutorials/introduction/#3.1-Heat-equation-using-finite-differences).
A third, standard boundary condition for PDEs is the Robin boundary condition; a linear combination of Neumann and Dirichlet BCs given by: 
$\left.(a \nabla u \cdot n + b u)\right|_{\partial\Omega} = g$, 
with $a,b$ some given constants, $u$ our unknown on $\Omega$, $n$ the normal vector, $\partial\Omega$ the boundary of the domain, and $g$ a function on said boundary.

Right now, as far as I understand it is not possible (or at least not documented and not obvious to me) how one could enforce such a boundary condition.
Robin-type boundary conditions appear in climate modeling, especially in simplified models suitable for testing and analysis.[^1]

**Describe the solution you'd like**
It should be possible to provide such a boundary condition for a diffusion-like PDE.
Since it consists of a combination of Dirichlet and Neumann BC, and since Dirichlet BC have to be implemented via `SetValue` for gradient operators, I think it makes sense to implement this for the gradient operators.
I personally need this for the combination `GradientC2F` x `DivergenceF2C`.

With the stencils used, for the lower boundary this will take the form:
```
a*(u[1] - u[0]) + b u[0] = g
a*u[1] + (b-a) * u[0] = g
=> u[0] = (g - a*u[1]) / (b-a)
```
and thus can be implemented as
```
G(1/2) = u[1] - (g - a*u[1]) / (b-a)    (1)
```

Note that the Dirichlet/Neumann cases are recovered for `a=0,b=1`, `a=1,b=0`, respectively.

**Describe alternatives you've considered**
These are not really alternatives but here are some thoughts I have on this and what make it a bit complicated to get running:
- Implementing this as `SetRobin` would deviate in spirit from the existing finite difference boundary conditions which use the terms "value" and "gradient" instead of "Dirichlet" and "Neumann". `SetLinearCombination` might be an alternative but it is not obvious to understand
[...]

PR #2589, opened 2026-08-17
Attempt implementing Robin boundary conditions for FD operators

valentinaschueller:
This PR is supposed to address #2485, i.e., provide support for Robin conditions for the finite difference operators.
It adds a `SetRobin` boundary condition for the `GradientF2C` operator.

The implementation is not correct yet: It recovers Neumann (`SetGradient`), but not Dirichlet (`SetValue`) boundary conditions. I have added [a script](https://github.com/valentinaschueller/ClimaCore.jl/blob/robin_bc/test_robin.jl) to test this (which can be turned into unit tests later): 
The first row shows the comparison to three different Dirichlet conditions (`SetValue` vs. `SetRobin` with `a=0, b=1`, following the notation in #2485 ).
The second row shows the comparison to three different Neumann conditions (`SetGradient` vs. `SetRobin` with `a=1, b=0`).

image

It seems that the issue is related to the spaces I work with: `SetValue` works directly in the reference element space (with a `Covariant3Vector`), while `SetGradient` takes a boundary condition in physical space (a `WVector`) and projects it.

I am a bit stuck and would appreciate some help/input on what's a good next step!

### DIFF
--- a/src/Operators/finitedifference.jl
+++ b/src/Operators/finitedifference.jl
@@ -111,6 +111,15 @@ struct SetValue{S} <: AbstractBoundaryCondition
 end
 
 """
+    SetRobin(val)
+
+Set a Robin BC.
+"""
+struct SetRobin{S} <: AbstractBoundaryCondition
+    val::S
+end
+
+"""
     SetGradient(val)
 
 Set the gradient at the boundary to be `val`. In the case of gradient operators
@@ -2848,7 +2857,7 @@ The following boundary conditions are supported:
 struct GradientC2F{BC} <: GradientOperator
     bcs::BC
     function GradientC2F(; kwargs...)
-        assert_valid_bcs("GradientC2F", kwargs, (SetValue, SetGradient))
+        assert_valid_bcs("GradientC2F", kwargs, (SetValue, SetGradient, SetRobin))
         new{typeof(NamedTuple(kwargs))}(NamedTuple(kwargs))
     end
     GradientC2F(bcs) = GradientC2F(; bcs...)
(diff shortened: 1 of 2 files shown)


petebachant:
Hi @valentinaschueller, thanks for contributing!

This might be a good question for @dennisYatunin or @imreddyTeja. What do you guys think?

akshaysridhar:
Hi @valentinaschueller ; 
assuming one sided boundary stencils ; we have 
`G^{3} = 2 (u_{center} - u_{face})` (bottom boundary, similarly for the top with indices flipped) (so there is an incorrect scaling term in the case of the Dirichlet boundary limit) 

That scaling term can be recovered with the `Geometry.project()` function via 
```
scaling = Geometry.project(Geometry.WAxis(), Geometry.Covariant3Vector(1), <face local geometry>)[1]
```
and in our case is ~ 1/Δz ; 

with this scaling fix - I can recover : 

Screenshot 2026-08-20 at 11 44 36 AM

Thanks for adding support for the Robin boundary type! 

@imreddyTeja perhaps you could comment on whether we should include this before or after your boundary condition refactor is merged.

valentinaschueller:
@akshaysridhar Cool! I knew I was missing the factor two, but I did not know how to properly recover it with `Geometry.project`. I'll try doing what you did and push an update.

Regarding merging: There are definitely two things that should be changed:
1. Rescaling to avoid singularities in case $a=b$, $a=-b$. (See the last bullet boint in #2485 )
2. I think this way of instantiating the boundary condition is not obvious, I should probably at least add some typing constraints

Besides, I have not written documentation or a test case 🙂 

I guess my question for you is: Is this a feature you would be willing to include? Then I'm happy to prepare a proper PR, in coordination with the BC refactoring you mentioned. 
E.g., let me know if I should rebase to a different branch.

Note also this comment I made in #2485:
> Implementing this as SetRobin would deviate in spirit from the existing finite difference boundary conditions which use the terms "value" and "gradient" instead of "Dirichlet" and "Neumann". SetLinearCombination might be an alternative but it is not obvious to understand

Do you have thoughts on this?

valentinaschueller:
@akshaysridhar okay, it turns out I am still not smart enough to understand how to apply the scaling properly :(

How did you do it?
The way I understand it, we have to apply some factor 
`robin_value = Geometry.WVector(c*(u_center-u_face))`
that switches between `c = 2*scaling` in the Dirichlet limit and `c = 1` in the Neumann limit.

But what value does it take otherwise, what am I missing?

imreddyTeja:
@valentinaschueller Thanks for opening a PR! I'm hoping to merge PR #2544 soon, which makes some changes to the finite difference operators. Notably, the gradient operator will only be represented by the "operator matrix" form. This means affine boundary conditions that modify the input/use ghost points (non-zero Dirichlet) will no longer be supported as primitives. Instead, they can be represented using the existing `SetGradient`. I'm also adding helper functions that automatically do this. I made a quick example of this for robin boundary conditions that can be added in a follow up. It also appears to fix the scaling issue. The code looks like:

```
            x₁ = Fields.level(x, 1)
            z₁ = Fields.level(z_center, 1)
            # the bottom-face z, as a field on the first center level
            # (`LeftBiasedF2C(z)[1] = z[1/2]`)
            z_f = Fields.level(LeftBiasedF2C().(z_face), 1)
            g = @. (γ - α * x₁) / (β - α * (z₁ - z_f))
            SetGradient(Geometry.WVector.(g))
```

Here is my understanding of the scaling:

Starting with:
$(a u_\text{face} + b \nabla u_\text{face}) = g$ and $\nabla u_\text{face} = \frac{u_1 - u_\text{face}}{z_1 - z_\text{face}}$
$a (u_1 - (z_1 - z_\text{face}) * \nabla u_\text{face} ) + b \nabla u_\text{face}  = g$
$\nabla u_\text{face}  = \frac{g - a x_1}{b - a (z_1 - z_\text{face})}$

I believe this is equivalent to scaling through projections, as `∂x∂ξ[3,3]` is computed using the distance between faces or centers.

valentinaschueller:
Cool, thanks! I will close this PR and create a new one that's based on #2544, since it appears to be merged now :)

PR #2590, opened 2026-08-17
Clean up imports flagged by ExplicitImports

imreddyTeja:
Address every finding from `print_explicit_imports(ClimaCore; warn_implicit_imports = false)` that can be fixed from this repo:

- Remove stale explicit imports (imported but unused).
- Drop self-qualification, so names a module owns are referenced directly rather than through the module itself. Where a local variable shadowed the function being called (`mesh` in Topologies, `topology` in Spaces), rename the local instead.
- Access names through their owner module: `Base.rand`, `Base.inv`, `Base.unsafe_load`, `Base.read`, `DiffRules.diffrules`, and `StaticArraysCore.StaticArrayStyle`. DiffRules and StaticArraysCore become direct dependencies, since they were previously reached through ForwardDiff and StaticArrays.

`DataLayouts` still imports LLVM, which defines the `unsafe_load` methods for `Core.LLVMPtr`; only the name now comes from Base, which owns it.

Two sets of stale imports are kept because their uses are invisible to static analysis, and both are commented as such:

- `InputOutput` imports type names so that `read_type`'s `eval` can resolve unqualified type strings, which is what an HDF5 file records when its writer had those names in scope.
- `Fields` imports `@fused_direct` to re-expose it as `Fields.@fused_direct`, which tests and downstream code use.

The remaining findings are accesses to names that dependencies do not declare public (`Base.Broadcast` internals, `ClimaComms.device`, `Adapt.@adapt_structure`, ...), which this package cannot fix.

This was done by claude

### DIFF
--- a/src/DataLayouts/DataLayouts.jl
+++ b/src/DataLayouts/DataLayouts.jl
@@ -1,7 +1,7 @@
 module DataLayouts
 
-import Base: @propagate_inbounds
-import LLVM: unsafe_load
+import Base: @propagate_inbounds, unsafe_load
+import LLVM # for the `unsafe_load` methods on `Core.LLVMPtr`
 import StaticArrays
 import BlockArrays
 import Adapt
--- a/src/Fields/Fields.jl
+++ b/src/Fields/Fields.jl
@@ -5,6 +5,7 @@ import MultiBroadcastFusion as MBF
 import ..slab, ..slab_args, ..column, ..column_args, ..level, ..level_args
 import ..DebugOnly: call_post_op_callback, post_op_callback
 import ..DataLayouts: DataLayouts, DataLayout, DataStyle, PointIndex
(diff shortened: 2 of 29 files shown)


.buildkite/Manifest.toml
```diff
@@ -25,9 +25,9 @@ version = "0.4.5"
 
 [[deps.Accessors]]
 deps = ["CompositionsBase", "ConstructionBase", "Dates", "InverseFunctions", "MacroTools"]
-git-tree-sha1 = "2eeb2c9bef11013efc6f8f97f32ee59b146b09fb"
+git-tree-sha1 = "7063ad1083578215c7c4bf410368150abe8d5524"
```
ph-kev: I don't think the Manifest should be updated given the changes. If we are adding new packages, it might be good to do it in another PR or commit.

ph-kev (approved):

PR #2591, opened 2026-08-17
Fix warp-shuffles with tensors

imreddyTeja:
This was causing ClimaAtmos conservation checks to error. There might be a cleaner way of fixing this. I do not know why the offset needs to be wrapped in Int32, but this fix does not work without it.

Also adds a test for gpu reductions with tensors. This bug was not caught in climacore ci, climaatmos ci, or the climacoupler e2e perf test...

This PR will now release ClimaCore as well

### DIFF
--- a/ext/cuda/loops.jl
+++ b/ext/cuda/loops.jl
@@ -165,8 +165,11 @@ function shuffle_reduce(scope, op::O, value, num_values) where {O}
     num_offsets = num_reductions(scope)
     num_inactive = THREADS_PER_WARP - num_active_threads(ThisWarp())
     thread_index = DataLayouts.thread_rank(scope)
-    for offset in ntuple(Base.Fix1(>>, DataLayouts.num_threads(scope)), Val(num_offsets))
-        shuffled_value = CUDA.shfl_xor_sync(CUDA.FULL_MASK >> num_inactive, value, offset)
+    for offset in
+        ntuple(Base.Fix1(>>, DataLayouts.num_threads(scope)) ∘ Int32, Val(num_offsets))
+        # shfl_recurse fails if `offset` is not explicitly converted to an Int32
+        shuffled_value =
+            CUDA.shfl_xor_sync(CUDA.FULL_MASK >> num_inactive, value, offset)
         if thread_index <= num_values && xor(thread_index - 1, offset) + 1 <= num_values
             value = op(value, shuffled_value)
         end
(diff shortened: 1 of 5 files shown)


imreddyTeja:
This is an alt to #2587

szy21:
This works in ClimaAtmos.

ext/cuda/loops.jl
```diff
@@ -178,3 +179,5 @@ end
 # shuffling each value that appears in a multi-component reduction.
 CUDA.shfl_recurse(op::O, x::Utilities.AutoBroadcaster) where {O} =
     Utilities.AutoBroadcaster(UnrolledUtilities.unrolled_map(op, Utilities.unwrap(x)))
+# can only use map because unrolled_map is not defined for tensors
```
dennisYatunin: I think `map` for a `Tensor` should automatically be unrolled, but that would be worth double-checking at some point.
If you want consistency here, the right move would probably be to use `map` over the `AutoBroadcaster`, rather than use `unrolled_map` for both it and the `Tensor`.

ext/cuda/loops.jl
```diff
@@ -166,7 +166,8 @@ function shuffle_reduce(scope, op::O, value, num_values) where {O}
     num_inactive = THREADS_PER_WARP - num_active_threads(ThisWarp())
     thread_index = DataLayouts.thread_rank(scope)
     for offset in ntuple(Base.Fix1(>>, DataLayouts.num_threads(scope)), Val(num_offsets))
-        shuffled_value = CUDA.shfl_xor_sync(CUDA.FULL_MASK >> num_inactive, value, offset)
+        shuffled_value =
+            CUDA.shfl_xor_sync(CUDA.FULL_MASK >> num_inactive, value, Int32(offset))
```
dennisYatunin: The `Int32` conversion should probably be applied to `offset` everywhere, not just here. Maybe something like
```julia
offsets = ntuple(Base.Fix1(>>, DataLayouts.num_threads(scope)) ∘ Int32, Val(num_offsets))
```
I suppose then you'd also want to convert the `1`s on the line below for completeness. Not sure if anything else is secretly an `Int64` here.

dennisYatunin (approved):

tapios:
@imreddyTeja @dennisYatunin Please also check https://github.com/CliMA/ClimaCore.jl/pull/2587 to see if there's anything to be learned there. It was my attempt to deal with the same InvalidIRError.

ext/cuda/loops.jl
imreddyTeja: Good point. That comment was misleading because `map` over a `Tensor` does just use the unrolled version. I'll change it to one method for both `Tensors` and `AutoBroadcasters` that uses `map`

tapios:
> @imreddyTeja @dennisYatunin Please also check #2587 to see if there's anything to be learned there. It was my attempt to deal with the same InvalidIRError.

Nvm. What you have here looks good. Maybe we can add the test in #2587 later--that may still be useful.

ext/cuda/loops.jl
tapios: Maybe good to add an explicit comment that this fails for recursed types unless the offset is wrapped in Int32

tapios (approved):

PR #2595, opened 2026-08-21
Clean up lib packages and add extensions

ph-kev:
This PR cleans up the lib packages by deprecating ClimaCoreMakie and ClimaCorePlots. The CI for these are now tested as part of ClimaCore. ClimaCoreVTK is removed and ClimaCoreSpectra is updated to not have ClimaCore as a dependency.

This PR requires a breaking release of ClimaCore to avoid users loading both ClimaCoreMakie and the Makie extension which causes method overwrites. I haven't tested this, but I believe a similar thing would happen with Plots.

In this PR or in another PR, we can delete the lib packages for ClimaCoreMakie and ClimaCorePlots.

I asked a LLM to do this.


lib/ClimaCoreMakie/Project.toml
```diff
@@ -1,12 +1,12 @@
...
 uuid = "908f55d8-4145-4867-9c14-5dad1a479e4d"
-version = "0.4.8"
+version = "0.5.0"
 
 [compat]
-ClimaCore = "0.11, 0.12, 0.13, 0.14, 0.15"
+ClimaCore = "0.16"
 Makie = "0.24"
-julia = "1.7"
+julia = "1.10"
```
ph-kev, line 9: This was added to make any usage of ClimaCoreMakie work still. Do we want to have this or just force everyone to use the Visualize module instead when we make the breaking release of ClimaCore?

test/runtests.jl
```diff
@@ -176,6 +176,10 @@ unit_tests = [
     UnitTest("Aqua"                                     ,"aqua.jl"; tier = :misc, subsystem = :quality),
     UnitTest("Deprecations"                             ,"deprecations.jl"; tier = :misc, subsystem = :quality),
 
+    # Visualization
+    UnitTest("Makie extension"                          ,"Visualize/visualize.jl"; meta = :cpu_only, tier = :unit, subsystem = :other),
+    UnitTest("Plots extension"                          ,"Visualize/plots.jl"; meta = :cpu_only, tier = :unit, subsystem = :other),
+
```
ph-kev, line 188: These tests add method ambiguities if Makie and Plots are loaded.

nefrathenrici: Maybe these should be tagged as `subsystem = :io` or make a new `:viz` grouping. `:other` is not very descriptive.

ext/plots/recipes.jl
```diff
@@ -0,0 +1,522 @@
...
+    )
+    hspace_ortho = ClimaCore.Spaces.SpectralElementSpace1D(
+        htopo_ortho,
+        ClimaCore.Spaces.quadrature_style(hspace),
+    )
+    vspace_ortho = ClimaCore.Spaces.CenterFiniteDifferenceSpace(
+        Spaces.vertical_topology(space),
+    )
+
+    if space.staggering === ClimaCore.Spaces.CellFace
```
nefrathenrici: I think that this will always be false because it's comparing a type to an instance.

ext/plots/tripcolor.jl
```diff
@@ -0,0 +1,129 @@
...
+) where {T}
+    nv = length(x)
+    nt = size(t, 2)
+    @assert (size(x) == (nv,) && size(y) == (nv,) && size(z) == (nv,) && size(t) == (3, nt)) "Bad input sizes"
+
+    dg_z = similar(t, eltype(z))
+    for it in 1:nt
+        dg_z[:, it] .= @views z[t[:, it]]
+    end
+    dgtripcolor(x, y, dg_z, t, cmap; px, py, zmin, zmax, bg)
```
nefrathenrici:
```suggestion
    dgtripcolor(x, y, dg_z, t, cmap; px, py, zmin, zmax, yflip, bg)
```

This currently doesn't forward `yflip`.

nefrathenrici (commented):
Thank you for cleaning this up! I had a few minor comments.

ext/ClimaCoreMakieExt.jl
```diff
@@ -0,0 +1,14 @@
+module ClimaCoreMakieExt
+
+import Makie: Makie, @recipe, lift, GLTriangleFace, Point3f, Observable
+import ClimaCore
```
imreddyTeja, line 4: There are is a use of `Space.global_geometry`. Was this broken before?

ext/ClimaCoreMakieExt.jl
```diff
@@ -0,0 +1,14 @@
+module ClimaCoreMakieExt
+
+import Makie: Makie, @recipe, lift, GLTriangleFace, Point3f, Observable
```
imreddyTeja: Can't leave a comment in fieldline.jl so I'm leaving it here. I think there is a bug from when it was updated to work with the datalayouts refactor. `vertices` is indexed into as a 3 dim array, when it is now a 5 dim. I think this would throw an error. Could you fix that as well

ext/plots/recipes.jl
```diff
@@ -0,0 +1,522 @@
...
+
+    xdata = parent(field)[:, 1]
+    ydata = parent(Spaces.coordinates_data(space).z)[:, 1]
+
+    coord_symbols = propertynames(coord_field)
+
+    # set the plot attributes
+    xguide --> "value"
+    yguide --> (
+        if field isa Spaces.FaceFiniteDifferenceSpace
```
imreddyTeja: I think this is always false

Project.toml
```diff
@@ -35,9 +35,14 @@ UnrolledUtilities = "0fe1646c-419e-43be-ac14-22321958931b"
...
 GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
 Krylov = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 
 [extensions]
 ClimaCoreCUDAExt = "CUDA"
+ClimaCoreMakieExt = "Makie"
+ClimaCorePlotsExt = ["Plots", "RecipesBase"]
```
imreddyTeja: Do we need to require Plots.jl here? Or is that to make it clear to users that adding Plots will load the ext?

NEWS.md
```diff
@@ -208,6 +208,16 @@ main
   components) are example code. Operator tests are included as part of
   `test/runtests.jl`.
 
+- ![][badge-✨feature/enhancement] Makie plotting has moved from the
```
imreddyTeja: Should these be marked as breaking? And I think an entry should be added for the deleted VTK stuff. The entries could probably be combined into one bulltet point as well

ext/ClimaCoreMakieExt.jl
```diff
@@ -0,0 +1,14 @@
+module ClimaCoreMakieExt
+
```
imreddyTeja, line 2: Could you also delete the unused imports. I think `Observable` and `lift` are unused... I also think I may have been the one who added them

imreddyTeja (approved):
All of the changes look good. The only real comment I had was about the `Plots` weak-dep. The other comments are pre-existing issues that would be nice to fix here as well

imreddyTeja (approved):
Only important comments are about the plots extension weak dep and news entry. The others are pre-existing issues that would be good to clean up as well. thanks for dealing with this!

Project.toml
ph-kev: I guess we don't need Plots. I can make it to `ClimaCoreRecipesBaseExt`.

ext/ClimaCoreMakieExt.jl
ph-kev: The function using `Spaces.global_geometry` was never called and it wouldn't have work since there was never an import for `ClimaCore.Spaces`. I decided to have it fully qualified.

PR #2596, opened 2026-08-21
Add remapping for multiple column space

ph-kev:
This PR adds the functionality to do vertical interpolation for a multiple column space to get back an array whose size is the number of columns by the number of vertical levels. Horizontal interpolation is not supported for this space.

I asked a LLM to implement this.

### DIFF
--- a/ext/cuda/remapping_distributed.jl
+++ b/ext/cuda/remapping_distributed.jl
@@ -2,6 +2,7 @@ import ClimaCore: Topologies, Spaces, Fields, Quadratures
 import ClimaComms
 import CUDA
 using CUDA: @cuda
+import UnrolledUtilities
 import ClimaCore.Remapping: _set_interpolated_values_device!, bilinear, linear
 
 function ClimaCore.Remapping._set_interpolated_values_bilinear!(
@@ -393,15 +394,16 @@ function set_interpolated_values_kernel!(
         v_lo, v_hi = vert_bounding_indices[j]
         A, B = vert_interpolation_weights[j]
 
-        val = zero(eltype(out))
-        for t in 1:Nq
-            val +=
-                I[i, t] * (
-                    A * field_values[k][v_lo, t, 1, h] +
-                    B * field_values[k][v_hi, t, 1, h]
-                )
-        end
-        out[i, j, k] = val
+        out[i, j, k] =
+            UnrolledUtilities.unrolled_applyat(k, field_values) do fvals
+                val = zero(eltype(out))
+                for t in 1:Nq
+                    val +=
+                        I[i, t] *
+                        (A * fvals[v_lo, t, 1, h] + B * fvals[v_hi, t, 1, h])
+                end
+                val
+            end
     end
     return nothing
 end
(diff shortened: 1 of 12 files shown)


ph-kev:
This PR requires some more work for the `HDF5Writer` and other missing methods for the spaces.

src/Remapping/distributed_remapping.jl
```diff
@@ -569,7 +587,64 @@ function _Remapper(
...
+
+    num_columns = Spaces.ncolumns(space)
+
+    vert_interpolation_weights =
+        ArrayType(vertical_interpolation_weights(space, target_zcoords))
+    vert_bounding_indices =
+        ArrayType(vertical_bounding_indices(space, target_zcoords))
+
+    # Use all columns
+    local_target_hcoords_bitmask = trues(num_columns)
```
imreddyTeja: I think the `isa_vertical_space` type check should to be expanded in `interpolate!` below, otherwise this skips the path that single columns would normally take.

ext/cuda/remapping_distributed.jl
```diff
@@ -393,15 +394,16 @@ function set_interpolated_values_kernel!(
         v_lo, v_hi = vert_bounding_indices[j]
         A, B = vert_interpolation_weights[j]
```
imreddyTeja, line 396: why is this change needed? The other interpolation kernels should probably be modified to use `unrolled_applyat`

src/Spaces/multicolumn.jl
```diff
@@ -139,6 +141,7 @@ nlevels(space::MultiColumnFiniteDifferenceSpace) =
 
 horizontal_space(space::MultiColumnFiniteDifferenceSpace) =
     PointCloudSpace(grid(space).horizontal_grid)
+horizontal_space(space::PointCloudSpace) = space
```
imreddyTeja: I think you need to add a `default_target_hcoords` method, otherwise this will make the fallback recurse infinitely

ext/cuda/remapping_distributed.jl
ph-kev: When interpolating multiple fields of different types, I ran into the issue of multiple dispatch in the kernel.

imreddyTeja (approved):
Have you ran a test with this functionality using MPI yet?

ph-kev:
> Have you ran a test with this functionality using MPI yet?

I don't think it is possible to make the space with MPI enabled.

src/Remapping/distributed_remapping.jl
```diff
@@ -569,7 +587,63 @@ function _Remapper(
...
         buffer_length,
         colons,
-        SpectralElementRemapping(), # no horizontal interpolation
+        nothing, # horiz_method: no horizontal interpolation
+    )
+end
+
+# Constructor for the case with multiple column space (horizontal_method
+# accepted, ignored)
+function _Remapper(
```
nefrathenrici, line 596: No `_Remapper` method currently takes `target_hcoords::Nothing` together with `target_zcoords::Nothing`, so `level(cspace, k)` will error because it returns a `PointCloudSpace` which has `nothing` for both fields. Could you add a clear error or a _Remapper method to support this?

nefrathenrici (approved):
Thanks! One non-blocking comment

src/Remapping/distributed_remapping.jl
ph-kev: I added a `_Remapper` method for `Union{Spaces.AbstractPointSpace, Spaces.PointCloudSpace}` to throw an error.

PR #2599, opened 2026-08-22
Adds infra to mark spaces as DG

tapios:
Makes it possible to mark discontinuous-Galerkin function spaces, which makes DSS and related CG operations a no-op (and ClimaAtmos and others can use that property to gate DSS).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,9 @@ main
   computed. The new queries `Grids.is_continuous(grid)` /
   `Spaces.is_continuous(space)` report the discretization, so downstream
   models can gate DSS on the space itself rather than on the quadrature type.
-  The flag is part of the grid cache key.
+  The flag is part of the grid cache key and is serialized by `InputOutput`
+  (grids in files written before it existed read back as continuous).
+  [2599](https://github.com/CliMA/ClimaCore.jl/pull/2599)
 
 v0.15.3
 -------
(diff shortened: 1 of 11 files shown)


akshaysridhar:
Thanks @tapios ; I've stacked and tested this against the simple unit tests in `as/dg-fluxes` branch and the change looks largely reasonable. 

One concern I have is whether the `restart` feature would capture this kwarg consistently (In the HDF5 write and read steps in the `InputOutput` module it seems we [explicitly write properties](https://github.com/CliMA/ClimaCore.jl/blob/505982708ffd1bdb60738614895d59c2b4f7cfc9/src/InputOutput/writers.jl#L394)), which means we should add the `discontinuous` attribute to the `write_new!` methods corresponding to the `SpectralElementGrid1D / 2D`

tapios:
> Thanks @tapios ; I've stacked and tested this against the simple unit tests in `as/dg-fluxes` branch and the change looks largely reasonable.
> 
> One concern I have is whether the `restart` feature would capture this kwarg consistently (In the HDF5 write and read steps in the `InputOutput` module it seems we [explicitly write properties](https://github.com/CliMA/ClimaCore.jl/blob/505982708ffd1bdb60738614895d59c2b4f7cfc9/src/InputOutput/writers.jl#L394)), which means we should add the `discontinuous` attribute to the `write_new!` methods corresponding to the `SpectralElementGrid1D / 2D`

Thank you. Good point, now addressed in the branch.

akshaysridhar (approved):

PR #2601, opened 2026-08-24
Target occupancy with `__launch_bounds__` instead of a register cap

petebachant:
Trying to speed up the very slow `set_microphysics_tendency_cache` 1M microphysics kernel without changing Atmos/CM.

A kernel whose register pressure holds it below a target occupancy is now recompiled with .maxntid/.minnctapersm set to that target, asking ptxas to fit it into the corresponding register budget.

Claude-Session: https://claude.ai/code/session_01A5D7Y4Ux6yvvYgsUipJEX7

### DIFF
--- a/ext/cuda/cuda_utils.jl
+++ b/ext/cuda/cuda_utils.jl
@@ -132,6 +132,25 @@ function max_active_blocks(threads_per_block, regs_per_thread, shmem_per_block)
     return (max_blocks_per_sm * attrs.sm_count, limit)
 end
 
+# The number of warps per multiprocessor that a given register count allows,
+# which is the quantity `__launch_bounds__` lets us ask ptxas to target. Register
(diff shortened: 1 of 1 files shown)


petebachant:
@Mikolaj-A-Kowalski FYI, I'm also experimenting with ideas for speeding up that kernel that you're working on. https://github.com/CliMA/CloudMicrophysics.jl/pull/769 was significantly beneficial until a few days ago (not sure where in the stack the benefits were diminished).

petebachant:
Closing for now. Using occupancy as a target does work with this approach and helps improve some kernels but it doesn't help the hot microphysics cache kernel ([more info](https://calkit.io/petebachant/clima-gpu-profiling?expanded_question=3&ref=342ba0c)) and AMIP speedup is negligible.

Could revisit in the future but I think I'm going to head back to looking at CM.

PR #2602, opened 2026-08-26
Add horizontal DG operators, CUDA kernels, and unit tests

akshaysridhar:
1) Add horizontal DG operators (includes CUDA extensions) 
2) Add unit tests for DG operators on extruded sphere grids 
3) Move Roe / Shallow-water flux helpers to utils_dg.jl 

This is the first in a series of PR adding horizontal DG operator support in ClimaCore. 
Co-author : @tapios  (This branch combines ts/dg + as/dg-fluxes)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,20 @@ main
   (grids in files written before it existed read back as continuous).
   [2599](https://github.com/CliMA/ClimaCore.jl/pull/2599)
 
+- Introduces horizontal discontinuous-Galerkin (DG) operator support. The
+  discretization-generic machinery lives in `src/Operators/numericalflux.jl`:
+  interior- and boundary-face numerical fluxes and symmetric face liftings on
+  pure-2D and extruded (1D and 2D horizontal) spectral-element spaces, the
+  flux-differencing (split-form / FDDG) volume divergence of Souza et al.
+  (2023), and the device-resident `DGConnectivity` face buffer (with
+  CUDA implementations in the `ClimaCoreCUDAExt` extension). A generic flux
+  library lives in `src/Operators/dg_fluxes.jl`: `CentralNumericalFlux`,
+  `RusanovNumericalFlux`, central-lifting and jump-penalty face functions, and
+  an LDG/SIPG Laplacian for optional scale-selective dissipation.
+  Equation-set-specific fluxes (compressible Euler with Cartesian momentum
+  components) are example code. Operator tests are included as part of
+  `test/runtests.jl`.
+
 v0.15.3
 -------
 
--- a/ext/cuda/operators_dg.jl
+++ b/ext/cuda/operators_dg.jl
@@ -1,0 +1,326 @@
+#=
+CUDA implementations of the DG internal-face and flux-differencing volume
+operators (see src/Operators/numericalflux.jl for the CPU methods and the
+operator contracts).
+
+Kernel design:
+- Flux-differencing volume (`_add_flux_differencing_divergence!`), element-local.
+- Internal-face fluxes (`_add_numerical_flux_internal!`,
+  `_add_lifting_flux_internal!`): two-pass staging + gather, follows `DSS` GPU kernels.
+- `tensor_product!` (cutoff filter): per-element enabling in-place assignment.
+=#
+
+import ClimaCore: Operators, Topologies, Quadratures, Grids, DataLayouts
+import ClimaCore.Operators: DGConnectivity
+
+# ---------------------------------------------------------------------------
+# Flux-differencing volume divergence
+# ---------------------------------------------------------------------------
+
+function Operators._add_flux_differencing_divergence!(
+    ::ClimaComms.CUDADevice,
+    fn2pt::F,
+    dydt,
+    y,
+) where {F}
+    space = axes(dydt)
+    grid = Spaces.grid(space)
+    if grid isa Grids.ExtrudedFiniteDifferenceGrid
+        @assert grid.horizontal_grid isa Grids.SpectralElementGrid2D
+        Nv = Spaces.nlevels(space)
+    else
+        @assert grid isa Grids.SpectralElementGrid2D
+        Nv = 1
+    end
+    quadrature_style = Spaces.quadrature_style(space)
+    Nq = Quadratures.degrees_of_freedom(quadrature_style)
+    FT = Spaces.undertype(space)
+    D = Quadratures.differentiation_matrix(FT, quadrature_style)
+    (_, w) = Quadratures.quadrature_points(FT, quadrature_style)
+    Nh = Topologies.nlocalelems(Spaces.topology(space))
+
+    dydt_data = Fields.field_values(dydt)
+    y_data = Fields.field_values(y)
+    lg_data = Spaces.local_geometry_data(space)
+
+    Nvt = max(1, min(fld(_max_threads_cuda(), Nq * Nq), Nv))
+    args = (dydt_data, y_data, lg_data, fn2pt, D, w, Val(Nq), Nv)
+    auto_launch!(
+        dg_fddg_volume_kernel!,
+        args;
+        threads_s = (Nq, Nq, Nvt),
+        blocks_s = (Nh, cld(Nv, Nvt)),
+    )
+    return dydt
+end
+
+function dg_fddg_volume_kernel!(
+    dydt_data,
+    y_data,
+    lg_data,
+    fn2pt::F,
+    D,
+    w,
+    ::Val{Nq},
+    Nv,
+) where {F, Nq}
+    i = threadIdx().x
+    j = threadIdx().y
+    h = blockIdx().x
+    v = threadIdx().z + (blockIdx().y - Int32(1)) * blockDim().z
+    if v ≤ Nv
+        CI = CartesianIndex
+        y_at = (a, b) -> y_data[CI(v, a, b, h)]
+        lg_at = (a, b) -> lg_data[CI(v, a, b, h)]
+        total = Operators._fd_volume_node_total(
+            fn2pt,
+            y_at,
+            lg_at,
+            D,
+            w,
+            Val(Nq),
+            i,
+            j,
+        )
+        I = CI(v, i, j, h)
+        dydt_data[I] = Operators._fd_add(dydt_data[I], total)
+    end
+    return nothing
+end
+
+# ---------------------------------------------------------------------------
+# Internal-face numerical flux and symmetric lifting (staging + gather)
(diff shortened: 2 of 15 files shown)


test/Operators/spectralelement/utils_dg.jl
```diff
@@ -51,3 +58,86 @@ function dg_divergence(uv)
...
+roe_average(ρ⁻, ρ⁺, v⁻, v⁺) =
+    (sqrt(ρ⁻) * v⁻ + sqrt(ρ⁺) * v⁺) / (sqrt(ρ⁻) + sqrt(ρ⁺))
+
+"""
+    sw_roeflux(normal, (y⁻, params⁻), (y⁺, params⁺))
+
+Shallow-water Roe numerical flux for the conserved state `(; ρ, ρu, ρθ)`.
+Central flux of `sw_flux` plus Roe dissipation (wave strengths `w1`–`w5`).
+Used by `unit_two_point_fluxes.jl` as a custom functor in place of a packaged
+`RoeNumericalFlux`.
```
tapios:
```suggestion
Used by `unit_two_point_fluxes.jl` as a custom functor.
```
(I don't think we have a packaged Roe flux.

tapios (approved):
LGTM.

Would be good if @dennisYatunin/@imreddyTeja can have a look at operators_dg.jl and spectralement.jl.

test/Operators/spectralelement/utils_dg.jl
akshaysridhar: Leftover from Roe being moved out of this PR branch (it's still available in `as/dg-fluxes` while I test the full system of equations in climacore).

src/Operators/dg_fluxes.jl
```diff
@@ -0,0 +1,141 @@
...
+central_gradient_lift(normal, (q⁻,), (q⁺,)) = ((q⁺ - q⁻) / 2) * normal
+
+"""
+    central_curl3_lift(normal, (u⁻, v⁻), (u⁺, v⁺))
+
+Central lifting for the radial component of the horizontal curl:
+``r̂ ⋅ (n̂ × (u^* - u_{side}))`` from the tangential jumps of the orthonormal
+velocity components `(u, v)`.
+"""
+central_curl3_lift(normal, (u⁻, v⁻), (u⁺, v⁺)) =
```
imreddyTeja, line 63: this is untested

src/Operators/dg_fluxes.jl
```diff
@@ -0,0 +1,141 @@
...
+        normal.components.data.:2 * (u⁺ - u⁻)
+    ) / 2
+
+"""
+    jump_penalty_lift(normal, (q⁻, λ⁻), (q⁺, λ⁺))
+
+λ-scaled interface penalty: each side relaxes toward its neighbor at rate
+``\\max(λ⁻, λ⁺)/2``.
+"""
+jump_penalty_lift(normal, (q⁻, λ⁻), (q⁺, λ⁺)) = max(λ⁻, λ⁺) / 2 * (q⁺ - q⁻)
```
imreddyTeja, line 75: same comment about being untested

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+    conn = Operators.dg_connectivity(space)
+    conn.nfaces == 0 && return dydt
+
+    dydt_data = Fields.field_values(dydt)
+    args_data =
+        map(a -> a isa Fields.Field ? Fields.field_values(a) : a, args)
+    T = eltype(dydt_data)
+    DA = ClimaComms.array_type(Spaces.topology(space))
+    nsides = mode isa Val{:lifting} ? 2 : 1
+    staging = DA{T}(undef, Nq, Nv, nsides, conn.nfaces)
```
imreddyTeja, line 130: this re-allocates every call. Should it be cached?

NEWS.md
```diff
@@ -16,6 +16,20 @@ main
   (grids in files written before it existed read back as continuous).
   [2599](https://github.com/CliMA/ClimaCore.jl/pull/2599)
 
+- Introduces horizontal discontinuous-Galerkin (DG) operator support. The
```
imreddyTeja, line 19: Should docs be added for this?

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+    nfaces,
+    mode,
+) where {F, Nq}
+    gidx = threadIdx().x + (blockIdx().x - Int32(1)) * blockDim().x
+    if gidx ≤ Nq * Nv * nfaces
+        (q, v, f) = Utilities.cart_ind((Nq, Nv, nfaces), gidx).I
+        elem⁻ = Int(faces[1, f])
+        face⁻ = Int(faces[2, f])
+        elem⁺ = Int(faces[3, f])
+        face⁺ = Int(faces[4, f])
```
imreddyTeja: Are these explicit conversions to `Int`s needed? If so, maybe they shoukd be `Int32`s for consistency

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+                a isa DataLayouts.DataLayout ?
+                a[CI(v, i⁻, j⁻, elem⁻)] : a,
+            args_data,
+        )
+        argvals⁺ = map(
+            a ->
+                a isa DataLayouts.DataLayout ?
+                a[CI(v, i⁺, j⁺, elem⁺)] : a,
+            args_data,
+        )
```
imreddyTeja, line 208: It should be fine as it is now, but these might need to be converted to unrolled_map in the future if `args` becomes complex

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+            conn.nbnodes,
+            mode,
+        );
+        threads_s = pB.threads,
+        blocks_s = pB.blocks,
+    )
+    return dydt
+end
+
+function dg_face_flux_kernel!(
```
imreddyTeja: for performance it might be a good idea to use `Base.@propagate_inbounds` and @inbounds to prevent bounds checking of array accesses

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+        e = Int(node_elem[n])
+        i = Int(node_i[n])
+        j = Int(node_j[n])
+        I = CartesianIndex(v, i, j, e)
+        acc = dydt_data[I]
+        for c in Int(node_offset[n]):(Int(node_offset[n + 1]) - 1)
+            f = Int(contrib_face[c])
+            q = Int(contrib_q[c])
+            side = Int(contrib_side[c])
+            if mode isa Val{:numflux}
```
imreddyTeja, line 249: I like this "mode" pattern. Would it make sense to do the same thing in the cpu code to avoid duplication? A lot of the functions in numericalflux.jl look very similar to each other (they only differ by ~8 lines)

ext/cuda/operators_dg.jl
```diff
@@ -0,0 +1,323 @@
...
+    auto_launch!(
+        dg_tensor_product_kernel!,
+        (out, indata, M, Val(Nij), Val(Nvt), Nv);
+        threads_s = (Nij, Nij, Nvt),
+        blocks_s = (Nh_runtime, cld(Nv, Nvt)),
+    )
+    return out
+end
+
+function Operators.tensor_product!(
```
imreddyTeja, line 287: I think the cpu version of the in-place tensor product needs to be updated too (but that is probably outside this PRs scope)

src/Operators/numericalflux.jl
```diff
@@ -138,3 +365,597 @@ function add_numerical_flux_boundary!(fn, dydt, args...)
...
+        (total = _fd_add(total, fn2pt(-w[i] * Ja2, -w[i] * Ja2, y_ij, y_ij)))
+    j == Nq &&
+        (total = _fd_add(total, fn2pt(w[i] * Ja2, w[i] * Ja2, y_ij, y_ij)))
+    return total
+end
+
+function _fd_divergence_slab!(fn2pt::F, dydt_slab, y_slab, lg_slab, D, w, Nq) where {F}
+    y_at = (a, b) -> y_slab[1, a, b, 1]
+    lg_at = (a, b) -> lg_slab[1, a, b, 1]
+    vNq = Val(Nq)
```
imreddyTeja: Have you checked this function for allocations? If not, I can do that. I'm thinking the closures might cause allocations, unless a let block is used. The `vNq = Val(Nq)` also won't be known at compile time because `Nq` is passed in to this function as an Int.

imreddyTeja (approved):
I left some minor comments. The gpu code looks good to me. I only looked at the software functionality of this, and not the mathematical correctness. I did have claude look at the math, and it claimed there is a minor issue:


`ldg_laplacian_tendency` does not converge because it is missing a consistency term.


I'm not sure if that is a real issue or not, but I'm leaving it here.

ext/cuda/operators_dg.jl
tapios: Yes, it's fixed here: https://github.com/CliMA/ClimaCore.jl/pull/2603. @akshaysridhar you can copy it from the MPI branch, or merge this and it'll be fixed in the next PR.

ext/cuda/operators_dg.jl
akshaysridhar: Yes - I'll do that : the general formulation of these fluxes is indeed similar, f(q+, q-, c)

src/Operators/numericalflux.jl
akshaysridhar: Not in addition to allocation checks that already exist in the ClimaCore unit tests. I can introduce this check.

ext/cuda/operators_dg.jl
akshaysridhar: I'm aiming to address the remaining comments and merge this PR first - so #2603 can contain additional refactor / allocation reductions without duplication here

ext/cuda/operators_dg.jl
akshaysridhar: @imreddyTeja cpu-version has been updated in this PR.

ext/cuda/operators_dg.jl
akshaysridhar: `faces[]` is already Int32 ; so I think I can safely drop `Int()` here.

ext/cuda/operators_dg.jl
akshaysridhar: Largely unified (will look over this in future PRs once I start building DG examples for extruded meshes in the upcoming PRs - I'll also add to the docs)

PR #2603, opened 2026-08-27
Enable MPI for DG and improve performance by limiting ghost exchange to faces

tapios:
Adds multi-rank distributed MPI support for DG operators for CPU and CUDA, along with communication optimizations and zero-allocation improvements:

1. **Distributed DG Support (CPU & GPU)**
   - Implemented distributed halo exchange and ghost-face kernels for DG operators (`add_numerical_flux_internal!`, `add_lifting_flux_internal!`, and `add_numerical_flux_boundary!`) across CPU and CUDA backends.
2. **Compact Face-Strip Halo Exchanges**
   - Replaced full-element slab communication with face-only halo buffers (`Topologies.GhostFaceExchange`), exchanging only the $N_q$ boundary face nodes rather than $N_q \times N_q$ full elements.
3. **Computation/Communication Overlap**
   - Structured halo exchanges so non-blocking communication is initiated before interior-face loops/kernels and completed just before ghost-face application, overlapping communication with local compute (20-30% performance improvement on 4 ranks).
4. **Shared Halo Exchanges Across Operators**
   - Added `Operators.start_dg_ghost_exchange(args...)` and the `ghost_exchange` keyword argument, enabling multiple face operators in the same stage (e.g., numerical flux and lifting flux) to reuse a single round of halo exchange.
5. **Zero-Allocation Operators**
   - Replaced tuple `map` calls with `UnrolledUtilities.unrolled_map` and introduced a constant no-op exchange handle (`NO_DG_GHOST_EXCHANGE`) for singleton comms contexts, achieving 0 heap allocations across CPU DG volume, numerical flux, lifting flux, and boundary flux operators.

### DIFF
--- a/.buildkite/Manifest.toml
+++ b/.buildkite/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.11"
+julia_version = "1.10.10"
 manifest_format = "2.0"
 project_hash = "19194cb825213fa6ff692385eb4c7eed7f946f0f"
 
@@ -347,9 +347,9 @@ weakdeps = ["CUDA", "MPI"]
 
 [[deps.ClimaCore]]
 deps = ["Adapt", "BandedMatrices", "BlockArrays", "ClimaComms", "ClimaInterpolations", "CubedSphere", "DataStructures", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "KrylovKit", "LLVM", "LazyBroadcast", "LinearAlgebra", "MultiBroadcastFusion", "NVTX", "PkgVersion", "Random", "RootSolvers", "SparseArrays", "StaticArrays", "Statistics", "UnrolledUtilities"]
-path = ".."
+path = "/Users/tapio/Desktop/Code/ClimaCore.jl"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
-version = "0.15.1"
+version = "0.15.3"
 weakdeps = ["CUDA", "GPUCompiler", "Krylov"]
 
     [deps.ClimaCore.extensions]
@@ -1321,7 +1321,7 @@ version = "1.1.10"
 [[deps.MbedTLS_jll]]
 deps = ["Artifacts", "Libdl"]
 uuid = "c8ffd9c3-330d-5841-b78e-0817d7145fa1"
-version = "2.28.1010+0"
+version = "2.28.2+1"
 
 [[deps.Measures]]
 git-tree-sha1 = "b513cedd20d9c914783d8ad83d08120702bf2c77"
@@ -1351,7 +1351,7 @@ uuid = "a63ad114-7e13-5084-954f-fe012c677804"
 
 [[deps.MozillaCACerts_jll]]
 uuid = "14a3606d-f60d-562e-9121-12d972cd8159"
-version = "2025.12.2"
+version = "2023.1.10"
 
 [[deps.MultiBroadcastFusion]]
 git-tree-sha1 = "4157ffb36526a275f644a3fbe378850e5eb1b470"
@@ -1429,7 +1429,7 @@ version = "0.3.24+0"
 [[deps.OpenBLAS_jll]]
 deps = ["Artifacts", "CompilerSupportLibraries_jll", "Libdl"]
 uuid = "4536629a-c528-5b80-bd46-f80d51c5b363"
-version = "0.3.23+5"
+version = "0.3.23+4"
 
 [[deps.OpenLibm_jll]]
 deps = ["Artifacts", "Libdl"]
@@ -2381,7 +2381,7 @@ version = "1.52.0+1"
 [[deps.p7zip_jll]]
 deps = ["Artifacts", "Libdl"]
 uuid = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
-version = "17.6.1+0"
+version = "17.4.0+2"
 
 [[deps.x264_jll]]
 deps = ["Artifacts", "JLLWrappers", "Libdl"]
--- a/.buildkite/Manifest.toml
+++ b/.buildkite/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.10"
+julia_version = "1.10.11"
 manifest_format = "2.0"
 project_hash = "19194cb825213fa6ff692385eb4c7eed7f946f0f"
 
@@ -347,9 +347,9 @@ weakdeps = ["CUDA", "MPI"]
 
 [[deps.ClimaCore]]
 deps = ["Adapt", "BandedMatrices", "BlockArrays", "ClimaComms", "ClimaInterpolations", "CubedSphere", "DataStructures", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "KrylovKit", "LLVM", "LazyBroadcast", "LinearAlgebra", "MultiBroadcastFusion", "NVTX", "PkgVersion", "Random", "RootSolvers", "SparseArrays", "StaticArrays", "Statistics", "UnrolledUtilities"]
-path = "/Users/tapio/Desktop/Code/ClimaCore.jl"
+path = ".."
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
-version = "0.15.3"
+version = "0.15.1"
 weakdeps = ["CUDA", "GPUCompiler", "Krylov"]
 
     [deps.ClimaCore.extensions]
@@ -1321,7 +1321,7 @@ version = "1.1.10"
 [[deps.MbedTLS_jll]]
 deps = ["Artifacts", "Libdl"]
 uuid = "c8ffd9c3-330d-5841-b78e-0817d7145fa1"
-version = "2.28.2+1"
+version = "2.28.1010+0"
 
 [[deps.Measures]]
 git-tree-sha1 = "b513cedd20d9c914783d8ad83d08120702bf2c77"
@@ -1351,7 +1351,7 @@ uuid = "a63ad114-7e13-5084-954f-fe012c677804"
 
 [[deps.MozillaCACerts_jll]]
 uuid = "14a3606d-f60d-562e-9121-12d972cd8159"
-version = "2023.1.10"
+version = "2025.12.2"
 
 [[deps.MultiBroadcastFusion]]
 git-tree-sha1 = "4157ffb36526a275f644a3fbe378850e5eb1b470"
@@ -1429,7 +1429,7 @@ version = "0.3.24+0"
 [[deps.OpenBLAS_jll]]
 deps = ["Artifacts", "CompilerSupportLibraries_jll", "Libdl"]
 uuid = "4536629a-c528-5b80-bd46-f80d51c5b363"
-version = "0.3.23+4"
+version = "0.3.23+5"
 
 [[deps.OpenLibm_jll]]
 deps = ["Artifacts", "Libdl"]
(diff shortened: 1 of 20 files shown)


src/Operators/dg_fluxes.jl
```diff
@@ -144,6 +144,15 @@ end
 
 Add consistent LDG/SIPG face coupling
 ``−\\{\\!\\{κ G\\}\\!\\}·n̂ + τ[[q]]`` to a WJ-weighted Laplacian residual.
-"""
-add_ldg_laplacian_flux_internal!(dydt, q, G, κ, τ) =
-    add_numerical_flux_internal!(LDGLaplacianFlux(τ), dydt, q, G, κ)
+Accepts a shared [`start_dg_ghost_exchange`](@ref) handle started on
+`(q, G, κ)`.
+"""
+add_ldg_laplacian_flux_internal!(dydt, q, G, κ, τ; ghost_exchange = nothing) =
```
imreddyTeja: nit: If possible, I prefer to avoid kwargs because they do not participate in method dispatch. If the tests say this function (and the others that now use kwargs) don't allocate, then it is probably fine

src/Operators/numericalflux.jl
```diff
@@ -5,13 +5,6 @@ Abstract type for numerical flux functions used in DG methods.
 """
 abstract type AbstractNumericalFlux end
 
-"""
-    AbstractBoundaryCondition
-
-Abstract type for boundary conditions in DG methods.
-"""
-abstract type AbstractBoundaryCondition end
```
imreddyTeja, line 13: good catch, I didn't notice this.

src/Operators/numericalflux.jl
```diff
@@ -337,7 +367,301 @@ function _add_interior_face_flux_extruded_2d!(
...
+        arg -> arg isa Fields.Field ? Fields.field_values(arg) : arg,
+        args,
+    )
+    recv = unrolled_map(
+        ex -> isnothing(ex) ? nothing : ex.recv_data,
+        ghost_bufs,
+    )
+    # The slot schedule is identical across the exchanged arguments; with no
+    # exchanged argument (all `recv` entries `nothing`), slots are never read.
+    exs = filter(!isnothing, ghost_bufs)
```
imreddyTeja: nit: `unrolled_filter` can be used here

src/Operators/numericalflux.jl
```diff
@@ -337,7 +367,301 @@ function _add_interior_face_flux_extruded_2d!(
...
+            )
+            recv_slabs⁺ = unrolled_map(
+                r -> isnothing(r) ? nothing : slab(r, v, slot),
+                recv,
+            )
+            for q in 1:Nq
+                i⁻, j⁻ = Topologies.face_node_index(face⁻, Nq, q, false)
+                q′ = reversed ? Nq - q + 1 : q
+
+                lg⁻ = lg_slab⁻[1, i⁻, j⁻, 1]
```
imreddyTeja: these could probably be annotated as inbounds

src/Operators/numericalflux.jl
```diff
@@ -337,7 +367,301 @@ function _add_interior_face_flux_extruded_2d!(
...
+# `Utilities.Cache.clean_cache!` releases it). The exchange is created once
+# per key, matching the create-once pattern of the limiter and DSS buffers:
+# each `ClimaComms.graph_context` draws a fresh MPI tag from a counter that
+# wraps at 32767, so an exchange per call would reallocate the strips every
+# tendency evaluation and eventually alias the tag of a live graph context.
+# The argument position `i` is part of the key because same-typed arguments in
+# one call each need their own send buffer. `nothing` (no ghost faces) is a
+# valid cached value, hence the `:missing` sentinel.
+function _dg_face_exchange(space, data, i)
+    key = (Topologies.GhostFaceExchange, Spaces.grid(space), typeof(data), i)
```
imreddyTeja, line 658: This key might not be specific enough. If `a` and `b` are fields with the same space and eltype, then `start_dg_ghost_exchange(a)` followed by `start_dg_ghost_exchange(b)` would write into the same exchange buffer.

src/Operators/numericalflux.jl
```diff
@@ -375,33 +699,68 @@ function ghost_state(::ReflectingWallBC, normal, argvals⁻)
...
+function _add_numerical_flux_boundary!(
+    ::ClimaComms.AbstractCPUDevice,
+    fn::F,
+    dydt,
+    args...,
+) where {F}
     space = axes(dydt)
     Nq = Quadratures.degrees_of_freedom(Spaces.quadrature_style(space))
     topology = Spaces.topology(space)
     boundary_surface_geometries = Spaces.grid(space).boundary_surface_geometries
```
imreddyTeja: I think this would error on an ExtrudedFiniteDifference space

test/runtests.jl
```diff
@@ -99,6 +99,7 @@ unit_tests = [
     UnitTest("Spectral elem - DG stability properties"  ,"Operators/spectralelement/unit_dg_stability.jl"; meta = :cpu_only, tier = :unit, subsystem = :operators),
     UnitTest("Spectral elem - DG boundary fluxes"       ,"Operators/spectralelement/unit_dg_boundary_fluxes.jl"; meta = :cpu_only, tier = :unit, subsystem = :operators),
```
akshaysridhar:
```suggestion
    UnitTest("Spectral elem - DG boundary fluxes"       ,"Operators/spectralelement/unit_dg_boundary_fluxes.jl"; tier = :unit, subsystem = :operators),
```

akshaysridhar: CUDA extensiosn for the dg boundary fluxes exist so I think we can relax the cpu restriction.

akshaysridhar (approved):
Tested against multi-GPU (2, 4) runs on `clima` A100 devices - the changes work with downstream copies of the Roe / Kennedy-Gruber / Ranocha fluxes. Fixed 2 tests to relax a `cpu_only` tag, re-running buildkite checks now.

ext/cuda/operators_dg.jl
```diff
@@ -96,18 +104,75 @@ Operators._add_numerical_flux_internal!(
...
+    space,
+    args,
+)
+    topology = Spaces.topology(space)
+    ClimaComms.context(topology) isa ClimaComms.SingletonCommsContext &&
+        return Operators.NO_DG_GHOST_EXCHANGE
+    grid = Spaces.grid(space)
+    Nv =
+        grid isa Grids.ExtrudedFiniteDifferenceGrid ? Spaces.nlevels(space) :
+        1
```
imreddyTeja: I don't think this is needed. `Spaces.nlevels` should return 1 for a spectral element space. Same comment for other uses of this pattern

src/Operators/numericalflux.jl
imreddyTeja: I think this would error on an ExtrudedFiniteDifference space

imreddyTeja (commented):
mostly looks good to me. The only one that might be an issue is the comment about the cache key. I'll take another look after I review the horizontal fusion PR

ext/cuda/operators_dg.jl
```diff
@@ -220,6 +370,68 @@ Base.@propagate_inbounds function dg_face_flux_kernel!(
...
+# (the single-kernel analog of `Topologies.fill_face_send_buffer!`): strip `s`
+# holds the `Nq` face-node values of local face `(slot_lidx[s], slot_face[s])`,
+# in the face's natural node order.
+function dg_face_pack_kernel!(
+    send_data,
+    data,
+    slot_lidx,
+    slot_face,
+    ::Val{Nq},
+    Nv,
```
imreddyTeja, line 383: might be good to Val Nv here to prevent a runtime non const division

src/Utilities/Utilities.jl
```diff
@@ -204,15 +204,15 @@ with special handling of `DataType` fields to avoid errors during compilation.
 
 # Examples
 
-```jldoctest; setup = :(import ClimaCore.Utilities: new), filter = r"\\d+"
-julia> new(Int)
-4889520192
+```jldoctest; setup = :(import ClimaCore.Utilities: new)
```
dennisYatunin:
```suggestion
```jldoctest; setup = :(import ClimaCore.Utilities: new), filter = r"\d+"
```
I don't think you should change the examples here; I just got the `filter` syntax wrong. The examples demonstrate that `new` can return any arbitrary value, not just 0.

.buildkite/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.11"
+julia_version = "1.10.10"
```
dennisYatunin: Was the downgrade on purpose?

tapios:
> mostly looks good to me. The only one that might be an issue is the comment about the cache key. I'll take another look after I review the horizontal fusion PR

Thanks for the helpful review, @imreddyTeja!

src/Operators/numericalflux.jl
```diff
@@ -309,18 +339,18 @@ function _add_interior_face_flux_extruded_2d!(
...
+                        slab_ isa DataLayouts.AbstractData ?
+                        slab_[1, i⁻, j⁻, 1] : slab_,
+                    arg_slabs⁻,
+                )
+                argvals⁺ = unrolled_map(
+                    slab_ ->
+                        slab_ isa DataLayouts.AbstractData ?
+                        slab_[1, i⁺, j⁺, 1] : slab_,
+                    arg_slabs⁺,
+                )
```
dennisYatunin, line 370: The chain of `unrolled_map` calls leading up to here will likely have higher latency than a single call. I'd recommend inlining the `Field->DataLayout` and `DataLayout->slab` maps into this `slab->value` map.

src/Operators/numericalflux.jl
tapios: Done

src/Operators/numericalflux.jl
tapios: Done

src/Operators/numericalflux.jl
tapios: Yes, good point. Fixed by adding an `in_flight` latch to `GhostFaceExchange`

src/Operators/numericalflux.jl
```diff
@@ -337,7 +367,339 @@ function _add_interior_face_flux_extruded_2d!(
...
+operators — see [`start_dg_ghost_exchange`](@ref). The first operator that
+consumes the handle completes the exchange (`ClimaComms.finish`); later
+consumers read the same recv strips.
+"""
+struct DGGhostExchange{B}
+    bufs::B
+    finished::Base.RefValue{Bool}
+end
+DGGhostExchange(bufs) = DGGhostExchange(bufs, Ref(false))
+const NO_DG_GHOST_EXCHANGE = DGGhostExchange(nothing, Base.RefValue{Bool}(true))
```
dennisYatunin:
```suggestion
const NO_DG_GHOST_EXCHANGE = DGGhostExchange(nothing, Ref(true))
```
(to match the syntax on the line above)

dennisYatunin (approved):

src/Operators/numericalflux.jl
tapios: Good point. Fixed by mimicking the GPU boundary path (and added tests to catch this).

ext/cuda/operators_dg.jl
tapios: Done

ext/cuda/operators_dg.jl
tapios: Correct, done

src/Operators/dg_fluxes.jl
tapios: It was compiled away here, but (with Claude's help), made it a positional argument anyway.

src/Utilities/Utilities.jl
tapios: Ok

.buildkite/Manifest.toml
tapios: No, mistake from local development. Reverted.

src/Operators/numericalflux.jl
tapios: collapsed into single map

PR #2598, opened 2026-08-21
Enable fusion of spectral element operators

dennisYatunin:
## Purpose

This PR turns spectral element operators into ordinary broadcast expressions built on the loop primitives from #2522, completing the second task outlined in #2468. A `SpectralBroadcasted` is now just a `Broadcasted` whose function is a `SpectralElementOperator`, so spectral operators compose with pointwise broadcasts, run through the same scope-dispatched `foreach_slice` machinery on CPUs and GPUs, and no longer need hand-written CUDA kernels.

## Disclosure

The second commit in this PR was generated using Claude. The PR description was also partially generated by Claude.

## Content

- Rewrite `SpectralElementOperator`s as slab-wise applications of the DataLayouts loop primitives
  - `SpectralBroadcasted` is now `Broadcasted{SpectralStyle}`; each operator is an `apply_operator(op, args...)` on one slab, driven by `foreach_slab` — one fused loop on CPUs, one kernel launch on GPUs, replacing `ext/cuda/operators_spectral_element.jl` and `ext/cuda/operators_sem_shmem.jl`
  - Weak-form operators are parameterizations of the strong-form ones (e.g. `Divergence{WeakForm}`), with `WeakDivergence` etc. as deprecated aliases; `NEWS.md` records the interface changes
  - Shape queries replace operator nodes with constants of their result eltypes (`drop_operators`), which keeps the mutually recursive queries below Julia's recursion limiter inside fused kernels
- Support spectral operators in fused loop bodies
  - Later statements may read earlier results, both pointwise and through operators; cross-statement spectral reads publish through synchronized shared buffers
  - In-body temporaries live in per-thread registers, so they stay live across later spectral statements instead of aliasing the equal-size shared operator buffers; `Restrict` cannot read a register-resident argument and throws a descriptive error
  - `Utilities.@drop_recursion_limits` and `Utilities.@drop_constprop` centralize the `Method`-level inference controls, covering `Core.kwcall` methods and their hidden body functions, where keyword arguments put the loop bodies
  - A Tuple-only `Utilities.Unrolled.unrolled_tuple_map` carries the hot broadcast-tree walks (about a tenth of a spectral expression's compilation memory); `UnrolledUtilities` imports are unchanged from `main`
- Keep shared memory unaliased and the launch configuration consistent with the slice loops
  - Operator destinations live in per-thread registers; shared-memory globals are named by byte size, since same-named globals from separately compiled functions resolve to one
[...]

PR #2606, opened 2026-08-29
GPU performance optimizations

tapios:
## Summary

Stacked on top of #2598.

This PR introduces GPU performance optimizations targeting large-scale cubed-sphere simulations (evaluated on an NVIDIA A100 with $h_\text{elem}=30, z_\text{elem}=63$), plus horizontal Laplacian atoms that centralize the hyperdiffusion pattern:

1. **Flattened `FieldVector` broadcasts on GPU** (the bulk of the speedup):
   - When every array in a `FieldVector` broadcast has the same contiguous layout as the destination, the expression is evaluated over `vec`s of the arrays with plain linear indexing, bypassing CUDA.jl's N-dimensional Cartesian index computation (4 sequential integer divisions/modulos per thread for these 4-D arrays). Vector update bandwidth reached **1.31 TB/s** (**3.04× faster per update**).
   - In-place updates like `x .-= dx` preserve destination object identity through the flattening, so `Base.Broadcast`'s aliasing check no longer allocates a defensive copy of the destination on every stage update — previously **7.85 GB of device allocations per step** on the 63-level sphere; time-step integration is now zero-allocation, worth another ~5% per step (241.0 → 228.2 ms).
   - The GPU extension marks its array types via a `Fields.is_gpu_array_type` trait (SubArrays recurse into their parent type); CPU broadcasts keep the standard path unchanged. Unit-tested: the gate, the transform, the destination-identity invariant, aliased in-place updates, and the end-to-end `FieldVector` broadcast (which exercises the flattened path on GPU CI).
2. **Horizontal Laplacian atoms + overlapped intermediate DSS in hyperdiffusion:**
   - Adds `Operators.scalar_laplacian` and `Operators.vector_laplacian`, the building blocks of ∇⁴ hyperdiffusion. On CG spaces they return lazy, element-local operator expressions that fuse into consuming broadcasts; on DG spaces `scalar_laplacian` includes the interior-penalty face corrections itself and `weighted_dss!` is a no-op, so one prep → `weighted_dss!` → apply sequence serves both discretizations.
   - The hybrid example's `hyperdiffusion_tendency!` is rewritten on the atoms: both intermediate Laplacians are computed before a single `Spaces.weighted_dss!(ᶜχ => ..., ᶜχuₕ => ...)` call, batching the scalar and vector exchanges into one communication phase, and the 2D/3D branching disappears into the atoms.
3. **Launch configuration & occupancy tuning:**
   - Capped `MAX_SUBBLOCK_LAUNCH_THREADS = 128` (4 warps/block) for 2D spectral slabs, halving shared-memory allocation per block to allow higher concurrent block occupancy per SM on large meshes.
   - Eager FD launch configurations pack 2 columns per block (128 threads) instead of 256, reducing register pressure, and query the SM count from the device instead of hardcoding the A100's 108.
   - The launch-configuration search keeps its **single-wave default** for now: with the allocation fix in place, a realistic-problem sweep found 4–8 waves ~6% faster per step (215 vs 228 ms), but the previously measured reduction slowdown at higher wave counts is unresolved, so the safest setting stands pending broader validation.
   - `CLIMA_CUDA_MAX_WAVES`, `CLIMA_FD_MAX_THREADS`, `CLIMA_DSS_MAX_THREADS`, and `CLIMA_COLLECT_KERNEL_STATS` environment variables (read once at module load into `Ref`s, never on the launch path) override the defaults; `perf/sweep_kernel_configs.jl` sweeps them on a benchmark case.
4. **Coalesced DG face-kernel memory access:**
   - Reshaped the DG staging buffer and `DGConnectivity` surface geometry to level-first layouts (`(Nv, Nq, nsides, nfaces)` and `(Nv, Nq, nfaces)`) and reordered thread indexing in the interior, boundary, ghost-face, and pack kernels to `(v, q, f)`, so warp threads share the face-node indices derived from `(q, f)` and read fields, geometry, and staging in contiguous 128-byte transactions; face kernels run in 128-thread blocks (`DG_FACE_BLOCK_THREADS`). Face flux evaluation is **1.72× faster**, accelerating DG flux computations by ~30% overall.

---

## Performance Metrics

*Benchmark: Baroclinic wave (`sphere/baroclinic_wave_rhoe`), $h_\text{elem}=30, z_\text{elem}=63$, $N_q=5$ (sweeps at $N_q=4$), $dt=400\,\text{s}$, NVIDIA A100 (40 GB SXM4).*

| Configuration | Median Step Time | Min Step Time | SYPD (Median) | SYPD (Max) | Speedup |
| :--- | :---: | :---: | :---: | :---: | :---: |
| **Baseline (#2598)** | 293.6 ms | 291.8 ms | 3.73 | 3.75 | Baseline |
| **+ FieldVector 1D Flattening** | 238.0 ms | 237.3 ms | 4.60 | 4.62 | +23.4% |
| **+ Overlapped DSS & Tuning** | 237.5 ms | 237.1 ms | 4.615 | 4.622 | +23.7% |
| **+ Aliasing fix (zero device allocs/step)** | **228.2 ms** | — | — | — | **+28.7%** |

- The DG face-kernel coalescing (item 4) does not affect this CG benchmark; its 1.72× face-flux speedup was measured on the DG unit benchmarks.
- With the experimental `CLIMA_CUDA_MAX_WAVES=8` override, the same case reaches 214.9 ms/step (see the sweep results in the commit messages).

Co-produced with Gemini and Claude.


examples/hybrid/hyperdiffusion.jl
```diff
@@ -19,33 +19,25 @@ function hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
+    # ∇⁴ is two ∇² passes with a DSS between them, so that the second pass
+    # differentiates continuous intermediate fields. Both intermediates are
+    # computed before a single weighted_dss! call, which batches the scalar
+    # and vector exchanges into one communication phase. The Laplacian atoms
+    # handle the horizontal-dimension distinction (no curl-curl term with one
+    # horizontal dimension), and on DG spaces they include the face
+    # corrections themselves (where weighted_dss! is a no-op), so this
+    # sequence is discretization-agnostic.
+    ᶜh_tot =
+        Base.Broadcast.broadcasted((ρe, p′, ρ) -> (ρe + p′) / ρ, Y.c.ρe, ᶜp, ᶜρ)
```
dennisYatunin:
```suggestion
    ᶜh_tot = lazy.((Y.c.ρe .+ ᶜp) ./ ᶜρ)
```
This would be much easier to read with `lazy` from `LazyBroadcast.jl`, which is already a dependency of ClimaCore.

ext/cuda/cuda_utils.jl
```diff
@@ -168,11 +168,16 @@ function uncached_launch_configuration(
...
+    # distributions. An A100 sweep over max_waves in (1, 2, 4) on small unit
+    # benchmarks found one wave at least as fast throughout and clearly fastest
+    # for reductions (the equal-shape lazy reduction went from 145 to 129 us).
+    # After the FieldVector aliasing fix removed the per-step device
+    # allocations, a realistic-problem sweep (baroclinic wave, h_elem = 30,
+    # z_elem = 63, A100; see perf/sweep_kernel_configs.jl) found 4-8 waves ~6%
+    # faster per step (215 vs 228 ms). The default stays at one wave until the
+    # reduction slowdown seen at higher wave counts is revisited;
+    # CLIMA_CUDA_MAX_WAVES overrides it for tuning experiments.
+    max_waves = something(default_max_waves, MAX_WAVES[])
```
dennisYatunin, line 177: Since reductions and iterations scale differently, they'll probably need different wave counts for optimal performance. If you don't need to add a `MAX_WAVES` global in this PR, maybe hold off on that for now.

ext/cuda/cuda_utils.jl
```diff
@@ -350,15 +356,39 @@ function _getenv_bool(var::AbstractString; default::Bool = false)
...
 # Always reload when module is imported so precompilation doesn't make it "stick"
 function __init__()
     NAME_KERNELS_FROM_STACK_TRACE[] = _getenv_bool(
         "CLIMA_NAME_CUDA_KERNELS_FROM_STACK_TRACE"; default = false,
     )
+    COLLECT_KERNEL_STATS[] =
+        _getenv_bool("CLIMA_COLLECT_KERNEL_STATS"; default = false)
+    MAX_WAVES[] = max(1, _getenv_int("CLIMA_CUDA_MAX_WAVES", 1))
+    FD_MAX_THREADS[] = max(1, _getenv_int("CLIMA_FD_MAX_THREADS", 128))
+    DSS_MAX_THREADS[] = max(1, _getenv_int("CLIMA_DSS_MAX_THREADS", 256))
```
dennisYatunin: If the user accidentally passes 0 or a negative number, it's probably better to fail loudly than to silently clip with `max(1, _)`.

ext/cuda/operators_dg.jl
```diff
@@ -21,6 +21,12 @@ import ClimaCore: Operators, Topologies, Quadratures, Grids, DataLayouts
 import ClimaCore.Operators: DGConnectivity
 import UnrolledUtilities: unrolled_map
 
+# Block size of the face flux, pack, and gather kernels. They are small,
+# memory-bound, and use no shared memory, so 128-thread blocks (4 warps) keep
+# the launch fine-grained enough to spread the (v, q, f)-ordered work across
+# SMs, matching the 128-thread caps used elsewhere for occupancy.
+const DG_FACE_BLOCK_THREADS = 128
```
dennisYatunin, line 28: Should this and `FD_MAX_THREADS` be unified into a single constant, since they have the same value?

ext/cuda/scopes.jl
```diff
@@ -1,19 +1,27 @@
...
+# allocation is compiled into the kernel before its launch configuration is
+# known, so the two have to agree on a bound. Without a cap the bound is a full
+# 1024-thread block, which makes every slab loop reserve four times the shared
+# memory it can use -- and shared memory is what limits the occupancy of spectral
+# element kernels. The kernels that slice loops replaced used 256; halving the
+# cap to 128 halves the shared memory reserved per block, which raised occupancy
+# and measured faster on an A100 baroclinic wave (h_elem = 30, z_elem = 63).
+# Slabs with more points than the cap fall back to multiple points per thread
+# (see DataLayouts.slice_subscope).
+const MAX_SUBBLOCK_LAUNCH_THREADS = 128
```
dennisYatunin, line 24: Maybe this should also be unified with `FD_MAX_THREADS`, if they are determined by the same optimization strategy?

ext/cuda/scopes.jl
```diff
@@ -61,18 +69,29 @@ Support for multidimensional grids may be added in a future release.
...
-# chain, since it is the scope of every shared-memory allocation.
+# A kernel is partitioned into sub-blocks of MAX_SUBBLOCK_LAUNCH_THREADS threads
+# rather than into blocks, so that every scope in the chain that slice_subscope
+# descends through has a compile-time thread count (see static_num_threads
+# below): DataLayouts.register_similar needs that count to be constant to size a
+# thread's register storage, and a dynamically-sized ThisBlock at the top of the
+# chain would make it a run-time value for every slice with more than
+# THREADS_PER_WARP points. ThisBlock stays a subscope of ThisKernel even though
+# it left the partition chain, since it is the scope of every shared-memory
+# allocation (see DataScope(::Type{<:CUDA.CuDeviceArray}) above).
```
dennisYatunin:
```suggestion
# it is not part of the static partitioning chain; it is the scope of every
# shared-memory allocation (see DataScope(::Type{<:CUDA.CuDeviceArray}) above).
```

ext/cuda/scopes.jl
```diff
@@ -86,23 +105,40 @@ Support for multidimensional blocks may be added in a future release.
...
+# a 4x4 GLL slab then occupies a whole 16-thread block, wasting half of the
+# block's only warp and capping occupancy at the hardware limit of 32 blocks per
+# multiprocessor (512 of 2048 threads), instead of packing 16 slabs into one
+# 256-thread block. Beyond a warp there is nothing to gain.
+#
+# The chain is only this long because every num_slice_points method is a function
+# of an argument's type. When a slice's point count was a run-time value, every
+# scope the descent could stop at ended up in the Union that slice_subscope
+# returned, and inference widened a Union of more than three scopes to Any, which
+# made every slice loop dispatch dynamically and allocate inside GPU kernels.
```
dennisYatunin: This is describing dead code and can be removed from here. If it needs to be put somewhere, the main `src/DataLayouts/scopes.jl` file would be a better location.

ext/cuda/topologies_dss.jl
```diff
@@ -1,9 +1,12 @@
 import ClimaCore: DataLayouts, Topologies
 
+# Generic block-size cap, also used by the DG volume and remapping kernels.
 _max_threads_cuda() = 256
```
dennisYatunin, line 4: If you've dropped all uses of `_max_threads_cuda`, this method should probably be removed.

perf/sweep_kernel_configs.jl
```diff
@@ -0,0 +1,129 @@
+EXAMPLE_DIR = joinpath(dirname(@__DIR__), "examples")
+using CUDA
+using Statistics
+
+ENV["CI_PERF_SKIP_RUN"] = true
+ENV["TEST_NAME"] = "sphere/baroclinic_wave_rhoe"
```
dennisYatunin, line 10: Baroclinic wave isn't very representative of the more complex broadcast expressions launched in ClimaAtmos. This is a great benchmark for demonstration purposes, but I'd be cautious about using it to formalize general-purpose rules. If you want something that generalizes, you'll probably need a test with some nontrivial parametrizations, like the end-to-end AMIP test.

src/Fields/fieldvector.jl
```diff
@@ -344,15 +344,83 @@ end
...
+# everything else keeps the standard path.
+is_gpu_array_type(::Type) = false
+is_gpu_array_type(::Type{<:SubArray{<:Any, <:Any, P}}) where {P} =
+    is_gpu_array_type(P)
+
+@inline is_contiguous_gpu_array(a::DenseArray) =
+    is_gpu_array_type(typeof(a)) && ndims(a) > 0
+@inline is_contiguous_gpu_array(a::SubArray) =
+    is_gpu_array_type(typeof(a)) && ndims(a) > 0 && Base.iscontiguous(a)
+@inline is_contiguous_gpu_array(a::AbstractArray) = false
```
dennisYatunin: This might be simpler if you reuse the `IndexStyle` machinery of `AbstractArrays` and `DataLayouts`.

src/Operators/laplacians.jl
```diff
@@ -0,0 +1,104 @@
...
+        curlcurl = Base.Broadcast.broadcasted(
+            Geometry.Covariant12Vector,
+            Base.Broadcast.broadcasted(
+                wcurl,
+                Base.Broadcast.broadcasted(
+                    Geometry.Covariant3Vector,
+                    Base.Broadcast.broadcasted(curl, u),
+                ),
+            ),
+        )
```
dennisYatunin: The expressions here would also be much simpler if they were rewritten using `LazyBroadcast: lazy`.

test/Fields/unit_fieldvector_flatten.jl
```diff
@@ -0,0 +1,131 @@
...
+# unit-tested without a GPU.
+struct FlatTestArray{T, N} <: DenseArray{T, N}
+    data::Array{T, N}
+end
+Base.size(a::FlatTestArray) = size(a.data)
+Base.getindex(a::FlatTestArray, i::Int) = a.data[i]
+Base.setindex!(a::FlatTestArray, v, i::Int) = (a.data[i] = v)
+Base.IndexStyle(::Type{<:FlatTestArray}) = IndexLinear()
+Base.strides(a::FlatTestArray) = strides(a.data)
+Fields.is_gpu_array_type(::Type{<:FlatTestArray}) = true
```
dennisYatunin, line 24: If the interface just checked for `IndexLinear`, you wouldn't need to add methods for internal functions like `is_gpu_array_type` here.

test/Fields/unit_fieldvector_flatten.jl
```diff
@@ -0,0 +1,131 @@
...
+    smaller = FlatTestArray(rand(FT, 3, 4, 1))
+    cpu = rand(FT, 3, 4, 5)
+
+    @test is_flat_compatible(dest, same)
+    @test is_flat_compatible(dest, FT(2))
+    # Broadcasts that expand a smaller argument keep Cartesian indexing.
+    @test !is_flat_compatible(dest, smaller)
+    # Non-GPU arrays and destinations keep the standard path.
+    @test !is_flat_compatible(dest, cpu)
+    @test !is_flat_compatible(cpu, same)
```
dennisYatunin: Is there a benefit to keeping the Cartesian indexing path? I would expect linear indexing to be faster across all devices.

test/Fields/unit_fieldvector_flatten.jl
```diff
@@ -0,0 +1,131 @@
...
+        same,
+    )
+    @test is_flat_compatible(dest, bc)
+    bc_mixed = Base.Broadcast.broadcasted(+, same, cpu)
+    @test !is_flat_compatible(dest, bc_mixed)
+
+    # The gate and the transform must accept exactly the same argument types:
+    # anything the gate rejects has no flatten_bc_arg method, so a divergence
+    # fails loudly instead of flattening incorrectly.
+    @test !hasmethod(flatten_bc_arg, Tuple{Array{FT, 3}, Vector{FT}, String})
```
dennisYatunin, line 62: It's not very useful to test unexported internal helper functions like `flatten_bc_arg`; doing so locks us into specific designs and makes it harder to tinker with them.

dennisYatunin (approved):
Great followup to #2598; I only had a few style/testing suggestions.

perf/sweep_kernel_configs.jl
tapios: True. But it's what's native to CC, and I'd like to stay within what we have here for now. We'll redo optimizations in ClimaAtmos once we are further along with fusing operations there.

tapios: Also, the big win here was the flattening of VectorField broadcasts, which is independent of the specific optimizations here (which made relatively little, <~ 10%, difference.

test/Fields/unit_fieldvector_flatten.jl
tapios: I disagree. It's important that internal helpers that are consequential are tested. It doesn't lock us in; just change the test if you want to change the helper.

src/Operators/laplacians.jl
tapios: done, rewritten with lazy

ext/cuda/cuda_utils.jl
tapios: done; the caps now error on values < 1.

ext/cuda/scopes.jl
tapios: removed; that type-stability rationale belongs at num_slice_points

ext/cuda/operators_dg.jl
tapios: kept separate; it's coincidental that the values are the same. Memory-bound DG face kernels vs FD column packing. Also FD_MAX_THREADS is a runtime env while DG_FACE_BLOCK_THREADS is a compile-time const.

ext/cuda/scopes.jl
tapios: Same as above; it's coincidental they are the same value: sub-block shared-memory occupancy cap vs FD column packing

ext/cuda/cuda_utils.jl
tapios: True. But the default stays 1 wave; the global exists only so the sweep can override it via env var. Agreed that reductions and iterations scale differently; that's why the default is left conservative rather than raised (larger would have been better for BW)

ext/cuda/topologies_dss.jl
tapios: still used by the DG volume kernels (operators_dg.jl) and remapping (remapping_interpolate_array.jl)

src/Fields/fieldvector.jl
tapios: Done. IndexStyle is helpful here.

test/Fields/unit_fieldvector_flatten.jl
tapios: Yes, in part. The IndexStyle half of the gate now comes from Base, so I removed the test's Base.strides override. But the is_gpu_array_type override stays because the device gate is still there.

test/Fields/unit_fieldvector_flatten.jl
tapios: The beenfit is zero-allocation CPU broadcasts, although linear is indeed faster everywhere. We'd need to make the wrappers allocation-free, which is a good follow-up.

PR #2607, opened 2026-08-30
Reduce compilation time

tapios:
Stacked on top of #2606.

## Summary

Reduces ClimaCore's compilation latency. Profiling showed ~81% of initial space-creation time was Julia type inference over the grid/space constructors and operator broadcasts; this PR caches that work at package precompilation time and restructures the largest constructor.

*Note: based on #2606 (`ts/perf`); only the last two commits are new.*

1. **Precompilation workload** (`src/precompile_workload.jl`, no new dependencies): exercises small Float32/Float64 instances of interval meshes and FD spaces, cubed-sphere spectral-element grids, extruded center/face spaces, coordinate fields, scalar and covariant-vector broadcasts, the hyperdiffusion Laplaed, vector), continuous DSS, the common vertical

Reduces ClimaCore's first-use (TTFX) latency. Profiling showed ~81% of initial space-creation time was Julia type inference over the grid/space constructors and operator broadcasts; this PR caches that work at package precompilation time and restructures the largest constructor.

*Note: based on #2606 (`ts/perf`); only the last two commits are new.*

1. **Precompilation workload** (`src/precompile_workload.jl`, no new dependencies): exercises small Float32/Float64 instances of interval meshes and FD spaces, cubed-sphere spectral-nter/face spaces, coordinate fields, scalar andcovariant-vector broadcasts, the hyperdiffusion Laplacian atoms (scalar, weighted, vector), continuous DSS, the common vertical stencil operators, and FieldVector broadcasts (includdate). Guarded by the 1.10-compatible`ccall(:jl_generating_output, ...)`; `CLIMA_SKIP_PRECOMPILE_WORKLOAD=true` skips it during development. The global object cache is emptied afterwards so no workload objects are serialized into the package image.
2. **`_SpectralElementGrid2D` modularization**: the monolithic ~200-line constructor is split into `@noinline` units (`compute_nodal_local_geometries!`, `apply_bubble_correction!`, `compute_surface_geometries`), each inferred and cached separately. Also removes wasted work: the high-order quadrature area loop (4× the nodal evaluations) previously ran for every element even with
the bubble correction disabled, its result unused.

## Measurements

Representative downstream first use (fresh session, `z_elem = 10`, `h_elem = 4` — sizes *different* from the workload's):

| | before | after |
|---|---|---|
| Total (load + spaces + fields + operators) | 16.9 s (pre-workload) / 7.3 s | **4.5 s** |
| Scalar / vector spectral operator first use | 1.0 s / 2.3 s | 0.13 s / 0.35 s |
| At workload-matched sizes | — | fully cached (~0 s per operation) |

Package precompilation grows ~28 → ~32 s. No runtime code paths change (construction and broadcast results are bit-identical; sphere, bubble-correction, and allocation test tiers pass).

## Known limitation

`Nv`, `Ni`, `Nj` are static type parameters of the data layouts (only `Nh` is dynamic), so column-dependent code re-infers per new vertical extent — the remaining ~3 s at a novel `z_elem`, dominated by the extruded constructor's two `product_geometry` broadcasts (~1.2 s). The fix would be `Nv`-generic construction ayout variant); proposed as a follow-up item for #2605rather than attempted here.

Co-produced with Gemini and Claude

### DIFF
--- a/src/Grids/spectralelement.jl
+++ b/src/Grids/spectralelement.jl
@@ -246,6 +246,236 @@ function get_CoordType2D(topology)
     end
 end
 
+# The "epsilon bubble" correction: the numerical area of the domain (the sum
+# of nodal integration weights times their Jacobians) is not exactly equal to
+# the geometric area (e.g. 4π radius² for the sphere), but the two are
+# required to match. The correction modifies the interior weights of each
+# element so that its numerical area equals the geometric area, approximated
+# by `high_order_elem_area` from a quadrature of twice the order. Linear
+# elements (Nq == 2) have no interior nodes, so the deficit is spread
+# uniformly over all nodes; higher-order elements use the HOMME bubble
+# correction, scaling J at interior nodes only. `@noinline` keeps this branchy
+# block a separately inferred and cached unit of the grid constructor.
+@noinline function apply_bubble_correction!(
+    local_geometry,
+    topology,
+    quadrature_style,
+    global_geometry,
+    autodiff_metric,
+    lidx,
+    elem,
+    elem_area::FT,
+    high_order_elem_area::FT,
+    quad_weights,
+) where {FT}
+    Nq = Quadratures.degrees_of_freedom(quadrature_style)
+    lg_args =
+        (global_geometry, topology, quadrature_style, autodiff_metric, elem)
+    if abs(elem_area - high_order_elem_area) ≤ eps(FT)
+        for i in 1:Nq, j in 1:Nq
+            u, ∂u∂ξ = local_geometry_at_nodal_point(lg_args..., i, j)
+            J = det(parent(∂u∂ξ))
+            WJ = J * quad_weights[i] * quad_weights[j]
+            local_geometry[1, i, j, lidx] = Geometry.LocalGeometry(u, J, WJ, ∂u∂ξ)
+        end
+    else
+        Δarea = high_order_elem_area - elem_area
+        if Nq == 2
+            for i in 1:Nq, j in 1:Nq
+                u, ∂u∂ξ = local_geometry_at_nodal_point(lg_args..., i, j)
+                J = det(parent(∂u∂ξ)) + Δarea / Nq^2
+                WJ = J * quad_weights[i] * quad_weights[j]
+                local_geometry[1, i, j, lidx] =
+                    Geometry.LocalGeometry(u, J, WJ, ∂u∂ξ)
+            end
+        else
+            interior_elem_area = zero(FT)
+            for i in 2:(Nq - 1), j in 2:(Nq - 1)
+                u, ∂u∂ξ = local_geometry_at_nodal_point(lg_args..., i, j)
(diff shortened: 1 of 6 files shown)


src/precompile_workload.jl
```diff
@@ -0,0 +1,103 @@
...
+   get(ENV, "CLIMA_SKIP_PRECOMPILE_WORKLOAD", "false") != "true"
+    let
+        for FT in (Float64, Float32)
+            context = ClimaComms.SingletonCommsContext()
+            vdomain = Domains.IntervalDomain(
+                Geometry.ZPoint(FT(0)),
+                Geometry.ZPoint(FT(1));
+                boundary_names = (:bottom, :top),
+            )
+            vmesh = Meshes.IntervalMesh(vdomain, nelems = 2)
```
dennisYatunin: Since `Nv` is a type parameter, everything you compile here is specialized on `Nv = 2`, so it won't be as useful for our `Nv = 63` and `Nv = 64` fields in AMIP configurations. When using precompilation, all type parameters need to exactly match the ones we use in production runs. You also need to be careful about how you extract subfields, since a field of scalars compiles into different kernel code than a scalar view of a field of `NamedTuple`s. And on top of that, every individual broadcast expression compiles into its own kernel.

In general, I think it would be a good idea to precompile a loop over many different layouts, beyond just the `Float64`/`Float32` distinction you have here. And looping over more broadcast expressions would also be good, though of course there's a limit to how much you can do this before code loading time becomes a problem.

src/Grids/spectralelement.jl
```diff
@@ -246,6 +246,236 @@ function get_CoordType2D(topology)
...
+    Nq = Quadratures.degrees_of_freedom(quadrature_style)
+    _, quad_weights = Quadratures.quadrature_points(FT, quadrature_style)
+
+    if enable_bubble
+        high_order_quadrature_style = Quadratures.GLL{Nq * 2}()
+        high_order_Nq =
+            Quadratures.degrees_of_freedom(high_order_quadrature_style)
+        _, high_order_quad_weights =
+            Quadratures.quadrature_points(FT, high_order_quadrature_style)
+        for (lidx, elem) in enumerate(Topologies.localelems(topology))
```
dennisYatunin, line 340: This would be simpler and more GPU-friendly if written using `foreach_slab`. But if that gives you compilation woes you can just leave this unchanged.

dennisYatunin (approved):
The details of the precompilation workload could use some more consideration, but regardless I expect this will reduce the latency of interactive development.

src/precompile_workload.jl
tapios: Fair point. Added additional precompile workload along the lines suggested.

src/Grids/spectralelement.jl
tapios: Left it unchanged. Doesn't seem to affect GPU (`local_geometry` is host array). It improves CPU thread parallelism but slows down single-threaded path by ~60%.

PR #2608, opened 2026-08-30
Naming cleanup for CG and DG operators

tapios:
Stacked on #2607 (`ts/ttfx`).

Naming cleanup for the DG/spectral-element operators. No behavior change except
the rename of unexported names (no known downstream users, no shims).

- **`Operators.Outflow(; order = 0)`** — physically named convenience
  constructor returning `Extrapolate{order}()` (zero-normal-gradient outflow).
- **DG interior-face rename** `internal → interior`:
  `add_numerical_flux_interior!`, `add_lifting_flux_interior!`,
  `add_ldg_laplacian_flux_interior!` (and the CPU/CUDA device seams), plus the
  private grid field `internal_surface_geometry → interior_surface_geometry`.
- **`LeftBiased` → `BottomBiased` and `RightBiased` → `TopBiased`** (leaving 
   deprecation shims)
- **Weak-form aliases → parameterized form** repo-wide:
  `WeakDivergence()`/`WeakGradient()`/`WeakCurl()` →
  `Divergence{WeakForm}()`/`Gradient{WeakForm}()`/`Curl{WeakForm}()`. The
  aliases stay in `deprecated.jl` for downstream only.
- **Docs**: weak-form math folded into the `Divergence`/`Gradient`/`Curl`
  docstrings; `operators.md` documents the parameterized operators.


src/Operators/finitedifference.jl
```diff
@@ -193,6 +193,24 @@ struct Extrapolate{N} <: AbstractBoundaryCondition
...
+fewer than `order + 1` interior points are in range, as documented for
+[`Extrapolate`](@ref); `Outflow()` is the zero-order (constant-value) closure.
+
+On DG spaces no boundary condition object is needed: the one-sided boundary
+flux applied by [`add_numerical_flux_boundary!`](@ref) evaluates the numerical
+flux from the interior state alone, which is the same zero-order outflow
+closure. If DG later grows ghost-state boundary policies (inflow / wall),
+`Outflow()` becomes one marker in that family.
+"""
+Outflow(; order = 0) = Extrapolate{order}()
```
imreddyTeja, line 207: Should the Divergence operator docstring be updated to use `Outflow` instead of `Extraploate`? I'm fine with either

src/Operators/finitedifference.jl
```diff
@@ -193,6 +193,24 @@ struct Extrapolate{N} <: AbstractBoundaryCondition
...
+advection operators: returns [`Extrapolate{order}()`](@ref Extrapolate), which
+pads the ghost points the interior stencil reaches with an order-`order`
+extrapolation from the interior. Near a boundary the order is reduced when
+fewer than `order + 1` interior points are in range, as documented for
+[`Extrapolate`](@ref); `Outflow()` is the zero-order (constant-value) closure.
+
+On DG spaces no boundary condition object is needed: the one-sided boundary
+flux applied by [`add_numerical_flux_boundary!`](@ref) evaluates the numerical
+flux from the interior state alone, which is the same zero-order outflow
+closure. If DG later grows ghost-state boundary policies (inflow / wall),
```
imreddyTeja: Will this boundary condition ever be used for the DG operators? If so, a ` ghost_state` method should be added, and this should be moved `common` or somewhere shared

docs/src/operators.md
```diff
@@ -126,8 +125,14 @@ SetValue
 SetGradient
 SetDivergence
 Extrapolate
+Outflow
 ```
 
+[`Outflow`](@ref) is a physically named convenience constructor for
+[`Extrapolate`](@ref) (the zero-normal-gradient outflow closure). The same
+outflow closure is the default boundary behavior on DG spaces, applied by
+[`add_numerical_flux_boundary!`](@ref) directly from the interior state.
```
imreddyTeja: nit: To me, this comment implies that `add_numerical_flux_boundary!` is automatically called.

imreddyTeja (approved):
Left some minor comments

src/Operators/finitedifference.jl
tapios: No. Outflow is a finite-difference advection BC; DG boundary faces don't take a BC object (they apply a one-sided flux directly). So no `ghost_state`/`common` move is needed, But I removed the midleading DG paragraph from the docstring so as not to create confusion.

src/Operators/finitedifference.jl
tapios: Good point. Outflow is now listed together with Extrapolate.

docs/src/operators.md
tapios: Agreed, and removed the DG sentence, which didn't really belong here: Extrapolate/Outflow are vertical BCs while the DG boundary flux is horizontal element-boundary treatment.

akshaysridhar (approved):
Tested changes downstream (WIP climacore branch) ; LGTM

NEWS.md
```diff
@@ -4,6 +4,35 @@ ClimaCore.jl Release Notes
 main
 -------
 
+- ![][badge-💥breaking] Renamed the DG interior-face assembly operators from
```
imreddyTeja, line 7:
```suggestion
- ![][badge-💥breaking] Renamed the DG interior-face assembly operators from
```

src/Operators/numericalflux.jl
```diff
@@ -811,14 +813,14 @@ function _add_numerical_flux_boundary!(
...
-    add_numerical_flux_boundary!(numflux::AbstractNumericalFlux, bc::AbstractBoundaryCondition, dydt, args...)
+    add_numerical_flux_boundary!(numflux::AbstractNumericalFlux, bc::HorizontalBoundaryCondition, dydt, args...)
 
 Add numerical flux at boundaries using a typed boundary condition.
 Constructs the ghost state via `ghost_state(bc, normal, argvals⁻)` and applies the numerical flux.
 """
 function add_numerical_flux_boundary!(
     numflux::AbstractNumericalFlux,
-    bc::AbstractBoundaryCondition,
+    bc::HorizontalBoundaryCondition,
```
imreddyTeja, line 823: I think a method for `add_numerical_flux_boundary` that takes a `VericalBoundaryCondition` and throws an error should be added. At the moment, if someone incorrectly calls this with a vertical BC, it ends up dispatching to the above methods instead of erroring out on `ghost_state`

src/Operators/deprecated.jl
```diff
@@ -1,145 +1,35 @@
...
-```
-"""
 const WeakCurl = Curl{WeakForm}
 
-# Nodal getter for Fields, used by the Remapping module and its CUDA extension.
+# The finite-difference biased interpolation operators are named for the
+# vertical direction they lean toward: `BottomBiasedC2F`/`BottomBiasedF2C` take
+# the value below a node, `TopBiasedC2F`/`TopBiasedF2C` the value above. The
+# `LeftBiased`/`RightBiased` spellings are retained for downstream packages; the
+# `false` suppresses the deprecation warning while they migrate.
```
imreddyTeja: I think the thrid arg controls of the deprecated binding get exported
```
    @deprecate old new [export_old=true]

Deprecate method `old` and specify the replacement call `new`, defining a new method `old`
with the specified signature in the process.

To prevent `old` from being exported, set `export_old` to `false`.
```

src/MatrixFields/matrix_multiplication.jl
```diff
@@ -250,8 +250,8 @@ const ⋅ = MultiplyColumnwiseBandMatrixField()
 
 Operators.strip_space(op::MultiplyColumnwiseBandMatrixField, _) = op
 
-struct TopLeftMatrixCorner <: Operators.AbstractBoundaryCondition end
-struct BottomRightMatrixCorner <: Operators.AbstractBoundaryCondition end
+struct TopLeftMatrixCorner <: Operators.VerticalBoundaryCondition end
+struct BottomRightMatrixCorner <: Operators.VerticalBoundaryCondition end
```
imreddyTeja, line 259: nit: now that we have `bottom` and `top` instead of `left` and `right`,  TopLeftMatrixCorner now corresponds to `bottom`. The name made a bit more sense when it corresponded to `left`. Now the "top" of the domain in the "bottom" of the matrix

imreddyTeja: I'm not sure what a good solution to that would be, but this is minor and not user facing. Maybe it would be a good idea to leave a clarifying comment

src/Operators/numericalflux.jl
tapios: Done.

src/Operators/deprecated.jl
tapios: Good catch. The comment was wrong. COrrected now.

src/MatrixFields/matrix_multiplication.jl
tapios: Good point. Added clarifying comment what the matrix corners mean.

Issue #2605, opened 2026-08-29
ClimaCore remaining refactor (including v1)


tapios:
# ClimaCore v1.0 Roadmap: Path to Release with CG and DG

## Context & Purpose

Succeeds #2554 and #2468. Tracks the remaining work to release **ClimaCore v1.0**:
a stable, unified, performant dynamical-core infrastructure supporting both
**Continuous Galerkin (CG)** and **Discontinuous Galerkin (DG)** horizontal
discretizations, with vertical finite-difference infrastructure.

## Definition of "v1.0"

Carried forward from #2554, these are the release *gates*:

1. **API frozen.** Every exported symbol is one we are prepared to support
   through v1.x; deprecation shims are removed or explicitly documented.
2. **Correctness.** CG and DG pass conservation and CPU/GPU-equivalence tests on
   cubed-sphere and extruded spaces, single- and multi-rank, under
   `--check-bounds=yes`.
3. **Performance.** No allocation or significant launch-latency regressions
   against the current benchmarks; masked and fused paths covered.
4. **Docs.** Architecture docs following Diataxis principles.
5. **Downstream green.** ClimaAtmos, ClimaLand, ClimaCoupler, ClimaDiagnostics,
   ClimaUtilities pass against the release.

---

## Merged Since #2554

- [x] **Phase 1 — DataLayouts unification.** 4 core layout types, universal
  `(v, i, j, h)` indexing, `DataScope`/`foreach_slice` loop abstractions (#2522).
- [x] **Field views in `foreach_slice`**, letting Field-level broadcasts fuse
  through the DataLayout loop primitives (#2571). Related GPU hardening: the
  inference-widening fix for broadcast-argument traversal (#2575) and linear
  striding / singleton broadcasting in GPU indexing (#2568).
- [x] **Phase 2 — FD operator stencil abstraction.** Operator deduplication,
  interior/boundary stencil consolidation, linear operators as matrix
  multiplies (#2544, #2526, #2548).
- [x] **Phase 3.2 — Strong/weak spectral form unification** via type-parameter
  dispatch (#2556).
- [x] **Test & example consolidation** (#2570, #2576).
- [x] **DG infrastructure (initial).**
  - Spaces can be marked discontinuous (`discontinuous = true` at grid
    construction); `weighted_dss!` becomes a no-op on them, and the flag
    round-trips through HDF5 serialization (#2599).
[...]

PR #2609, opened 2026-08-31
Add model-level CG↔DG switching via tendency completions

tapios:
Model-level CG↔DG switching (#2605): a tendency written once — an element-local
weak form plus one completion call — runs on both discretizations, selected by
the space.

- **CG/DG as discretization alternatives.** A spectral-element grid stores its
  Galerkin discretization as the singleton `Grids.CG()` (default) or
  `Grids.DG()`, constructed with `discretization = Grids.DG()` and read back with `Grids.discretization(grid)` / `Spaces.discretization(space)` for
  dispatch.
- `Operators.tendency_completion(dydt; numflux, boundary_numflux = nothing)` built from the space and dispatched on `Spaces.discretization`: a DSS buffer
  on `CG()`, interface/boundary numerical fluxes on `DG()`. `numflux` is unused
  on CG, so models pass it unconditionally.
- `Operators.complete_tendency!(completion, dydt, args...)` — `weighted_dss!`
  on CG; on DG the mass-weighted surface term (`*= WJ` → interface/boundary
  flux → `/= WJ`).
- Remove allocations in DG Laplacian
- The Bickley-jet CG/DG smoke test runs both discretizations.

`vector_laplacian` still errors on DG (needs grad-div/curl-curl lifting), so
the switch covers flux-form tendencies for now.

Co-produced with Claude


akshaysridhar:
In the docbuild `examples.md` file there are (at least 2) references/links to the `weak form` operators which still redirect to the strong-form docs (the weak and strong forms are split by the `SplitDivergence` section ; so either providing direct links or adding `SplitDivergence` after the `WeakDivergence` section would clean up the docs section (I built docs off commit `8010a9`, will retry from the latest commit shortly) 

e.g. in `examples.md`

Screenshot 2026-08-31 at 10 54 22 AM

src/Operators/completion.jl
```diff
@@ -0,0 +1,129 @@
...
+    dydt /= WJ
+
+Returns `dydt`.
+"""
+function complete_tendency!(completion::DSSCompletion, dydt, args...)
+    Spaces.weighted_dss!(dydt, completion.buffer)
+    return dydt
+end
+
+function complete_tendency!(completion::NumericalFluxCompletion, dydt, args...)
```
akshaysridhar: This is reasonable - right now my ClimaAtmos tests have the tendency evaluation conditional on the `do_dss!` method (false for `discontinuous=true`) - needs to be changed to follow the `Discretization` type.

src/Operators/laplacians.jl
```diff
@@ -21,24 +21,22 @@ corrections (see [`ldg_laplacian_tendency`](@ref)), and `weighted_dss!` is a
...
-        return ldg_laplacian_tendency(
-            q,
-            weight,
-            κ,
-            ldg_penalty_parameter(κ, space),
-        )
-    end
+    q = Base.Broadcast.materialize(χ)
+    κ = one(Spaces.undertype(space))
+    return ldg_laplacian_tendency(q, weight, κ, ldg_penalty_parameter(κ, space))
```
akshaysridhar: I'll add a density current example which exercises the LDG formulation after this PR is merged.

tapios: Good idea. Thank you!

tapios:
> In the docbuild `examples.md` file there are (at least 2) references/links to the `weak form` operators which still redirect to the strong-form docs (the weak and strong forms are split by the `SplitDivergence` section ; so either providing direct links or adding `SplitDivergence` after the `WeakDivergence` section would clean up the docs section (I built docs off commit `8010a9`, will retry from the latest commit shortly)
> 
> e.g. in `examples.md`
> 
> Screenshot 2026-08-31 at 10 54 22 AM

I updated this in the ts/rename branch (#2608)

akshaysridhar (approved):
Paying attention largely to the `complete_tendency!` method since this defines ClimaAtmos DG functionality  -
```
### Simple example - downstream tested 
@. Yₜ.c.ρ = -wdivₕ(ρu)                    # element-local weak-form
Operators.complete_tendency!(completion, Yₜ.c.ρ, Y, p)  
### e.g. Rusanov flux on DG; with assertions on Spaces.is_continuous(space) to support CG DSS / hyperdiffusion / diagnostics . 
```

NEWS.md
```diff
@@ -4,6 +4,60 @@ ClimaCore.jl Release Notes
...
+  DG spaces (each call built ~5 device fields): the gradient and a
+  materialized lazy argument now live in scratch fields keyed on the space's
+  grid (released by the zero-argument `Utilities.Cache.clean_cache!()`), and
+  the new in-place `Operators.ldg_laplacian_tendency!` writes the tendency
+  into a caller-owned field.
+
+- ![][badge-✨feature/enhancement] Added `Operators.scalar_laplacian!(out, χ;
+  weight)`, which writes the Laplacian into a caller-owned field and is
+  allocation-free on both discretizations; `out` may alias `χ`. This is the
+  form to use in a tendency. Correspondingly, `Operators.scalar_laplacian` on
```
imreddyTeja: I think `Operators.scalar_laplacian` should be `Operators.scalar_laplacian!` here

src/Grids/spectralelement.jl
```diff
@@ -664,10 +719,9 @@ end
 @inline _orth_axis(::Geometry.LocalGeometry{I}) where {I} =
     Geometry.Components{Geometry.Orthonormal, I}()
 
-function compute_dss_weights(local_geometry, topology, quadrature_style, discontinuous)
-    !discontinuous && Quadratures.requires_dss(quadrature_style) ||
-        return nothing
-
+compute_dss_weights(local_geometry, topology, quadrature_style, ::DG) =
+    nothing
+function compute_dss_weights(local_geometry, topology, quadrature_style, ::CG)
```
imreddyTeja: `quadrature_style` now unused

imreddyTeja (approved):

NEWS.md
imreddyTeja, line 1: At some point after follow up PRs are merged, it would be a good idea to merge all of the DG additions into one (or a few) entries. We probably don't need to mention fixes to DG because it was never released as broken

docs/src/APIs/dss_api.md
```diff
@@ -23,5 +23,10 @@ Spaces.weighted_dss_ghost!
 Spaces.weighted_dss!
 Spaces.is_continuous
 Grids.is_continuous
+Spaces.discretization
+Grids.discretization
```
dennisYatunin: This is a bit of a confusing pattern, though I know we've used it elsewhere in the past. `Spaces.discretization` could be an import of `Grids.discretization` instead of a new function, so they wouldn't need separate docstrings. If you want to be more generic than `Grids`, it could also go in the top-level `interface.jl`, and be imported in both modules.

docs/Project.toml
```diff
@@ -16,3 +16,12 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
...
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+TaylorSeries = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
+
+[compat]
+# TaylorSeries is an indirect dependency (via CubedSphere); left to the
+# resolver it lands on 0.13, whose Requires-based IntervalArithmetic hook
+# errors at load time against the IntervalArithmetic 1.0 pulled in by the
+# Makie stack. 0.17+ uses package extensions, which are silently skipped on
+# an incompatible IntervalArithmetic instead of erroring.
+TaylorSeries = "0.17, 0.18, 0.19, 0.20, 0.21, 0.22"
```
dennisYatunin:
```suggestion
TaylorSeries = "0.17 - 0.22"
```

src/Grids/spectralelement.jl
```diff
@@ -2,60 +2,110 @@
...
 } <: AbstractSpectralElementGrid
     topology::T
     quadrature_style::Q
     global_geometry::GG
     local_geometry::LG
     dss_weights::D
-    discontinuous::Bool
+    # A type parameter rather than a `Union{CG, DG}` field, so that
+    # `discretization` is inferrable and the methods dispatching on it
+    # (`scalar_laplacian`, `tendency_completion`) resolve statically.
```
dennisYatunin: This is just standard Julia practice, so I don't think the reasoning here needs to be documented. Same for `SpectralElementGrid2D` below.

src/Grids/spectralelement.jl
```diff
@@ -2,60 +2,110 @@
...
+    discretization::Disc
 end
 
 Adapt.@adapt_structure SpectralElementGrid1D
 
 local_geometry_type(
-    ::Type{SpectralElementGrid1D{T, Q, GG, LG}},
-) where {T, Q, GG, LG} = eltype(LG) # calls eltype from DataLayouts
+    ::Type{SpectralElementGrid1D{T, Q, GG, LG, D, Disc}},
+) where {T, Q, GG, LG, D, Disc} = eltype(LG) # calls eltype from DataLayouts
```
dennisYatunin:
```suggestion
    ::Type{SpectralElementGrid1D{<:Any, <:Any, <:Any, LG}},
) where {LG} = eltype(LG) # calls eltype from DataLayouts
```
No need to capture all these extra parameters here, and same for `SpectralElementGrid2D` below.

src/Grids/spectralelement.jl
```diff
@@ -2,60 +2,110 @@
...
+            discretization,
+        )
     end
 end
 
+# The stored discretization is normalized against the quadrature: nodes shared
+# between elements exist only for `requires_dss` quadratures (e.g. GLL), so
+# any other quadrature yields a DG grid regardless of the keyword.
+normalize_discretization(discretization::Discretization, quadrature_style) =
+    Quadratures.requires_dss(quadrature_style) ? discretization : DG()
```
dennisYatunin: This fallback behavior could also lead to silent mistakes. It might be better to throw an `ArgumentError`:
```julia
Quadratures.requires_dss(quadrature_style) || discretization isa DG || throw(...)
```

src/Operators/dg_fluxes.jl
```diff
@@ -87,20 +87,43 @@ WJ-normalized interior-penalty Laplacian tendency approximating
...
-    return residual ./ lgeom.WJ
+    # Face normals are UVVector; G is built in that basis for G·n̂, and the
+    # volume term takes the divergence of the same field (the covariant-to-UV
+    # conversion is exact, so the result differs only by roundoff from a
+    # divergence of the covariant gradient).
+    if ρ_weight === nothing
+        @. G_uv = Geometry.UVVector(grad(q))
+    else
+        @. G_uv = Geometry.UVVector(ρ_weight * grad(q))
+    end
```
dennisYatunin, line 122:
```suggestion
    G = isnothing(ρ_weight) ? lazy.(grad.(q)) : lazy.(ρ_weight .* grad.(q))
    G_uv = lazy.(Geometry.UVVector.(G))
```
Does `G_uv` need to be materialized here? You can reduce the number of kernel launches if it doesn't.

src/Operators/numericalflux.jl
```diff
@@ -70,6 +70,12 @@ const NO_DG_GHOST_EXCHANGE = DGGhostExchange(nothing, Ref(true))
 # Device-dispatch seam (DSS-style): CPU methods live here; the
 # `ClimaComms.CUDADevice` methods are provided by the ClimaCoreCUDAExt
 # extension (ext/cuda/operators_dg.jl).
+#
+# Throughout the CPU face-operator call chain, `args` is declared as
+# `Vararg{Any, N}` with `N` a type parameter: the arguments are mostly just
+# forwarded, so Julia's Vararg heuristic would otherwise compile the chain
+# unspecialized on them, heap-allocating the argument tuple at every call
+# boundary and the index closures at every face node.
```
dennisYatunin, line 78: The `Vararg` heuristic doesn't bite us elsewhere like this, so I'm not entirely convinced we need it here. You might be able to achieve the same type stability using `@inline`?

src/precompile_workload.jl
```diff
@@ -19,7 +19,16 @@ if ccall(:jl_generating_output, Cint, ()) == 1 &&
...
         for FT in (Float64, Float32)
-            context = ClimaComms.SingletonCommsContext()
+            # The device is pinned rather than taken from `ClimaComms.device()`,
+            # which reads `CLIMACOMMS_DEVICE` and throws on `CUDA` unless the
+            # CUDA extension is loaded. This workload runs inside ClimaCore's
+            # own precompilation, where CUDA.jl is a weak dependency and so is
+            # never loaded, so a run script exporting `CLIMACOMMS_DEVICE=CUDA`
+            # before `using ClimaCore` would otherwise fail to precompile.
+            # Pinning the single-threaded device also keeps the compiled
+            # specializations independent of `Threads.nthreads()`.
```
dennisYatunin: `Threads.nthreads()` should never enter type parameter space, so we don't need to worry about accidentally specializing on it. Since the default CPU device is `CPUMultiThreaded`, that should probably be used instead of `CPUSingleThreaded`?

test/Spaces/inference_spaces.jl
```diff
@@ -59,9 +59,15 @@ end
...
             enable_mask = false,
+            discretization = Grids.CG(),
         )
         n_found = length(JET.get_reports(result.analyzer, result.result))
-        n_allowed = 0
+        # `discretization` has no default, so omitting it makes the analyzed
+        # frame throw `UndefKeywordError` unconditionally; JET skips the
+        # resulting no-return frame and reports nothing, and the assertion
+        # then passes having analyzed no code at all. Measured at Float32,
+        # GLL{4}, on a CenterExtrudedFiniteDifferenceSpace.
```
dennisYatunin: This sounds like a mistake made during test development; no need to document it.

test/Operators/spectralelement/unit_laplacians.jl
```diff
@@ -159,6 +163,40 @@ end
...
+    # Each call owns its result, so two results may be live at once. A shared
+    # output buffer would make the second call overwrite the first, and
+    # `∇²f_live + ∇²g_live` would silently evaluate to `2∇²g`.
+    g = @. cosd(coord.long) * sind(coord.lat)
+    ∇²f_live = Operators.scalar_laplacian(f)
+    ∇²g_live = Operators.scalar_laplacian(g)
+    @test ∇²f_live !== ∇²g_live
+    @test parent(∇²f_live) ≈ parent(expected)
+    @test parent(∇²g_live) ≈
+          parent(Operators.ldg_laplacian_tendency(g, nothing, κ, τ))
```
dennisYatunin: This is only a potential failure mode inside kernel code, so it doesn't need to be tested here. When running host-side code, I don't think we encounter duplicate array pointers by accident (unless we do something like returning pointers saved as global constants, which is how the issue arises for `CuStaticSharedArray` in the first place).

test/Operators/spectralelement/unit_laplacians.jl
```diff
@@ -159,6 +163,40 @@ end
...
+    @test parent(out) ≈ parent(expected)
+
+    aliased = copy(f)
+    Operators.scalar_laplacian!(aliased, aliased)
+    @test parent(aliased) ≈ parent(expected)
+
+    # A lazy argument is materialized into a scratch field; the operator is
+    # linear, so the result of doubling the input is exactly double.
+    got_lazy =
+        Operators.scalar_laplacian(Base.Broadcast.broadcasted(x -> 2 * x, f))
```
dennisYatunin:
```suggestion
        Operators.scalar_laplacian(Base.Broadcast.broadcasted(*, 2, f))
```
The anonymous function is unnecessary here.

test/Operators/spectralelement/unit_laplacians.jl
```diff
@@ -159,6 +163,40 @@ end
...
+    @test parent(got_lazy) ≈ 2 .* parent(expected)
+
+    # Feeding a result straight back in (a DG ∇⁴ needs no DSS between the
+    # passes) is well defined, since each call owns its result.
+    ∇²f = Operators.scalar_laplacian(f)
+    expected_∇⁴ = Operators.ldg_laplacian_tendency(∇²f, nothing, κ, τ)
+    got_∇⁴ = Operators.scalar_laplacian(Operators.scalar_laplacian(f))
+    @test parent(got_∇⁴) ≈ parent(expected_∇⁴)
```
dennisYatunin: Is this also testing the "accidentally duplicated array pointers" failure mode? If so, it can probably be removed, since this is only running host-side code.

src/Grids/spectralelement.jl
```diff
@@ -678,21 +721,28 @@ end
...
 
+"""
+    Grids.discretization(grid)
+
+The [`Discretization`](@ref) of `grid`: [`CG`](@ref)`()` or [`DG`](@ref)`()`,
+as stored at grid construction (normalized against the quadrature; see
+[`Discretization`](@ref)). Grids with no horizontal spectral elements (e.g.
+column grids) are `CG()`.
+"""
+discretization(grid::AbstractGrid) = CG()
```
dennisYatunin: This fallback is similar to those that used to exist in `interface.jl`, which I removed in #2598. They had previously caused a lot of developer headache through silently incorrect behavior, and I'm thinking we should avoid a fallback here as well. You can instead add methods for `FiniteDifferenceGrid` and `PointCloudGrid`, and assume that future grids will need their own methods.

If you want to leave the fallback, `nothing` might be a more sensible return value than `CG()`, given the potential failure mode described [here](https://github.com/CliMA/ClimaCore.jl/pull/2609/changes#diff-578d9f1891c7bdbde6f5e15efe7d9c7358fcdfa466c04f8cd6281112f40e02a1R18-R20).

src/Grids/spectralelement.jl
tapios: removed from both methods and both call sites

NEWS.md
tapios: Correct, thank you.

NEWS.md
tapios: Yes, I have rewritten the entries here.

dennisYatunin (approved):

docs/src/APIs/dss_api.md
tapios: Done. Spaces now imports `discretization` and `is_continuous` from `Grids` and adds space methods, so each is one function with one docstring.

docs/Project.toml
tapios: Done.

src/Grids/spectralelement.jl
tapios: Removed

src/Grids/spectralelement.jl
tapios: Done

src/Grids/spectralelement.jl
tapios: Agreed; silent downgrade removed.

test/Spaces/inference_spaces.jl
tapios: Removed.

test/Operators/spectralelement/unit_laplacians.jl
tapios: Done

test/Operators/spectralelement/unit_laplacians.jl
tapios: Kept: it checks a DG ∇⁴ (two passes, no DSS between) composes to the expected value.

src/Grids/spectralelement.jl
tapios: Fallback removed.

src/Operators/dg_fluxes.jl
dennisYatunin: Would you mind confirming if `G_uv` can be lazified here? I might do that in a future PR if it can.

src/Operators/numericalflux.jl
tapios: These CPU face loops allocated ~630 KB/call and now measure 0 B. @inline does not help in that case.

src/Operators/dg_fluxes.jl
tapios: Yes, sorry I didn't reply earlier. Two issues FWIK: `G_uv` is passed to `add_numerical_flux_interior!`, which reads arguments per face node; a `Broadcasted` G_uv then encounters the `_face_node_value(arg, ...) = arg` fallback and would be passed through as a scalar, which is wrong. The form here also seemed essential to make `ldg_laplacian_tendency!` allocation-free, though perhaps there's a cleverer way of doing this. 

This is merged now. But you can try separately.

src/precompile_workload.jl
tapios: Done, both.

PR #2611, opened 2026-08-31
Use multi prefix naming instead of ensemble

### LINKED ISSUE #2597: Change name of `PointCloud` types to "Multi"
I find it confusing that we have types like `Spaces.MultiColumnFiniteDifferenceSpace` and `Spaces.CenterMultiColumnFiniteDifferenceSpace`, but also `Grids.PointCloudGrid` and `Grids.ExtrudedPointCloudGrid`. I think it is cleaner to just use `Multi` for this instead. This can be done when the next breaking release of ClimaCore is made.

ph-kev:
closes https://github.com/CliMA/ClimaCore.jl/issues/2597 - I asked a LLM to look through the codebase and rename the space of multiple points/columns to `MultiPoint*` or `MultiColumn*` where `*` is a `Space` or a `Grid`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -299,6 +299,15 @@ main
   Equation-set-specific fluxes (compressible Euler with Cartesian momentum
   components) are example code. Operator tests are included as part of
   `test/runtests.jl`.
+- The "point cloud" types have been renamed to use the `Multi` prefix:
+  - `Grids.PointCloudGrid` is now `Grids.MultiPointGrid`,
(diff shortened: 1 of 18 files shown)


imreddyTeja (approved):
Would it be worth using `Base.deprecate` for this?

ph-kev:
> Would it be worth using `Base.deprecate` for this?

I used `Base.@deprecate_binding` for this.

Issue #1936, opened 2024-08-16
_coordinate not defined for `LatLong` points


Sbozzolo:
This leads to problems with Remapper.

```
  MethodError: no method matching _coordinate(::ClimaCore.Geometry.LatLongPoint{Float64}, ::Val{1})
  
  Closest candidates are:
    _coordinate(::ClimaCore.Geometry.XZPoint, ::Val{1})
     @ ClimaCore ~/.julia/packages/ClimaCore/ANgUC/src/Geometry/coordinates.jl:171
    _coordinate(::ClimaCore.Geometry.YZPoint, ::Val{1})
     @ ClimaCore ~/.julia/packages/ClimaCore/ANgUC/src/Geometry/coordinates.jl:176
    _coordinate(::ClimaCore.Geometry.Abstract1DPoint, ::Val{1})
     @ ClimaCore ~/.julia/packages/ClimaCore/ANgUC/src/Geometry/coordinates.jl:162
    ...
  
  Stacktrace:
    [1] coordinate(pt::ClimaCore.Geometry.LatLongPoint{Float64}, ax::Int64)
      @ ClimaCore.Geometry ~/.julia/packages/ClimaCore/ANgUC/src/Geometry/coordinates.jl:191
    [2] containing_element(mesh::ClimaCore.Meshes.RectilinearMesh{ClimaCore.Meshes.IntervalMesh{ClimaCore.Domains.IntervalDomain{ClimaCore.Geometry.LongPoint{Float64}, Tuple{Symbol, Symbol}}, LinRange{ClimaCore.Geometry.LongPoint{Float64}, Int64}}, ClimaCore.Meshes.IntervalMesh{ClimaCore.Domains.IntervalDomain{ClimaCore.Geometry.LatPoint{Float64}, Tuple{Symbol, Symbol}}, LinRange{ClimaCore.Geometry.LatPoint{Float64}, Int64}}}, coord::ClimaCore.Geometry.LatLongPoint{Float64})
```

I found that I had to define them like this:
```julia
    ClimaCore.Geometry._coordinate(
        pt::ClimaCore.Geometry.LatLongPoint,
        ::Val{1},
    ) = ClimaCore.Geometry.LongPoint(pt.long)
    ClimaCore.Geometry._coordinate(
        pt::ClimaCore.Geometry.LatLongPoint,
        ::Val{2},
    ) = ClimaCore.Geometry.LatPoint(pt.lat)
```
[...]

PR #2612, opened 2026-08-31
Geometry: fix YZPoint typo, make Points composable

### LINKED ISSUE #1476: Make `Point`s (more) composable
It would be nice to provide additional constructors to compose points, e.g., if `x` is an `XPoint` and `y` is a `YPoint`, `XYPoint(x, y)` should work. Currently, this fails because `XYPoint` expects floats.

I think that this operation is natural and is helpful when one is working with points of different ranks (e.g., creating a 2D grid of points from 1D lists of points).

imreddyTeja:
**Stack 1/4** — base `main`. Next: #2613 (`tr/meshes-fixes`).

Two self-contained `Geometry/coordinates.jl` fixes, one commit each.

> **Note:** this previously also carried a fix for #1936 (`_coordinate` /
> `_coordinate_type` for `LatLong` points). That commit has been dropped —
> the lat/long axis order needs a decision that is being deferred. See below.

### 1. `YZPoint` field typo (not filed)

`_coordinate(p::YZPoint, ::Val{1})` read `p.x`, but `YZPoint`'s fields are `y`
and `z`:

```julia
julia> Geometry.coordinate(Geometry.YZPoint(1.0, 2.0), 1)
ERROR: type YZPoint has no field x
```

That breaks `Meshes.containing_element`/`reference_coordinates` for any
`RectilinearMesh` built from a Y interval and a Z interval. The 2D point testset
only exercised `XYPoint` and `XZPoint`, which is why it was never caught;
`YZPoint` is added to it.

### 2. Make `Point`s composable (#1476)

`XYPoint(XPoint(1.0), YPoint(2.0))` was a `MethodError`. Added constructors that
compose a point from its lower-dimensional parts for the `XY`/`XZ`/`YZ`, `XYZ`
and `LatLong`/`LatLongZ`/`LatLongP` families. They forward to the existing
`product_coordinates`, so mixed component types promote
(`XYPoint(XPoint(1.0f0), YPoint(2.0)) === XYPoint(1.0, 2.0)`) and composing
mismatched axes stays a `MethodError`. `LatLongPoint` takes its two parts in
either order, mirroring `product_coordinates`.

Cartesian points are left alone: `product_coordinates` has no methods for them,
so composing them is a larger change than this issue asks for.

### Why #1936 was dropped

`LatLongPoint`'s axis order is genuinely ambiguous, because both
`product_coordinates(::LatPoint, ::LongPoint)` and the reverse produce a
`LatLongPoint`, and the two real consumers disagree:

| | `intervalmesh1` | needs `Val{1}` → |
| --- | --- | --- |
| ClimaDiagnostics (`islatlonbox`) | Lat | `LatPoint` |
| #1936's reporter (their stack trace) | Long | `LongPoint` |

ClimaDiagnostics also *pirates* these exact signatures as an explicit
"Workaround for ClimaCore #1936"
([netcdf_writer_coordinates.jl:781-784](https://github.com/CliMA/ClimaDiagnostics.jl/blob/3e5bad4556a13c6bb6ffc9619a539caf4fab1315/src/netcdf_writer_coordinates.jl#L781-L784)),
mapping `Val{1}` to `LatPoint`. Whichever order ClimaCore picks, loading
ClimaDiagnostics silently redefines `_coordinate` and leaves it disagreeing with
`_coordinate_type`. So fixing #1936 needs a decision on the canonical order plus
a coordinated removal of that workaround, which is out of scope here.

### Testing

`test/Geometry/unit_geometry.jl` passes (new testsets: `2D XY,XZ,YZ Points`
99/99, `Composable Points` 37/37). `Domains`, `Meshes`, `Spaces`, `Fields`,
`InputOutput` and `Remapping` unit tests all pass.

Closes #1476

### DIFF
--- a/src/Geometry/coordinates.jl
+++ b/src/Geometry/coordinates.jl
@@ -161,7 +161,7 @@ _coordinate(p::XZPoint, ::Val{2}) = ZPoint(p.z)
 
 _coordinate_type(::Type{YZPoint{FT}}, ::Val{1}) where {FT} = YPoint{FT}
 _coordinate_type(::Type{YZPoint{FT}}, ::Val{2}) where {FT} = ZPoint{FT}
-_coordinate(p::YZPoint, ::Val{1}) = YPoint(p.x)
+_coordinate(p::YZPoint, ::Val{1}) = YPoint(p.y)
 _coordinate(p::YZPoint, ::Val{2}) = ZPoint(p.z)
 
 _coordinate_type(::Type{XYZPoint{FT}}, ::Val{1}) where {FT} = XPoint{FT}
--- a/src/Geometry/coordinates.jl
+++ b/src/Geometry/coordinates.jl
@@ -139,6 +139,29 @@ product_coordinates(latlongp::LatLongPoint, zp::ZPoint) =
 product_coordinates(latlongp::LatLongPoint, pressure_p::PPoint) =
     LatLongPPoint(promote(latlongp.lat, latlongp.long, pressure_p.p)...)
 
+XYPoint(xp::XPoint, yp::YPoint) = product_coordinates(xp, yp)
+XZPoint(xp::XPoint, zp::ZPoint) = product_coordinates(xp, zp)
+YZPoint(yp::YPoint, zp::ZPoint) = product_coordinates(yp, zp)
+
+XYZPoint(xyp::XYPoint, zp::ZPoint) = product_coordinates(xyp, zp)
+XYZPoint(xp::XPoint, yp::YPoint, zp::ZPoint) =
+    product_coordinates(product_coordinates(xp, yp), zp)
+
+LatLongPoint(latp::LatPoint, longp::LongPoint) =
+    product_coordinates(latp, longp)
+LatLongPoint(longp::LongPoint, latp::LatPoint) =
+    product_coordinates(longp, latp)
+
+LatLongZPoint(latlongp::LatLongPoint, zp::ZPoint) =
+    product_coordinates(latlongp, zp)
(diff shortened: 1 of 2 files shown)


ph-kev:
For point #2, ClimaDiagnostics commits type piracy and define it here: https://github.com/CliMA/ClimaDiagnostics.jl/blob/3e5bad4556a13c6bb6ffc9619a539caf4fab1315/src/netcdf_writer_coordinates.jl#L781-L784 with the order reversed.

src/Geometry/coordinates.jl
```diff
@@ -139,6 +139,29 @@ product_coordinates(latlongp::LatLongPoint, zp::ZPoint) =
 product_coordinates(latlongp::LatLongPoint, pressure_p::PPoint) =
     LatLongPPoint(promote(latlongp.lat, latlongp.long, pressure_p.p)...)
 
+XYPoint(xp::XPoint, yp::YPoint) = product_coordinates(xp, yp)
+XZPoint(xp::XPoint, zp::ZPoint) = product_coordinates(xp, zp)
+YZPoint(yp::YPoint, zp::ZPoint) = product_coordinates(yp, zp)
+
+XYZPoint(xyp::XYPoint, zp::ZPoint) = product_coordinates(xyp, zp)
```
ph-kev, line 146: Do we need to worry about the `XYZ(x::XPoint, yz::YZPoint)` case?

ph-kev (approved):

src/Geometry/coordinates.jl
imreddyTeja: probably not because non of the other point compose methods have support for cases like that

PR #2614, opened 2026-08-31
Add compile-time ndims for spaces and grids

### LINKED ISSUE #2206: Add simple compile-time way to determine the number of dimensions of a `Space`
The number of spatial dimensions (1-3) of a Space is fully determined by its type.

For example, a `CenterFiniteDifferenceSpace` is guarenteed to have 1 dimension.
```julia
ndims(::CenterFiniteDifferenceSpace) = 1
```
An `ExtrudedFiniteDifferenceSpace` could have 2 or 3, and we can find which one of the two by looking at the type of grid
```julia
ndims(::ExtrudedFiniteDifferenceSpace{G}) where {G <: ExtrudedSpectralElementGrid1D} = 2`
ndims(::ExtrudedFiniteDifferenceSpace{G}) where {G <: ExtrudedSpectralElementGrid2D} = 3`
```

An alternative to this might to return a tuple with `(n_horizontal, n_vertical)` dimensions

This would help with writing code that more easily supports more configurations.

imreddyTeja:
**Stack 3/4** — base `tr/meshes-fixes` (#2613). Next: `tr/gradc2f-docs`.
Review only the last three commits.

### `ndims` for a `Space` (#2206)

The number of spatial dimensions of a space is fixed by its type, but there was
no way to ask for it — `ndims(space)` was a `MethodError`.

`Base.ndims` is now defined on both `Spaces.AbstractSpace` and
`Grids.AbstractGrid`, and on their **types**, so it is a compile-time constant
(the tests assert `@inferred` and `@allocated == 0`).

Spaces built on a grid forward to the grid, mirroring how `local_geometry_type`
is already defined per space type. The generic grid method reads the number of
components off the grid's coordinate type, via a new `Geometry.coordinate_type`
accessor on `LocalGeometry`/`CoordinateOnlyGeometry` that mirrors the existing
`undertype`. That makes it correct for every space without enumerating them.

`ColumnGrid` and `LevelGrid` override it. Both keep the coordinate type of the
grid they slice, so the generic rule would report the **ambient** dimension: a
column of a 3D space would be 3 rather than 1, contradicting the
`ndims(::CenterFiniteDifferenceSpace) == 1` the issue asks for.

I went with a single `Int` rather than the `(n_horizontal, n_vertical)` tuple the
issue floats as an alternative — `has_horizontal`/`has_vertical` already answer
that question, and `ndims` matching `Base`'s meaning seems more useful.

### Drive-by: `local_geometry_type` for `SpectralElementGrid1D` (separate commit)

Found while wiring this up. The method was written for `SpectralElementGrid1D{T,
Q, GG, LG}`, but the struct carries a fifth `D` (dss weights) parameter, so it
**never matched** and every call fell through to the
`local_geometry_type(::Type{T}) = Union{}` fallback. It is matched on `<:` now,
so a future parameter cannot silently reintroduce this. Currently latent — the
one caller that would notice, `InputOutput`, only uses it for point spaces — but
`ndims` depends on it.

### Testing

New `ndims` testset in `test/Spaces/unit_spaces.jl`, 38/38: explicit 1D/2D/3D
spaces, columns and levels, the compile-time checks, and a sweep over every space
`TestUtilities.all_spaces` knows about, cross-checked against the number of
components of each space's coordinates. `Spaces`, `Meshes`, `Domains`,
`Geometry`, `Fields`, `InputOutput` and `Remapping` unit tests pass.

Closes #2206

### DIFF
--- a/src/Spaces/Spaces.jl
+++ b/src/Spaces/Spaces.jl
@@ -224,6 +224,28 @@ get_mask(space::ExtrudedFiniteDifferenceSpace) =
     get_mask(horizontal_space(space))
 
 """
+    ndims(space::AbstractSpace)
+
+The number of spatial dimensions of `space`.
+
+This is fixed by the type, so it is available at compile time: a
+`CenterFiniteDifferenceSpace` is always 1, a `SpectralElementSpace2D` always 2,
+and an `ExtrudedFiniteDifferenceSpace` is 2 or 3 depending on its horizontal
+space.
+
+```julia
+julia> ndims(Spaces.CenterFiniteDifferenceSpace(grid))
+1
+```
+
+Spaces that are built on a `Grids.AbstractGrid` forward to `ndims` of that
+grid; the rest read the number of components off their coordinate type.
+"""
+Base.ndims(space::AbstractSpace) = ndims(typeof(space))
+Base.ndims(::Type{S}) where {S <: AbstractSpace} =
+    Geometry.ncomponents(Geometry.coordinate_type(local_geometry_type(S)))
+
+"""
     has_vertical(::AbstractSpace)
 
 Returns a bool indicating that the space has a vertical grid.
--- a/src/Geometry/localgeometry.jl
+++ b/src/Geometry/localgeometry.jl
@@ -116,3 +116,14 @@ end
 undertype(::Type{<:LocalGeometry{I, C, FT}}) where {I, C, FT} = FT
 undertype(::Type{SurfaceGeometry{FT, N}}) where {FT, N} = FT
 undertype(::Type{<:CoordinateOnlyGeometry{C}}) where {C} = eltype(C)
+
+"""
+    coordinate_type(geom)
+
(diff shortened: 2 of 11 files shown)


src/Spaces/Spaces.jl
```diff
@@ -223,6 +223,28 @@ get_mask(space::SpectralElementSpaceSlab) = DataLayouts.NoMask()
...
 
+"""
+    ndims(space::AbstractSpace)
+
+The number of spatial dimensions of `space`.
+
+This is fixed by the type, so it is available at compile time: a
+`CenterFiniteDifferenceSpace` is always 1, a `SpectralElementSpace2D` always 2,
+and an `ExtrudedFiniteDifferenceSpace` is 2 or 3 depending on its horizontal
+space.
```
ph-kev: I don't think this is relevant to say that it is available at compile time. It might be better just to give this as a list of all spaces and the number of dimensions.

imreddyTeja:
Closing because defining dims for spaces and grids is non-trivial. This is a problem that should be thought about more

PR #2615, opened 2026-08-31
Document GradientC2F's behaviour with no boundary condition

### LINKED ISSUE #2473: Clarify default boundary conditions for `GradientC2F`
In [§2.2 of the tutorial in the docs](https://clima.github.io/ClimaCore.jl/stable/tutorials/introduction/#2.2-Finite-difference-operators), the following is stated:

> **Center to face gradient**
>
> Uses the same stencil, but doesn't work directly:
> ```julia
> sinz = sin.(column_center_coords.z)
> gradc2f = ClimaCore.Operators.GradientC2F()
> # ∇sinz = gradc2f.(sinz) ## this would throw an error
> ```
> 
> This throws an error because face values at the boundary are _not_ well-defined:

I am on v0.14.50, and this does not seem to be true, `∇sinz = gradc2f.(sinz)` will not throw an error.
I have tried this with some experiments and it seems that the following boundary condition is applied:
At indices `0`, `n+1`, homogeneous Dirichlet data is assumed, that is, at (nonexisting) cell centers _outside the domain_.[^1]
This leads to a boundary gradient value that is _half the size_ of what you would get by applying homogeneous Dirichlet boundary conditions the standard way (i.e., `SetValue(0.0)`).
If I understand the code behavior correctly, it seems that with `SetValue()`, the Dirichlet data is placed at the faces (indices `1/2`, `n+1/2`).
Below I have included a script that I wrote to understand what's going on, maybe it is helpful.

I am not so familiar with the code, so I am not sure that this is actually what's going on.
But perhaps you can clarify this in the documentation/the tutorial.

[^1]: I am following the numbering convention from [here](https://clima.github.io/ClimaCore.jl/dev/operators/#Finite-difference-operators)

---

Example script:

```julia
import ClimaCore as CC
using ClimaComms
using ClimaCorePlots
using Plots

function get_coord(lower_boundary::Float64, upper_boundary::Float64, nelems::Int)
    device = ClimaComms.device()
    domain = CC.Domains.IntervalDomain(
        CC.Geometry.ZPoint(lower_boundary),
        CC.Geometry.ZPoint(upper_boundary);
        boundary_names=(:bottom, :top),
    )
    mesh = CC.Meshes.IntervalMesh(domain, nelems=nelems)
    cspace = CC.Spaces.CenterFiniteDifferenceSpace(device, mesh)
    coord = CC.Fields.coordinate_field(cspace)
    return coord
end

function plot_gradient(lower_boundary::Float64, upper_boundary::Float64, nelems::Int, color)
    coord = get_coord(lower_boundary, upper_boundary, nelems)
    sinz = sin.(coord.z)

    bottom_bc = CC.Operators.SetValue(0.0)
    top_bc = CC.Operators.SetValue(0.0)
    gradc2f = CC.Operators.GradientC2F(bottom=bottom_bc, top=top_bc)
    # uncomment the line below to switch to GradientC2F without boundary conditions
    # gradc2f = CC.Operators.GradientC2F()
    ∇sinz = gradc2f.(sinz)
    @info parent(CC.Geometry.WVector.(∇sinz))[end]
    plot!(
        map(x -> x.w, CC.Geometry.WVector.(∇sinz)), 
        color=color, 
        label="nelems = $nelems",
        xlabel="z",
        ylabel="∇sin(z) approximation"
    )
end

lower_boundary = 0.0
upper_boundary = 10.0
nelems = 10

plot()
plot_gradient(0.0, 10.0, 10, :black)
plot_gradient(0.0, 10.0, 5, :blue)
plot_gradient(0.0, 10.0, 20, :green)
plot_gradient(0.0, 10.0, 100, :red)
savefig("grad_comparison.png")

```

imreddyTeja:
The tutorial claimed `gradc2f.(sinz)` "throws an error when bounds checking is
enabled", and otherwise had "undefined behavior at boundary ... it will read some
random memory outside the array". 

Neither is true any more. Since the switch to
NaN boundary rows for missing boundary conditions (`1b17a043e`), the broadcast
runs and fills the two boundary faces with `NaN`:

```
no BC:            [NaN, 0.91795, 0.70257, 0.38023, 0.0, -0.38023, -0.70257, -0.91795, NaN]
SetValue(0.0):    [0.99359, 0.91795, ..., -0.91795, -0.99359]
SetGradient(1.0): [1.0, 0.91795, ..., -0.91795, 1.0]
```

- Rewrite the tutorial section to say what actually happens, and let the snippet
  run instead of leaving it commented out.
- Add an "Omitting a boundary condition" section to the `GradientC2F` docstring,
  including that `SetGradient` and `SetValue` place their data in different
  places — the actual confusion in the issue — and that a boundary only needs a
  condition if the enclosing broadcast reads that face.
- Add the same point, operator-generic, to the boundary-conditions section of
  `docs/src/operators.md`, which is the page the issue links to.
- Fix an unrelated bug for local geometry type method

### Verification

Both claims were checked in a REPL rather than inferred: the boundary faces are
`NaN` with no boundary condition, and `divf2c.(gradc2f.(x))` contains no `NaN`s,
so the missing inner boundary really is harmless. The tutorial snippet was run
against the tutorial's own column setup. The example block is fenced as `julia`,
not `jldoctest`, so it is not executed by `doctest = true`.

Closes #2473

### DIFF
--- a/docs/tutorials/introduction.jl
+++ b/docs/tutorials/introduction.jl
@@ -271,17 +271,7 @@ plot(map(x -> x.w, ClimaCore.Geometry.WVector.(∇cosz)), ylim = (0, 10))
 
 # **Center to face gradient**
 #
-# Uses the same stencil, but doesn't work directly:
-
-sinz = sin.(column_center_coords.z)
-gradc2f = ClimaCore.Operators.GradientC2F()
-# this throws an error when julia is launched with --check-bounds=yes
-# if ran with nomrally (auto bounds checking), behavior at the boundary is undefined
-# (i.e. it will read some random memory outside the array)
-# ∇sinz = gradc2f.(sinz) ## undefined behavior at boundary
-#----------------------------------------------------------------------------
-
-# This throws an error when bounds checking is enabled because face values at the boundary are _not_ well-defined:
+# Uses the same stencil, but the boundary faces are _not_ well-defined:
 #
 # ```
 # ...
@@ -297,6 +287,17 @@ gradc2f = ClimaCore.Operators.GradientC2F()
 #         ????
 # ```
 #
+# The stencil for a face needs a center value on each side of it, and the two
+# boundary faces only have one. Broadcasting the operator without a boundary
+# condition does _not_ throw: it runs, and fills the two boundary faces with
+# `NaN`.
+
+sinz = sin.(column_center_coords.z)
+gradc2f = ClimaCore.Operators.GradientC2F()
+∇sinz_nan = gradc2f.(sinz)
+first(parent(ClimaCore.Geometry.WVector.(∇sinz_nan))) ## NaN
+#----------------------------------------------------------------------------
+
 # To handle boundaries we need to *modify the stencil*. This is done by providing the *gradient* $\nabla\theta^*$ of $\theta$ at the boundary:
 #
 # ```math
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -114,6 +114,28 @@ upwind_biased_product_c2f_dirichlet
 
 ## Finite difference boundary conditions
 
+A boundary condition is attached to an operator by boundary name, e.g.
+`GradientC2F(; bottom = SetGradient(v₀), top = SetGradient(v₁))`. Which
+conditions an operator accepts is listed in its own docstring.
+
+A boundary left without a condition is not an error, but what it does depends
+on the operator:
+
+| Operator                                                                                                                       | Boundary left without a condition                           |
+|:------------------------------------------------------------------------------------------------------------------------------ |:----------------------------------------------------------- |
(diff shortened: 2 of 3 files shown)


docs/tutorials/introduction.jl
```diff
@@ -297,6 +287,18 @@ gradc2f = ClimaCore.Operators.GradientC2F()
 #         ????
 # ```
 #
+# The stencil for a face needs a center value on each side of it, and the two
+# boundary faces only have one. Broadcasting the operator without a boundary
+# condition does _not_ throw: it runs, and fills the two boundary faces with
+# `NaN`. That way a forgotten boundary condition shows up in the output rather
+# than silently producing a plausible number.
```
ph-kev: I think the last sentence can be removed or better explain why someone would care about this.

ph-kev (approved):

PR #2617, opened 2026-08-31
Make QuasiMonotoneLimiter work on GPU and MPI

### LINKED ISSUE #2610: Restarts with MPI and QuasiMonotoneLimiter do not work
## Describe the bug
An `isbits` error is thrown when trying to restart an mpi simulation in ClimaAtmos that uses the QuasiMonotoneLimiter. This is not exercised in ClimaCore or ClimaAtmos ci.

## To Reproduce

I have not had time to make a minimal reproducer. [This](https://github.com/CliMA/ClimaAtmos.jl/commit/f1e626160292e2981e18ba9cb8b29626cf01bad4) is a commit on a ClimaAtmos branch that can reproduce this bug.

ph-kev:
closes https://github.com/CliMA/ClimaCore.jl/issues/2610 - This PR fixes the issue by adding support for CUDA + MPI for the `QuasiMonotoneLimiter`.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -613,8 +613,10 @@ steps:
         command: "srun julia --color=yes --check-bounds=yes --project=.buildkite test/Limiters/distributed/dlimiter.jl"
         env:
           CLIMACOMMS_CONTEXT: "MPI"
+          CLIMACOMMS_DEVICE: "CUDA"
         agents:
           slurm_ntasks: 3
+          slurm_gpus: 1
           modules: mpiwrapper/2024_05_27 climacommon/2024_10_10
 
   # TODO: improve performance label: [inference, allocs, flops, latency, benchmark, boundscheck]
--- a/src/Limiters/quasimonotone.jl
+++ b/src/Limiters/quasimonotone.jl
@@ -248,19 +248,21 @@ end
 Update the field `limiter.q_bounds_nbr` based on `limiter.q_bounds` in the ghost
 neighbors. This should be called after the ghost exchange has completed.
 
-!!! note
-
-    This loop indexes slabs of the receive buffer from the host, so the
-    distributed limiter is only supported on CPUs. Running it with a
-    `CUDADevice` and an `MPICommsContext` would trigger scalar indexing of a
-    `CuArray`. Making distributed limiters work on GPUs requires replacing
-    this loop (and the `q_bounds_nbr` update below) with a kernel.
-
 Part of [`compute_bounds!`](@ref).
 """
+compute_neighbor_bounds_ghost!(
+    limiter::QuasiMonotoneLimiter,
+    topology::Topologies.AbstractTopology,
+) = compute_neighbor_bounds_ghost!(
+    limiter,
+    topology,
+    ClimaComms.device(topology),
+)
+
 function compute_neighbor_bounds_ghost!(
     limiter::QuasiMonotoneLimiter,
     topology::Topologies.AbstractTopology,
+    dev::ClimaComms.AbstractCPUDevice,
 )
     q_bounds_nbr = limiter.q_bounds_nbr
     (Nv, _, _, Nh) = size(q_bounds_nbr)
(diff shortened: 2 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -613,8 +613,10 @@ steps:
         command: "srun julia --color=yes --check-bounds=yes --project=.buildkite test/Limiters/distributed/dlimiter.jl"
         env:
           CLIMACOMMS_CONTEXT: "MPI"
+          CLIMACOMMS_DEVICE: "CUDA"
```
imreddyTeja, line 616: probably a good idea to keep an MPI + CPU test instead of replacing it

src/Limiters/quasimonotone.jl
```diff
@@ -248,19 +248,21 @@ end
 Update the field `limiter.q_bounds_nbr` based on `limiter.q_bounds` in the ghost
 neighbors. This should be called after the ghost exchange has completed.
```
imreddyTeja, line 250: I think the `min_tracer_mass = Inf` in `apply_limiter!` below might be a bug. Could you look into that if possible? It would also be good to update the PR description with a list of things fixed

.buildkite/pipeline.yml
ph-kev: Do we want to keep both MPI + GPU and MPI + CPU tests?

imreddyTeja: up to you. you probably know the most about this now

src/Limiters/quasimonotone.jl
tapios: Yes, it's a bug. It is passed into a max, that then always defaults to Inf. Would be good to fix here.

tapios (commented):
A few other bugs in limiters that would be good to fix in this PR:

1) quasimonotone.jl (l. 379). The warn is unconditional and warns whenever the `warn` flag is present. Should be conditional on non-convergence: `if warn && !converged`

2) `apply_limiter!` on GPU returns `nothing` (ext/cuda/limiters.jl, l. 145); on CPU it returns `ρq`. Would be good to make consistent (I'm not sure why `ρq` is returned on CPU, maybe it can also be nothing?).

3) `QuasiMonotoneLimiter` should probably a subtype of `AbstractLimiter`, like the other limiters, i.e., `struct QuasiMonotoneLimiter{D, G, FT, CS} <: AbstractLimiter` in quasimonotone.jl, l. 98.

tapios:
> A few other bugs in limiters that would be good to fix in this PR:
> 
> 1. quasimonotone.jl (l. 379). The warn is unconditional and warns whenever the `warn` flag is present. Should be conditional on non-convergence: `if warn && !converged`
> 2. `apply_limiter!` on GPU returns `nothing` (ext/cuda/limiters.jl, l. 145); on CPU it returns `ρq`. Would be good to make consistent (I'm not sure why `ρq` is returned on CPU, maybe it can also be nothing?).
> 3. `QuasiMonotoneLimiter` should probably a subtype of `AbstractLimiter`, like the other limiters, i.e., `struct QuasiMonotoneLimiter{D, G, FT, CS} <: AbstractLimiter` in quasimonotone.jl, l. 98.

Correction to previous version of the above comment: The quasi-monotone limiter must be fixed in a different way. 

a) Change quasimonotone.jl (l. 372) to `min_tracer_mass = min(min_tracer_mass, slab_min_tracer_mass)`
b) Change the logic in quasimonotone.jl (l. 502 ff)

```
converged_now = abs(Δtracer_mass) <= rtol * abs(tracer_mass)
rel_err = iszero(tracer_mass) ? zero(FT) : abs(Δtracer_mass) / abs(tracer_mass)
max_rel_err = max(max_rel_err, rel_err)
min_tracer_mass = min(min_tracer_mass, abs(tracer_mass))
converged_now && break
```

The problem is the division by the tracer mass in the current code. It NaNs for zero tracer, which is not uncommon, e.g., for specific humidity. 

The per-iteration update of `min_tracer_mass` inside the loop is also redundant since tracer_mass is constant for the slab, and can be hoisted above the loop.

@akshaysridhar can help here. Would be good add/revise tests so this works reliably (e.g., add a unit test that runs the limiter on a field with an all-zero element and asserts convergence and a finite `max_rel_err`--it would have caught this bug).

PR #2620, opened 2026-09-01
Unify the CG/DG bickleyjet examples + enforce symmetry in DG laplacian operator

akshaysridhar:
Fixes laplacian operator convergence (Adds term to satisfy symmetry property). Updates corresponding unit tests.

Merges CG / DG Bickley jet (flux form) example into a single unified driver applying the new `complete_tendency` API. 

(Bickley example refactor + docs + formatter via Claude)


src/Operators/dg_fluxes.jl
```diff
@@ -75,101 +75,190 @@ central_curl3_lift(normal, (u⁻, v⁻), (u⁺, v⁺)) =
...
 
-In-place form of [`ldg_laplacian_tendency`](@ref): writes the tendency into
-`out`, using `G_uv` as scratch for the gradient. Neither `out` nor `G_uv` may
-alias `q` or `ρ_weight`. Returns `out`.
+In-place [`ldg_laplacian_tendency`](@ref): writes into `out`, using `G_uv` and
+`R_uv` as scratch. None of the three may alias `q` or `ρ_weight`. Returns
+`out`.
 """
-function ldg_laplacian_tendency!(out, G_uv, q, ρ_weight, κ, τ)
+function ldg_laplacian_tendency!(out, G_uv, R_uv, q, ρ_weight, κ, τ)
```
imreddyTeja: question that is semi-related to these changes: If boundaries exist for the horizontal domain, do those need to be dealt with seperately?

imreddyTeja: Does `κ` need to be a constant here?

imreddyTeja (approved):
`bickleyjet.jl` looks good to me!

src/Operators/dg_fluxes.jl
akshaysridhar: For non-periodic boundaries we would need to account for boundary fluxes via `add_numerical_flux_boundary!` ; see https://github.com/CliMA/ClimaCore.jl/blob/803457513d1d1c9235201930451f01c5e88c8acf/src/Operators/completion.jl#L66

κ need not be a constant : 
```Favg = -((κ⁻ * G⁻ + κ⁺ * G⁺) / 2)' * normal```
where  κ⁻ , κ⁺ can be fields ( let me know if I've missed some broadcast rule here) - for the simpler examples these tend to be constants (e.g. bickley jet/ density current problems)

tapios (approved):
LGTM.

Two details:

1) It may be good to add a symmetry test with a non-uniform weight.
2) The switch uses `Spaces.CG()/Spaces.DG()` but the docstrings and NEWS use `Grids.CG()`--would be good to be consistent

PR #2625, opened 2026-09-04
Assorted test clean up

imreddyTeja:
I told claude to look for some common cleanup items I noticed in the tests.

- Tests in for loops that incorrectly loop over only one value
- unused domain setup
- unused gpu_broken
- deprecated name usage
- lines that are testing something but missing the `@test`

### DIFF
--- a/test/Operators/spectralelement/unit_rectilinear.jl
+++ b/test/Operators/spectralelement/unit_rectilinear.jl
@@ -24,410 +24,373 @@ domain = Domains.RectangleDomain(
 Nq = 5
 quad = Quadratures.GLL{Nq}()
 device = ClimaComms.CPUSingleThreaded()
-grid_mesh = Meshes.RectilinearMesh(domain, 17, 16)
-grid_topology =
-    Topologies.Topology2D(ClimaComms.SingletonCommsContext(device), grid_mesh)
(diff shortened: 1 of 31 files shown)


test/Operators/spectralelement/unit_rectilinear.jl
```diff
@@ -24,410 +24,373 @@ domain = Domains.RectangleDomain(
...
-    Topologies.Topology2D(ClimaComms.SingletonCommsContext(device), grid_mesh)
-grid_space = Spaces.SpectralElementSpace2D(grid_topology, quad)
-grid_coords = Fields.coordinate_field(grid_space)
-
-grid_test_setup = (grid_topology, grid_space, grid_coords)
+mesh = Meshes.RectilinearMesh(domain, 17, 16)
+topology =
+    Topologies.Topology2D(ClimaComms.SingletonCommsContext(device), mesh)
+space = Spaces.SpectralElementSpace2D(topology, quad)
+coords = Fields.coordinate_field(space)
```
ph-kev: I am not sure if this is true, but these lines should be wrapped in a `@testset`, so the GC can reclaim them. This is a problem with `@safetestset`, but ClimaCore isn't using that and is hand rolling their own solution.

ph-kev (approved):

PR #2630, opened 2026-09-09
Directly use GPUCompiler in cuda ext

imreddyTeja:
I won't merge until I confirm downstream works with CUDA v6.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,7 +39,7 @@ Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 
 [extensions]
-ClimaCoreCUDAExt = "CUDA"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -39,7 +39,7 @@ Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 
 [extensions]
-ClimaCoreCUDAExt = "CUDA"
+ClimaCoreCUDAExt = ["CUDA", "GPUCompiler"]
```
ph-kev, line 42: I am assuming that CUDA has `GPUCompiler` as a dependency so users don't need to load both themselves?

ph-kev (approved):

Project.toml
imreddyTeja: yes. I tested that loading only CUDA.jl and ClimaCore triggers the extension load

PR #2635, opened 2026-09-10
Change metric of multiple column grid to identity matrix

ph-kev:
This PR adds the necessary functionality to support ClimaAtmos simulation of multiple columns. This includes allowing the user to pass the `deep` keyword argument to `Grids.ExtrudedFiniteDifferenceGrid` when using the convenience constructor for a multiple column grid/space and missing functionality for `MultiPointSpace` that ClimaAtmos need.

In addition, the metric of a `MultiPointSpace` was changed to be the identity matrix which is consist with a `PointSpace`.


test/Operators/unit_multicolumn_column_equivalence.jl
```diff
@@ -0,0 +1,107 @@
+# Applying the same operator to a MultiColumnFiniteDifferenceSpace or a
```
imreddyTeja, line 1: should deep atmosphere equivalence be tested here as well?

src/Grids/multipoint.jl
```diff
@@ -12,12 +12,8 @@ basis, DSS, or horizontal operators are supported on this grid.
 This is the horizontal component used by a multi-column extruded space (N
 independent columns at user-chosen sphere locations).
 
-The `local_geometry` is stored as a `VIJFH{LG, 1, 1, 1, N}` data layout, with
-each of the `N` locations represented by an element with one nodal point. Based
-on the [metric
-tensor](https://en.wikipedia.org/wiki/Metric_tensor#The_round_metric_on_a_sphere)
-of a sphere, the horizontal Jacobian `∂x∂ξ` is given by the diagonal matrix
-`diag(R·π/180, R·cosd(lat)·π/180)`, with the determinant `J = R²·cosd(lat)·(π/180)²`.
+Every location has unit horizontal metric term. The sphere radius is only used for the global
```
imreddyTeja: The top of this docstring says no horiontal ops are supported. That should be updated

imreddyTeja: I also think it should be
"horizontal metric term" -> "horizontal metric terms"

imreddyTeja (approved):
Looks good, other than some minor docstring issues

src/CommonGrids/CommonGrids.jl
```diff
@@ -695,8 +695,9 @@ vertical columns at given latitude-longitude locations on a sphere: a
 [`Grids.FiniteDifferenceGrid`](@ref) vertical grid, with `Grids.Flat()`
 hypsography and a shallow spherical global geometry.
 
-The columns have no horizontal connectivity, so horizontal operators are not
-defined on the grid; `Fields.bycolumn` iterates over the columns.
+The columns have no horizontal connectivity: DSS is not supported, and
```
imreddyTeja, line 698: Should the `MultiColumnGrid` have a `deep` arg?

imreddyTeja: Asking because the ` MultiPointGrid` docstring says `radius` is used for deep-atmos scaling

imreddyTeja (approved):

src/CommonGrids/CommonGrids.jl
ph-kev: The `radius` is used in ClimaAtmos to do scaling for the radiation but it is not used in ClimaCore. I don't think the `MultiColumnGrid` should have a `deep` kwarg.

Issue #2640, opened 2026-09-14
Cannot use DivergenceC2F with SetValue boundary conditions in an operator matrix


valentinaschueller:
## Describe the bug

Using the `DirichletC2F` operator with a `SetValue` boundary condition inside implicit timesteppers seems to be no longer possible, since the call to `operator_matrix()` fails.
I believe this bug is a result of the FD operator revamp, #2544, and thus present since v0.16.0.

The [docs section on boundary conditions](https://clima.github.io/ClimaCore.jl/stable/howto/boundary_conditions/#Vertical-boundaries-(finite-difference-operators)) explicitly mentions that `SetValue` boundary conditions are supported for `DirichletC2F`, so this is unexpected.

_Side note:_ I have tried using an equivalent `SetGradient` condition on the `GradientF2C` operator instead.
This silently leads to `NaN` values in the operator matrix and the solution becomes `NaN` when using an implicit timestepping method.
I believe this change of behavior can also be inferred from the [release notes for 0.16.0](https://github.com/CliMA/ClimaCore.jl/blob/main/NEWS.md#v0160).

## Reproducer

I have adapted this from [the cubed sphere example from the docs](https://clima.github.io/ClimaCore.jl/stable/tutorials/extruded_sphere/#3.-Explicit-horizontal-and-implicit-vertical-tendencies) (by the way, very nice with the new structure!):

```julia
using ClimaComms
ClimaComms.@import_required_backends
import ClimaCore: Domains, Meshes, Spaces, Fields, Geometry, Operators
import ClimaCore.MatrixFields
import ClimaCore.MatrixFields: @name, ⋅

vert_domain = Domains.IntervalDomain(
    Geometry.ZPoint(0.0),
    Geometry.ZPoint(1.0);
    boundary_names = (:bottom, :top),
)
vert_mesh = Meshes.IntervalMesh(vert_domain; nelems = 10)
device = ClimaComms.device()
vert_space = Spaces.FaceFiniteDifferenceSpace(device, vert_mesh)

C3 = Geometry.WVector
divᵥ = Operators.DivergenceC2F(;
    bottom = Operators.SetValue(C3(0.0)),
    top = Operators.SetValue(C3(0.0)),
)
divᵥ_matrix = MatrixFields.operator_matrix(divᵥ)
```

Error message (latest version of the main branch, corresponding to v0.16.2):
```julia
[...]

PR #2646, opened 2026-09-17
Add unit test for DivergenceC2F with SetValue boundary condition

valentinaschueller:
Unit test for the `DivergenceC2F` operator with a `SetValue` (Dirichlet) boundary condition, baseline check to showcase the bug reported in #2640


valentinaschueller:
(I am a bit surprised that the unit tests pipeline succeeds, the test I added fails on my end. On the other hand, I do not find [the name of the corresponding unit test](https://github.com/CliMA/ClimaCore.jl/blob/80e3ce17802277306f08c09a6114be2eb780b46b/test/runtests.jl#L153) "MatrixFields - operator matrices" in the logs, so it seems like it was not run?)

PR #2650, opened 2026-09-18
Avoid unnecessary errors in broadcast axes checks

dennisYatunin:
This PR fixes two bugs in `Field`/`DataLayout` broadcasting:
- As mentioned [here](https://github.com/CliMA/ClimaLand.jl/pull/1882#discussion_r4041855083), the `foreach_slice` axis check was flagging `data .= slice_of_data` as a `DimensionMismatch`. This PR combines the left-hand side and right-hand side of a broadcast expression into one `Broadcasted` object before they are passed to `foreach_slice`, so there is no longer any ambiguity from mismatched axes.
- As seen [here](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1517) and [here](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1519), the `issubspace` check was returning `false` when comparing a horizontal space to a single level from an extruded space. This PR reimplements `issubspace` in terms of a lower-level `issubgrid` function, which recurses through grid wrappers like `LevelGrid` and `ColumnGrid`. To avoid inference failures for spaces without grids, `issubspace` now also has methods for `PointSpace` and `SpectralElementSpaceSlab`.

Both bugs were introduced during the recent ClimaCore refactoring. Unit tests have been added to prevent similar regressions in the future.

This PR was tested against nightly AMIP [here](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1520), using the latest main branches of all other CliMA packages.


szy21 (commented):
It looks good to me, but it's probably better for Nat and Teja to look at it.

src/Grids/column.jl
```diff
@@ -47,3 +47,9 @@ vertical_topology(colgrid::ColumnGrid) = vertical_topology(colgrid.full_grid)
 local_geometry_data(colgrid::ColumnGrid, staggering::Staggering) =
     column(local_geometry_data(colgrid.full_grid, staggering), colgrid.indices...)
 global_geometry(colgrid::ColumnGrid) = global_geometry(colgrid.full_grid)
+
+issubgrid(subgrid::ColumnGrid, grid::ColumnGrid) = subgrid === grid
+issubgrid(subgrid::ColumnGrid, grid::AbstractGrid) =
+    subgrid === grid || subgrid.full_grid === grid
+issubgrid(subgrid::AbstractGrid, grid::ColumnGrid) =
+    subgrid === grid || issubgrid(subgrid, grid.full_grid)
```
nefrathenrici, line 55: I know I tried to add this in the coupler, but I don't think this method should be used either:  recursing `issubgrid(subgrid, grid::ColumnGrid)` onto the full extruded grid would count the extruded grid as a subgrid of the column. I think this should check `issubgrid(subgrid, grid.full_grid.vertical_grid)`.

src/Spaces/Spaces.jl
```diff
@@ -146,7 +147,9 @@ Return `true` if fields on `subspace` can be broadcast against fields on
 extruded `space`, or the vertical space or a column of it. Two spaces built on
 the same grid with different staggering are not subspaces of each other.
 """
-issubspace(subspace::AbstractSpace, space::AbstractSpace) = subspace === space
+issubspace(subspace::AbstractSpace, space::AbstractSpace) =
```
nefrathenrici, line 150: I don't think this will work for all spaces, since some don't have `grid` or `staggering` defined.

